PR #9, opened 2022-07-28
CMake improvements with plugin example

DraTeots:
- **ecce is replaced with epic (BREAKING)**
- Some meta improvements in CMake system
- Plugin boilerplate macros-es are introduced:
   - plugin_add, 
   - plugin_glob_all, 
   - plugin_include_direcotries, 
   - plugin_link_libraries
- Example plugin is made

So I've made that plugin_* macros that minimizes plugins boilerplate to: 

```cmake
# Automatically set plugin name the same as the directory name
# Don't forget string(REPLACE " " "_" PLUGIN_NAME ${PLUGIN_NAME}) if this dir has spaces in its name
get_filename_component(PLUGIN_NAME ${CMAKE_CURRENT_LIST_DIR} NAME)

print_header(">>>>   P L U G I N :   ${PLUGIN_NAME}    <<<<")       # Fancy printing

# Function creates ${PLUGIN_NAME}_plugin and ${PLUGIN_NAME}_library targets
# Setting default includes, libraries and installation paths
plugin_add(${PLUGIN_NAME})

# Find dependencies
find_package(JANA REQUIRED)
find_package(EDM4HEP REQUIRED)
find_package(podio REQUIRED)
find_package(DD4hep REQUIRED)
find_package(ROOT REQUIRED)

# The macro grabs sources as *.cc *.cpp *.c and headers as *.h *.hh *.hpp
# Then correctly sets sources for ${_name}_plugin and ${_name}_library targets
# Adds headers to the correct installation directory
plugin_glob_all(${PLUGIN_NAME})

# Add include directories
# (same as target_include_directories but for both plugin and library)
plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${JANA_INCLUDE_DIR} ${podio_INCLUDE_DIR} ${EDM4HEP_INCLUDE_DIR} ${DD4hep_INCLUDE_DIRS} ${ROOT_INCLUDE_DIRS})

# Add libraries
# (same as target_include_directories but for both plugin and library)
plugin_link_libraries(${PLUGIN_NAME} ${JANA_LIB})
```
It creates both plugin and a library removing `${plugin_name}.cc` from a library list. 
It also provides kind of nice output like: 

```
>>>>   P L U G I N :   data_flow_test    <<<<
---------------------------------------------
-- Including DD4hepBuild.cmake
-- DD4hep uses Geant4
-- Found Python: /usr/include/python3.8 (found suitable exact version "3.8.10") found components: Development Development.Module Development.Embed 
-- |++> Using python executable:  /usr/bin/python3.8
-- Source files:
--      OccupancyAnalysis.cc
--      data_flow_test.cc
-- Plugin-only file is: data_flow_test.cc
-- Header files:
--      OccupancyAnalysis.h
-- 
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,7 @@ plugins/*
 
 # CMake CLion style
 cmake-build-*/
+src/cmake-build-*/
 
 # File-based project format
 *.iws
@@ -86,7 +87,6 @@ tags
 # EMACS
 
 # -*- mode: gitignore; -*-
-*~
 \#*\#
 /.emacs.desktop
 /.emacs.desktop.lock
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -48,16 +48,17 @@ if(NOT DEFINED PLUGIN_OUTPUT_DIRECTORY)
     set(PLUGIN_OUTPUT_DIRECTORY "plugins")
     message(STATUS "${CMAKE_PROJECT_NAME}: Set default PLUGIN_OUTPUT_DIRECTORY")
 endif()
+message(STATUS "${CMAKE_PROJECT_NAME}: PLUGIN_OUTPUT_DIRECTORY: ${PLUGIN_OUTPUT_DIRECTORY}")
 
 # Default plugins static libraries installation directory is 'lib'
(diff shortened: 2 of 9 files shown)


faustus123 (approved):
This looks reasonable to me.

PR #32, opened 2022-08-07
Pass EDM4HEP include directory to make_datamodel_glue.h from cmake using value found.

faustus123:


wdconinc (approved):
Not tested but looks good.

wdconinc:
Should have tested more, I guess... This now fails with the following:
```console
>>>>   P L U G I N :   podio    <<<<
------------------------------------
CMake Error at /opt/local/cmake/RootMacros.cmake:442 (message):
  Cannot find header datamodel_includes.h to generate dictionary
  G__datamodel_vectors for.  Did you forget to set the INCLUDE_DIRECTORIES
  property for the current directory?
Call Stack (most recent call first):
  services/io/podio/CMakeLists.txt:29 (root_generate_dictionary)

-- Configuring incomplete, errors occurred!
See also "/home/wdconinc/git/EICrecon/build/CMakeFiles/CMakeOutput.log".
```

https://github.com/eic/EICrecon/blob/5cad2a399ccd5fe98300420191f39ccbd5e36e39/src/services/io/podio/CMakeLists.txt#L29 We probably should remove the include here, which is non-existent at cmake time.

PR #37, opened 2022-08-08
CI workflows for EICrecon build tests

wdconinc:
Build and install against the most recent jana2@2.0.6 in the eic-shell environment on cvmfs.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -25,7 +25,7 @@ jobs:
         setup: /opt/detector/epic-nightly/setup.sh
         run: |
           # install this repo
-          cmake -B build -S src \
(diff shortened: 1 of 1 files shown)


wdconinc:
This is currently still failing because of an issue with eic-shell (we were stripping our libraries for smaller container sizes, but static libraries become mostly useless that way). That should be resolved overnight as Open Science Grid syncs with docker hub.

wdconinc:
Now this is failing due to jana2@2.0.6 being a hard requirement, not just 2.0.5.

wdconinc:
@faustus123 There are lot of fast-moving PRs getting merged with (one fears) only testing on the immediate development machine of the submitter/merger. I'm hoping a simple CI workflow that attempts to compile on at least 1 other system will be helpful.

PR #50, opened 2022-08-16
Draft: Use generic `algorithms` library in the the EICrecon plugins

wdconinc:
**Note: this is a work in progress branch that will aim to track the `main` branch until it is ready to merge into `main`. That may not happen for some time, though.**

The `algorithms` library is a thin shared library that only depends on DD4hep and EDM4hep (and our extensions of it), and maintains the full commit and development history of all juggler algorithms. It is intended as a compatibility layer between EICrecon and key4HEP efforts and to provide a smooth transition path where continued algorithm development can happen even while algorithms are being ported to EICrecon. The goal here is to remove the need for a lot of code porting from gaudi to jana2, and instead end up with algorithms that are small, modular, self-contained, and rely by design only on the data model for interfacing.

### DIFF
--- a/src/detectors/BEMC/BEMCRawCalorimeterHit_factory.cc
+++ b/src/detectors/BEMC/BEMCRawCalorimeterHit_factory.cc
@@ -12,6 +12,10 @@
 #include <fmt/format.h>
 using namespace dd4hep;
 
+#ifdef USE_ALGORITHMS
+#include "JugDigi/CalorimeterHitDigi.h"
+#endif
+
 //
 // This algorithm converted from:
 //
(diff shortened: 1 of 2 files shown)


DraTeots:
To some degree I was doing this work - moved some Juggler files like 

image

How to avoid redundancy in what we are doing and not do double work? Are you going to provide a thin wrapper replacing Gaudi base classes? 

P.S. There is \<algorithm\> C++ library, we have **algorithms** folder in EICrecon (which just one letter from conflicting with the above) and now there is **algorithms* side library. Could we agree to rename everything (except for stdlib algorith probably) to be less ambiguous?

faustus123:
I'm closing this for now since it is 379 commits behind main. I understand work is being done on this elsewhere and will likely be pulled in after the Oct. campaign gets underway.

PR #57, opened 2022-08-18
EICD support

faustus123:
This puts in support for EICD. This includes build instructions and integration into the podio event source.

This PR also includes:

1. fixes to cmake_wrapper.sh for integrating with CLion
2. Fix to BEMC_test install location
3. Disabling of the podio version checking (which seems to be broken on the podio side, yet, seemed to work previously?)
4. Bug fixes to the generated eicrecon.sh script
5. Up the version number for JANA to v2.0.7 in the documentation (no in cmake configuration yet)
6. Installation of the cmake tools to the EICrecon install directory so they can be used for out-of-repository plugin builds.
7. Updated README with instructions for building spdlog

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,11 @@ print_subdirectory_tree()
 message(STATUS "\n-------------------------------")
 
 
+# Install all cmake helpers
+file(GLOB EICRECON_CMAKE_FILES cmake/*)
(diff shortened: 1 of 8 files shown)


T-Britton:
the eicd instructions still contain a typo. The cmake line contains a reference to ${eicd_ROOT}.  The actual variable is ${EICD_ROOT}

T-Britton (changes requested):
line 110 EICD_ROOT

torrijeske (approved):

Issue #72, opened 2022-09-03
Build unexpectedly succeeds inside of eic-shell


faustus123:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? RHEL7
- Any special versions of ROOT or Geant4?

### Steps to reproduce: (give a step by step account of how to trigger the bug)
Trying to build EICrecon on REHL7 outside of eic-shell container is giving different results from inside container. Steps to build packages currently listed in top-level README.md for EICrecon. 

### Expected Result: (what do you expect when you execute the steps above)
Building in eic-shell seems to work just fine with no errors. I expect the same when building locally using similar package versions. 

Looking at the CMakeLists.txt however, I cannot see where ROOT is introduced to the _rootfile_ project. (It is not in the top-level CMakeLists.txt, nor in the CMakeLists.txt in the services/rootfile directory, nor in the cmake/jana_plugin.cmake file that it uses). Thus, I guess the error is the "expected" behavior and the success of the container build is unexpected(?)

### Actual Result: (what do you get when you execute the steps above)
I get the following error:
~~~
In file included from /home/davidl/work/2022.09.02.EICrecon/EICrecon/src/services/rootfile/rootfile.cc:6:
/home/davidl/work/2022.09.02.EICrecon/EICrecon/src/services/rootfile/RootFile_service.h:15:10: fatal error: TFile.h: No such file or directory
   15 | #include <TFile.h>
~~~

Here are some of the details:
1. I'm using root 6.26.04 which is newer than the container version of 6.22.06, but I don't think that is the issue
2. I'm using cmake 3.24.1 which is newer than the container version of 3.23.1, but I don't think that is the issue
3. cmake does not complain about being unable to find root despite several _find\_package(ROOT REQUIRED)_ lines

If I add these lines to _src/services/rootfile/CMakeLists.txt_, then it compiles just fine.
~~~
find_package(ROOT REQUIRED)
plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${ROOT_INCLUDE_DIRS})
~~~

So basically, why does it work in the container without those lines?

faustus123:
Note: I just submitted a draft PR #73 that gets around the problem, but I'm not convinced this is the right solution

faustus123:
Scratch that. Commit did not include the change that fixes it. Working on it ...

[...]

PR #73, opened 2022-09-03
Davidl cmake fixes

faustus123:
This addresses  issue #72 , though maybe not in a good permanent way.
This allows EICrecon to be built outside of the eic-shell container where you have access to more advanced runtime debugging tools.

The README.md has updated build instructions for fmt and spdlog. Specifically, downgrading fmt to version 8.1.1 (instead of 9.0.0) and building spdlog with -fPIC and having it use external fmt package rather than bundled.

- [X] Bug fix (issue #72)
- [ ] New feature (issue #__)
- [X] Documentation update
- [ ] Other: __

No

No(?)


faustus123:
I believe this is all fixed now. I am able to build and run with both manually built and eic-shell environments. 

The one thing @DraTeots may want to look closely at is that ROOT is now added by default to all plugins built with the jana_plugin.cmake facility. This may may customizing the root module list more difficult for specific directories.

DraTeots:
Looked closely at all changes and merged the commit

Issue #61, opened 2022-08-30
Tool/option to list available types and collection names.


faustus123:
We need a tool to make it easy to list the available data types/name available in a eicrcon process. This should probably just be an option in eicrecon itself. It simply needs to loop over factories and print the data type and tag for each. This should be similar to the podio:print_type_table option in the podio plugin. The tool should actually be able to combine the lists of factories and of collections from the podio file.

The limitation here will be objects placed in the event with Insert() since those will not be known at program start up.

nathanwbrei:
Here's a hint to help get you started: Take a look at JComponentManager::get_component_summary. All JANA components are registered in the JComponentManager after JApplication::Initialize() is called. The JComponentManager doesn't keep a pointer to factories directly (because JANA spins up multiple copies of each); instead, you have a set of JFactoryGenerators which populate a JFactorySet. Once the factory set is populated, you can query it using JFactorySet::Summarize(), or do what JFactorySet::Summarize does and iterate over all the factories to get the data you want.

nathanwbrei:
@faustus123 If we always register the dummy factories with JANA the same way that GlueX does, we can query them just fine. Alternatively, we run JANA for a couple of events on the input file and then check to see what factories showed up.

faustus123:
I assume you are referring to factories that are created merely as containers via JEvent::Insert() operations which don't happen until other factories are processed for an event.

I think having an option to run a few events would be good. Maybe a config parameter that is set to 0 by default, but the user can adjust?

cissieAB:
Ask some dummy questions here: 
is this a tool to call JANA JComponentManager::get_component_summary and print out all the JFactory information?

If it is a tool to call JANA2, should the changes go to JANA repo instead of this one?

nathanwbrei:
Not a silly question. JANA already prints this information out when you call JApplication::Init. My guess as to what David wants is to have a command-line option for the eicRecon executable that prints _only_ that table and then exits.

DraTeots:
But it does only for activated plugins. What about scanning other plugins?

faustus123:
We could have an option to scan all plugins in JANA_PLUGIN_PATH, but I would not make it the default. I would have it print the table of attached plugins and then a brief message saying something like:

Only attached plugins are shown
To scan all available plugins add "-Pprint_factories=scan_all"
To list all available plugins run with --list_plugins

My concern with scanning all of them is that there may be stale or conflicting plugins in the path they would never want to use.

[...]

PR #94, opened 2022-09-07
[discussion only] draft

cissieAB:

Add a cli option "--list_factories" to list all the factories without truly running the JANA processors.

- [ ] Bug fix (issue #__)
- [ x ] New feature (issue #61 __)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ src/services/io/podio/datamodel_glue.h
 # Indirectory builds
 include/*
 plugins/*
+build/*
 
 ##### C++ basic gitignore #####
 
--- a/src/utilities/eicrecon/eicrecon.cc
+++ b/src/utilities/eicrecon/eicrecon.cc
@@ -7,7 +7,7 @@
 
 #include <TFile.h>
 
-#include <JANA/CLI/JMain.h>
+#include "eicrecon_cli.h"
 
 int main( int narg, char **argv)
 {
@@ -34,6 +34,7 @@ int main( int narg, char **argv)
 
     if(const char* env_p = std::getenv("EICrecon_MY")) japp->AddPluginPath( std::string(env_p) + "/plugins" );
 
+    // TODO: add by command line paras
     japp->AddPlugin( "podio"           );
     japp->AddPlugin( "dd4hep"          );
     japp->AddPlugin( "acts"        );
--- a/src/utilities/eicrecon/eicrecon_cli.cpp
+++ b/src/utilities/eicrecon/eicrecon_cli.cpp
@@ -40,7 +40,7 @@ namespace jana {
         std::cout << "   -l   --loadconfigs <file>    Load configuration parameters from file" << std::endl;
         std::cout << "   -d   --dumpconfigs <file>    Dump configuration parameters to file" << std::endl;
         std::cout << "   -b   --benchmark             Run in benchmark mode" << std::endl;
-        std::cout << "   --list_factories             List all the factories without running" << std::endl;
+        std::cout << "   -L   --list_factories        List all the factories without running" << std::endl;
         std::cout << "   -Pkey=value                  Specify a configuration parameter" << std::endl << std::endl;
     }
 
@@ -161,6 +161,7 @@ namespace jana {
         tokenizer["--dumpconfigs"] = DumpConfigs;
         tokenizer["-b"] = Benchmark;
         tokenizer["--benchmark"] = Benchmark;
+        tokenizer["-L"] = ListFactories;
         tokenizer["--list_factories"] = ListFactories;
 
         if (nargs == 1) {
--- a/src/utilities/eicrecon/eicrecon_cli.cpp
+++ b/src/utilities/eicrecon/eicrecon_cli.cpp
@@ -3,6 +3,7 @@
 //
 
 #include "eicrecon_cli.h"
+#include "print_info.h"
 
 #include <JANA/CLI/JVersion.h>
 #include <JANA/CLI/JBenchmarker.h>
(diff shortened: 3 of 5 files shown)


faustus123:
Thanks for putting this in Cissie. A couple of notes:

1. It is preferred for you to submit PRs from a branch within the eic/EICrecon repository rather than from a fork. (I can give you the reasons offline)
2. There should be a short version for the --list_factories option (use "-L" since "-l" is already in use). 
3. Is it possible to suppress printing of the table with all of the workers/cores listed? This fills a lot of space and is distracting when running in the list_factories mode where the information being sought does not include this info.
4. Have the factory table print such that the columns are aligned. It is much easier to read and eventually, this list is going to be quite long.
5. By default, the types from the input file (if given) should be included. The user, will generally be interested in the full list of available types and not just those generated by factory algorithms. This will be a little trickier, since it can only be done if the user supplies a filename AND if that happens to be a podio file. Ideally, JANA will decide on whether there is a source to open then the right JEventSource subclass to instantiate. Then you can try doing something like downcasting to JEventSourcePODIOsimple to check if it is a PODIO source. Look here for how to get at the types/names contained in it: [JEventSourcePODIOsimple:: PrintCollectionTypeTable](https://github.com/eic/EICrecon/blob/dd2f4f8d3a2abbc98f7cf72101177214bb602538/src/services/io/podio/JEventSourcePODIOsimple.cc#L223)

For item 5. it will be worth talking to Nathan. What we should really do is have a virtual method in JEventSource to get this information. Then, you would not need to know anything about JEventSourcePODIOsimple. It could then also work with other sources. However, that will require a new tagged version of JANA.

nathanwbrei:
Why not print _all_ of the possible podio types, regardless whether they are present in the file? We have that information in make_datamodel_glue.py, we'd just need to write some python to propagate it to datamodel_glue.h.

nathanwbrei:
Although, I do agree that it would be very nice if we had a generic way of reporting all of the types that are output from the JEventSource. However, it is very hard to come by this information automatically because we have to account for  JEvent::Insert (from either JEventSource::GetEvent or JEventSource::Preprocess) as well as JEventSource::GetObjects, and there's no guarantee that any particular event will include all of the information. 

On the other hand, it would be extremely reasonable to manually register the space of all possible outputs, analogously to how we should be manually registering the space of all possible inputs for JFactory and JEventProcessor.

nathanwbrei:
Try something like this: 

1. JApplication::Initialize().  // This will turn the resource string (e.g. input filename) into some JEventSource object
2. JComponentManager::get_evt_srces(). // This will give you pointers to each event source
3. Use dynamic_cast to check whether your event source is a PODIO one
4. If it is, you need to open it. You have two options. Either
    A. Set parameter jana::nevents=1 and then call JApplication::Run().
    B. Call JEventSource::DoInitialize().     // I recommend this option
5. Now that your event source is initialized, you can ask it for the PODIO collections it claims it can provide.

nathanwbrei:
To turn off the jana messages, set the parameter log:off=JApplication,JArrowProcessingController,JComponentManager

nathanwbrei:
Remember that you have to do this _before_ you create the JApplication object

cissieAB:
> Thanks for putting this in Cissie. A couple of notes:
> 
> 1. It is preferred for you to submit PRs from a branch within the eic/EICrecon repository rather than from a fork. (I can give you the reasons offline)

I am not allowed to push to eic directly. I guess my permission is restricted.

> 2. There should be a short version for the --list_factories option (use "-L" since "-l" is already in use).

Commit https://github.com/eic/EICrecon/pull/94/commits/95087d10b44e7cd57a1cd13b98162d2fac8064ca

> 3. Is it possible to suppress printing of the table with all of the workers/cores listed? This fills a lot of space and is distracting when running in the list_factories mode where the information being sought does not include this info.

The info lines come from initializing the Japplication. It can be turned off by tunning the log level to OFF manually before creating the JApplication, as mentioned by Nathan's comment above.

I am working on this for option [ListFactories](https://github.com/cissieAB/EICrecon/blob/cissie_cli_options/src/utilities/eicrecon/eicrecon_cli.h#L20), but I guess you may require it for other options as well?


> 4. Have the factory table print such that the columns are aligned. It is much easier to read and eventually, this list is going to be quite long.

Commit https://github.com/eic/EICrecon/pull/94/commits/e7cd33d120efd24136a0decf9d94b61db1c7b9b5

> 5. By default, the types from the input file (if given) should be included. The user, will generally be interested in the full list of available types and not just those generated by factory algorithms. This will be a little trickier, since it can only be done if the user supplies a filename AND if that happens to be a podio file. Ideally, JANA will decide on whether there is a source to open then the right JEventSource subclass to instantiate. Then you can try doing something like downcasting to JEventSourcePODIOsimple to check if it is a PODIO source. Look here for how to get at the types/names contained in it: [JEventSourcePODIOsimple:: PrintCollectionTypeTable](https://github.com/eic/EICrecon/blob/dd2f4f8d3a2abbc98f7cf72101177214bb602538/src/services/io/podio/JEventSourcePODIOsimple.cc#L223)
> 
> For item 5. it will be worth talking to Nathan. What we should really do is have a virtual method in JEventSource to get this information. Then, you would not need to know anything about JEventSourcePODIOsimple. It could then also work with other sources. However, that will require a new tagged version of JANA.

WIP.

cissieAB:
> 2. JComponentManager::get_evt_srces(). // This will give you pointers to each event source

@nathanwbrei I am stuck with this step. I do not find the way to access the JComponetManager directly. It seems [JApplication::m_component_manager](https://github.com/cissieAB/JANA2/blob/master/src/libraries/JANA/JApplication.h#L136) is a private member and member function [JApplication::GetService](https://github.com/cissieAB/JANA2/blob/master/src/libraries/JANA/JApplication.h#L122) is not defined. Any clues on that? Thanks!

Issue #83, opened 2022-09-06
Refactor algorithms library and make a space for global algorithms


DraTeots:
**A problem**

Currently Tracking in algorithms includes both independent algorithms and Jana factory (generic but framework dependent). It was a temporary move while splitting algorithms to framework independent and framework dependent parts and it is time to clean this up. 

**Proposed solution**

- Change CMake to have a single `algorithms` library with both static and shared entities. 
- For now create `srd/global` directory where things like global tracking/PID/etc framework dependent factories will live (any better naming and directory structure suggestions?)
- Move tracking parts correctly between folders (see #81)
- Synchronize with current other ongoing developments

faustus123:
So if I understand this correctly, the idea is to have:

src/algorithms  - generic, non-JANA algorithm classes
src/detectors  - detector specific JANA algorithm classes (mostly wrappers for things in src/algorithms)
src/global  - JANA algorithm classes for global reconstruction like tracking, PID, etc. (also with wrappers for src/algorithms)

It is possible that the src/algorithms directory could be replaced by the [eic/algorithms repository](https://github.com/eic/algorithms)  or by  something similar from key4hep in the future.

Is this a correct description?

DraTeots:
Yes, eventually I also see it like this.  Certainly there should be a community defined API for switch which is not existent at this point. But it is a change exactly in this direction.

DraTeots:
To remove naming collision I renamed algorithms libraries targets as: 

- algorithms_tracking_library
- algorithms_calorimetry_library
- algorithms_digi_library

PR #96, opened 2022-09-08
Refactor algorithms library to be framework independent

DraTeots:
Full description is in #83


faustus123:
This includes updates to the documentation for edm4eic which has not been integrated yet. Even though that is inevitable, it might be better to include those updates in that specific PR.

DraTeots:
Ah, that is a mistake. I actually converted, then converted back and probably left the documentation. I'll fix that. Reviews are good!

faustus123:
I just pushed a small change to fix the plugin list in eicrecon to reflect the new plugin names and additional plugins this  brings in. With that change, things compile and run just fine with the EEMCcheck plugin for calorimetry.

PR #99, opened 2022-09-09
Add cli option "-L"/"--list_factories"

cissieAB:
- Add "-L"/"--list_factories" as a cli option. Print the factories given by [JComponentSummary](https://github.com/cissieAB/JANA2/blob/master/src/libraries/JANA/JApplication.h#L98). Use [JTablePrinter](https://github.com/cissieAB/JANA2/blob/master/src/libraries/JANA/Utils/JTablePrinter.h) to do the work.
- Turn off the irrelevant logger information given this option.
- Keep "-Ppodio:print_type_table=1" working with this option

- [ ] Bug fix (issue #__)
- [x] New feature (issue #61 _)
- [ ] Documentation update
- [ ] Other: __

Unclear. It better to split the jana::Flag into two categories: run and not run app. Otherwise when the option number increases, the logic will become too complicated to manage.
Eg. the behavior of "-b -L"

If this option is given, the application will not run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ src/services/io/podio/datamodel_glue.h
 # Indirectory builds
 include/*
 plugins/*
+build/*
 
 ##### C++ basic gitignore #####
 
--- a/src/utilities/eicrecon/eicrecon.cc
+++ b/src/utilities/eicrecon/eicrecon.cc
@@ -7,7 +7,7 @@
 
 #include <TFile.h>
 
-#include <JANA/CLI/JMain.h>
+#include "eicrecon_cli.h"
 
 int main( int narg, char **argv)
 {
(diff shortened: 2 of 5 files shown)


cissieAB:
cont. of #94 


- [x] Add "-L"/"--list_factories" as a cli option. Print the factories given by [JComponentSummary](https://github.com/cissieAB/JANA2/blob/master/src/libraries/JANA/JApplication.h#L98). Use [JTablePrinter](https://github.com/cissieAB/JANA2/blob/master/src/libraries/JANA/Utils/JTablePrinter.h) to do the work.
- [ ] Turn off the irrelevant logger information given this option.
- [ ] Keep "-Ppodio:print_type_table=1" working.

Current status: stuck at getting JComponetManager (https://github.com/eic/EICrecon/pull/94#issuecomment-1241421836)

PR #100, opened 2022-09-09
Add ability to merge background events in with the primary.

faustus123:
This allows the user to specify a second podio file containing background events that are to be merged in with the primary events at the event source. User may also specify how many background events to merge in.

This currently merges ALL objects in the background event, including generated particles. 

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No. One needs to specify the podio:background_filename to activate the new feature.

### DIFF
--- a/src/services/io/podio/JEventSourcePODIOsimple.cc
+++ b/src/services/io/podio/JEventSourcePODIOsimple.cc
@@ -83,11 +83,18 @@ JEventSourcePODIOsimple::~JEventSourcePODIOsimple() {
 void JEventSourcePODIOsimple::Open() {
 
     // Allow user to specify to recycle events forever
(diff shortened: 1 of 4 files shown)


faustus123:
Don't look at this just yet. I realized there is a memory leak and it needs to have an offset for the event numbers read for each of the background readers.

nathanwbrei:
Looks good. 

I'm a little bit confused why we are loading the same file `num_background_events` times, but hey.

One issue I can imagine popping up is that object relations could get mangled if someone attempts to create a reference between podio objects that span two different `podio::EventStore`s (particularly if they contain otherwise identical collections). However, I don't see any reason for people to do that in practice, and I think the correct response is to replace `podio::EventStore` for a multitude of other reasons.

Issue #53, opened 2022-08-18
Multi-threaded podio


faustus123:

nathanwbrei:
Is the (short-term) goal to make the event source (reader) multi-threaded, or just to make the event writer's critical section independent from the event reader's?

faustus123:
Well, actually first goal needs to be making the writer work again as it is currently broken. The second goal will be to make the reader/writer work in a way that allows multi-threaded event processing in between.

Issue #48, opened 2022-08-12
Implement podio Event writer (single threaded)


faustus123:

faustus123:
PR#49 implements this, but is limited. It will write an output file, applying a filter to collections that came from the input file. Still needs to add new collections from reconstructed objects.

faustus123:
This was addressed in PR#135.

PR #135, opened 2022-09-16
PODIO persistence

nathanwbrei:
This PR introduces a JEventProcessor that persists data from any PODIO-producing JFactory down to a ROOT file. 
Previously, EICrecon could only persist data that had been originally read via a PODIO event source. It does so in a way which allows multithreading.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #48, issue #53)
- [ ] Documentation update
- [ ] Other: __

No

No. 
JEventProcessorPODIO does nothing by default, unless the user specifies an output file. (This output file will now contain more branches.) It is also worth nothing that this PR removes the barrier that prevents multiple events from being in-flight at once. This may uncover unrelated multithreading bugs if the user runs JANA with `nthreads != 0`.


nathanwbrei:
Please don't merge yet! Remaining things to do:
[ ] Get rid of the event barrier
[ ] Use spdlog instead of jout
[ ] Uncomment inclusion/exclusion list logic

Also want to get rid of calls to clone() by making the EventStore a little smarter, but that may be out of scope for this PR.

nathanwbrei:
Turns out we can't get rid of the event barrier just yet because without it we get errors elsewhere, particularly JEventSourcePODIOSimple. So that will go in a future PR instead.

faustus123 (approved):
This compiles and runs. It suffers from the same issue as other attempts that causes it to be unusably slow on writing, but this is an issue with podio and not something we can easily resolve. The changes are worth locking in at this point though.

PR #141, opened 2022-09-20
fix: eicd -> edm4eic

wdconinc:
EICD is defunct and not included in the containers anymore. It is only a matter of time before this fails (unless it's dead code).

@faustus123 The containers without eicd are probably still rolling out to cvmfs servers around the world. I check to make sure you are not depending on this anymore in CMakeLists.txt, but apparently it still has a few occurences in source files, likely included 'by accident' through some other convoluted path.

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


faustus123 (approved):
Good catch.

Issue #87, opened 2022-09-06
-PNThreads!=1 sometimes causes a crash


T-Britton:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? no
- Any special versions of ROOT or Geant4? no

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. run a plugin (maybe using the geometry) with -PNThread!=1
2. do that a few times to trigger a crash (usually a segfault of some kind)

### Expected Result: (what do you expect when you execute the steps above)
it not to crash

### Actual Result: (what do you get when you execute the steps above)
it crashes sometimes

faustus123:
I confirmed that this is an issue. I see two problems:

1. it seems not all JANA threads are honoring jana:warmup_timeout since with multiple threads, I am getting timeout errors after 8 seconds and before 30 seconds. Setting jana:timeout to a larger number fixes this.

2. When timing out, a fatal error is printed and then the program proceeds to exit, but it ends up seg faulting in the end. Actually, there are multiple errors observed if run multiple times suggesting corrupted memory. The seg fault at the end is most often.

This can be reproduced in eic-shell with
~~~
eicrecon -Pnthreads=4 -Pjana_timeout=1 -Pplugins=EEMCcheck 2022-09-04_pgun_e-_podio-0.15_edm4hep-0.6_0-30GeV_alldir_1k.edm4hep.root
~~~

n.b. EEMCcheck can be found here: https://jeffersonlab-my.sharepoint.com/:u:/g/personal/davidl_jlab_org/EbsqzQgxJ5dPsDVr4hph5W4ByHf43wd6P04f6t81qk6OAA

the input file can be found here: https://eicaidata.s3.amazonaws.com/2022-09-04_pgun_e-_podio-0.15_edm4hep-0.6_0-30GeV_alldir_1k.edm4hep.root

faustus123:
The thread timeouts are now increased when running eicrecon. This was introduced in PR#149.

PR #149, opened 2022-09-25
PODIO writer now activates explicitly included factories

nathanwbrei:
Fixes two problems encountered recently:
1. PODIO writer wouldn't call factory->Create() even if user explicitly asked for it via `podio:output_include_collections`
2. eicrecon timed out when nthreads>1 (See issue #87)
 
- [x] Bug fix (issue #87)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

It changes the default thread timeout to 60 seconds.

### DIFF
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -115,6 +115,7 @@ void JEventProcessorPODIO::FindCollectionsToWrite(const std::vector<JFactory*>& 
     }
     else {
         m_log->debug("Persisting podio types from includes list");
+        m_user_included_collections = true;
         // We match up the include list with what is actually present in the JFactorySet
         for (const auto& col : m_output_include_collections) {
             auto it = all_collections_to_types.find(col);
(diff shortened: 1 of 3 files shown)


nathanwbrei:
While I was testing this, I randomly picked a collection:
`-Ppodio:output_include_collections=EcalBarrelClusters`

When the podio writer tried to call `Cluster_factory_EcalBarrelClusters::Process()`, it crashed, saying:

`[2022-09-24 22:37:03.681] [JEventProcessorPODIO] [error] Exception adding PODIO type 'edm4eic::Cluster:EcalBarrelClusters': Could not find JFactoryT<edm4eic::CalorimeterHit> with tag=.`

I dug around a little bit and found that the JFactory call stack ultimately got to 
`TruthCluster_factory_EcalBarrelTruthProtoClusters::Process`, 

which attempted to load some CalorimeterHits using a collection tag that was never initialized:

`m_inputHits = event->Get<edm4eic::CalorimeterHit>(m_inputHit_tag);`

I'm not sure whether this is worth filing an issue over, so let me know. Would you have expected this to work?

faustus123:
This is a known issue. The calorimeter cluster chain is currently broken for this reason. It needs to be fixed, in two ways. the main one is that it needs to generate clusters from hits and not Truth hits.

There is a wider issue that these kind of errors pop up a lot at run time because we are using tags as collection names. There is currently no way to verify that a tag corresponds to the correct typeid either at compile time or program start up. We will need to think how to best handle it, but for now, we will just need to keep working with it the way that it is.

faustus123 (approved):
This looks correct.

PR #164, opened 2022-09-28
Fix bug that causes crashes when trying to write certain collections

faustus123:
This addresses an issue that causes podio to throw an exception when collections that have one to many relations are being written to the output file, but the relation objects themselves are marked as "untracked".

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes. Program will proceed with an error message now, but does not crash.

### DIFF
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -83,7 +83,7 @@ void JEventProcessorPODIO::Init() {
     m_log = app->GetService<Log_service>()->logger("JEventProcessorPODIO");
     m_log->set_level(spdlog::level::debug);
     m_store = new eic::EventStore();
-    m_writer = std::make_shared<eic::ROOTWriter>(m_output_file, m_store);
+    m_writer = std::make_shared<eic::ROOTWriter>(m_output_file, m_store, m_log);
 
 }
 
--- a/src/services/io/podio/RootWriter.cc
+++ b/src/services/io/podio/RootWriter.cc
@@ -14,7 +14,7 @@
 
 namespace eic {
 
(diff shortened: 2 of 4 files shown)


nathanwbrei:
Looks reasonable. I assume this PR means that if you specify -Ppodio_output_collections= EcalEndcapNIslandProtoClusters, it will log an error message and write an empty output file, but if you specify
-Ppodio_output_collections= EcalEndcapNIslandProtoClusters,EcalEndcapNRawHits, it will correctly write both?

Another (very debatable) approach would be to clear all the references and try to write it again without them. 

Obviously not for this week, but in the future, if the user specifies -Ppodio_output_collections= EcalEndcapNIslandProtoClusters, should we try to figure out that this depends on some CalorimeterHit collections, 
identify those, and write those out as well? I don't think that PODIO constrains references to all belong to a single collection though...

faustus123:
Actually, if you specify `-Ppodio_output_collections= EcalEndcapNIslandProtoClusters,EcalEndcapNRawHits` it should only write out the `EcalEndcapNRawHits`. This is because we are creating the `edm4eic::RawCalorimeterHit` objects with "new" rather than as part of a collection. Thus, these are "untracked". That seems to be what podio has an objection to. 

The only way to make it work is to use a collection to generate the objects (you may even need to use an EventStore to create the collection. I'm not actually sure). This goes back to the discussion at today's Developer's meeting about having all JANA factories work directly with the EventStore. It's either that, or have the podio authors fix podio, or replace the podio RootWriter and RootReader with something that can deal directly with objects rather than through collections. As you know, we tried that last one, but they have structured the podio classes with private and friend classes so as to prevent doing anything like that without going through collections. (sigh)

DraTeots:
Please merge then! I have to write updates to Shayam

PR #167, opened 2022-09-29
Fix bugs in SiliconTrackerDigi

faustus123:
This fixes two bugs in the SiliconTrackerDigi algorithm.

1. The value for charge was being written into the timestamp and vice versa
2. The time units for the central value were changed from fs to ps. This also fixes an issue with the Gaussian smearing being in ps while the central value was in fs.

The digitized units should be revisited since 1fs LSB seems way too good. Especially for a devise that expects to have ~8ns resolution.

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No


DraTeots:
How tracking worked at all?! (I know how, but still a good catch). Hopefully we will have time at some point to work more towards unit testing - it is good in terms of catching such things on the spot.

PR #168, opened 2022-09-29
Roman Pots

faustus123:
This adds digitized roman pot hits as well as a port of the juggler algorithm.

The ForwardRomanPotParticles are created, but not many and the algorithm may need some work to be really useful.

These particles are not added to the global list of ReconstructedParticles 

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [X] Tests for the changes have been added (in RPOTScheck benchmark plugin)

No

Yes, adds ability to digitize and reconstruct Roman Pot particles


ajentsch (approved):
It looks like the basic functionality should be okay, but I will need to change this to be able to handle "local hits", rather than global ones. As it stand, I put in some faux offset values to shift from global to local, but this was always not a good solution. Using the geometry service to get local coordinates is pretty hefty for something which really should be the default.

I will work on the changes after this is merged into the master branch.

Issue #65, opened 2022-09-02
Make sure file headers are correctly transferred and all credits are saved


DraTeots:

DraTeots:
Will be checking it

wdconinc:
Not my job.

DraTeots:
> Not my job.

Just to be clear. I'm not suggesting you to sit and fix headers but rather make sure in the end, that everything is correct and e.g. credits to your inputs are labelled/transferred correctly.

DraTeots:
Said done in corresponding PR  #183

PR #183, opened 2022-10-05
feat: add LGPL as license

wdconinc:
There is now so much LGPL code copied here that it is simply not a sustainable situation to keep this under anything else than an open source license.

Explicit pproval needed from:
- [x] @faustus123
- [ ] @DraTeots 
- [ ] @nathanwbrei
- [ ] @T-Britton 
- [ ] @cissieAB 
- [ ] @veprbl 
- [x] @wdconinc 

- [ ] Bug fix (issue #__)
- [X] New feature (issue #2)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
I approve switch to LGPL for my contribution to EICrecon

faustus123:
So this may actually be a very complicated issue. Those of us working at national labs and writing software as part of our paid jobs automatically have it placed in the _public domain_. This means it cannot be licensed or copyrighted, which means the LPGL would be legally inapplicable. The codes taken from juggler that have an LPGL license should, in principle, require that the license be honored. However, any of that code written by people like Sylvester or Whit who work for ANL should have been _public domain_ at the time of creation and therefore un-licensable. It is a confusing mess. Have a look at the OSDI FAQ on _public domain_ [here](https://opensource.org/faq#osd) for some details. The relevant paragraph is:

~~~
There are certain circumstances, such as with U.S. government works as described above, where it is not easy
to apply a license, and the software must be released into the public domain. In these cases, while it would be 
inaccurate to display the OSI logo or say that the license is OSI-approved (since there is no license), nevertheless 
we think it is accurate to say that such software is effectively open source, or open source for most practical 
purposes, even though it is not officially released under an open source license. (This is assuming, of course, 
that in the laws of releasing jurisdiction the meaning of "public domain" is compatible with the Open Source 
Definition.) After all, the freedoms guaranteed by open source licenses are still present, and it is possible for 
the familiar dynamics of open source collaboration to arise around the software.
~~~

This means that even the placeholder _LICENSE_ file we have claiming some Copyright restrictions should not (nor ever have been) there.

Let me check with legal to figure out what we are actually allowed to do and what we are required to do.

nathanwbrei:
Let's definitely ask legal. However, I'd bet that the answer lies in multi-licensing. [1] There's nothing stopping us from releasing software under multiple licenses, in this case "LGPL" and "public domain". The end user then gets to choose which license they are using. I don't think there would be a logical contradiction since the logical "OR" of "LGPL" and "public domain" is still "public domain".

[1] https://en.wikibooks.org/wiki/FOSS_Licensing/How_to_Make_the_Source_Free_or_Open#Multiple_Licensing

faustus123:
I'll admit to being confused. Licenses/copyrights are generally used to restrict the use of something. Having one that is completely unrestricted and another with some restrictions means it is always completely unrestricted. I don't think they issue the multi-licenses as an OR that combines the two. Rather It sounds like the main use-case for multi-licenses is for the controlling organization to sell a version of the license that doesn't force the licensee to publish their derived work. You are using one or the other.

IMHO we should just put a public domain declaration in the repository. Everyone contributing to any of the EIC software (in any repository) should be made to clearly understand that they are _donating_ it to the public domain. I don't have a problem putting the LPGL into the src/algorithms directory for now since that should be 99% of the code copied from juggler.

I'm in communications with some higher-ups at JLab to find out what our options are.

wdconinc:
If you copy LGPL code into a project, you must pick one (or more) license(s) that is(are) compatible with that practice. Dual licensing a project that includes LGPL code requires that any license you allow use under must be compatible with the LGPL, and that means you cannot just pick public domain as one of the licenses and be done with it since releasing LGPL code in a public domain project is counter to the LGPL licensing conditions.

You can have different licenses for different parts of the code base. In that case it must be clear which part of the code base is provided under which license. So, here you could do something like everything under `algorithms` is LGPL, but everything elsewhere is public domain (assuming there was no copying of LGPL code in those parts).

You can also just not copy LGPL into your code base, keep the LGPL code in a separate LGPL-licensed library, and release everything that you wrote into the public domain. That is exactly what the LGPL was designed to allow.

sly2j:
At Argonne we are explicitly allowed to use common open source licenses such as GPL, LGPL (and of course the more permissive ones). You can see another example of LGPL code here on the official Argonne Github repository https://github.com/Argonne-National-Laboratory/deer/blob/master/LICENSE

We chose LGPL for Juggler and our other projects, and hence what is happening here is a blatant copyright violation. In light of not holding up the production I decided not to make a big deal out of this yet. But we have to address this going forward. There are two paths forward:

1. License EICRecon under LGPL. This is required if you want to keep the current code in the repo. I think licensing of this repo should be a collaboration decision, not just a decision of the EICRecon team
2. `algorithms` will fix this. Using the external `algorithms` library, which is an offshoot of Juggler and licensed under LGPL past the first production will give us more freedom for EICRecon itself.

I would be in favor of just going for solution (2), and then having an open discussion on software licenses to decide what we want to do as collaboration.

faustus123:
To make sure we are clear, this not "a decision of the EICrecon team". This is a question of whether we can even legally use the LGPL.  Regardless of ANL's policy, they may not have had a legal ability to place their software under LPGL in the first place. It is clearly something that needs to be discussed in the broader software group, but we need to know what, if any limitations we have.

I've already put in a PR to this PR's branch that moves the LPGL to the algorithms directory. It also places a placeholder at the top of the repository declaring public domain (which I have good reason to believe will be our only real option going forward).

faustus123:
So I was just looking over some of the code in the algorithms directory and realize that many of those files have lost the header comments that declared the license the original file was developed under and the list of authors. See for example these two. The first has the lines, but the second does not.

[src/algorithms/digi/SmearedFarForwardParticles.h]()
[src/algorithms/calorimetry/CalorimeterHitDigi.h]()

This is clearly not right and apologies are owed to the original authors. I think I understand better why you were upset. All I can say is that this was done inadvertently while working in haste to meet a deadline. I will put in an issue so this can get rectified soon.

DraTeots:
Yes. No-one intentionally trying to erase anyone credits. If this happens, it is only because of the to many things have to be done fast. We have an issue to fix that. I haven't gone through all algorithms checking that yet, wanted to wait for the moment when things are settled after the campaign start. 

https://github.com/eic/EICrecon/issues/65

wdconinc:
> No-one intentionally trying to erase anyone credits.

They just disappeared on their own?

T-Britton:
As per the tutorial jana-generate factory produces the correct jana form. It has no knowledge of the underlying comments. In haste while testing jana-generate and trying to port the volume of algorithms in short order I can see top level comments being unintentionally erased. I am not sure that the porter is even flagged….

I’ll go through the calo algorithms this morning (in about an hour) and restore/ensure the top level comment blocks are there with a sincere apology for any missing ones.

Thomas Britton

On Oct 6, 2022, at 12:12 AM, Wouter Deconinck ***@***.***> wrote:

﻿

No-one intentionally trying to erase anyone credits.

They just disappeared on their own?

faustus123:
Individual files have had their declarations and author credits restored. (This was done a while back in main and the changes merged into this branch.) 

The LPGL file at the top was moved into the src/algorithms directory since many files in this repository were not developed under that but most of the files in that directory were. A NOTICE.md file was added to the topmost directory to declare code submitted to EICrecon as being under the public domain unless otherwise specified. It explains that some areas of the repository were developed under different licenses and carry the applicable license in that directory or one of the parent directories.

Since this is a change from the original PR I will uncheck the boxes at the top where approval was already given so those people will need to re-approve these changes.

wdconinc:
GitHub doesn't let me approve since this is my own pull request, but I approve of @faustus123's proposal and have checked my box.

wdconinc:
> those people will need to re-approve these changes.

And so much for that...

T-Britton:
My apologies if I misread things as:
Wouter started the PR (not a draft PR)
David made edits
You approve of those edits to a non draft PR but can’t do the merge because GitHub. This pr passes the tests. 

If we need to revert say the word and give me 30seconds

T-Britton:
yeah my apologies.  Rushing a little too much reverting now

T-Britton:
And a recommendation:
When you reissue the PR mark it as draft. (Upper right button). Then when everyone checks off their box you can flip the switch to ready. That way no one can merge in a rush 😉

DraTeots:
I am confused? Is it open, closed, dead, alive or undead?

T-Britton:
I was performing necromancy. But David put a stop to my black magic.

Thomas Britton

On Oct 19, 2022, at 9:17 PM, Dmitry Romanov ***@***.***> wrote:

﻿

I am confused? Is it open, closed, dead, alive or undead?

PR #246, opened 2022-10-19
BEMC (sciglass) parameter update

wdconinc:


wdconinc:
@veprbl I can't approve, but you can :-D

Issue #230, opened 2022-10-17
Missing plugin dependencies


veprbl:
EICrecon does not specify dependencies between the libraries correctly.

Building on macOS leads to linker errors due to undefined symbols, as ld64 doesn't allow that by default.

This can be reproduced with binutils on Linux `LDFLAGS=-Wl,--no-undefined`. Here is a build log:
https://gist.github.com/veprbl/c1512e87438c43449923803aa17d8d44

Runtime error on Linux:
```
[ERROR] JPluginLoader: Couldn't load plugin 'acts.so'
  Make sure that JANA_HOME and/or JANA_PLUGIN_PATH environment variables are set correctly.
  Paths checked:
    .../lib/EICrecon/plugins/dd4hep.so  =>  Loaded successfully
    .../lib/EICrecon/plugins/acts.so  =>  Loading failure: .../lib/EICrecon/plugins/acts.so: undefined symbol: _ZN11Log_service6loggerERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
    ./acts.so  =>  File not found
    .../plugins/acts.so  =>  File not found

[FATAL] Exception: Couldn't find plugin 'acts.so'
```

DraTeots:
Confirm. Fixing

veprbl:
Oh, and there will be ~~dragons~~cyclical dependencies too.

nathanwbrei:
@DraTeots Keep me in the loop on this one

DraTeots:
I believe today in the morning I was young and naive thinking that I quickly fix all the issues. 

The PR does a lot of cleanup and speedup for CMake that was in plans for a while. It should also ~kill dragons~ remove some cyclic dependencies. 

But in the end of the day, when I became old and wise, I'm not sure it is possible to compile everything with `--no-undefined` because of the dynamic idea of plugins, that are dlopen-ed depending on user parameters. So...

1. Please try the changes by this PR
2. Maybe @faustus123 could help with MAC. Unfortunately I don't have ability to try and experience here. 
3. Could we point out ~where live the dragons~ what cyclic dependencies we talk about . Because I know at least one place where it exist and is difficult to exterminate for stupid historical reasons.

faustus123:
[...]

PR #252, opened 2022-10-20
Add explicit dependencies on libraries

veprbl:
This intended to address #230 in the direct way by resolving undefined symbols with explicit linking. This is to be done until or if linking with `--no-undefined` succeeds.

I understand, that this does not align with the aim of having replaceable services/algorithms, at least not in in an obvious way. However, it does improve on the status quo, where some dependencies are defined but not in a systematic way.

### DIFF
--- a/src/algorithms/tracking/CMakeLists.txt
+++ b/src/algorithms/tracking/CMakeLists.txt
@@ -31,7 +31,7 @@ plugin_glob_all(${PLUGIN_NAME})
 
 # Add libraries
 # (same as target_include_directories but for both plugin and library)
-plugin_link_libraries(${PLUGIN_NAME} Eigen3::Eigen)
+plugin_link_libraries(${PLUGIN_NAME} Eigen3::Eigen ROOT::EG)
 
 #
 ## Find dependencies
--- a/src/benchmarks/reconstruction/tracking_efficiency/CMakeLists.txt
+++ b/src/benchmarks/reconstruction/tracking_efficiency/CMakeLists.txt
@@ -31,4 +31,4 @@ plugin_include_directories(${PLUGIN_NAME}
 
 # Add libraries
 # (same as target_include_directories but for both plugin and library)
-plugin_link_libraries(${PLUGIN_NAME} ${ROOT_LIBRARIES} ActsCore EDM4EIC::edm4eic)
+plugin_link_libraries(${PLUGIN_NAME} ${ROOT_LIBRARIES} ActsCore log_library)
--- a/src/services/io/podio/CMakeLists.txt
+++ b/src/services/io/podio/CMakeLists.txt
@@ -8,18 +8,14 @@ print_header(">>>>   P L U G I N :   ${PLUGIN_NAME}    <<<<")       # Fancy prin
 project(podio_project)
 
 # Find dependencies (JANA and spdlog automatically added already)
-find_package(EDM4HEP REQUIRED)
 find_package(EDM4EIC REQUIRED)
 find_package(podio REQUIRED)
 find_package(ROOT REQUIRED COMPONENTS Core Tree Hist RIO EG)
 find_package(fmt REQUIRED)
-if(NOT DD4hep_FOUND)
-    find_package(DD4hep REQUIRED)
-endif()
 
 set(EDM4EIC_INCLUDE_DIR ${EDM4EIC_DIR}/../../include)
 set(fmt_INCLUDE_DIR ${fmt_DIR}/../../../include)
-set(INCLUDE_DIRS ${podio_INCLUDE_DIR} ${EDM4HEP_INCLUDE_DIR} ${DD4hep_INCLUDE_DIRS} ${ROOT_INCLUDE_DIRS} ${EDM4EIC_INCLUDE_DIR} ${fmt_INCLUDE_DIR})
+set(INCLUDE_DIRS ${fmt_INCLUDE_DIR})
 
 # Function creates ${PLUGIN_NAME}_plugin and ${PLUGIN_NAME}_library targets
 # Setting default includes, libraries and installation paths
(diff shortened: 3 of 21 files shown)


src/algorithms/tracking/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ plugin_glob_all(${PLUGIN_NAME})
 
 # Add libraries
 # (same as target_include_directories but for both plugin and library)
-plugin_link_libraries(${PLUGIN_NAME} Eigen3::Eigen)
+plugin_link_libraries(${PLUGIN_NAME} Eigen3::Eigen ROOT::EG)
```
DraTeots, line 34: I think ROOT::EG should be added to `plugin_add_cern_root`

https://github.com/eic/EICrecon/blob/main/cmake/jana_plugin.cmake#L219

There are several reasons for this: 

1. ROOG::EG is used for TDatabasePDG, which in turn is used to get charge, and invocations are scattered around many places in code. Moreover, the more algorithms will be added, the more TDatabasePDG is going to be used. So adding ROOG::EG to "global" cmake macro seems reasonable. 

2. You see commented lines there - ROOT::EG is actually a regression to previous problems. And the problems are that some compilations/versions of root have it by default in `${ROOT_LIBRARIES}` while others - don't. And I hoped not to list particular packages.

src/benchmarks/reconstruction/tracking_efficiency/CMakeLists.txt
```diff
@@ -31,4 +31,4 @@ plugin_include_directories(${PLUGIN_NAME}
```
DraTeots, line 31: Ok, instead of putting it to each factory, lets put it in global plugin_add macro then

Somewhere around

https://github.com/eic/EICrecon/blob/main/cmake/jana_plugin.cmake#L31

Shouldn't be too bad.

src/services/io/podio/CMakeLists.txt
```diff
@@ -8,18 +8,14 @@ print_header(">>>>   P L U G I N :   ${PLUGIN_NAME}    <<<<")       # Fancy prin
 project(podio_project)
 
 # Find dependencies (JANA and spdlog automatically added already)
-find_package(EDM4HEP REQUIRED)
 find_package(EDM4EIC REQUIRED)
 find_package(podio REQUIRED)
 find_package(ROOT REQUIRED COMPONENTS Core Tree Hist RIO EG)
 find_package(fmt REQUIRED)
```
DraTeots, line 14: I didn't want to touch it previously because it is the place Nathan is working on right now and it is high probably that one will need to clean it up twice. But since it is started, lets then cleanup this file:

- fmt - is not needed (we use spdlog, that embeds fmt)
- find_package(EDM4EIC REQUIRED) and find_package(podio REQUIRED) are not needed as plugin_add_event_model adds them
- find_package(ROOT... should be replaced with plugin_add_cern_root

src/services/io/podio/CMakeLists.txt
```diff
@@ -8,18 +8,14 @@ print_header(">>>>   P L U G I N :   ${PLUGIN_NAME}    <<<<")       # Fancy prin
 project(podio_project)
```
DraTeots, line 8: This should be removed as plugin_add adds a project

src/services/io/podio/CMakeLists.txt
```diff
@@ -8,18 +8,14 @@ print_header(">>>>   P L U G I N :   ${PLUGIN_NAME}    <<<<")       # Fancy prin
...
 find_package(ROOT REQUIRED COMPONENTS Core Tree Hist RIO EG)
 find_package(fmt REQUIRED)
-if(NOT DD4hep_FOUND)
-    find_package(DD4hep REQUIRED)
-endif()
 
 set(EDM4EIC_INCLUDE_DIR ${EDM4EIC_DIR}/../../include)
 set(fmt_INCLUDE_DIR ${fmt_DIR}/../../../include)
-set(INCLUDE_DIRS ${podio_INCLUDE_DIR} ${EDM4HEP_INCLUDE_DIR} ${DD4hep_INCLUDE_DIRS} ${ROOT_INCLUDE_DIRS} ${EDM4EIC_INCLUDE_DIR} ${fmt_INCLUDE_DIR})
+set(INCLUDE_DIRS ${fmt_INCLUDE_DIR})
```
DraTeots, line 18: Those sets are not needed as decribed above

src/services/rootfile/CMakeLists.txt
```diff
@@ -17,3 +17,6 @@ plugin_glob_all(${PLUGIN_NAME})
 # Add cern root
 plugin_add_cern_root(${PLUGIN_NAME})
 
+# Add libraries
+# (same as target_include_directories but for both plugin and library)
+plugin_link_libraries(${PLUGIN_NAME} log_library)
```
DraTeots, line 22: Same as with other log_libraries

DraTeots (changes requested):
Thank you for starting this! 

- We can't add `--no-undefined` globally but I thought about adding `--no-undefined` at least to some CMake projects that don't rely on others. In particular: 
   - all projects in `algorithms` directory
   - log service
   - dd4hep service
   - podio service? 
   could you do these changes? 
  
- IDK if you have seen [this documentation on CMake](https://eic.github.io/EICrecon/#/design/cmake). It might help to understand, what is going on

If you would like, we can have a quick phone/zoom/whatever call

src/algorithms/tracking/CMakeLists.txt
DraTeots: So what I mean here, is if you think reasonable, please switch back to required components in plugin_add_cern_root

veprbl:
The `--no-undefined` is the way to go for now

PR #260, opened 2022-10-21
Add hits merging and update digitization for barrel HCAL

johnlajoie:
This PR adjusts the Barrel HCAL hits merging, digitization, and reconstruction to match what is expected for the re-use of the sPHENIX barrel HCAL 

- [ ] Bug fix (issue #__)
- [ X ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ X ] Tests for the changes have been added

Because the hits output is different for the ATHENA HCAL and the sPHENIX HCAL, this will break the merging for the ATHENA HCAL geometry.  That shouldn't be a showstopper since I don't think we plan to carry that forward.

See above it is adapted to the hits output format for the sPHENIX HCAL.


faustus123:
This all looks reasonable.

PR #285, opened 2022-10-26
Davidl warnings errors

faustus123:
This cleans up the output of eicrecon a bit to make it easy to see real problems. This includes:

- This turns a lot of the non-fatal error messages into warning messages.
- Check that XML geometry file exists and quit with informative error if not.
- Make printing of summary all config. parameters at end of job optional with default not to print.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: __

No

Yes, but mainly cosmetic.

### DIFF
--- a/src/services/geometry/dd4hep/JDD4hep_service.cc
+++ b/src/services/geometry/dd4hep/JDD4hep_service.cc
@@ -8,6 +8,8 @@
 #include <vector>
 #include <sstream>
 #include <algorithm>
+#include <filesystem>
+#include <fmt/color.h>
 
 #include "JDD4hep_service.h"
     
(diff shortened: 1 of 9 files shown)


src/services/geometry/dd4hep/JDD4hep_service.cc
```diff
@@ -105,6 +107,19 @@ void JDD4hep_service::Initialize() {
...
+            // Check that this XML file actually exists.
+            if( ! std::filesystem::exists(filename) ){
+                // Here we go against the standard practice of throwing an error and print
+                // the message and exit immediately. This is because we want the last message
+                // on the screen to be that this file doesn't exist.
+                auto mess = fmt::format(fmt::emphasis::bold | fg(fmt::color::red),"ERROR: ");
+                mess += fmt::format(fmt::emphasis::bold, "file: {} does not exist!",  filename);
+                mess += "\nCheck that your DETECTOR and DETECTOR_CONFIG environment variables are set correctly.";
+                std::cerr << std::endl << std::endl << mess << std::endl << std::endl;
+                _exit(-1);
```
DraTeots, line 120: Could you actually make it compatible with this functionality

https://github.com/eic/EICrecon/blob/ff19490a6bdd4c95a7d0cb9e6897e532e69905a8/src/services/geometry/dd4hep/JDD4hep_service.cc#L127

Described here:

https://eic.github.io/EICrecon/#/design/common_flags_env?id=flags

DraTeots (changes requested):

src/services/geometry/dd4hep/JDD4hep_service.cc
DraTeots: Because it is very convenient to give just a file name, assuming DETECTOR_PATH is given. 

Like 

```
eicrecon ... -Pdd4hep:xml_files=epic.xml

# now try tracking only
eicrecon ... -Pdd4hep:xml_files=epic_tracking_only.xml

# or switching between major geometry variants
```

DraTeots: If you are confused now, I can fix that (file I mean)

faustus123: I guess I am confused. The values in `m_xml_files` should already have the DETECTOR_PATH prepended (see line 79). Thus, it looks like the second try in the catch clause is just prepending it a second time. Or am I missing something?

Issue #279, opened 2022-10-25
EICrecon reports incorrect version


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): 0.2.6
- Which version (or `HEAD` for the most recent on git): 0.2.6
- Any specific OS or system where the issue occurs? eic-shell
- Any special versions of ROOT or Geant4? no

### Steps to reproduce: (give a step by step account of how to trigger the bug)
```console
jug_xl> wdconinc@menelaos:~$ eicrecon -v
      EICrecon version: 0.2.3
jug_xl> wdconinc@menelaos:~$ eic-info |grep eicrecon
 - 7scprls eicrecon@0.2.6~ipo build_type=Release
```

### Expected Result: (what do you expect when you execute the steps above)
Version reported by `eicrecon -v` should be actual version.

### Actual Result: (what do you get when you execute the steps above)
Version disagrees. This may lead to confusion when debugging with users ("I'm running eicrecon 0.2.3", "you probably aren't").

PR #287, opened 2022-10-27
eicrecon version is set by root CMake project version

DraTeots:

The result of  `eicrecon --version` is set from root CMakeLists.txt

- [x] Bug fix (Fixes https://github.com/eic/EICrecon/issues/279)

The version is now connected to CMake global project version

The version is now connected to CMake global project version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@
 
 cmake_minimum_required(VERSION 3.16)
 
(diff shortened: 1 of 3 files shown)


faustus123:
This is fine, but doesn't it just move the problem to the CMakeLists.txt file which still needs to be updated manually? It would be nice if there were some way to pull this from the git branch name and possibly include a commit hash.

Issue #275, opened 2022-10-24
No factory for EcalEndcapPTruthClusters output tag


jizhongling:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? No
- Any special versions of ROOT or Geant4? No

The JEventProcessorPODIO requests EcalEndcapPTruthClusters at https://github.com/eic/EICrecon/blob/main/src/services/io/podio/JEventProcessorPODIO.cc#L92 . However, no factory will produce it. Instead, both ProtoCluster_factory_EcalEndcapPTruthProtoClusters and TruthCluster_factory_EcalEndcapPTruthProtoClusters will produce EcalEndcapPTruthProtoClusters. We should let ProtoCluster_factory_EcalEndcapPTruthProtoClusters produce EcalEndcapPTruthClusters and remove TruthCluster_factory_EcalEndcapPTruthProtoClusters.

PR #291, opened 2022-10-27
Combined bug fixing for EEMC and corresponding algorithms

jizhongling:
This is a combined bug fixing for EEMC and corresponding algorithms.

- [X] Bug fix (issue #275 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No breaking changes.

No.

This is a combined bug fixing for EEMC and corresponding algorithms.

(1) Corrected the scaleResponse of raw hits in digitization for pECal and pECal insert in reco_flags.py.

(2) The digitization in signal_sum_digi was fixed similar to that in single_hits_digi in CalorimeterHitDigi.cc.

(3) Added factories to produce truth clusters from proto-truth clusters.

(4) Corrected the proto-cluster tag-name for pECal insert in reco_flags.py.

(5) The nested clusters in CalorimeterClusterMerger.cc, which prevented merged clusters to be written out,
      was temporally commented out. This is a global issue with podio reference.

(6) Added dummy factories to let cluster associations written out.

(7) Renamed all tag-names that end with "ClusterAssociations" to "ClustersAssociations".
     This makes the naming uniform by just adding "Associations" without removing "s" in "Clusters".

(8) Fixed some typos of filenames for the pECal insert.

The code has been tested by 100 events.

(1) It did not produce any error related to EEMC.

(2) The energy responses of truth, reco, and merged clusters were reasonable.

(3) The cluster associations were successfully written out.


src/detectors/EEMC/Cluster_factory_EcalEndcapNClusters.h
```diff
@@ -15,6 +15,19 @@
 
 
 
+// Dummy factory for JFactoryGeneratorT
+class Association_factory_EcalEndcapNClustersAssociations : public JFactoryT<edm4eic::MCRecoClusterParticleAssociation> {
```
faustus123, line 19: For future reference:

1. We try to use a naming convention of <type_produced>_factory_<tag> so this class should probably be:
`MCRecoClusterParticleAssociation_factory_EcalEndcapNClusterAssociations`
(and yes, this is quite long!)

2. The `JFactoryGeneratorT` class actually takes a tag as the argument so you don't really need to define this class. You could do it in EEMC.cc as:

`app->Add( new JFactoryGeneratorT<JFactoryT<edm4eic::MCRecoClusterParticleAssociation>>("EcalEndcapNClustersAssociations"));`

I forgot to put the tag argument in the Mattermost message so sorry about that.

As a I say, this is for future reference. What you've done will work so I don't see a need to change it now.

jizhongling: Thanks for mentioning the tag name in `JFactoryGeneratorT`. I think I can change it to make the code looks more clear.

PR #295, opened 2022-10-28
reco_flags.py now accepts -P parameter flags

DraTeots:

Now one can pass -P flags to reco_flags.py or run_eicrecon_reco_flags.py

```bash
python3 reco_flags.py -Pjana:timeout=0 input.edm4hep.root output_base_name
```

- [x] New feature (issue #__)

NO

NO

### DIFF
--- a/src/utilities/dump_flags/DumpFlags_processor.h
+++ b/src/utilities/dump_flags/DumpFlags_processor.h
@@ -54,7 +54,7 @@ private:
     std::string m_json_file_name = "";
 
     /// Print parameter summary to screen at end of job
-    bool m_print_to_screen = false;
+    bool m_print_to_screen = true;
 
     /// Print only reconstruction flags
     bool m_only_reco = true;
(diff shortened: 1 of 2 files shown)


src/utilities/dump_flags/DumpFlags_processor.h
```diff
@@ -54,7 +54,7 @@ class DumpFlags_processor: public JEventProcessor, public eicrecon::SpdlogMixin<
     std::string m_json_file_name = "";
 
     /// Print parameter summary to screen at end of job
-    bool m_print_to_screen = false;
+    bool m_print_to_screen = true;
```
faustus123, line 57: Ughh. Why turn this back on? I found it highly annoying to have to scroll past pages of parameters when trying to see useful messages for debugging. This is why I set the default to off. Maybe we could dump this to a config file by default instead? If you find it that useful to have it print to the screen we can leave it on. I was just loving it being disabled by default though. Just my 2 cents.

PR #305, opened 2022-11-02
Deadlock Fix

faustus123:

Catch the dd4hep exception that was causing a deadlock.

Limit number of VolumeManager errors due to bad cellID and print useful message.

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Yes.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitReco.cc
@@ -35,6 +35,16 @@ void CalorimeterHitReco::AlgorithmInit(std::shared_ptr<spdlog::logger>& logger) 
         return;
     }
 
+    // First, try and get the IDDescriptor. This will throw an exception if it fails.
+    try{
+        auto id_spec = m_geoSvc->detector()->readout(m_readout).idSpec();
+    }catch(...){
+        m_logger->warn("Failed to get idSpec for {}", m_readout);
+        return;
+    }
+
+    // Get id_spec again, but here it should always succeed.
+    // TODO: This is a bit of a hack so should be cleaned up.
     auto id_spec = m_geoSvc->detector()->readout(m_readout).idSpec();
     try {
         id_dec = id_spec.decoder();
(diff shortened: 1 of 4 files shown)


src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -35,6 +35,16 @@ void CalorimeterHitReco::AlgorithmInit(std::shared_ptr<spdlog::logger>& logger)
         return;
     }
 
+    // First, try and get the IDDescriptor. This will throw an exception if it fails.
+    try{
+        auto id_spec = m_geoSvc->detector()->readout(m_readout).idSpec();
```
wdconinc, line 40: Doesn't this limit scope to the try block? Or at least complain about unused id_spec? And shadowing? How strict are you running, `-Werror`?

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -106,6 +116,16 @@ void CalorimeterHitReco::AlgorithmChangeRun() {
 //------------------------
 void CalorimeterHitReco::AlgorithmProcess() {
 
+    // For some detectors, the cellID in the raw hits may be broken
```
wdconinc, line 119: Wondering if this is after hcal hit merging. In juggler, that used to not leave the resulting hit with a valid cellID, but was modified over the summer to get the cellID of the highest energy hit. All hits should have a valid cellID.

src/detectors/HCAL/CalorimeterHit_factory_HcalBarrelRecHits.h
```diff
@@ -95,6 +95,7 @@ class CalorimeterHit_factory_HcalBarrelRecHits : public JFactoryT<edm4eic::Calor
         hits.clear(); // not really needed, but better to not leave dangling pointers around
     }
 
+    std::shared_ptr<spdlog::logger> m_log;
```
wdconinc, line 98: Why not in base of factory instead of leaving up to each factory to implement?

wdconinc (approved):

PR #307, opened 2022-11-04
Davidl eicrecon fixes

faustus123:

- Clean up the spdlog implementation in Calorimetery by removing a lot of duplicate code and instead use the centralized service. Now, all of the factories (except tracking) define their specific log level config parameters during their constructor call so they can be listed with eicrecon -c. The convention is to use: <collectionName>:LogLevel to set the log level of a specific factory. The global log level is set with eicrecon:LogLevel as before and is inherited by all factory loggers by default.

- Cleanup how eicrecon prints config parameters. The JANA built-in PrintParameters method uses a fixed column width for the value and description which results in some exceptionally wide columns due to the very long podio:output_include_collections parameter. This is still not perfect, but is much more readable.

- Fix jana:nskip behavior. The code now uses thread id to determine if GetEvent should declare itself busy. This is not actually working to block multi-thread use, but does allow nskip to work as expected. 

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes. If a user specifies to use multiple threads, the program will likely crash. This is acceptable for now in order to regain the jana:nskip functionality.

No.

### DIFF
--- a/src/services/io/podio/JEventSourcePODIOsimple.cc
+++ b/src/services/io/podio/JEventSourcePODIOsimple.cc
@@ -225,8 +225,16 @@ void JEventSourcePODIOsimple::GetEvent(std::shared_ptr<JEvent> event) {
     }
 
     // The podio supplied RootReader and EventStore are not multi-thread capable so limit to a single event in flight
-    if( m_inflight ) throw RETURN_STATUS ::kBUSY;
-    m_inflight = true;
+    // Since the JANA skip events mechanism does not seem to call FinishEvent when skipping, any flag for
+    // in-flight events will not be cleared. Thus, use the JEvent pointer so that
+    std::thread::id no_thread; // default value is no thread
+    if( processing_thread_id == no_thread ){
+        // No thread is processing event
+        processing_thread_id = std::this_thread::get_id();
+    }else if( processing_thread_id != std::this_thread::get_id() ){
+        // Another thread is already processing the event
+        throw RETURN_STATUS ::kBUSY;
+    }
 
     // Read the specified event into the EventStore and make the EventStore pointer available via JANA
     store.clear();
@@ -282,7 +290,7 @@ void JEventSourcePODIOsimple::GetEvent(std::shared_ptr<JEvent> event) {
 //------------------------------------------------------------------------------
 void JEventSourcePODIOsimple::FinishEvent(JEvent &event){
 
-    m_inflight = false;
+    processing_thread_id = std::thread::id(); // reset to no thread value
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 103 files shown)


src/services/io/podio/JEventSourcePODIOsimple.cc
```diff
@@ -225,8 +225,16 @@ void JEventSourcePODIOsimple::GetEvent(std::shared_ptr<JEvent> event) {
     }
 
     // The podio supplied RootReader and EventStore are not multi-thread capable so limit to a single event in flight
-    if( m_inflight ) throw RETURN_STATUS ::kBUSY;
-    m_inflight = true;
+    // Since the JANA skip events mechanism does not seem to call FinishEvent when skipping, any flag for
```
nathanwbrei, line 228: I'm pretty sure this is a bug in JANA. I'll fix this for the next release.

src/services/io/podio/JEventSourcePODIOsimple.h
```diff
@@ -42,7 +42,7 @@ class JEventSourcePODIOsimple : public JEventSource {
     std::set<std::string> m_INPUT_EXCLUDE_COLLECTIONS;
     bool m_run_forever=false;
 
-    bool m_inflight = false; // is an event currently in flight (and therefore using the EventStore)?
+    std::thread::id processing_thread_id;  // defaults to a "no thread" value
 
 };
```
nathanwbrei, line 48: I understand what you are doing and why. However, it is going to make people very nervous so we should get rid of it in the next JANA release if at all possible. 

My preferred solution to this problem is to simply set 
`event->SetSequential(true)` and/or `jana:limit_events_in_flight=1`. 

Of course, I'm pretty sure this will have the same problems with `nskip` as what we had here before, so this is another item for the next JANA release.

src/services/log/Log_service.cc
```diff
@@ -13,30 +13,37 @@ Log_service::Log_service(JApplication *app) {
...
             logger = spdlog::default_logger()->clone(name);
+
+            // Set log level for this named logger allowing user to specify as config. parameter
+            // e.g. EcalEndcapPRecHits:LogLevel
+            std::string log_level_str = m_log_level_str;
+            m_application->SetDefaultParameter(name+":LogLevel", log_level_str, "log_level for "+name+": trace, debug, info, warn, error, critical, off (def. inherits from eicreconLogLevel)");
+            logger->set_level(eicrecon::ParseLogLevel(log_level_str));
         }
         return logger;
     }
```
nathanwbrei, line 49: Thank you, we've really needed this!

PR #338, opened 2022-11-12
Restore use of _exit

faustus123:

This reverses the change in PR #331 so that _exit is used unless running on macos. The use of _exit was very deliberately chosen so that the root exit handlers are bypassed, ensuring the last message on the screen is the one indicating the cause of the exit. When just using std::exit, you will get messages like the following printed *after* the message that the file doesn't exist:

~~~
Error in <TClass::LoadClassInfo>: no interpreter information for class TVirtualStreamerInfo is available even though it has a TClass initialization routine.
~~~

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: __

No
No

### DIFF
--- a/src/services/geometry/dd4hep/JDD4hep_service.cc
+++ b/src/services/geometry/dd4hep/JDD4hep_service.cc
@@ -143,7 +143,11 @@ std::string JDD4hep_service::resolveFileName(const std::string &filename, char *
                 mess += fmt::format(fmt::emphasis::bold, "file: {} does not exist!", filename);
                 mess += "\nCheck that your DETECTOR and DETECTOR_CONFIG environment variables are set correctly.";
                 std::cerr << std::endl << std::endl << mess << std::endl << std::endl; // TODO standard log here!
-                std::exit(-1);  // TODO isn't unhandled JException is a better way to abort the application
+#ifdef __APPLE__
+                std::exit(-1);  // macos does notsupport _exit()
+#else
+                _exit(-1);  // exit while bypassing exit handlers
+#endif
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/services/geometry/dd4hep/JDD4hep_service.cc
```diff
@@ -143,7 +143,11 @@ std::string JDD4hep_service::resolveFileName(const std::string &filename, char *
                 mess += fmt::format(fmt::emphasis::bold, "file: {} does not exist!", filename);
                 mess += "\nCheck that your DETECTOR and DETECTOR_CONFIG environment variables are set correctly.";
                 std::cerr << std::endl << std::endl << mess << std::endl << std::endl; // TODO standard log here!
-                std::exit(-1);  // TODO isn't unhandled JException is a better way to abort the application
+#ifdef __APPLE__
+                std::exit(-1);  // macos does notsupport _exit()
+#else
+                _exit(-1);  // exit while bypassing exit handlers
+#endif
```
veprbl, line 150:
```suggestion
                std::_Exit(EXIT_FAILURE);
```

src/services/io/podio/JEventSourcePODIOsimple.cc
```diff
@@ -131,7 +131,11 @@ void JEventSourcePODIOsimple::Open() {
             auto mess = fmt::format(fmt::emphasis::bold | fg(fmt::color::red),"ERROR: ");
             mess += fmt::format(fmt::emphasis::bold, "file: {} does not exist!",  GetResourceName());
             std::cerr << std::endl << std::endl << mess << std::endl << std::endl;
-            std::exit(EXIT_FAILURE);
+#ifdef __APPLE__
+            std::exit(EXIT_FAILURE);  // macos does notsupport _exit()
+#else
+           _exit(-1);  // exit while bypassing exit handlers
+#endif
```
veprbl, line 138:
```suggestion
            std::_Exit(EXIT_FAILURE);
```

faustus123:
So I take it this means _Exit is defined on both Linux and macos? Geez, if there is one thing guaranteed its that whatever standard exists, they will have to change it at some point in a way that is utterly confusing!

Using a single exit call is certainly better than conditionally compiled based on OS. I'll go ahead and accept the changes.

PR #349, opened 2022-11-15
feat: SiPM pixel gap cuts

c-dilks:
Add the ability to cut on the boundaries of an XY-segmented SiPM sensor, to simulate gaps between SiPM pixels. Also adds an overall safety factor, to account for the non-ideal reality.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
no

### DIFF
--- a/src/algorithms/digi/PhotoMultiplierHitDigi.cc
+++ b/src/algorithms/digi/PhotoMultiplierHitDigi.cc
@@ -24,9 +24,14 @@ using namespace dd4hep;
 //------------------------
 // AlgorithmInit
 //------------------------
-void eicrecon::PhotoMultiplierHitDigi::AlgorithmInit(std::shared_ptr<spdlog::logger>& logger) 
+void eicrecon::PhotoMultiplierHitDigi::AlgorithmInit(dd4hep::Detector *detector, std::shared_ptr<spdlog::logger>& logger) 
 {
+    // services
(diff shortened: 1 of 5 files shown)


c-dilks:
@DraTeots I'm getting a strange warning here. If I change this `pixelSize` from `4.0` to `3.0` here
https://github.com/eic/EICrecon/blob/4c0c87cb888cae2b97500f90096c319e83e7efa5/src/algorithms/digi/PhotoMultiplierHitDigiConfig.h#L21
I get the following `JANA` warning:
```
[WARN] Parameter 'DRICH:DRICHRawHits:pixelSize' loses equality with itself after stringification
```
The warning also prints when it is (erroneously) set to `3.0*dd4hep::ns`, but other parameters that are defined in that header in `ns` units do not have this warning.

Since these parameters will be user-overridden anyway (e.g., `reco_flags.py`), do we care enough about this to open an issue?

DraTeots:
> @DraTeots I'm getting a strange warning here. If I change this `pixelSize` from `4.0` to `3.0` here
> ...
> 
> Since these parameters will be user-overridden anyway (e.g., `reco_flags.py`), do we care enough about this to open an issue?

After some investigation, I believe you can safely ignore this warning. Jana2 checks how it does stringification/parsing of parameters and floating point numbers sometimes just can't be equal to them selves by their nature. I created an issue in Jana2 on your behalf: 

https://github.com/JeffersonLab/JANA2/issues/175

Issue #351, opened 2022-11-15
Imaging clusters missing in reconstructed files


mariakzurek:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Check the newest produced test files in Actions, for example here: https://github.com/eic/EICrecon/actions/runs/3450751526

### Expected Result: (what do you expect when you execute the steps above)

Expected clusters from imaging layers

### Actual Result: (what do you get when you execute the steps above)

No clusters from imaging layers reconstructed 

I've reviewed again the parameters in [reco_flags.py](https://github.com/eic/EICrecon/blob/main/src/tools/default_flags_table/reco_flags.py#L144)

and I've spotted one issue. The min calo energy is 0.5 * MeV and it should be 0.5 * MeV/sampling_fraction, as we discussed here: 
https://github.com/eic/EICrecon/pull/253#discussion_r1001288982

I don't think that this will cause all clusters to disappear, because it's a lower limit, but it should be fixed anyhow. 

The rest of the parameters look ok

`    ('BEMC:EcalBarrelImagingProtoClusters:input_tag',           'EcalBarrelImagingRecHits',        'Name of input collection to use'),
    ('BEMC:EcalBarrelImagingProtoClusters::localDistXY',        '2.0,2.0',                         '* [mm]'),
    ('BEMC:EcalBarrelImagingProtoClusters::layerDistEtaPhi',    '0.01,0.01',                       '* [radian]'),
    ('BEMC:EcalBarrelImagingProtoClusters::neighbourLayersRange', '2.0',                           ''),
    ('BEMC:EcalBarrelImagingProtoClusters::sectorDist',         '3.0*cm',                          ''),
    ('BEMC:EcalBarrelImagingProtoClusters::minClusterHitEdep',  '0.',                              ''),
    ('BEMC:EcalBarrelImagingProtoClusters::minClusterCenterEdep', '0.',                            ''),
    ('BEMC:EcalBarrelImagingProtoClusters::minClusterEdep',     '0.5*MeV',                         ''),
    ('BEMC:EcalBarrelImagingProtoClusters::minClusterNhits',    '5',                               ''),`

But are they really passed with the proper units? Is, e.g., `localDistXY` passed as `2*mm` or as `2`? or is `0.5*MeV` passed as `0.5` (GeV?)?

faustus123:
I will take a look at this later today or this evening.

[...]

PR #354, opened 2022-11-16
Fix Imaging Barrel Ecal Clustering

faustus123:

This fixes the Imaging Barrel calorimeter so that the `EcalBarrelImagingClusters` and `EcalBarrelImagingMergedClusters` are written out. This was tested with single 10GeV e- events (see attached plots). 

The main issue was with podio refusing to write collections that references "untracked" objects. This is a known issue with the podio+JANA2 implementation. The "fix" was to disable adding the references to the cluster objects.

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes, it makes it work.

image

### DIFF
--- a/src/algorithms/calorimetry/ImagingClusterReco.h
+++ b/src/algorithms/calorimetry/ImagingClusterReco.h
@@ -96,7 +96,7 @@ public:
             for (const auto &layer: cl_layers) {
                 auto layer_ptr = new edm4eic::Cluster( layer );
                 layers.push_back(layer_ptr);
-                cl.addToClusters(*layer_ptr);
+//                cl.addToClusters(*layer_ptr);
             }
             clusters.push_back(new edm4eic::Cluster( cl ));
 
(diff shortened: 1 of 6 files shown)


mariakzurek:
Can we change the [imaging cluster threshold](https://github.com/eic/EICrecon/blob/ea1e92e8d9a5f05147993a414f1c693ea43b6988/src/tools/default_flags_table/reco_flags.py#L147) according to the comment in the Issue #351

mariakzurek:
I think that this fixes the issue with non-existing imaging clusters! 

However when I look at the SciFi clusters I am seeing a bit different picture then you. The low energy threshold for the cluster is currently 10 MeV, and this can "catch" low-energy outliers (peak at ~0). However in your plot for 10 GeV I see that there is also non negligible number of clusters in between ~0.01 GeV peak and ~8-10 GeV peak. I show the similar plot below produced with the same parameters from juggler. There is a small energy peak, but much less "in between". 
10GeV-test

mariakzurek:
I think that we can push the current changes, and work on SciFi clusters separatelly

PR #365, opened 2022-11-18
fix: use DD4hep units in reco_flags.py

wdconinc:
There are too many unit conversions that are not happening, and it all goes back to the fact that reco_flags it not sending parameters in the DD4hep unit system that is used in the factories and algorithms. This aims to simplify at least one part of this by removing the need for an interface.

- [X] Bug fix (issue: units not always converted correctly)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No, not towards users who shoudl specify values with their units.


DraTeots (approved):

DraTeots:
@wdconinc is this still work in progress or it is a draft by mistake?

wdconinc:
> @wdconinc is this still work in progress or it is a draft by mistake? 

This doesn't currently produce the same results as main.

DraTeots:
Thank you for the clarification!

c-dilks:
Shouldn't this have been merged before production?

wdconinc:
No consensus.

DraTeots:
As soon as this get merged I can start fixing default parameters in code to get rid of reco_flags.py

c-dilks:
Rebased

c-dilks:
I'm adding a way for `eicrecon` to parse parameters as DD4hep expressions, automating the conversion from 'with-units' to 'unitless'. Needs more work, but the results are promising:

2022-12-07-011830_925x488_scrot

faustus123:
It is certainly very appealing from the "setter" point of view to be able to use this kind of syntax. My one concern is that without having the set line in front of you, how would one know if ``test:param`` was in GeV or MeV? The only way would be to know that all parameters 100% agree on the internal units and that they define GeV=1. IMHO giving the parameter a name like "test:param_GeV" makes it unambiguous in all contexts. It won't look quite as elegant as what you've done but:

``eicrecon -Ptest:param_GeV=1``

Doesn't seem too bad. What do you think?

c-dilks:
I like the idea of unambiguous units, but I fear this will cause formulas to be plagued with factors of 10, which is a common source for mistakes. It is also harder to update units, if there is ever a reason to.

For below, let's define a "system of units" as a map of unit names to `double` powers of 10. 

There are times when an algorithm will need to know about units, to know which power of 10 to use, for example converting a photon energy to a wavelength. The algorithm itself does not care what system of units is being used, and can be configured to use any; let's call that system "algorithm units".

The only time when it matters *which* system of algorithm units is used happens when numbers are passed into or out of the algorithm, from or to somewhere else, say an `eicrecon` factory. Suppose `eicrecon` factories use the "eicrecon units" system. Our parameter parser maps human-specified "user units" to "eicrecon units" (by multiplying a `double` power of 10).

We allow both "eicrecon units" and "algorithm units" to be freely configurable, but to allow compatibility, we configure them to be equivalent to each other. Then in the algorithm side, we only need to divide the "algorithm units" (divide a `double` power of 10) whenever we care about units, but otherwise most formulas can be written without caring about any SI-prefixes.

Having a configurable system of units is more consistent with other software "modules", such as Geant4 and DD4hep. As long as the systems are configured equivalently, the modules are compatible. There is no need to lock a module into having a specific system of units, but of course we want all of the modules to agree on a single (and easily configurable) system.

c-dilks:
The parser is now fully compatible with `reco_flags.py`, and its results have been cross checked. Consistency with previous versions of EICrecon was verified by comparing `run_eicrecon_reco_flags.py` output, with the `-c` option used for `eicrecon`. The only differences are minor, typically dropping trailing zeros (e.g., `"4.0"` becomes `"4"`).

There are 2 unit system choices included:

### Geant4 units:
```cpp
1 = millimeter = MeV = nanosecond = nanoampere (via e+ charge) = kelvin = mole = candela = radian = steradian
```
This is what the current `main` branch `reco_flags.py` is using, and is consistent with parsing using
```cpp
dd4hep::tools::Evaluator::Object(
    1.e+3, // meter
    1./1.60217733e-25, // kilogram
    1.e+9, // second
    1./1.60217733e-10, // ampere
    1.0, // kelvin
    1.0, // mole
    1.0, // candela
    1.0  // radian
    );
);
```


### DD4hep units:
Our current `eic-shell` build of DD4hep uses the following system:
```cpp
1 = centimeter = GeV = second = nanoampere = kelvin = mole = candela = radian = steradian
```
Since this can be re-configured, instantiate the `Evaluator` with whatever the current `dd4hep::` `double` values are:
```cpp
dd4hep::tools::Evaluator::Object(
    dd4hep::meter,
    dd4hep::kilogram,
    dd4hep::second,
    dd4hep::ampere,
    dd4hep::kelvin,
    dd4hep::mole,
    dd4hep::candela,
    dd4hep::radian
    );
```
Parsing with this system is consistent with Wouter's last commit https://github.com/eic/EICrecon/pull/365/commits/d71dee8a8c20e608c84a9feeb5d91041bf953e04, since he modified `reco_flags.py` to use the `eic-shell` DD4hep system.

c-dilks:
Attempt to correct for cases where a parameter that should have units does not. For example, if `-Pdistance=1` is used, assuming this means 1 mm, and is used in a formula like `distance / dd4hep::mm`, it will become 10 mm. Instead, use `-Pdistance=1*mm`.

faustus123:
Thanks again for working on this Chris and for putting a lot of thought into it. 

There is currently a proposal to make the next iteration of the reconstruction software in a form that allows for a certain class of user. This user would modify dd4hep geometry files to add a new detector and then modify configuration files to run hits from that detector through algorithms, all without having to look at or even recompile the reconstruction code. In other words, they will be completely blind to anything in C++. 

There are two places where I see this user interacting with the configuration parameters:

1. When they set them in the configuration file. You have handled this very nicely as they can specify the units in an intuitive way.

2. When they see a dump of the configuration parameters in a log. This will then just be a number, converted using whatever unit system was used.

So my question is how do you see handling case 2 so that they know the units the number is in when all they see is the name of the parameter and a bare number?

wdconinc:
> So my question is how do you see handling case 2 

If you have a system of units, it's of course possible to export that system of units too. Eg printing out the numerical value of `mm` as `1` or `0.1` (and similar for other units), fully specifies the necessary info in any output.

faustus123:
> If you have a system of units, it's of course possible to export that system of units too. Eg printing out the numerical value of mm as 1 or 0.1 (and similar for other units), fully specifies the necessary info in any output.

This is certainly true. It will require the user though to:

A. Know that the values all adhere to the same set of units
B. Search through the log to find the table of values (and know it is there)
C. Know the type of unit the value they are looking at is in

Most of the time "C" will be obvious. Things like "dynamicRangeADC" may not necessarily be so obvious that it is in units of energy vs. say "8192" or "1V".

My suggestion is that a name like "dynamicRangeADC_GeV" would make it completely clear without requiring A,B, or C. i.e. one would not require any prior knowledge or need to search for additional information in the log.

The downside is that on the C++ side, a unit conversion would have to be applied explicitly as opposed to the automatic conversion in Chris' solution. e.g.

`m_dynamicRangeADC = dynamicRangeADC_GeV * dd4hep::GeV`

Maybe what is needed is to add an optional unit field to `JParameter` so it can be also expressed whenever the value is printed?

wdconinc:
Suggestion: take the unit parser out of this PR, put the parser only in a new PR, have this discussion in a meeting where EPIC CompSW/SimQA can express a preference, and in the meantime fix the units in EICrecon as in original scope. It is not a tenable situation when we have to define parameters intentionally incorrectly to have the reconstruction work, as in https://github.com/eic/EICrecon/pull/389.

c-dilks:
@faustus123 could the user's parameter input be tracked? Then the user does not have to separate the unit(s) from the number and can use the DD4hep `Evaluator` as is.

This would be similar to DD4hep, which stores the name, the numerical "unitless" value, and the user's input (from the compact `xml` file). For example, dumping constants from `epic`:
```
DRICH_aerogel_material         = "Aerogel_DRICH" = Aerogel_DRICH
DRICH_aerogel_thickness        =        4.000    = 4.0*cm
DRICH_airgap_thickness         =        0.001    = 0.01*mm
```

c-dilks:
Rebased this PRs branch, and rolled it back to @wdconinc's last commit. This PR thus now focuses on its original goal. A new PR for the `Evaulator`-based PR will be opened and we can continue the discussion there

faustus123:
So this is still marked as a "Draft". Is it ready to be merged?

wdconinc:
> Is it ready to be merged?

Maybe. I had marked it as draft mainly because the imaging calorimeter clusters were not reconstructed correctly in this branch, but neither are they in the main branch apparently, so that's no regression.

It's not tested to the extent I would want a PR to the reconstruction code to be tested, but I think it's as well tested as many other PRs that have gotten merged.

If the testing strategy here is mainly based on whether it compiles and runs, and if all tests on whether things actually reconstruct correctly are pushed to the validation phase by the working groups when we run a first pass of large simulations, then yeah just merge it and glue the pieces later.

PR #367, opened 2022-11-18
fix: add default Ecal Endcap N params in factories

wdconinc:
This propagates the EcalEndcapN parameters from reco_flags.py to the factories.

### DIFF
--- a/src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
+++ b/src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
@@ -26,15 +26,15 @@ public:
         m_input_tag = "EcalEndcapNRawHits";
 
         // digitization settings, must be consistent with digi class
-        m_capADC=8096;//{this, "capacityADC", 8096};
-        m_dyRangeADC=100. * MeV;//{this, "dynamicRangeADC", 100. * MeV};
-        m_pedMeanADC=400;//{this, "pedestalMean", 400};
-        m_pedSigmaADC=3.2;//{this, "pedestalSigma", 3.2};
+        m_capADC=16384;//{this, "capacityADC", 8096};
+        m_dyRangeADC=20. * GeV;//{this, "dynamicRangeADC", 100. * MeV};
+        m_pedMeanADC=100;//{this, "pedestalMean", 400};
+        m_pedSigmaADC=1;//{this, "pedestalSigma", 3.2};
         m_resolutionTDC=10 * dd4hep::picosecond;//{this, "resolutionTDC", 10 * ps};
 
         // zero suppression values
         m_thresholdFactor=4.0;//{this, "thresholdFactor", 0.0};
-        m_thresholdValue=0.0;//{this, "thresholdValue", 0.0};
+        m_thresholdValue=3.0;//{this, "thresholdValue", 0.0};
 
         // energy correction with sampling fraction
         m_sampFrac=0.998;//{this, "samplingFraction", 1.0};
--- a/src/detectors/EEMC/Cluster_factory_EcalEndcapNTruthClusters.h
+++ b/src/detectors/EEMC/Cluster_factory_EcalEndcapNTruthClusters.h
@@ -45,8 +45,8 @@ public:
         m_input_simhit_tag="EcalEndcapNHits";
         m_input_protoclust_tag="EcalEndcapNTruthProtoClusters";
     
-        m_sampFrac=1.0;//{this, "samplingFraction", 1.0};
-        m_logWeightBase=3.6;//{this, "logWeightBase", 3.6};
+        m_sampFrac=0.03;//{this, "samplingFraction", 1.0};
+        m_logWeightBase=4.6;//{this, "logWeightBase", 3.6};
         m_depthCorrection=0.0;//{this, "depthCorrection", 0.0};
         m_energyWeight="log";//{this, "energyWeight", "log"};
         m_moduleDimZName="";//{this, "moduleDimZName", ""};
--- a/src/detectors/EEMC/RawCalorimeterHit_factory_EcalEndcapNRawHits.h
+++ b/src/detectors/EEMC/RawCalorimeterHit_factory_EcalEndcapNRawHits.h
@@ -39,10 +39,10 @@ public:
         m_input_tag = "EcalEndcapNHits";
         u_eRes = {};
         m_tRes = 0.0 * ns;
-        m_capADC = 8096;
-        m_dyRangeADC = 100 * MeV;
-        m_pedMeanADC = 400;
-        m_pedSigmaADC = 3.2;
+        m_capADC = 16384;
+        m_dyRangeADC = 20 * GeV;
+        m_pedMeanADC = 100;
+        m_pedSigmaADC = 1;
         m_resolutionTDC = 10 * picosecond;
         m_corrMeanScale = 1.0;
         u_fields={};


src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
```diff
@@ -26,15 +26,15 @@ class CalorimeterHit_factory_EcalEndcapNRecHits : public JFactoryT<edm4eic::Calo
         m_input_tag = "EcalEndcapNRawHits";
 
         // digitization settings, must be consistent with digi class
-        m_capADC=8096;//{this, "capacityADC", 8096};
-        m_dyRangeADC=100. * MeV;//{this, "dynamicRangeADC", 100. * MeV};
-        m_pedMeanADC=400;//{this, "pedestalMean", 400};
-        m_pedSigmaADC=3.2;//{this, "pedestalSigma", 3.2};
+        m_capADC=16384;//{this, "capacityADC", 8096};
+        m_dyRangeADC=20. * GeV;//{this, "dynamicRangeADC", 100. * MeV};
+        m_pedMeanADC=100;//{this, "pedestalMean", 400};
```
johnny8266, line 31: I did the quick estimation of pedestal in unit of GeV, if I think correctly:
100[mean pedestal in ADC] / 16384[updated ADC dynamic range] * 20GeV = 0.12 GeV
which looks really high.

src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
```diff
@@ -26,15 +26,15 @@ class CalorimeterHit_factory_EcalEndcapNRecHits : public JFactoryT<edm4eic::Calo
...
+        m_pedSigmaADC=1;//{this, "pedestalSigma", 3.2};
         m_resolutionTDC=10 * dd4hep::picosecond;//{this, "resolutionTDC", 10 * ps};
 
         // zero suppression values
         m_thresholdFactor=4.0;//{this, "thresholdFactor", 0.0};
-        m_thresholdValue=0.0;//{this, "thresholdValue", 0.0};
+        m_thresholdValue=3.0;//{this, "thresholdValue", 0.0};
```
johnny8266, line 37: Just out of curiosity, what is the unit of m_thresholdValue?

src/detectors/EEMC/Cluster_factory_EcalEndcapNTruthClusters.h
```diff
@@ -45,8 +45,8 @@ class Cluster_factory_EcalEndcapNTruthClusters : public JFactoryT<edm4eic::Clust
         m_input_simhit_tag="EcalEndcapNHits";
         m_input_protoclust_tag="EcalEndcapNTruthProtoClusters";
     
-        m_sampFrac=1.0;//{this, "samplingFraction", 1.0};
-        m_logWeightBase=3.6;//{this, "logWeightBase", 3.6};
+        m_sampFrac=0.03;//{this, "samplingFraction", 1.0};
+        m_logWeightBase=4.6;//{this, "logWeightBase", 3.6};
```
johnny8266, line 49: Just out of curiosity as well here, what is the unit of these 2 values?
what is the difference the m_sampFrac and the sampling fraction we used to correct energy?
Thanks!

src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
wdconinc: Those are the values in https://github.com/eic/EICrecon/blob/main/src/tools/default_flags_table/reco_flags.py#L182-L209.

wdconinc: Those are the values in https://github.com/eic/EICrecon/blob/main/src/tools/default_flags_table/reco_flags.py#L182-L209.

src/detectors/EEMC/Cluster_factory_EcalEndcapNTruthClusters.h
wdconinc: Sampling fraction is a number between 0 and 1: how much energy is deposited in sensitive elements (as opposed to radiators). For homogeneous approximations of microstructure, this is expected to be different from reality.

The log weight base is a dimensionless constant that ln(E/Etotal) is added to in the center of gravity cluster weighting. A larger values means more uniform weighting; a smaller value means more logarithmic weighting.

src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
wdconinc: It seems in units of ADC.

src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
```diff
@@ -26,15 +26,15 @@ class CalorimeterHit_factory_EcalEndcapNRecHits : public JFactoryT<edm4eic::Calo
         m_input_tag = "EcalEndcapNRawHits";
 
         // digitization settings, must be consistent with digi class
-        m_capADC=8096;//{this, "capacityADC", 8096};
-        m_dyRangeADC=100. * MeV;//{this, "dynamicRangeADC", 100. * MeV};
-        m_pedMeanADC=400;//{this, "pedestalMean", 400};
-        m_pedSigmaADC=3.2;//{this, "pedestalSigma", 3.2};
+        m_capADC=16384;//{this, "capacityADC", 8096};
+        m_dyRangeADC=20. * GeV;//{this, "dynamicRangeADC", 100. * MeV};
```
DraTeots, line 30: Where is the GeV multiplier comes from and how it is related to GeV in reco_flags.py?

DraTeots (commented):
Where GeV in C++ coming from? No namespace or explicit header.

src/detectors/EEMC/CalorimeterHit_factory_EcalEndcapNRecHits.h
wdconinc: JDD4hep_service typically

wdconinc: It's the same now.

DraTeots: I still don't like that there is an implicit "GeV" without namespace, which means that it is getting slipped in through a header. But that is probably for another issue. 

@wdconinc

wdconinc: A long list of headers have a `using namespace dd4hep`...

wdconinc: The latest clang has `-Wheader-hygiene` which flags exactly this in headers (to be used with `-Werror`).

johnny8266:
> @johnny8266 note that I simply copied what is in `reco_flags.py` and that these are the values we are already using. If they need changing, that should be part of a separate PR.

Thanks for the detailed reply. That's really helpful. At the beginning, we just keep those parameters unchanged, now, we are looking into the simulation results and compare the hits and reconstruction clusters. we would make another PR, if we want to changed those parameters.

src/detectors/EEMC/Cluster_factory_EcalEndcapNTruthClusters.h
johnny8266: Very thanks, it is much more clear now. Final question, do you know how to estimate the log weight? By the ratio of center seed energy and rest of hit energy?

wdconinc: No idea. Maybe @chao1009 or @mariakzurek knows.

Chao1009: Use b = -ln(x), with x = hit_energy/cluster_energy to estimate the base.
For example, b = 4.6 means those hits with energy less than 1% of the cluster energy won't participate in coordinate reconstruction.

johnny8266: Thanks for answering!

PR #377, opened 2022-11-19
ci: build as matrix; use ccache

wdconinc:
The build takes 15 minutes, and likely most files don't change frequently. We can use ccache to speed up. Rather than implementing the caching twice, this moves the builds with gcc and clang into a matrix build. This improves build times to 3 minutes:
image

- [ ] Bug fix (issue #__)
- [x] New feature (issue: faster CI builds)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -50,7 +50,7 @@ jobs:
         platform-release: "jug_xl:nightly"
         run: |
           # install this repo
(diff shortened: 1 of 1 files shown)


wdconinc:
Note: GitHub Caches are per repository so this won't pick up the cache I created for testing on my fork.

DraTeots (approved):

PR #380, opened 2022-11-19
Fix CI failure on redundatnt flags python generation

### LINKED ISSUE #374: CI flags generation is not working
It became obsolete as @sqrd-max made table driven by vue

DraTeots:

Removes "report" step and generation of python files with flags and flags.md page. This all has been replaced by using output flags.json directly (from eicrecon run with dump_flags). Report step can be re-instantiated later... when there will be something else to report. 

- [x] Bug fix (fixes #374)

NO

NO

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -166,25 +166,7 @@ jobs:
         path: jana.dot
         if-no-files-found: error
 
-  diff-flags:
-    runs-on: ubuntu-latest
-    needs:
-      - run_eicrecon_reco_flags-gcc
-      - eicrecon-gcc
-    strategy:
-      matrix:
-        detector_config: [arches, brycecanyon]
-    steps:
-      - uses: actions/download-artifact@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -184,39 +184,12 @@ jobs:
       - run: |
           diff ${{ matrix.detector_config }}_right_flags.json ${{ matrix.detector_config }}_flags.json
```
wdconinc, line 185: Isn't this diff line what fails?

DraTeots: IDK whose line is that =)

Json file has both default valus and user values. So single file from reco-flags has all that is needed to diff. Then there is a GUI now.

DraTeots: Sorry, from phone, cant run blame and investigate

DraTeots: fixed

PR #382, opened 2022-11-21
Add simple dispatch

DraTeots:
Experimenting with running benchmarks in a separate repository


DraTeots:
You probably want to ingore this PR for now as it is used for experiments with triggering benchmarks after eicrecon tests are done

PR #388, opened 2022-12-02
Fixes to LFHCal simulation digi/reco

steinber:

This includes some changes (hard won) which fixes inconsistencies in the digi and reco steps in the LFHCal (previously they were far out of sync, and so many of the energy deposits were not being reconstructed) 

- [X ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [] Tests for the changes have been added - I have tested it locally.  Added to where?

Nothing broke before, but most of the energy deposits were truncated and so no clusters were dfound
The LFHCal reconstructs now reconstructs energies (although the scale has changed since 2 weeks ago - under investigation).


DraTeots (changes requested):
Please, use DD4Hep units explicitly units for config parameter values. We now experience hard time to solve problems as values fly around without proper units set.

src/detectors/HCAL/CalorimeterHit_factory_HcalEndcapPRecHits.h
```diff
@@ -26,9 +26,9 @@ class CalorimeterHit_factory_HcalEndcapPRecHits : public JFactoryT<edm4eic::Calo
         m_input_tag = "HcalEndcapPRawHits";
 
         // digitization settings, must be consistent with digi class
-        m_capADC=8096;//{this, "capacityADC", 8096};
-        m_dyRangeADC=100. * dd4hep::MeV;//{this, "dynamicRangeADC", 100. * dd4hep::MeV};
-        m_pedMeanADC=400;//{this, "pedestalMean", 400};
+        m_capADC=4096;//{this, "capacityADC", 8096};
+        m_dyRangeADC=200.;//{this, "dynamicRangeADC", 100. * dd4hep::MeV};
```
DraTeots, line 30: Please use units here

src/detectors/HCAL/RawCalorimeterHit_factory_HcalEndcapPRawHits.h
```diff
@@ -39,10 +39,10 @@ class RawCalorimeterHit_factory_HcalEndcapPRawHits : public JFactoryT<edm4hep::R
         m_input_tag = "HcalEndcapPHits";
         u_eRes = {};
         m_tRes = 0.0 * dd4hep::ns;
-        m_capADC = 1024;
-        m_dyRangeADC = 3.6 * dd4hep::GeV;
-        m_pedMeanADC = 20;
-        m_pedSigmaADC = 0.8;
+        m_capADC = 4096;
+        m_dyRangeADC = 200. ;
```
DraTeots, line 43: Please ensure units are used here

DraTeots (changes requested):

src/detectors/HCAL/RawCalorimeterHit_factory_HcalEndcapPRawHits.h
steinber: Hi - I had to remove the units since dyRangeADC secretly adds them in CalorimeterHitReco.cc in L27 like this: 

`dyRangeADC = m_dyRangeADC * dd4hep::MeV;`

So If I presented it with MeV, then it would be scaled twice (which completely threw me off).  I would prefer dyRangeADC to be assumed to have energy units, and remove this hidden rescaling (unless I missed something).

DraTeots: Hidden conversions are bad. Let me look into that. The problem is that we have issue with units. @wdconinc started to fix those and filter out what happens in calorimetry algorithms with units here:  https://github.com/eic/EICrecon/pull/365. But it has not yet been finished. And I started to put the correct default values from python to C++. With all that we decided that the correct implicit units shold be DD4Hep ones. @wdconinc is it known and fixed in your PR?

wdconinc: #365 does indeed remove the explicit addition of MeV in CalorimeterHitReco.cc, https://github.com/eic/EICrecon/pull/365/files#diff-7a64db59fabe7d6949bed432036c5f5250a349e7e4d50fc7d4d20a39a96c6b36L27.

steinber:
Hi - just to clarify: i'm waiting for #365, so I can test it properly?

DraTeots (approved):

DraTeots:
@steinber Lets commit it as is (as it works in main now) and then probably it would be a critical place to watch for error to reappear. I hope on your help here and validation of what is being done.

steinber:
Thanks - i'm not remotely done with this, so I'm happy to keep up with it.

PR #399, opened 2022-12-16
feat: use modified true momentum as tracking seed parameter

wdconinc:
When tracking fails, we bleed through initial momentum seed settings to the tracking results. This is hard to debug or even notice. This branch just serves as a debugging vehicle.

In this branch we pick the initial momentum as 90%, 100% or 110% of the initial momentum, with equal probabilities. If the dp/p starts to look like 🤘 then we have a problem...


wdconinc:
@rahmans1 This could help figure out if the tracking is failing. Not in acts-20.3.0 branch, so it would require a custom install into your container.

sly2j:
Seeing as this is in general a useful debugging tool, wouldn't it make sense to ship this with the main truth seeder (enabled by a flag)? Something where we have different truth seeding modes:
* Exact (default)
* Gaussian
* Bifurcated

Simple-Shyam:
I want to study it as discussed in ACTS meeting. When will it merge to EICRecon?

src/global/tracking/TrackParamTruthInit_factory.cc
```diff
@@ -15,12 +15,20 @@
...
+    auto app = GetApplication();
+
     // This prefix will be used for parameters
     std::string param_prefix = GetDefaultParameterPrefix();
 
     InitLogger(param_prefix);
     InitDataTags(param_prefix);
 
+    // Algorithm configuration
+    m_truth_track_seeding_algo.m_momentum_split = 0.0;
```
DraTeots, line 27: It is suggested, instead of using class memebers directly (which is supposed to be somewhat prived) it is suggested to use <className>Config POD structure to store algorithms configs. And then you can inherit from WithPODConfig to get some helper methods. To be more precise: 

- [Example of simple config TrackerHitReconstructionConfig.h](https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackerHitReconstructionConfig.h)
- [Example of CFKTrackingConfig](https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/CKFTrackingConfig.h)
   
   ```c++
       // Initialized with default values
       struct CKFTrackingConfig {
        std::vector<double> m_etaBins = {};  // {this, "etaBins", {}};
        std::vector<double> m_chi2CutOff = {50.}; //{this, "chi2CutOff", {15.}};
        std::vector<size_t> m_numMeasurementsCutOff = {10}; //{this, "numMeasurementsCutOff", {10}};
    };
   ```
   
- [Example of using WithPODConfig for CFKTracking](https://github.com/eic/EICrecon/blob/cb09f2e59edaab38e3ce348a64d9ef01c4328531/src/algorithms/tracking/CKFTracking.h#L45)

   ```c++
       class CKFTracking: public WithPodConfig<eicrecon::CKFTrackingConfig> {
   ```
   
   It makes m_cfg (or getConfig()) available inside CKFTracking.

- Then if EICRecon specific JChainFactoryT is used (and they are used everywhere for tracking in EICrecon) it also provides helper methods. [Example with JChainFactoryT  definition for CKFTracking:](https://github.com/eic/EICrecon/blob/cb09f2e59edaab38e3ce348a64d9ef01c4328531/src/global/tracking/CKFTracking_factory.h#L21) 

   ```c++
       class CKFTracking_factory :
            public JChainFactoryT<eicrecon::TrackingResultTrajectory, CKFTrackingConfig>,
            public SpdlogMixin<CKFTracking_factory>
   ```

- Config parameters are simply accessible [example](https://github.com/eic/EICrecon/blob/main/src/global/tracking/TrackerHitReconstruction_factory.cc#L19)

   ```c++
       // Set time resolution
    app->SetDefaultParameter(param_prefix + ":TimeResolution", m_reco_algo.getConfig().time_resolution, "threshold");
    ```
  
- For more complex cases DefaultConfig() and then ApplyConfig methods could be used. [Example](https://github.com/eic/EICrecon/blob/cb09f2e59edaab38e3ce348a64d9ef01c4328531/src/global/tracking/CKFTracking_factory.cc#L40)

   ```c++
       // Algorithm configuration
    auto cfg = GetDefaultConfig();
    app->SetDefaultParameter(param_prefix + ":EtaBins", cfg.m_etaBins, "Eta Bins for ACTS CKF tracking reco");
    app->SetDefaultParameter(param_prefix + ":Chi2CutOff", cfg.m_chi2CutOff, "Chi2 Cut Off for ACTS CKF tracking");
    app->SetDefaultParameter(param_prefix + ":NumMeasurementsCutOff", cfg.m_numMeasurementsCutOff, "Number of measurements Cut Off for ACTS CKF tracking");

    // Initialize algorithm
    m_tracking_algo.applyConfig(cfg);
  ```

src/algorithms/tracking/TrackParamTruthInit.h
```diff
@@ -26,6 +27,13 @@ namespace eicrecon {
         std::shared_ptr<spdlog::logger> m_log;
         TrackParamTruthInitConfig m_cfg;
         std::shared_ptr<TDatabasePDG> m_pdg_db;
+
+        double m_momentum_split{0.0};
```
DraTeots, line 31: It is suggested to use Config structure for this class. See further.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -72,6 +72,12 @@ eicrecon::TrackParameters *eicrecon::TrackParamTruthInit::produce(const edm4hep:
         return nullptr;
     }
 
+    // modify initial momentum to avoid bleeding truth to results when fit fails
+    // this picks uniformly between [1-eps,1,1+eps] times true momentum, then smeared
+    const auto pinit = pmag 
+                     * (1.0 + m_momentum_split * m_uniformIntDist(generator))
```
DraTeots, line 78: It is suggested to use config POD for parameters such as m_momentum_split

DraTeots (changes requested):

DraTeots:
@wdconinc If you would like, I can add this config thing to this PR, just tell.

src/global/tracking/TrackParamTruthInit_factory.cc
wdconinc: Out of scope of this PR.

src/algorithms/tracking/TrackParamTruthInit.h
wdconinc: Out of scope.

src/algorithms/tracking/TrackParamTruthInit.cc
wdconinc: Out of scope.

wdconinc:
@DraTeots There will be a discussion (being organized by @faustus123 and @nathanwbrei) about how to do configuration in EICrecon. There are currently two approaches. Without discussion, I don't think it is clear that this struct-based approach that you suggest is the right one to use.

DraTeots:
1. First the whole tracking part is written like this. I think it is good manner to follow surrounding code. 
2.  > Without discussion, I don't think it is clear that this struct-based approach that you suggest is the right one to use.
   there going to be @sly2j algorithms which would replace all this, so what? Now we use what is there.

DraTeots:
Moreover. [TrackParamTruthInitConfig](https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackParamTruthInitConfig.h) is there. Why would one ignore that?

wdconinc:
> Moreover. [TrackParamTruthInitConfig](https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackParamTruthInitConfig.h) is there. Why would one ignore that?

That's an excellent point. I simply came to this from the user point of view (i.e. the factory) where none of the configuration setters are implemented and I missed the fact that a choice had already been made for this class. I will modify and make sure all configuration settings are settable. Thanks for pointing this out.

src/algorithms/tracking/TrackParamTruthInitConfig.h
```diff
@@ -15,6 +15,8 @@ struct TrackParamTruthInitConfig {
     double m_minMomentum      = 100 * Acts::UnitConstants::MeV;
     double m_maxEtaForward    = 4.0;
     double m_maxEtaBackward   = 4.1;
+    double m_momentumSplit    = 0.0;
```
osbornjd, line 18: Shouldn't we set these to be nonzero by default? Since we know that the kalman filter can be biased by too good of input seed parameters?

osbornjd (changes requested):
Looks fine to me, other than I think we should set the default smearing values to something nonzero so that the CKF by default does not use the truth track parameters. This will change the default behavior but I think it should be changed, unless the intent was to provide this functionality first to then test what the smearing value should be and then implement that as the default in a future PR.

src/algorithms/tracking/TrackParamTruthInitConfig.h
DraTeots: I believe this file should be modified. And used. That is exactly what I wrote above to @wdconinc

osbornjd: I think it is used, my suggestion was just to set the default behavior to use nonzero smearing

wdconinc: > the intent was to provide this functionality first to then test what the smearing value should be and then implement that as the default in a future PR

Yes :-)

> I think [this file] is used

Yes.

osbornjd:
Okay fine by me, I'll approve this once it passes the CI

wdconinc:
This still seems to work as before :-)

rec_eicrecon_reco_flags_rec_e_1GeV_20GeV_arches.tree.edm4eic.root:
image

osbornjd (approved):
Looks good to me

osbornjd:
We also need @DraTeots approval to merge

DraTeots (approved):

DraTeots:
Done

Simple-Shyam:
May I ask how do you we use these values with eicrcon as an input? Is there any example for that or we need to modify in the class?

Issue #401, opened 2022-12-18
Random segfaults in ~MCParticle


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): v0.3.6.2
- Any specific OS or system where the issue occurs? eic-shell 
- Any special versions of ROOT or Geant4? podio-0.15

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Second-most common production job failure mode related to EICrecon 

### Expected Result: (what do you expect when you execute the steps above)
Segfaults should not happen.

### Actual Result: (what do you get when you execute the steps above)
```
Thread model: pthreads
47974610482880:
       `- eicrecon (0x55725bfd0bf5)
        `- /lib/x86_64-linux-gnu/libc.so.6 (0x2ba1cd093f90)
         `- edm4hep::MCParticle::~MCParticle() (0x2ba1cf3e2048)
          `- edm4hep::CaloHitContributionObj::~CaloHitContributionObj() (0x2ba1cf3eb3e0)
           `- edm4hep::CaloHitContributionObj::~CaloHitContributionObj() (0x2ba1cf3eb409)
            `- edm4hep::CaloHitContributionCollectionData::clear(bool) (0x2ba1cf3ec9e3)
             `- edm4hep::CaloHitContributionCollection::clear() (0x2ba1cf3eb571)
              `- podio::EventStore::clear() (0x2ba1d151f942)
               `- JEventSourcePODIOsimple::GetEvent(std::shared_ptr<JEvent>) (0x2ba1d1405b3e)
                `- eicrecon (0x55725c00348c)
                 `- eicrecon (0x55725c00c08f)
                  `- eicrecon (0x55725c011209)
                   `- /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x2ba1cce134a3)
                    `- /lib/x86_64-linux-gnu/libc.so.6 (0x2ba1cd0e0fd4)
                     `- __clone (0x2ba1cd1608d0)
```
The segfaults are always in ~MCParticle, typically due to CaloHitContributions, but rarely due to other linked reference tables.

wdconinc:
Rerunning these failing jobs typically does not trigger the same segfault.

wdconinc:
Other instances are resulting in
```
Thread model: pthreads                                                                    47496643311296:
       `- eicrecon (0x55b0a4c73bf5)
[...]

PR #403, opened 2022-12-19
ci: valgrind memcheck on eicrecon

wdconinc:
Introduce CI job that runs `valgrind --tool=memcheck`. No automatic CI pipeline failure on memory leaks.

### TODO
- [x] change required jobs in repository settings since build jobs renamed with more matrix dimensions

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -107,7 +107,7 @@ jobs:
           export DETECTOR_CONFIG=${DETECTOR}_${{ matrix.detector_config }}
           export JANA_PLUGIN_PATH=$PWD/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
           chmod a+x bin/*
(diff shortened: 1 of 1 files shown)


c-dilks:
Looks good. It would be great if we could automatically parse the results, but I'm not aware of any (good) tools to read `valgrind` logs.

Does this PR really resolve #401?

For the bigger picture, applicable to other CI jobs, I have a few questions:
- Is the `ddsim` particle gun job sufficient for testing, or should we use a sample `hepmc` file? Perhaps this will be better addressed by benchmarks.
- Do we need to include the `dump_flags` and `janadot` plugins, or will the defaults suffice?
- Similarly, are we okay just testing with whatever the default `podio:output_include_collections` are, or should we try including something like `MCParticles`?

wdconinc:
> Does this PR really resolve https://github.com/eic/EICrecon/issues/401?

No. It is related, but I don't think it will close #401 because I didn't use the keywords 'fixes' or 'resolves'. I've disconnected it to make sure #401 is not closed.

> Is the ddsim particle gun job sufficient for testing, or should we use a sample hepmc file? Perhaps this will be better addressed by benchmarks.

The ddsim job is not even remotely sufficient for testing. We don't even look at the outputs. This still requires a human to look at the ROOT file. Preparing the ROOT file is all the CI job automates.

> Do we need to include the dump_flags and janadot plugins, or will the defaults suffice?

We run production with dump_flags and janadot plugins, so I want to know if they start leaking. Again, not like this CI test will catch it since it requires someone actually looking at the outputs. Preparing the valgrind log is all this CI job automates.

> Similarly, are we okay just testing with whatever the default podio:output_include_collections are, or should we try including something like MCParticles?

The philosophy behind eicrecon has been that it should do 'the sensible thing'/'what we want' in the default configuration (when just calling `eicrecon <input.root>`. If MCParticles needs to be added to the output, it probably should be done in the cpp code that steers the output.

c-dilks (approved):

Issue #404, opened 2022-12-24
ACTS tracking with B0 trackers


rahmans1:
**Is your feature request related to a problem? Please describe.**
When running reconstruction, getting the following error
```
[2022-12-24 12:05:38.657] [JEventProcessorPODIO] [debug] Factory producing type 'podio::EventStore' has an empty collection name, omitting.
[2022-12-24 12:05:38.658] [JEventProcessorPODIO] [debug] Persisting podio types from includes list
[2022-12-24 12:05:38.658] [JEventProcessorPODIO] [warning] Explicitly included collection 'B0TrackerRecHits' not present in JFactorySet, omitting.
```
**Describe the solution you'd like**
B0TrackerRecHits to be included in output collection and B0 hits be used in reconstruction

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rahmans1:
@DraTeots Following up from discussion in ACTS meeting. Here's the detailed steps I followed for B0 Reconstruction:

Clone epic
```
git clone https://github.com/eic/epic
cd epic
git checkout 161-acts-tracking-for-B0
```
Create a file inside the configurations folder called tracking.yml containing 
```
features:
  fields:
    marco:
  tracking:
    definitions:
    vertex_barrel:
    silicon_barrel:
    mpgd_barrel:
    mpgd_dirc:
    silicon_disks:
    support_service_assembly:
    tof_barrel:
    tof_endcap:
  far_forward:
    far_forward:
```

Create a steering file called proton.py containing
```

from DDSim.DD4hepSimulation import DD4hepSimulation
from g4units import mm, GeV, MeV, mrad
SIM = DD4hepSimulation()

SIM.enableGun = True
SIM.gun.thetaMin = 6*mrad
SIM.gun.thetaMax = 22*mrad
SIM.gun.momentumMin = 275*GeV
SIM.gun.momentumMax = 275*GeV
SIM.gun.distribution = 'cos(theta)'
SIM.gun.particle = 'proton'
SIM.outputFile = 'proton_275GeV_6mrad_22mrad_costheta.edm4hep.root'

```
Then build epic and set environment
```
cmake -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX=install -S . -B build
cmake --build build -j 8
cmake --install build
source install/setup.sh
export DETECTOR_CONFIG=epic_tracking
export BEAMLINE_CONFIG=epic_tracking
```

Run sim

```
[...]

PR #405, opened 2022-12-25
404 acts tracking with b0 trackers

rahmans1:
Add acts tracking for B0 trackers

- [ ] Bug fix (issue #__)
- [X] New feature (issue #404)
- [ ] Documentation update
- [ ] Other: __

N/A
Use B0 tracker hits in reconstruction


src/detectors/B0TRK/B0TRK.cc
```diff
@@ -1,3 +1,4 @@
+
```
wdconinc, line 1:
```suggestion
```

src/detectors/B0TRK/B0TRK.cc
```diff
@@ -26,15 +24,10 @@ void InitPlugin(JApplication *app) {
...
+    app->Add(new JChainFactoryGeneratorT<SiliconTrackerDigi_factory>({"B0TrackerHits"}, "B0TrackerRawHits", digi_default_cfg));
 
     // Convert raw digitized hits into hits with geometry info (ready for tracking)
-    TrackerHitReconstructionConfig hit_reco_cfg;
-    hit_reco_cfg.time_resolution = 8;
-    app->Add(new JChainFactoryGeneratorT<TrackerHitReconstruction_factory>(
-            {"B0TrackerHitRawHit"},     // Input data collection tags
-            "B0TrackerRecHits",         // Output data tag
-             hit_reco_cfg));            // Hit reco default config for factories
+    app->Add(new JChainFactoryGeneratorT<TrackerHitReconstruction_factory>({"B0TrackerRawHits"},"B0TrackerRecHits"));
```
wdconinc, line 30: I'd keep this config option even if not used now.

src/algorithms/tracking/TrackerSourceLinker.cc
```diff
@@ -80,19 +80,24 @@ eicrecon::TrackerSourceLinkerResult *eicrecon::TrackerSourceLinker::produce(std:
 
         Acts::Vector2 loc = Acts::Vector2::Zero();
         Acts::Vector2 pos;
+        auto hit_det = hit->getCellID()&0xFF;
+        auto onSurfaceTolerance = 0.0001;      // By default, ACTS uses 0.1 micron as the on surface tolerance
+        if (hit_det==150){
+         onSurfaceTolerance = 0.001;           // Ugly hack for testing B0. Should be a way to increase this tolerance in geometry.
```
wdconinc, line 86:
```suggestion
         onSurfaceTolerance = 0.001;           // FIXME Ugly hack for testing B0. Should be a way to increase this tolerance in geometry.
```

wdconinc: Also add units.

wdconinc (commented):
Looks fine to me.

src/detectors/B0TRK/B0TRK.cc
rahmans1: You mean to keep the hit_reco_cfg option in there with the time resolution?

rahmans1: Done

src/algorithms/tracking/TrackerSourceLinker.cc
```diff
@@ -80,19 +80,24 @@ eicrecon::TrackerSourceLinkerResult *eicrecon::TrackerSourceLinker::produce(std:
 
         Acts::Vector2 loc = Acts::Vector2::Zero();
         Acts::Vector2 pos;
+        auto hit_det = hit->getCellID()&0xFF;
+        auto onSurfaceTolerance = 0.1*Acts::UnitConstants::um;      // By default, ACTS uses 0.1 micron as the on surface tolerance
+        if (hit_det==150){
+         onSurfaceTolerance = 1*Acts::UnitConstants::um;           // FIXME Ugly hack for testing B0. Should be a way to increase this tolerance in geometry.
+        }
```
wdconinc, line 87: As discussed, I think this can be modified to use the `ActsGeometryProvider` instead of just the `dd4hep::rec::CellIDPositionConverter`. Then we can add a `dd4hep::rec::CellIDPositionConverter` hook in the `ActsGeometryProvider` but also use it to access the constants with something like
```suggestion
        if (hit_det==acts_geo->detector()->constant<int>("B0Tracker_Station_1_ID")){
         onSurfaceTolerance = 1*Acts::UnitConstants::um;           // FIXME Ugly hack for testing B0. Should be a way to increase this tolerance in geometry.
        }
```

rahmans1: I was thinking of using the [dd4hep geometry service](https://github.com/eic/EICrecon/blob/158e3e0364a95965b249b0ffac16534e8123c6ad/src/services/geometry/dd4hep/JDD4hep_service.cc#L35) because the ActsGeometryProvider doesn't seem to provide a hook to retrieve the DD4hep geo. Other places seem to use JServiceLocator or GetApplication() to grab the DD4hep geometry service but i am having a hard time  getting it to work (https://github.com/eic/EICrecon/blob/158e3e0364a95965b249b0ffac16534e8123c6ad/src/global/tracking/CKFTracking_factory.cc#L30).

wdconinc: Either way would be fine. I was just thinking `ActsGeometryProvider::dd4hepDetector()` ,https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/ActsGeometryProvider.h#L62 since that's already in theat class.

rahmans1: Never mind. You are right. I was looking in acts geo service instead of the acts geometry provider class and confused myself. I have applied the fix and seems to work I think. This is 1000 events with brycecanyon.
image

src/algorithms/tracking/TrackerSourceLinker.cc
```diff
@@ -80,19 +81,24 @@ eicrecon::TrackerSourceLinkerResult *eicrecon::TrackerSourceLinker::produce(std:
 
         Acts::Vector2 loc = Acts::Vector2::Zero();
         Acts::Vector2 pos;
+        auto hit_det = hit->getCellID()&0xFF;
+        auto onSurfaceTolerance = 0.1*Acts::UnitConstants::um;      // By default, ACTS uses 0.1 micron as the on surface tolerance
+        if (hit_det==dd4hepGeo->constant<int>("B0Tracker_Station_1_ID")){
```
wdconinc, line 86: Now I'd just move this lookup into initialization instead of doing this explicitly again on every hit.

wdconinc: I meant putting in the `init` method.

rahmans1: So, declare a variable for m_dd4hepGeo and m_b0detid? Something like that?

wdconinc: Yeah, for example. I'm worried that the lookup of the constant is expensive since it's a string comparison over all possible constants in the geometry, which are probably not stored in a structure that is optimized for these lookups (and maybe not even a single structure but distributed across detector elements).

rahmans1: Please check now

wdconinc (approved):
Looks good now.

Issue #410, opened 2022-12-28
Parameter 'HCAL:HcalBarrelMergedHits:input_tag' has conflicting defaults: 'HcalEndcapPInsertRecHits' vs 'HcalBarrelRecHits'


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? eic-shell
- Any special versions of ROOT or Geant4?

### Steps to reproduce: (give a step by step account of how to trigger the bug)
https://github.com/eic/EICrecon/actions/runs/3795680476/jobs/6455154937#step:5:580

### Expected Result: (what do you expect when you execute the steps above)
No warning.

### Actual Result: (what do you get when you execute the steps above)
```
[WARN] Parameter 'HCAL:HcalBarrelMergedHits:input_tag' has conflicting defaults: 'HcalEndcapPInsertRecHits' vs 'HcalBarrelRecHits'
```

PR #411, opened 2022-12-28
fix: use HcalEndcapPInsertMergedHits tag in factory

wdconinc:
Issue #410 is resolved by replacing the tag name with the correct name instead of defining it twice in different factories.


faustus123 (approved):
Looks good.

PR #416, opened 2022-12-29
fix: use `TDatabasePDG` as a singleton

c-dilks:
`TDatabasePDG` should be instantiated with its `Instance()` method, otherwise it will print a warning if you try to instantiate it twice with, e.g., `std::make_shared`.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
no


wdconinc:
https://github.com/root-project/root/issues/11580 Even with the singleton it may need a mutex. It probably should be a service to manage that.

c-dilks:
Oh no! Opened an issue so we can discuss how to proceed: https://github.com/eic/EICrecon/issues/417

DraTeots:
I guess this one should be closed in favor of a centralized service?

PR #420, opened 2022-12-30
fix: allow EICrecon to compile with fmt-9.1.0

wdconinc:
While fmt-8.1.1 seems to fall back to ostream operators for types without a formatter specialization, this does not seem to be thr case for fmt-9.1.0. This adds explicit ostream_formatters for the necessary types.

Ideally the template metaprogramming approach would be all that's needed, but it fails to work because I don't yet understand how the context should be specialized.

- [x] Bug fix (issue: compilation failure with fmt-9.1.0)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No


wdconinc:
@nathanwbrei You might enjoy the sfinae question here too.

DraTeots (approved):

Issue #440, opened 2023-01-16
Compilation error if EICrecon is used as cmake subproject


DraTeots:
jana_plugin.cmake using CMAKE_SOURCE_DIR which points to a root project thus making incorrect include directory

PR #441, opened 2023-01-16
Fix 440 EICRecon compilation errors if used as cmake subproject

DraTeots:
In Cmake not there is EICRECON_SOURCE_DIR  points to repository root which is used in jana_plugin.cmake and other places that need this particular place to be referenced

- [x] Bug fix (fixes issue #440)

no

no


wdconinc (commented):
I'm surprised there is a need for a new `EICRECON_SOURCE_DIR`. CMake provides so many `*_DIR` variables already that I doubt this is necessary...

veprbl:
https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html?highlight=project%20dir

DraTeots:
> I'm surprised there is a need for a new `EICRECON_SOURCE_DIR`. CMake provides so many `*_DIR` variables already that I doubt this is necessary...

Usually they point to root or current project. Not specific if EICRECON is in the middle of the chain. There is a named _SOURCE_DIR but I'll explain it further

> https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html?highlight=project%20dir

Indeed it is better to be renamed to EICRECON_ROOT_DIR or EICRECON_HOME_DIR (As things named in jana). From the user point or usage points (like specific plugin CMake) it is better to rely on a special variable rather than on some project name. First it shows the intent better (and the intent to get ROOT of EICRecon dir rather than some specific project EICRecon). And the same goes to reliability. Finally there are 2 eicrecon projects. One is the root project and the secon is executable eicrecon. So I think introducing EICRECON_ROOT_DIR is good from reliability, readability and unambiguousness points.

wdconinc:
Which project (besides your clion settings) is this expected to affect?

DraTeots:
Any project or setup that does ```add_subdirectory(EICRecon)``` will fail.

wdconinc (approved):
Since EICrecon will be reworked, let's just approve this and free up the queue for other issues and pull requests.

PR #444, opened 2023-01-19
fix: ensure sectorDist is compared with hit distance in right units.

wdconinc:
In CalorimeterIslandClustering, we compare the hit distances with the sector distance `sectorDist`. The EDM4hep units (hits) are mm = 1 while the dd4hep units (sectorDist parameter) are cm = 1. We need an explicit conversion.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ ] Changes have been communicated to collaborators @mariakzurek 

No.

Yes, some other calorimeter parameters would need changing if they are determined only on the basis of observed clusters.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterIslandCluster.h
+++ b/src/algorithms/calorimetry/CalorimeterIslandCluster.h
@@ -141,7 +141,7 @@ private:
           // different sector, local coordinates do not work, using global coordinates
         } else {
           // sector may have rotation (barrel), so z is included
-          return (edm4eic::magnitude(h1->getPosition() - h2->getPosition()) <= m_sectorDist / dd4hep::mm); # EDM4hep units are mm, so convert sectorDist to mm
+          return (edm4eic::magnitude(h1->getPosition() - h2->getPosition()) <= m_sectorDist / dd4hep::mm); // EDM4hep units are mm, so convert sectorDist to mm
         }
    }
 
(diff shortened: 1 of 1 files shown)


mariakzurek (approved):

wdconinc:
This will also affect (hopefully for the better) all detectors that use CalorimeterIslandClustering and that have set sectorDist based on physical sizes.

mariakzurek:
For the documentation purposes:

In the EICrecon flag dump file (-Pdump_flags), I see that `m_sectorDist=5.0 * dd4hep::cm` in https://github.com/eic/EICrecon/blob/main/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelScFiProtoClusters.h#L37 is translated to `["BEMC:ecalbarrelscfiprotoclusters:sectorDist", "5", "5", ""]` (dd4hep units: cm = 1). 

The sector distance is used here: https://github.com/eic/EICrecon/blob/main/src/algorithms/calorimetry/CalorimeterIslandCluster.h#L144 to compare to the hit position magnitude with is in edm4eic units of mm = 1.

Plots below are made with 1000 simulated photons at eta=0 (same simu file, diferent reconstruction)

- SciFi clusters w/o the fix (lots of low energy clusters, lots of clusters):

10GeV-photon_EICreco

- SciFi clusters w/ fix:

10GeV-photon_EICreco-withfix

- Jugler with all thesame paramaters(including sampling fraction, which is not exactly optimal):

10GeV-photon_Juggler

PR #446, opened 2023-01-21
feat(ci): run DIS test and produce outputs

wdconinc:
Running on single particles does not allow for doing smoke tests on things like truth vs reconstructed DIS. Therefore, run on DIS. These events are also more populated and more likely to show issues.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Right... matrices scale exponentially...

wdconinc:
And in a surprise to no one, EICrecon cannot handle equal signs in filenames since they are specified as key=value pairs...

wdconinc:
Strange, cannot reproduce the `std::length_error` locally... @nathanwbrei @faustus123 Any idea where `key`-`value` pair parsing can go wrong when passing a (quote) string with equal sign as the `value`?

wdconinc:
Turns out `=` is (probably) just fine when quoted, and I was fooled by another issue.

wdconinc:
Had a look at one of the DIS reconstruction artifacts for 5x41 and there was a peak at slightly less than 5 GeV in energy of the negative endcap clusters.

wdconinc:
Enabled auto-merge, but it will still require a change in the settings due to the name change of a required CI test job. Once this is approved, I'll change the settings and it will auto-merge.

PR #451, opened 2023-01-27
feat(ci): cache ddsim based on detector hash

wdconinc:
Instead of running a new simulation for every EICrecon change, this cache the ddsim outputs based on a detector hash (github sha for epic geometry). If there is ever an upgrade of the containers that makes the ddsim output binary incompatible for the same detector geometry hash, then we'll need to do a manual cache clear. This should speed up the pipelines.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
It is indeed reusing cached ddsim simulations 

Screenshot_20230127-183654

DraTeots (approved):

PR #459, opened 2023-02-03
fix: tree-wide pragma once

wdconinc:
This is less error-prone than ifndef/define include guards and (now) widely supported by all compilers we care about.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/utilities/eicrecon/print_info.h
```diff
@@ -49,4 +48,3 @@ void printJANAHeaderIMG() {
                  " YMMMM9 _dM_     _dMM_M_      \\M _dM_     _dMM_MMMMMMMM " << std::endl << std::endl;
 }
```
veprbl, line 50: Would be nice to not leave a trailing whitespace

veprbl (approved):
LGTM

src/utilities/eicrecon/print_info.h
wdconinc: Yeah, but my sed fu only goes so far and I was going to have pre-commit do that... I'll take a look later.

wdconinc: Resolved by #464.

PR #460, opened 2023-02-03
feat: add adjacency matrix in calorimeter clustering

wdconinc:
This introduces the adjacency matrix into calorimeter island clustering and is a port of https://github.com/eic/juggler/pull/2 into eicrecon. Copying the comments from original author @veprbl into this PR.

The issue with the current implementation is that `is_neighbour()` relies on geometrical dimensions of a calorimeter being used. This makes it cumbersome and error-prone to enable clustering for different detectors. The goal is to instead use readout cellID fields.

This PR suggests a following interface for doing that: A new parameter `adjacencyMatrix` is added to describe the adjacency of the sensitive elements. It can be used replace the existing settings like `localDistXY`. The value specified must be a string expression (to be evaluated by DD4hep evaluator) that returns a "boolean" value which is non-zero for adjacent cells. An example predicate for a calorimeter with a readout ID spec like `system:8,row:8,column:8` could look like

```python
cl = IslandCluster("cl",
  readoutClass="EcalBarrelSciGlassHits",
  adjacencyMatrix="(abs(row_1 - row_2) + abs(column_1 - column_2)) == 1",
  # ...
)
```

A calorimeter that has sectors might require a more sophisticated predicate like:
```
(
  abs(tower_1 - tower_2)
  + abs((sector_1 - sector2) * 5 + row_1 - row2)
) == 1
```

There are few minor issues with the current implementation:
1. The `5` is a magical constant for number of rows per sector. It is possible to scan the geometry and define some sort of `row_max` constant to be used in place.
2. Expressions are evaluated for every tower. This may be slow, but those values can be precomputed if needed.
3. The DD4hep evaluator is designed to work with doubles, there might be some floating points caveats. Should not affect us unless we get into large numbers?

- [ ] Bug fix (issue #__)
- [x] New feature (issue)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
+++ b/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
@@ -39,13 +39,13 @@ public:
         m_readout = "EcalBarrelSciGlassHits";
 
         // neighbour checking distances
-        m_sectorDist=5.0 * dd4hep::cm;             // from ATHENA reconstruction.py
+        m_sectorDist=0.0 * dd4hep::cm;             // not applicable
         u_localDistXY={};     //{this, "localDistXY", {}};
         u_localDistXZ={};     //{this, "localDistXZ", {}};
         u_localDistYZ={};     //{this, "localDistYZ", {}};
         u_globalDistRPhi={};  //{this, "globalDistRPhi", {}};
-        u_globalDistEtaPhi={};//{this, "globalDistEtaPhi", {}};
-        u_dimScaledLocalDistXY={1.8,1.8};// from ATHENA reconstruction.py
+        u_globalDistEtaPhi={}; //{this, "globalDistEtaPhi", {}};
+        u_dimScaledLocalDistXY={}; // not used
 
         app->SetDefaultParameter("BEMC:EcalBarrelSciGlassProtoClusters:input_tag", m_input_tag, "Name of input collection to use");
         app->SetDefaultParameter("BEMC:EcalBarrelSciGlassProtoClusters:splitCluster",             m_splitCluster);
--- a/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
+++ b/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
@@ -35,7 +35,10 @@ public:
 
         // adjacency matrix
         m_geoSvcName = "GeoSvc";
-        u_adjacencyMatrix = "(abs(tower_1 - tower_2) + abs((sector_1 - sector_2) * 5 + row_1 - row_2)) == 1";
+        // Magic constants:
+        //  24 - number of sectors
+        //  5  - number of towers per sector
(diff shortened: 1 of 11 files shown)


wdconinc:
Draft until CI succeeds (it wont).

src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
```diff
@@ -33,6 +33,11 @@ class ProtoCluster_factory_EcalBarrelSciGlassProtoClusters : public JFactoryT<ed
         m_minClusterHitEdep=1.0 * dd4hep::MeV;    // from ATHENA reconstruction.py
         m_minClusterCenterEdep=30.0 * dd4hep::MeV; // from ATHENA reconstruction.py
 
+        // adjacency matrix
+        m_geoSvcName = "GeoSvc";
+        u_adjacencyMatrix = "";
```
wdconinc, line 38: @veprbl Could you code-suggest what this should be and how I can see if it works here?

wdconinc:
Marked as ready and asked for review since I think this is functional but have no ability to test without the settings that should be used for the sciglass.

src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
veprbl:
```suggestion
        u_adjacencyMatrix = "(abs(row_1 - row_2) + abs(column_1 - column_2)) == 1";
```

veprbl: Oops, actually:
```suggestion
        u_adjacencyMatrix = "(abs(tower_1 - tower_2) + abs((sector_1 - sector2) * 5 + row_1 - row2)) == 1";
```

wdconinc:
@veprbl I'm thinking we can avoid the cost by doing a cached lookup and only recalculate if not already calculated before. That's probably more relevant than a full pre-calculation. But we can accept the performance hit here at first and keep that improvement for a future pr.

veprbl:
That is understood. Pre-calculation was also part of the idea that iterating over all of the geometry to figure out limits for the values.

veprbl:
How do we enable island clustering in JANA2 now? In Gaudi there was a steering from python.

wdconinc:
Isn't it already enabled? I thought it was. I'll look later, but I think this should be all that was necessary.

wdconinc:
https://github.com/eic/EICrecon/blob/main/src/detectors/BEMC/BEMC.cc#L37 is where the factory is included which will then produce the collections of clusters using island clustering.

wdconinc:
And https://github.com/eic/EICrecon/blob/main/src/services/io/podio/JEventProcessorPODIO.cc#L140 is where we tell Jana to require the sciglass clusters in the output, which means the factory will be asked to produce them.

veprbl:
clustering

wdconinc:
I just want to check if there were no regressions with calorimeter island clustering with the original method and then I'll merge.

wdconinc:
No regressions in island clustering with other options:

### before
image

### after
image

veprbl (approved):

PR #469, opened 2023-02-07
treewide: s/ClustersAssociations/ClusterAssociations/

veprbl:
Previously using plural:
 - EcalEndcapN
 - EcalEndcapP
 - EcalEndcapPInsert

Using singular:
 - B0ECal
 - EcalBarrelSciGlass
 - EcalBarrelSciImaging
 - HcalBarrel
 - HcalEncapN
 - ZDCEcal

Previously using both:
 - HcalEndcapP

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @FriederikeBock @rymilton @johnny8266 @jizhongling

Yes. Change name of the branch they'll be reading

No


wdconinc:
Hesitant to merge since breaking interface change.

wdconinc:
> Hesitant to merge since breaking interface change.

Hesitation less in the morning.

wdconinc (approved):

veprbl:
> Hesitant to merge since breaking interface change.

I don't think this one is such a big deal. However it would be nice to have some idea about how to provide backward-compatibility aliases for such changes in the future. Perhaps it would be a matter of adding some alias factory?

wdconinc:
Podio has support of reference collections: essentially it adds a reference to the objects in another collection. This is in particular useful for e.g. ReconstructedElectrons as a reference collection based on ReconstructedParticles, etc. Unfortunately this doesn't work in JANA2 because it has a different memory ownership model and therefore can't handle the references. Should be addressed by future changes, but unclear when those are coming.

PR #471, opened 2023-02-08
feat(ci): trigger eicweb container build for benchmarks

wdconinc:
After some initial succesful running of eicrecon, we pass on to building a full eicweb container and subsequent benchmarks.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No. Won't make this test required (for now) so shouldn't slow down the pipelines until completion.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -781,6 +781,33 @@ jobs:
           python3 ${{ github.workspace }}/epic-capybara/bara.py ref/rec_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4eic.root rec_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4eic.root
           echo "::endgroup::"
 
+  trigger-container:
+    runs-on: ubuntu-latest
+    needs:
+    - eicrecon-gun
+    steps:
+    - uses: eic/trigger-gitlab-ci@v2
+      id: trigger
+      with:
+        url: https://eicweb.phy.anl.gov
+        project_id: 290
+        token: ${{ secrets.EICWEB_CONTAINER_TRIGGER }}
+        ref_name: master
+        variables: |
+          GITHUB_REPOSITORY=${{ github.repository }}
+          GITHUB_SHA=${{ github.event.pull_request.head.sha || github.sha }}
+    - run: |
+        gh api \
+           --method POST \
+          -H "Accept: application/vnd.github+json" \
+          /repos/${{ github.repository }}/statuses/${{ github.event.pull_request.head.sha || github.sha }} \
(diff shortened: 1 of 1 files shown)


wdconinc:
Well, that worked better than I thought it was going to: container build triggered, but with standard eicrecon.

wdconinc:
Triggered correctly at https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/2018552#L278

wdconinc:
Requires https://eicweb.phy.anl.gov/containers/eic_container/-/merge_requests/735

wdconinc:
Seems to be working: eic_container triggered, which triggered detector benchmarks, which triggered both reconstruction and physics benchmarks. All are running so no final success or failure yet.

Screenshot_20231009-223341

.github/workflows/linux-eic-shell.yml
```diff
@@ -781,6 +781,36 @@ jobs:
...
+      with:
+        url: https://eicweb.phy.anl.gov
+        project_id: 290
+        token: ${{ secrets.EICWEB_CONTAINER_TRIGGER }}
+        ref_name: master
+        variables: |
+          GITHUB_REPOSITORY=${{ github.repository }}
+          GITHUB_SHA=${{ github.event.pull_request.head.sha || github.sha }}
+          GITHUB_PR=${{ github.event.pull_request.number }}
+          VERSION="eicrecon-${GITHUB_SHA::8}"
```
wdconinc, line 800: Doesn't seem like we can use variables that we are still setting in the same block...
```suggestion
          VERSION="eicrecon-${{ github.event.pull_request.head.sha || github.sha }}"
```

wdconinc:
@veprbl This now works, and the last CI run through was successful through running the pipelines after building the temporary container, with the exception that it pushed that temporary container to a docker hub eicweb/jug_xl:eicrecon--default where there should have been a hash between the two dashes... i.e. https://hub.docker.com/layers/eicweb/jug_dev/eicrecon--default/images/sha256-7446018a5c52b4b1982ced5ec437eb240862988334a59037f257c45a7e7051ce?context=explore

veprbl (approved):
Looks great!

veprbl:
I'm wondering, if we'd need to propagate GitHub PR info from container pipelines to downstream pipelines somehow.

wdconinc:
That used to be the case and it used to work. But changes in the way the benchmarks were reconfigured (https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/merge_requests/150) introduced circular dependencies that broke propagation (https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/merge_requests/154, https://eicweb.phy.anl.gov/containers/eic_container/-/merge_requests/736). So temporarily we don't have direct status callbacks from downstream benchmarks and only get the container status callback after the benchmark pipelines complete. I think that is probably ok for now.

PR #472, opened 2023-02-09
fix: remove spaces in adjacency matrix expression

wdconinc:
When there are spaces in the adjacency matrix string expression's default value, JANA2 chops at the first space. This results in many warnings, i.e. https://github.com/eic/EICrecon/actions/runs/4121127389/jobs/7116665443.

The issue seems to be in the `SetDefaultParameter` call, since modifying `u_adjacencyMatrix` in the algorithm didn't have any impact. At that point the string has already been chopped.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
+++ b/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
@@ -38,6 +38,7 @@ public:
         //  24 - number of sectors
         //  5  - number of towers per sector
         u_adjacencyMatrix = "(abs(tower_1 - tower_2) + (abs((sector_1 - sector_2) * 5 + row_1 - row_2) == 1) + (abs((sector_1 - sector_2) * 5 + row_1 - row_2) == (24 * 5 - 1))) == 1";
+        std::remove(u_adjacencyMatrix.begin(), u_adjacencyMatrix.end(), ' ');
         m_readout = "EcalBarrelSciGlassHits";
 
         // neighbour checking distances


wdconinc:
@faustus123 @nathanwbrei I am very surprised that JANA2 changes default values set in the C++ in a SetDefaultParameter call. Can't we even trust default values anymore?

```console
[2023-02-08 21:29:43.387] [EcalBarrelSciGlassProtoClusters] [error] (abs(tower_1 - tower_2) + (abs((sector_1 - sector_2) * 5 + row_1 - row_2) == 1) + (abs((sector_1 - sector_2) * 5 + row_1 - row_2) == (24 * 5 - 1))) == 1
[WARN] Parameter 'BEMC:EcalBarrelSciGlassProtoClusters:adjacencyMatrix' loses equality with itself after stringification
[2023-02-08 21:29:43.387] [EcalBarrelSciGlassProtoClusters] [error] (abs(tower_1
```

faustus123:
> @faustus123 @nathanwbrei I am very surprised that JANA2 changes default values set in the C++ in a SetDefaultParameter call. Can't we even trust default values anymore?

Ha! You're right. This is a bug where we need to handle strings differently than other data types. Spaces in strings are not being handled properly.

The issue is due to parameters being passed through `std::stringstream` to convert them into string form first and then back into the native type of the parameter (e.g. `double`). This is actually done on purpose so one can run a process using default values from the C++ code and have it dump the full configuration to a file. That configuration file should then be able to reproduce the job exactly. If there are round-off errors in the stringification, those will need to be included in the original run.

We will need to add a template specialization for strings types that skips the conversion using `std::stringstream`. We'll take care of that on the JANA2 end.

wdconinc:
I'd like to merge this soon. Anyone can review and approve?

- Before: https://github.com/eic/EICrecon/actions/runs/4121127389/jobs/7116665443
- After: https://github.com/eic/EICrecon/actions/runs/4130669007/jobs/7137702389

faustus123 (approved):
This looks like a reasonable and simple workaround.

AyanRoy16:
Hello @wdconinc , 
We have added a test case for JParameter to show that a c++ command that is parsed as a string remains the same after it is parsed, the white spaces are correctly parsed. Let me know if we are missing out on something. 
PR: https://github.com/JeffersonLab/JANA2/pull/248
cc. @faustus123 @nathanwbrei

PR #475, opened 2023-02-09
Removing momentum split part.

Simple-Shyam:

Removing momentum split part from the true momentum smearing as discussed.
- [x ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

It will be the correct way to smear the momentum as Gaussian.

### DIFF
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -73,10 +73,7 @@ eicrecon::TrackParameters *eicrecon::TrackParamTruthInit::produce(const edm4hep:
     }
 
     // modify initial momentum to avoid bleeding truth to results when fit fails
-    // this picks uniformly between [1-eps,1,1+eps] times true momentum, then smeared
-    const auto pinit = pmag
-                     * (1.0 + m_cfg.m_momentumSplit * m_uniformIntDist(generator))
-                     * (1.0 + m_cfg.m_momentumSmear * m_normDist(generator));
+    const auto pinit = pmag*(1.0 + m_cfg.m_momentumSmear * m_normDist(generator));
 
     // build some track cov matrix
     Acts::BoundSymMatrix cov                    = Acts::BoundSymMatrix::Zero();


osbornjd (dismissed):
I think this is probably the best strategy as the real seeds will be smeared by some amount around the actual truth value, not doubly smeared in some way.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -73,10 +73,7 @@ eicrecon::TrackParameters *eicrecon::TrackParamTruthInit::produce(const edm4hep:
     }
 
     // modify initial momentum to avoid bleeding truth to results when fit fails
-    // this picks uniformly between [1-eps,1,1+eps] times true momentum, then smeared
-    const auto pinit = pmag
-                     * (1.0 + m_cfg.m_momentumSplit * m_uniformIntDist(generator))
-                     * (1.0 + m_cfg.m_momentumSmear * m_normDist(generator));
+    const auto pinit = pmag*(1.0 + m_cfg.m_momentumSmear * m_normDist(generator));
```
wdconinc, line 76: This PR should also remove the `m_momentumSplit` parameter since it is now unused.

wdconinc: (from the configuration, that is)

Simple-Shyam: Yes @wdconinc you are right. Sorry, I selected the option as a bug but  I changed  it to an other option. It is kind of small changes we want it as discussed on email.

wdconinc: Any progress on removing the `m_momentumSplit` parameter?

osbornjd:
It looks like this now combines the work from https://github.com/eic/EICrecon/pull/476 into this PR. I suggest closing 476 and changing the description of this PR to encompass all of the changes (e.g. removing the momentum split and adding in the 10% smearing by default)

src/algorithms/tracking/TrackParamTruthInit.cc
Simple-Shyam: Thanks, Yes I removed it from the configurable and also from the global configurable. I also changed default momentum smearing to 10%.

Simple-Shyam:
Now all checks are passed successfully.

osbornjd (approved):
Looks good to me, if Wouter agrees he can merge

PR #489, opened 2023-02-16
fix(ci): keep .git and .ccache out of artifacts

wdconinc:
I'd been wondering why the artifacts kept increasing in size... Default directory even after checkout is in ~ which is also where .ccache lives.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
Looks fine

PR #490, opened 2023-02-16
ci: avoid duplicating entries in ccache.conf

veprbl:
It would seem like the first command should start an empty file

Found a very minor defect while staring at diff of #489

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc (approved):
Thanks. Maybe most of our ccache is actually lines in the conf file by now ;-)

PR #496, opened 2023-02-16
fix: disable momentum in reco/truth particle matching

### LINKED ISSUE #450: ReconstructedChargedParticlesAssociations with too-often missing associations
This was found in the EICrecon output from the Oct simulation campaign.

1. git clone https://github.com/KongTu/EICreconOutputReader.git
2. cd EICreconOutputReader
3. ./runTestAssociation.sh input/input-eicrecon-DIS_18x275_Q2-1_0001.root eicrecon-DIS_18x275_Q2-1_0001

I expect that for reconstructed scattered electron, it should almost always have an association to the MC. 

the printout will show events that a reconstructed level scattered electron is found, no association can be found, meaning the recID is not in the ReconstructedChargedParticlesAssociations. The issue here seems to me that it happens too often.

wdconinc:
This effectively disables the momentum in reconstructed vs truth particle matching, and increases the phi tolerance from 0.03 to 0.1 rad. Closes #450. 

- [x] Bug fix (issue #450)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @bspage912 

No.

Yes, less strict matching.


wdconinc:
@bspage912 is this something you can easily test?

bspage912:
Hey Wouter, in principle this should be easy to test, just run one MC file through EICrecon with the new algo. In practice, it may take a bit as I will have to go back and remind myself how to run things. This change is so simple, I can't imagine any unexpected behaviour ...

wdconinc:
@bspage912 Can you approve this here, just based on the changes and not testing if you agree with the changes? I can then tag a release and submit some jobs for more detailed testing.

bspage912 (approved):
While I have not explicitly tested this branch, I did emulate the matching algorithm in my offline tests without the relative momentum condition and with the phi tolerance changed to 0.1 and saw improved behaviour. This leads me to believe the changes in this branch should be fine to merge in.

PR #506, opened 2023-02-23
BUGFIX: ProtoCluster_factory_EcalBarrelSciGlassProtoClusters: fix adjacencyMatrix

veprbl:
A critical bug is present in the original expression:

`(abs(dx) + (abs(dy) == 1) + (abs(dy) = 1234)) == 1`

Allows for arbitrary `dy` in range `[2, 1233]` as long as `abs(dx) == 1`. That caused merging of the clusters in across the azimuthal direction.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
+++ b/src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
@@ -37,8 +37,17 @@ public:
         // Magic constants:
         //  24 - number of sectors
         //  5  - number of towers per sector
-        u_adjacencyMatrix = "(abs(tower_1 - tower_2) + min(abs((sector_1 - sector_2) * 5 + row_1 - row_2), 24 * 5 - abs((sector_1 - sector_2) * 5 + row_1 - row_2))) == 1";
-        std::remove(u_adjacencyMatrix.begin(), u_adjacencyMatrix.end(), ' ');
+        u_adjacencyMatrix = "
+        (
+          abs(tower_1 - tower_2)
+        + min(
+              abs((sector_1 - sector_2) * 5 + row_1 - row_2),
+              24 * 5 - abs((sector_1 - sector_2) * 5 + row_1 - row_2)
+          )
+        ) == 1";
+        u_adjacencyMatrix.erase(
+          std::remove_if(u_adjacencyMatrix.begin(), u_adjacencyMatrix.end(), ::isspace),
+          u_adjacencyMatrix.end());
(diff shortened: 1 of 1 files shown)


src/detectors/BEMC/ProtoCluster_factory_EcalBarrelSciGlassProtoClusters.h
```diff
@@ -37,7 +37,7 @@ class ProtoCluster_factory_EcalBarrelSciGlassProtoClusters : public JFactoryT<ed
         // Magic constants:
         //  24 - number of sectors
         //  5  - number of towers per sector
-        u_adjacencyMatrix = "(abs(tower_1 - tower_2) + (abs((sector_1 - sector_2) * 5 + row_1 - row_2) == 1) + (abs((sector_1 - sector_2) * 5 + row_1 - row_2) == (24 * 5 - 1))) == 1";
+        u_adjacencyMatrix = "(abs(tower_1 - tower_2) + min(abs((sector_1 - sector_2) * 5 + row_1 - row_2), 24 * 5 - abs((sector_1 - sector_2) * 5 + row_1 - row_2))) == 1";
         std::remove(u_adjacencyMatrix.begin(), u_adjacencyMatrix.end(), ' ');
```
wdconinc, line 41: For my own mental parser, this should strip out newlines too...
```suggestion
        u_adjacencyMatrix = "
        (
          abs(tower_1 - tower_2)
        + min(
              abs((sector_1 - sector_2) * 5 + row_1 - row_2),
              24 * 5 - abs((sector_1 - sector_2) * 5 + row_1 - row_2)
          )
        ) == 1";
        u_adjacencyMatrix.erase(
          std::remove_if(u_adjacencyMatrix.begin(), u_adjacencyMatrix.end(), ::isspace),
          u_adjacencyMatrix.end());
```

Is that the same as:
```
        u_adjacencyMatrix = "
        ( 
          abs(tower_1 - tower_2)
        + abs(((sector_1 - sector_2) * 5 + row_1 - row_2) % (5*24))
        ) == 1";
```

veprbl: No, it's not. The idea of this is that for each two hits, a cylinder has two geodesic curves: one wraps around without crossing from tower 119 to 0 and one that does. We are just picking a shortest distance.

veprbl: Basically it's a taxicab metric on a cylinder.

wdconinc (approved):

PR #393, opened 2022-12-11
feat!: add IRT PID to DRICH plugin

c-dilks:
### Briefly, what does this PR introduce?
Add Indirect Ray Tracing (IRT) particle identification to the `DRICH` plugin

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No

### Does this PR change default behavior?
Yes, dRICH PID output collection will be enabled by default

# Factorization Checklist
This is a very large PR, and therefore is being factorized into smaller, contained PRs for review purposes. The following checklist tracks all of the proposed changes. Boxes are checked when the PR is opened.

- DRICH plugin
  - [x] src/detectors/DRICH/DRICH.cc: common file, updated with each relevant PR below
  - [x] src/detectors/DRICH/README.md: #708
- algorithms and factories
  - digitization:
    - [x] https://github.com/eic/EICrecon/pull/599
    - [x] https://github.com/eic/EICrecon/pull/555
  - track projection: 
    - [x] https://github.com/eic/EICrecon/pull/656
    - [x] https://github.com/eic/EICrecon/pull/679
    - [x] https://github.com/eic/EICrecon/pull/619
    - [x] https://github.com/eic/EICrecon/pull/682
    - [x] #726
  - PID:
    - IRT:
      - [x] #707
    - Merging
      - [x] #730
    - Linking
      - [x] #829
        - plugins
          - [x] src/global/pid/pid.cc
          - [x] src/global/tracking/tracking.cc
        - algorithms
          - [x] src/algorithms/pid/ParticlesWithPID.cc
          - [x] src/algorithms/pid/ParticlesWithPIDConfig.h
          - [x] src/algorithms/pid/ParticlesWithPID.h
          - [x] src/algorithms/pid/ConvertParticleID.h
        - factories
          - [x] src/global/pid/ParticlesWithPID_factory.cc
          - [x] src/global/pid/ParticlesWithPID_factory.h
- benchmarks:
  - [x] https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/merge_requests/293
- `richgeo` service:
  - [x] https://github.com/eic/EICrecon/pull/608
  - [x] #727

PR #508, opened 2023-02-24
refactor: replace `RawPMTHit` with `RawTrackerHit`

c-dilks:
`edm4eic::RawPMTHit` is basically a duplicate `datatype` of `edm4eic::RawTrackerHit`, so let's use `RawTrackerHit` and deprecate `RawPMTHit`.

See also https://github.com/eic/EDM4eic/pull/22

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: datatype change

No (except in #393)

No


c-dilks:
CI failures are unrelated to this PR (and currently happening on `main`).

wdconinc:
Why is it segfaulting now? Just noticed in main too ...

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -131,13 +131,13 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
...
+        std::vector<edm4eic::RawTrackerHit*> raw_hits;
         for (auto &it : hit_groups) {
             for (auto &data : it.second) {
-                edm4eic::RawPMTHit* hit = new edm4eic::RawPMTHit{
+                edm4eic::RawTrackerHit* hit = new edm4eic::RawTrackerHit{
                   it.first,
-                  static_cast<uint32_t>(data.signal),
-                  static_cast<uint32_t>(data.time/m_cfg.timeStep)
+                  static_cast<int32_t>(data.signal),
+                  static_cast<int32_t>(data.time/m_cfg.timeStep)
```
wdconinc, line 140: :-/ this should probably be `uint32_t` in EDM4eic (tag @sly2j). It doesn't make sense for these data words to be signed.

It will just cause issues. If you want to be resilient (and explicitly screw up at the boundaries):
```suggestion
                  static_cast<decltype(hit.charge)>(data.signal),
                  static_cast<decltype(hit.timeStamp)>(data.time/m_cfg.timeStep)
```
(untested)

wdconinc (dismissed):
Looks good. Minor suggestion for future resiliency.

wdconinc:
Set to auto-merge, pending unresolved conversation.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
wdconinc: Maybe clearer if:
```suggestion
                  static_cast<decltype(edm4eic::RawTrackerHit::charge)>(data.signal),
                  static_cast<decltype(edm4eic::RawTrackerHit::timeStamp)>(data.time/m_cfg.timeStep)
```

c-dilks: Definitely, actually this is what I'm doing in #393 already, but I will do it here now.

c-dilks:
I also applied these changes to the [Juggler MR](https://eicweb.phy.anl.gov/EIC/juggler/-/merge_requests/509)

wdconinc (approved):

c-dilks:
Same exact CI failures are happening on unrelated PR https://github.com/eic/EICrecon/pull/524, so it seems like this is a general problem.

c-dilks:
Oddly enough, the last push to `main` ran fine...

wdconinc:
Can you restart a job on main? Might be due to jana upgrade...

c-dilks:
Same failures

wdconinc:
And it's one of those 'fun' failures that doesn't fail inside valgrind...

Issue #495, opened 2023-02-16
Reconstruction failures are not reported in exit code


veprbl:
### Environment: (where does this bug occur, have you tried other environments)
This is on CI

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Make a change that breaks reconstruction: https://github.com/eic/EICrecon/actions/runs/4170791240/jobs/7220213032

### Expected Result: (what do you expect when you execute the steps above)

Red checkmark

### Actual Result: (what do you get when you execute the steps above)

A green checkmark on a job that produced a single event instead of a thousand.
https://github.com/eic/EICrecon/actions/runs/4170791240/jobs/7220213032

```output
+--------------------------+------------+--------+-----+---------+-------+--------+---------+-------------+
  |           Name           |   Status   |  Type  | Par | Threads | Chunk | Thresh | Pending |  Completed  |
  +--------------------------+------------+--------+-----+---------+-------+--------+---------+-------------+
  | sim_dis_10x100_minQ2=1000_arches.edm4hep.root | Finished   | Source |  F  |       0 |    40 |      - |       - |           1 |
  | processors               | Finished   | Sink   |  T  |       0 |     1 |     80 |       0 |           1 |
  +--------------------------+------------+--------+-----+---------+-------+--------+---------+-------------+
  +--------------------------+-------------+--------------+----------------+--------------+----------------+
  |           Name           | Avg latency | Inst latency | Queue latency  | Queue visits | Queue overhead | 
  |                          | [ms/event]  |  [ms/event]  |   [ms/visit]   |    [count]   |     [0..1]     | 
  +--------------------------+-------------+--------------+----------------+--------------+----------------+
  | sim_dis_10x100_minQ2=1000_arches.edm4hep.root |          58 |           58 |         0.0006 |            1 |       1.04e-05 |
  | processors               |    1.08e+04 |     1.08e+04 |         0.0066 |            1 |        6.1e-07 |
  +--------------------------+-------------+--------------+----------------+--------------+----------------+
  +----+----------------------+-------------+------------+-----------+----------------+------------------+
  | ID | Last arrow name      | Useful time | Retry time | Idle time | Scheduler time | Scheduler visits |
  |    |                      |     [ms]    |    [ms]    |    [ms]   |      [ms]      |     [count]      |
  +----+----------------------+-------------+------------+-----------+----------------+------------------+
  |  0 | idle                 |    1.08e+04 |      0.002 |         0 |         0.0008 |                2 |
[...]

Issue #340, opened 2022-11-13
Non-unique type_info for `Jug::BField::DD4hepBField`


veprbl:
### Environment: (where does this bug occur, have you tried other environments)
- 0.3.5
- 0.3.5
- acts 19.6.0
- macOS
- Any special versions of ROOT or Geant4?

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. `run_eicrecon_reco_flags.py sim.edm4hep.root foo`
2.

### Expected Result: (what do you expect when you execute the steps above)
Reconstruction runs

### Actual Result: (what do you get when you execute the steps above)
```output
Thread model: pthreads
0x70000724b000: 
       `- _sigtramp (0x7fff5f37bb5d)
        `- 3   ???                                 0x15e710ab88f70018 0x0 + 1578248523339989016 (0x15e710ab88f70018)
         `- std::__1::vector<Acts::Result<Acts::CombinatorialKalmanFilterResult, std::__1::error_code>, std::__1::allocator<Acts::Result<Acts::CombinatorialKalmanFilterResult, std::__1::error_code> > > Acts::CombinatorialKalmanFilter<Acts::Propagator<Acts::EigenStepper<Acts::StepperExtensionList<Acts::detail::GenericDefaultExtension<double> >, Acts::detail::VoidAuctioneer>, Acts::Navigator> >::findTracks<boost::container::vec_iterator<std::__1::reference_wrapper<Jug::IndexSourceLink const>*, true>, std::__1::vector<Acts::SingleBoundTrackParameters<Acts::SinglyCharged>, std::__1::allocator<Acts::SingleBoundTrackParameters<Acts::SinglyCharged> > >, Acts::SingleBoundTrackParameters<Acts::SinglyCharged> >(std::__1::vector<Acts::SingleBoundTrackParameters<Acts::SinglyCharged>, std::__1::allocator<Acts::SingleBoundTrackParameters<Acts::SinglyCharged> > > const&, Acts::CombinatorialKalmanFilterOptions<boost::container::vec_iterator<std::__1::reference_wrapper<Jug::IndexSourceLink const>*, true> > const&) const (0x10b398f0c)
[...]

PR #511, opened 2023-02-25
cmake: enable __GXX_MERGED_TYPEINFO_NAMES globally

veprbl:


veprbl:
This was supposed to be a reproducer for #340, but it fails even earlier than that!

Also, notice how it throws an exception but exits successfully https://github.com/eic/EICrecon/actions/runs/4268038688/jobs/7430338286, which is another case of #495

wdconinc:
This will also fix some of the sanitizer warnings on duplicate vtable entries, I suspect.

veprbl:
The duplicate vtables will require #518, but also moving all virtual destructors from headers to their respective translation units.

faustus123:
I have to say that I am not familiar with `__GXX_MERGED_TYPEINFO_NAMES` and had to look it up. A lot of the web pages were pretty dense so I went to chatGPT to summarize it for me and asked why it isn't turned on by default. The response is below. I can say that internally JANA does have some dependence on typeinfo. I will say that this makes me a little nervous that we are planting a bomb that will express as a subtle bug later. Do we have some numbers on the benefit to binary sizes that support this?


> The reason why __GXX_MERGED_TYPEINFO_NAMES is not enabled by default in C++ compilers is because it can potentially cause issues with certain codebases and lead to subtle bugs.
> 
> The macro changes the behavior of the compiler by merging typeinfo names, which can result in smaller binary sizes and faster program execution. However, this optimization is not always safe, especially in large codebases with complex class hierarchies and multiple instantiations of templates.
> 
> Enabling __GXX_MERGED_TYPEINFO_NAMES can potentially cause issues with RTTI (Run-Time Type Identification) and other features that rely on type information, leading to unexpected behavior at runtime. Additionally, merging typeinfo names can make debugging more difficult, as it can be harder to identify the source of a problem when using a debugger.
> 
> Therefore, enabling __GXX_MERGED_TYPEINFO_NAMES should be done with caution and only after careful testing to ensure that it does not introduce any bugs or unexpected behavior in your codebase. As with any advanced compiler option, it is important to thoroughly understand its behavior and potential risks before using it.
>

nathanwbrei:
JANA indexes JFactories using std::type_index, but if that lookup fails, it (manually) uses the demangled typename string as a fallback. This functionality exists because RTTI+dlopen was horribly broken on some platform before my time. This means that JANA is tolerant to errors where one type maps to two or more different typeinfos. JANA will break if different types map to the same typeinfo.

std::type_index appears to be a trivial wrapper around typeinfo that provides a hash code (which hashes the mangled typename string) and the comparison operators (which call strcmp on the mangled typename string). If we enable __GXX_MERGED_TYPEINFO_NAMES, both of these will return the address of the typeinfo cstring because they assume it is now unique. I have no idea when or how this merge happens in the context of dlopen. 

Take a look at this code in typeinfo.h:
```c++
  #if !__GXX_MERGED_TYPEINFO_NAMES
    // Even with the new abi, on systems that support dlopen
    // we can run into cases where type_info names aren't merged,
    // so we still need to do string comparison.
    bool before(const type_info& __arg) const _GLIBCXX_NOEXCEPT
    { return (__name[0] == '*' && __arg.__name[0] == '*')
	? __name < __arg.__name
	: __builtin_strcmp (__name, __arg.__name) < 0; }

    bool operator==(const type_info& __arg) const _GLIBCXX_NOEXCEPT
    {
      return ((__name == __arg.__name)
	      || (__name[0] != '*' &&
		  __builtin_strcmp (__name, __arg.__name) == 0));
    }
  #else
    // On some targets we can rely on type_info's NTBS being unique,
    // and therefore address comparisons are sufficient.
    bool before(const type_info& __arg) const _GLIBCXX_NOEXCEPT
    { return __name < __arg.__name; }

    bool operator==(const type_info& __arg) const _GLIBCXX_NOEXCEPT
    { return __name == __arg.__name; }
  #endif
```

The phrase 'on some targets' worries me. However (and this may be naive) I doubt that it would merge two unrelated types, and is more likely to fail to merge two identical types. In which case JANA should be fine. It is extra super important to test this thoroughly, though.

Issue #483, opened 2023-02-14
Symbol not found: vtable for rich::IrtGeoDRICH


veprbl:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): v0.6.0
- Any specific OS or system where the issue occurs? macOS
- Any special versions of ROOT or Geant4? no

### Steps to reproduce: (give a step by step account of how to trigger the bug)
Compile on macOS with `LDFLAGS=-Wl,-undefined,dynamic_lookup`, run

### Expected Result: (what do you expect when you execute the steps above)
Plugin can load

### Actual Result: (what do you get when you execute the steps above)

```
    ./log.so  =>  File not found
    /tmp/eicrecon/lib/EICrecon/plugins//log.so  =>  Loaded successfully
    ./dd4hep.so  =>  File not found
    /tmp/eicrecon/lib/EICrecon/plugins//dd4hep.so  =>  Loaded successfully
    ./acts.so  =>  File not found
    /tmp/eicrecon/lib/EICrecon/plugins//acts.so  =>  Loaded successfully
    ./richgeo.so  =>  File not found
    /tmp/eicrecon/lib/EICrecon/plugins//richgeo.so  =>  Loaded successfully
    ./rich.so  =>  File not found
    /tmp/eicrecon/lib/EICrecon/plugins//rich.so  =>  Loading failure: dlopen(/tmp/eicrecon/lib/EICrecon/plugins//rich.so, 137): Symbol not found: __ZTVN4rich11IrtGeoDRICHE
  Referenced from: /tmp/eicrecon/lib/EICrecon/plugins//rich.so
  Expected in: flat namespace
```

veprbl:
The issue here is due to the default linker mode on macOS using the Two-level namespaces. For that, each symbol is identified not only by its mangled name, but also the associated library. The symbol itself is exported from richgeo.so:
```
% nm /tmp/eicrecon/lib/EICrecon/plugins//richgeo.so | grep __ZTVN4rich11IrtGeoDRICHE       (git)-[pr/root_macros_mv]-% 
00000000000175b8 S __ZTVN4rich11IrtGeoDRICHE
```
But the rich.so plugin doesn't know which library name it is, so it looks things up in the flat namespace instead. I haven't figured out how to compile EICrecon entirely with a flat namespace only.

veprbl:
[...]

PR #513, opened 2023-02-25
Merge geometry/rich/richgeo into geometry/rich

veprbl:

Remove standalone library that is not an actual JANA2 plugin, but pretends to be one.

Resolves: #483

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


c-dilks:
Thanks! Yes, this should be a single plugin.

Let's rename it to `richgeo` so it's not confused with a detector plugin. The README should also be updated (I can take care of these if you prefer).

Do we have to flatten it to one directory? It's not necessary, but I prefer to have a clearer separation between the JANA/EICrecon-dependent code and the JANA/EICrecon-independent code.

c-dilks:
> Let's rename it to `richgeo` so it's not confused with a detector plugin. The README should also be updated (I can take care of these if you prefer).

Done and done.

veprbl:
@c-dilks If you think this is good, please approve

c-dilks:
> I should also test this with https://github.com/eic/EICrecon/pull/393.

I can no longer do this thanks to new upstream issues, so I'll deal with that later after this PR is merged.

c-dilks (approved):

c-dilks:
(need another reviewer since I pushed to this branch)

veprbl:
> > Do we have to flatten it to one directory?
> 
> Awaiting feedback on this.

You could set sources manually instead of calling `plugin_glob_all`. I just don't see the point. The separation should be already apparent from the filenames.

PR #514, opened 2023-02-28
feat: add sanitizers in CMake; remove valgrind from CI

wdconinc:
Introduce address and thread sanitizers to aid in debugging.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl:
Seems like there is incorrect indexing of `cdim` in

https://github.com/eic/EICrecon/blob/5b1692ce6e7d39aaeaa301dd9d8d64faa62d1de2/src/algorithms/calorimetry/CalorimeterHitReco.cc#L210-L226

Perhaps, comes from the `converter->cellDimensions(cellID)` branch of the if.

wdconinc:
Probably should revert from https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L198 to https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L259 in our algorithms.

veprbl:
> Probably should revert from https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L198 to https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L259 in our algorithms.

I'd support that. It seems wasteful to carry extra fields in each event just to simplify the algorithms. Although, saving some geometry support information doesn't seem like such a bad idea.

veprbl:
> I don't think we can have valgrind work on the asan binaries. Should we drop valgrind?

veprbl (dismissed):

veprbl:
Couple UBs popped up:

```output
/opt/software/linux-debian-x86_64/gcc-12.2.0/jana2-2.0.8-xpabcq7aslj6yrnewutmoc63vwfy5pln/include/JANA/JFactorySet.h:52:16: runtime error: downcast of address 0x6120002f1940 which does not point to an object of type 'JFactoryT<edm4hep::SimCalorimeterHit>'
0x6120002f1940: note: object is of type 'JFactoryT<edm4hep::SimCalorimeterHit const>'
 01 00 00 00  f0 36 5e 5c c8 7f 00 00  58 19 2f 00 20 61 00 00  00 00 00 00 00 00 00 00  00 be be be
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'JFactoryT<edm4hep::SimCalorimeterHit const>'
```

```output
/home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/ImagingPixelReco.h:111:51: runtime error: implicit conversion from type 'std::uint64_t' (aka 'unsigned long') of value 18279547340559880805 (64-bit, unsigned) to type 'dd4hep::long64' (aka 'long long') changed the value to -167196733149670811 (64-bit, signed)
```

wdconinc:
My main worry right now is how do we avoid having 2.5GB builds...

wdconinc:
```
/home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterHitReco.cc:160:91: runtime error: implicit conversion from type 'std::uint64_t' (aka 'unsigned long') of value 18352169581181866857 (64-bit, unsigned) to type 'dd4hep::long64' (aka 'long long') changed the value to -94574492527684759 (64-bit, signed)
```

Hmm, `uint64_t cellID` in the [data model](https://github.com/key4hep/EDM4hep/blob/master/edm4hep.yaml#L254), but `typedef long long int CellID` in [DD4hep](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/include/DDSegmentation/Segmentation.h#L48). Arguably, DD4hep is wrong here.

veprbl:
This PR removes slow reconstruction runs that run EICrecon under valgrind and take 1h to complete. Instead, we are looking to run EICrecon with a different kind of instrumentation: ASAN, UBSAN, LEAKSAN. For now, only ASAN is able to pass, but we'd like to eventually support all. Another aspect is that we are now running clang build of EICrecon and not GCC. Historically clang was first to acquire sanitizers, but, from what I understand, those has been implemented in GCC in some capacity.

.github/workflows/linux-eic-shell.yml
```diff
@@ -263,52 +266,23 @@ jobs:
...
     needs:
     - build
     - ddsim-gun
     strategy:
       matrix:
+        CC: [clang]
         particle: [pi, e]
         detector_config: [arches, brycecanyon]
+    env:
+      ASAN_OPTIONS: malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=true:halt_on_error=false:detect_odr_violation=0
```
wdconinc, line 280: Should we use `halt_on_error=true` in these jobs? Any other ASAN options we should enable?

veprbl: Googling suggests that it may not even work without some `-fsanitize-recover` setting on the compiler side. I think, we should just drop that option (1 is its default value).

wdconinc:
Let's just set `new_delete_type_mismatch=0`, get this to pass, and get some sanitizers in main. That's better than nothing. I think we don't understand the reason for the new/delete type mismatch, but we also know we can't actually fix it here.

veprbl (dismissed):

wdconinc:
https://github.com/eic/EICrecon/actions/runs/5407023666/jobs/9825039519?pr=514#step:5:703 fixed by https://github.com/eic/EICrecon/pull/718 I think.

wdconinc:
> because of a missing output file

Yeah, the output file was missing because EICrecon didn't run to completion and didn't create an error code (not sure why, but I'd rather it did fail and return an error code, or that no error code is at least a sign of success).

wdconinc:
Passed all checks, updated required checks in settings, now updating branch and should auto-merge.

veprbl:
Do we merge without error codes fixed?

wdconinc:
> Do we merge without error codes fixed?

I'd propose so, in the interest of moving forward? It gets more eyes on it at least.

veprbl (approved):

PR #516, opened 2023-02-28
fix: OOB read of cellDimensions

wdconinc:
As hoc fix for now. We should evaluate this in more detail later: what is the dimension stored for, and why as a 3D bounding box in some cases and as a 2D surface-constrained coordinate in other cases?

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitReco.cc
@@ -1,4 +1,3 @@
-
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2022 Chao Peng, Sylvester Joosten, Wouter Deconinck, Chao, Whitney Armstrong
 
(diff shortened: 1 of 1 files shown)


wdconinc:
@veprbl This PR is currently set to merge into `sanitizers` to demonstrate it fixes the sanitizer issue. I'd rebase to `main` and merge there though.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -223,7 +222,7 @@ void CalorimeterHitReco::AlgorithmProcess() {
         const decltype(edm4eic::CalorimeterHitData::position) position(gpos.x() / m_lUnit, gpos.y() / m_lUnit,
                                                                     gpos.z() / m_lUnit);
         const decltype(edm4eic::CalorimeterHitData::dimension) dimension(cdim[0] / m_lUnit, cdim[1] / m_lUnit,
```
wdconinc, line 224:
```suggestion
        // `dimension` is 3D, but `cdim` can be 2D only
        const decltype(edm4eic::CalorimeterHitData::dimension) dimension(cdim[0] / m_lUnit, cdim[1] / m_lUnit,
```

veprbl: I assume that's not true for the `else` branch of the `if` above.

veprbl (approved):

Issue #474, opened 2023-02-09
CalorimeterClusterMerger should not be applied to calorimeters other than EcalBarrelImaging


veprbl:
As explained by Chao during Today's meeting, this is not intended to be used for other calorimeters. Possibly, this extra step crept in during porting from Juggler?

PR #519, opened 2023-03-01
treewide: remove truth-based MergedClusters for all calorimeters except Imaging

veprbl:
According to Chao, the MergedClusters algorithm is intended to do merging of ScFi and AstroPix clusters in Imaging. Other calorimeters need not to rely on this.

Resolves: #474

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes, removes certain outputs. Users are to migrate to use regular clusters.

Yes, removes certain outputs


veprbl:
There are still leftover references in docs[1], I can follow up with fixes to those if this is approved

[1] https://github.com/eic/EICrecon/blob/main/docs/design/gaudi-algorithms.md

sly2j:
There were two cluster-merging algorithms in Juggler. I'm not arguing to keep either, but I think there's some confusion there:

* ClusterMerger applies to all calorimeters to merge back clusters that were improperly reconstructed in parts. This was never really used that heavily. This is the algorithm you seem to be referring to here (https://eicweb.phy.anl.gov/EIC/juggler/-/blob/main/JugFast/src/components/ClusterMerger.cpp)
* TruthEnergyPositionClusterMerger is used to combine imaging clusters with scifi clusters in the imaging ecal. I'm not sure this was ported to EICRecon (never looked into it) https://eicweb.phy.anl.gov/EIC/juggler/-/blob/main/JugFast/src/components/TruthEnergyPositionClusterMerger.cpp

wdconinc:
There was also:
- EnergyPositionClusterMerger: which was also not ported (https://eicweb.phy.anl.gov/EIC/juggler/-/blob/main/JugReco/src/components/EnergyPositionClusterMerger.cpp)

veprbl:
Yes, so I may be still ignorant about which cluster merging the Imaging calo relies on, but the point stands that we now assume that nobody actually uses this algorithm, so it can be safely disabled.

sly2j (dismissed):

Chao1009 (dismissed):

wdconinc (approved):

Issue #515, opened 2023-02-28
Renaming of TPCHit in EDM4hep


jmcarcell:
This PR: https://github.com/key4hep/EDM4hep/pull/179 is going to be merged soon and it will rename one field that appears a few times in this repo:
```
src/utilities/podio_copy/podio_copy.cc
27:#include <edm4hep/TPCHitCollection.h>
292:    if( colltype=="edm4hep::TPCHit" ) return CopyCollectionT<edm4hep::TPCHit, edm4hep::TPCHitCollection>( store_in, store_out, name);

src/utilities/podio_read/podio_read.cc
27:#include <edm4hep/TPCHitCollection.h>
302:    if( colltype=="edm4hep::TPCHit" ) return CopyCollectionT<edm4hep::TPCHit, edm4hep::TPCHitCollection>( store_in, store_out, name);
```
The change (in this case, it's not the only one, see discussion and linked PRs in the EDM4hep PR) is easy to do, rename `TPCHit` to `RawTimeSeries` and for collections `TPCHitCollection` to `RawTimeSeriesCollection`. I open this issue in this repo but it's relevant for all repos that use this, which I haven't checked (`EDM4eic` seems to be unaffected).

wdconinc:
Thanks a lot, @jmcarcell. We really appreciate the heads up! It seems that we will be relatively little affected by this, and you have identified the places where we will be affected.

PR #522, opened 2023-03-03
feat: remove unused podio_read and podio_copy

wdconinc:
The two utilities `podio_read` and `podio_copy` are unused, uncompiled, untested in CI. They include hardcoded EDM4hep type collection enumerations, contrary to typical podio data model use, and probably should not be maintained in this repository. If necessary for future purposes, their existence is sufficiently hinted at by this PR and they can be found in the commit history. 

- [x] Bug fix (issue #515)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ ] Tests for the changes have been added: 
- [x] Changes have been communicated to collaborators @DraTeots @faustus123 @veprbl 

No.

No.


veprbl (approved):
`podio_read` looks like a generic benchmark, so can it go. The `podio_copy` is, probably, the original code. Not sure about its purpose, it could be doing some subtle transformation on ddsim output?

wdconinc:
> it could be doing some subtle transformation on ddsim output?

Could be, but then its calling must be both obfuscated and its results must be cached, since there is no reference to `podio_copy` in the tree and the tests complete fine without it :-D

@faustus123 @DraTeots Do we need `podio_copy`?

DraTeots:
IMO - no. I believe it exists from times when David was experimenting with how to implement PODIO service for JANA2. Don't know about any use of it since

PR #524, opened 2023-03-03
fix: forbidden string conversion warning from `reco_particles_track_matching.cc`

c-dilks:
Fixes the following warning:
```
EICrecon/src/examples/track_matching/reco_particles_track_matching.cc: In function ‘int main()’:
EICrecon/src/examples/track_matching/reco_particles_track_matching.cc:59:35: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
   59 |     reco_particles_track_matching("/home/romanov/eic/soft/eicrecon/main/src/examples/test_data_generator/2022-11-15_pgun_pi-_epic_arches_e0.01-30GeV_alldir_4prt_1000evt_reco.tree.edm4eic.root");
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no

no

### DIFF
--- a/src/examples/track_matching/reco_particles_track_matching.cc
+++ b/src/examples/track_matching/reco_particles_track_matching.cc
@@ -14,7 +14,7 @@
 
 using std::string;
 
-void reco_particles_track_matching(char *file_name) {
+void reco_particles_track_matching(TString file_name) {
     auto file = new TFile(file_name);
     auto tree = (TTree *) file->Get("events");
     TTreeReader tree_reader(tree);       // !the tree reader


src/examples/track_matching/reco_particles_track_matching.cc
```diff
@@ -14,7 +14,7 @@
 
 using std::string;
 
-void reco_particles_track_matching(char *file_name) {
+void reco_particles_track_matching(TString file_name) {
```
veprbl, line 17:
```suggestion
void reco_particles_track_matching(std::string file_name) {
```

c-dilks: `std::string` throws
```
error: no matching function for call to ‘TFile::TFile(std::string&)
```
Either `TString` or `const char*` is adequate.

src/examples/track_matching/reco_particles_track_matching.cc
```diff
@@ -14,7 +14,7 @@
...
     auto file = new TFile(file_name);
```
veprbl, line 18:
```suggestion
void reco_particles_track_matching(std::string &file_name) {
    auto file = new TFile(file_name.c_str());
```

c-dilks: Let's not overthink this and let the author, @DraTeots, decide.

DraTeots: The most ugliest thing in the code was a string with some file on my system. The second thing, which I don't like about root is that in 2023 we have to use TFile(file_name.**c_str()**);. Since it is a root macro and root is at play, let it be a canonical TString. While I don't really see any difference here and just follow "not overthink" with 50.0001% vs 49.9999% weights =)

DraTeots (approved):

src/examples/track_matching/reco_particles_track_matching.cc
veprbl: We should use one string class for everything. If this is a collaboration, authorship doesn't grant ownership over the code.

veprbl: FYI, TFile doesn't have native support for TString, it's just that TString has an implicit conversion operator.

wdconinc:
This could have been a good opportunity to remove the hardcoded paths in that same file... It's clear that building this leads to a rather useless example for anyone except maybe @DraTeots.

PR #545, opened 2023-03-15
Fixed RP code to use local coordinates for matrix reconstruction. 

ajentsch:

This PR uses local coordinates for reconstruction of hits in the Roman Pots - this is a required step since the detectors will have no clue about a global coordinate system.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No - it's all behind the scenes, and changes nothing about the collections passed to or from the reco code.

Only in the actual matrix calculation to obtain the momentum.

### DIFF
--- a/src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
+++ b/src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
@@ -19,8 +19,14 @@ namespace eicrecon {
 
 	m_log = app->GetService<Log_service>()->logger("ForwardRomanPotRecParticles");
 
-	/*
-        auto id_spec = detector->readout(m_readout).idSpec();
+	m_readout = "ForwardRomanPotHits";
+	app->SetDefaultParameter("RPOTS:ForwardRomanPotHits:readoutClass", m_readout);
+	m_geoSvc = app->GetService<JDD4hep_service>();
+
+	
+	if(m_readout.empty()){ std::cout << "READOUT IS EMPTY!" << std::endl;  return; }
+
+        auto id_spec = m_geoSvc->detector()->readout(m_readout).idSpec();
         try {
             id_dec = id_spec.decoder();
             if (!m_sectorField.empty()) {
@@ -36,6 +42,8 @@ namespace eicrecon {
             throw JException("Failed to load ID decoder");
         }
 
+	std::cout << "Decoding complete..." << std::endl;
+		
         // local detector name has higher priority
         if (!m_localDetElement.empty()) {
             try {
@@ -60,7 +68,7 @@ namespace eicrecon {
             //                      fmt::join(fields, ", "))
             //        << endmsg;
         }
-	*/
+	
         double det = aXRP[0][0] * aXRP[1][1] - aXRP[0][1] * aXRP[1][0];
 
         if (det == 0) {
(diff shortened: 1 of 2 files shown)


src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
```diff
@@ -19,8 +19,14 @@ namespace eicrecon {
...
 	m_log = app->GetService<Log_service>()->logger("ForwardRomanPotRecParticles");
 
-	/*
-        auto id_spec = detector->readout(m_readout).idSpec();
+	m_readout = "ForwardRomanPotHits";
+	app->SetDefaultParameter("RPOTS:ForwardRomanPotHits:readoutClass", m_readout);
+	m_geoSvc = app->GetService<JDD4hep_service>();
+
+
+	if(m_readout.empty()){ std::cout << "READOUT IS EMPTY!" << std::endl;  return; }
```
DraTeots, line 27: Could you please use m_log instead of std::cout

src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
```diff
@@ -19,8 +19,14 @@ namespace eicrecon {
 
 	m_log = app->GetService<Log_service>()->logger("ForwardRomanPotRecParticles");
 
-	/*
-        auto id_spec = detector->readout(m_readout).idSpec();
+	m_readout = "ForwardRomanPotHits";
+	app->SetDefaultParameter("RPOTS:ForwardRomanPotHits:readoutClass", m_readout);
```
DraTeots, line 23: It is better to use PluginName() function instead of directly writing "RPOTS". And also output tag instead of ForwardRomanPotHits

Here is an exmple: 
https://github.com/eic/EICrecon/blob/main/src/global/tracking/TrackParameters_factory.cc#L21

src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
```diff
@@ -36,6 +42,8 @@ namespace eicrecon {
             throw JException("Failed to load ID decoder");
         }
 
+	std::cout << "Decoding complete..." << std::endl;
```
DraTeots, line 45: m_log

src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
```diff
@@ -118,33 +121,35 @@ namespace eicrecon {
...
+
+	    //information is stored in cm, we need mm - multiply everything by 10 for now
 
-	    if(!goodHit2 && pos0.z > 27099.0 && pos0.z < 28022.0){
+	    if(!goodHit2 && 10*gpos.z() > 27099.0 && 10*gpos.z() < 28022.0){
 
-		goodHitX[1] = pos0.x;
-		goodHitY[1] = pos0.y;
-		goodHitZ[1] = pos0.z;
+		goodHitX[1] = 10*pos0.x();
```
DraTeots, line 142: If 10 relates to unit conversions, please use dd4hep units instead of raw 10 which is really difficult to interpret

src/detectors/RPOTS/RomanPotsReconstruction_factory.h
```diff
@@ -54,17 +55,18 @@ namespace eicrecon {
 	std::string m_readout;
 	std::string m_layerField;
 	std::string m_sectorField;
+	std::string m_geoSvcName;
```
DraTeots, line 58: It is not critical or a blocker of this PR, but could I kindly suggest that we have coding conventions

https://github.com/eic/EICrecon/blob/main/CONTRIBUTING.md

src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
ajentsch: Fixed.

src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
ajentsch: Okay, I fixed this. I am still not sure about the "readoutClass" part of this, so I have that written in by hand for the moment, but I did use the same setup as in the example for the first two pieces of the string.

src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
ajentsch: Fixed.

src/detectors/RPOTS/RomanPotsReconstruction_factory.h
ajentsch: I will need to review this more-carefully to catch everything which doesn't currently conform, will certainly do this for the next PR.

src/detectors/RPOTS/RomanPotsReconstruction_factory.cc
ajentsch: Fixed.

DraTeots (approved):

Issue #539, opened 2023-03-11
Add an RNG service and avoid using `TRandom`


c-dilks:
**Is your feature request related to a problem? Please describe.**
Problem: In the algorithm `PhotoMultiplierHitDigi`, if the seed for the `TRandomMixMax` object is set to 0 (use "random" seed, see [documentation](https://root.cern.ch/doc/master/classTRandom.html#a2007ae15ce828266aecb783ed3410e8b)), then the randomly generated number can sometimes remain the same for several generations, then start actually randomizing later; this is a possible symptom of thread unsafety. This issue does not appear to happen if the seed is non-zero (a fixed seed, useful for reproducibility, but less random than seed 0).

`TRandom` is used in the following algorithms and factories:
```
src/algorithms/digi/PhotoMultiplierHitDigi.h
src/algorithms/digi/SiliconTrackerDigi.h
src/algorithms/reco/MC2SmearedParticle.h
src/global/digi/SiliconTrackerDigi_factory.h
src/global/tracking/TrackerHitReconstruction_factory.h
```
and some tracking efficiency benchmarks.

**Describe the solution you'd like**
Ideally, we should implement a thread-safe random number generation service.

**Describe alternatives you've considered**
Use seed != 0, which appears to work okay, but is not really ideal for the long term, especially if we want more randomness.

DraTeots:
BTW, there is [random generator service](https://github.com/eic/EICrecon/tree/main/src/services/randomgenerator) but it is barely usable in the current implementation.

c-dilks:
Nice, let's make it a proper service (e.g., `src/RandGen_driver.cpp` should not live in `src`) and start using it! Unfortunately I don't have time to help do this, but this seems like a very good entry-level issue for a new person to get involved with.

wdconinc:
> this seems like a very good entry-level issue for a new person to get involved with.

Not quite. Getting random numbers isn't hard. It's getting the same random numbers in the same sequence to the same algorithms, even when algorithms may run out of sequence the next time you run it... And this must all work in multithreaded running.

c-dilks:
> BTW, there is [random generator service](https://github.com/eic/EICrecon/tree/main/src/services/randomgenerator) but it is barely usable in the current implementation.

Removed by https://github.com/eic/EICrecon/pull/640.

PR #555, opened 2023-03-17
Noise injection in DRICH simulation

DelloStritto:
In this PR, the noise injection is added in the dual RICH digitizer.

- [ ] New feature (issue #__)

No breaking changes introduced

No


src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -90,10 +90,11 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
             // pixel gap cuts
             // FIXME: generalize; this assumes the segmentation is `CartesianGridXY`
             auto id = ahit->getCellID();
-            dd4hep::Position pos_pixel, pos_hit;
+            dd4hep::Position pos_pixel, pos_hit, pos_hit_global;
             if(m_cfg.enablePixelGaps) {
               pos_pixel = get_sensor_local_position( id, m_cellid_converter->position(id) );
               pos_hit   = get_sensor_local_position( id, vec2pos(ahit->getPosition())     );
+              pos_hit_global =  m_cellid_converter->position(id);
```
c-dilks, line 97: Save some time and call `m_cellid_converter->position(id)` once
```suggestion
              pos_hit_global = m_cellid_converter->position(id);
              pos_pixel = get_sensor_local_position( id, pos_hit_global );
              pos_hit   = get_sensor_local_position( id, vec2pos(ahit->getPosition())     );
```

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -116,10 +117,10 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
                 }
                 // no hits group found
                 if (i >= it->second.size()) {
-                    it->second.emplace_back(HitData{1, amp + m_cfg.pedMean + m_cfg.pedError*m_rngNorm(), time, pos_hit});
+                    it->second.emplace_back(HitData{1, amp + m_cfg.pedMean + m_cfg.pedError*m_rngNorm(), time, pos_hit_global});
                 }
             } else {
-                hit_groups[id] = {HitData{1, amp + m_cfg.pedMean + m_cfg.pedError*m_rngNorm(), time, pos_hit}};
+                hit_groups[id] = {HitData{1, amp + m_cfg.pedMean + m_cfg.pedError*m_rngNorm(), time, pos_hit_global}};
```
c-dilks, line 123: Instead of changing the meaning of `HitData::pos`, add another `dd4hep::Position` member to `HitData` itself so you can store both `pos_hit` and `pos_hit_global`.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -277,3 +330,8 @@ dd4hep::Position eicrecon::PhotoMultiplierHitDigi::get_sensor_local_position(uin
 
   return pos_transformed;
 }
+
+bool  eicrecon::PhotoMultiplierHitDigi::Noise_Digits(float noiseRate, int timeWindow) const
+{
+  return (m_rngUni() < (noiseRate*timeWindow*dd4hep::ns));
+}
```
c-dilks, line 337: Does this need to be its own function? It's only called in one place.

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
```diff
@@ -74,6 +80,8 @@ namespace eicrecon {
         puts("enablePixelGaps",enablePixelGaps);
         puts("pixelSize",pixelSize);
         puts("safetyFactor",safetyFactor);
+        puts("noiseRate",noiseRate);
+        puts("timeWindow",timeWindow);
```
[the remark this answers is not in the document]
DelloStritto: Done

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -142,8 +143,60 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
...
+              if(de_name.find("sensor_de_"+secName)!=std::string::npos) {
+                // get sensors info
+                auto imodsec = detSensor.id();
+                uint64_t x_encoded = 0;
+
+                for (uint64_t x = 0; x < 7; x++){
+                  x_encoded = ( x << readoutCoder["x"].offset() )  & readoutCoder["x"].mask();
+                  uint64_t y_encoded = 0;
+                  for (uint64_t y = 0; y < 7; y++){
+                    y_encoded = ( y << readoutCoder["y"].offset() ) & readoutCoder["y"].mask();
```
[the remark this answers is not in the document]
DelloStritto: Done. But it will need further modification to make the code general

src/global/digi/PhotoMultiplierHitDigi_factory.cc
```diff
@@ -55,6 +55,10 @@ void eicrecon::PhotoMultiplierHitDigi_factory::ChangeRun(const std::shared_ptr<c
 
 void eicrecon::PhotoMultiplierHitDigi_factory::Process(const std::shared_ptr<const JEvent> &event) {
 
+  auto app = GetApplication();
+  // Services
+  auto geo_service = app->GetService<JDD4hep_service>();
+
```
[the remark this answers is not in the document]
DelloStritto: Done!

src/algorithms/digi/PhotoMultiplierHitDigi.cc
DelloStritto: The reason why I did it in this way is that we may need to add some feature to the function, making it longer and more complex. But I see your point.

c-dilks: Sounds good!

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -142,8 +145,62 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
...
+                  for (uint64_t y = 0; y < num_px; y++){
+                    y_encoded = ( y << readoutCoder["y"].offset() ) & readoutCoder["y"].mask();
+                    auto cellID = systemID + imodsec + x_encoded + y_encoded;
+                    auto it = hit_groups.find(cellID);
+                    if (Noise_Digits(m_cfg.noiseRate, m_cfg.timeWindow)){
+                      // cell time, signal amplitude
+                      double amp = m_cfg.speMean + m_rngNorm()*m_cfg.speError;
+                      double time = m_cfg.timeWindow*m_rngUni();
+                      dd4hep::Position pos_hit_global = m_cellid_converter->position(cellID);
+                      dd4hep::Position pos_hit = get_sensor_local_position( cellID, pos_hit_global );
```
c-dilks, line 181: Hmm... now I see why you originally wanted to change `HitData::pos`. The only reason I want to keep `pos_hit = get_sensor_local_position(...)` for the actual signal hits is to check if the pixel gap cuts are working correctly; but I fear the `get_sensor_local_position` function is kind of slow. Here in this noise injection loop, you don't have a need to apply pixel gap cuts, so you could just leave this as empty (and when we want to check pixel gap cuts, we just disable noise injection). So to speed things up, I suggest:
```suggestion
                      dd4hep::Position pos_hit{0,0,0}; // unused
```

c-dilks:
CI build job is having trouble; I suspect because this is a branch from a fork. Testing disabling the `eicrecon -version` check

.github/workflows/linux-eic-shell.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           # install this repo
           CC=${{ matrix.CC }} CXX=${{ matrix.CXX }} cmake -B build -S . -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ matrix.CMAKE_BUILD_TYPE }}
           cmake --build build -- -j 2 install
-          if [[ $(eicrecon --version) =~ unknown ]] ; then false ; fi
+          # if [[ $(eicrecon --version) =~ unknown ]] ; then false ; fi
```
wdconinc, line 62:
```suggestion
          eicrecon --version
          # if [[ $(eicrecon --version) =~ unknown ]] ; then false ; fi
```

c-dilks:
Opened https://github.com/eic/EICrecon/issues/567 and a test PR

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -142,8 +143,60 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
...
             }
         }
-        return raw_hits;
 
+        if (m_cfg.noiseInjection) {
+          //build noise raw hits
+          std::unordered_map<uint64_t, std::vector<HitData>> hit_groups_noise;
+
+          const auto& readoutCoder   = *detector->readout("DRICHHits").idSpec().decoder();
+          auto nSectors       = detector->constant<int>("DRICH_num_sectors");
```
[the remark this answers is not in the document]
c-dilks: Hi @DelloStritto, I have implemented this higher-order function to move the dRICH dependence into `richgeo`. Please check the pull request https://github.com/DelloStritto/EICrecon/pull/1, which targets your PR here, and merge it if it works (or make changes if it doesn't).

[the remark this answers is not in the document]
DelloStritto: Hi @c-dilks . Thanks for the implementation. It seems to work. I have already merged it

src/detectors/DRICH/DRICH.cc
```diff
@@ -34,6 +34,9 @@ extern "C" {
     digi_cfg.enablePixelGaps = true;
     digi_cfg.pixelSize       = 3.0; // [mm]
     digi_cfg.safetyFactor    = 0.7;
+    digi_cfg.enableNoise     = true;
```
c-dilks, line 37: Let's keep the noise injection off for now in this PR, so the IRT PID algorithm continues to have the same performance. Checking how well IRT does with noise is the next immediate step, however.
```suggestion
    digi_cfg.enableNoise     = false;
```

c-dilks:
This looks good to me, and works well under my local tests. I have not checked with the IRT algorithm yet, but that should be one of our next steps.

Unfortunately we cannot 100% remove the RICH-readout dependence from `PhotoMultiplierHitDigi`, but I think the way we have it written now is flexible enough to permit future generalization. We may even move toward a more general noise injection, such as those available in `DDDigi`, but I think this current approach is good enough for now.

src/algorithms/digi/PhotoMultiplierHitDigi.h
```diff
@@ -70,5 +76,6 @@ class PhotoMultiplierHitDigi : public WithPodConfig<PhotoMultiplierHitDigiConfig
     void qe_init();
     template<class RndmIter, typename T, class Compare> RndmIter interval_search(RndmIter beg, RndmIter end, const T &val, Compare comp) const;
     bool qe_pass(double ev, double rand) const;
+    bool Noise_Digits(float noiseRate, int timeWindows) const;
```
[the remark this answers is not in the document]
wdconinc: And this capitalization is just a local convention, it seems...

wdconinc (commented):
Main comment: avoid looping over all pixels, but instead thrown a Poisson number of them uniformly in sensors and x,y pixels. That leads to better computational scaling.

src/global/digi/PhotoMultiplierHitDigi_factory.cc
```diff
@@ -26,6 +26,12 @@ void eicrecon::PhotoMultiplierHitDigi_factory::Init() {
   auto geo_service = app->GetService<JDD4hep_service>();
   InitLogger(param_prefix, "info");
 
+  // get readout info (if a RICH)
+  if(plugin_name.find("RICH") != std::string::npos) {
```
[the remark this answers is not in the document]
c-dilks: changed to
```cpp
plugin_name=="DRICH" || plugin_name=="PFRICH"
```

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -142,8 +145,38 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
                 raw_hits.push_back(hit);
             }
         }
-        return raw_hits;
 
+        //build noise raw hits
+        if (m_cfg.enableNoise) {
+          m_log->trace("{:=^70}"," BEGIN NOISE INJECTION ");
+          std::unordered_map<uint64_t, std::vector<HitData>> hit_groups_noise;
```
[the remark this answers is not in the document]
c-dilks: See https://github.com/eic/EICrecon/blob/ba7ab65a1d54fc35f63138922d2998174ba4817e/src/services/geometry/richgeo/ReadoutGeo.cc#L47-L54
I wasn't sure how to correctly deal with this...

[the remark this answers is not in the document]
c-dilks: And in DD4hep's `BitFieldCoder.h`:
```cpp
typedef long long int long64   // (in namespace dd4hep)
```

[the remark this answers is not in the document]
c-dilks: vs. https://github.com/key4hep/EDM4hep/blob/5c48904b41f6ae5f04a1f0726ee9a7041ba67b2d/edm4hep.yaml#L212

src/services/geometry/richgeo/ReadoutGeo.cc
```diff
@@ -0,0 +1,76 @@
...
+
+    // get number of pixels along one side of a SiPM
+    m_num_px = m_det->constant<int>("DRICH_num_px");
+
+    // define cellID looper
+    m_loopCellIDs = [this] (std::function<void(uint64_t)> lambda) {
+      m_log.PrintLog("call ReadoutPixelLoop for systemID = {} = {}", m_systemID, m_detName);
+
+      // loop over sensors (for all sectors)
+      for(auto const& [deName, detSensor] : m_detRich.children()) {
```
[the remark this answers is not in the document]
c-dilks: Excellent idea! @DelloStritto perhaps we make a list of all the possible `cellID`s, then draw _k_ of them for each event. We could use the `ReadoutGeo::VisitAllReadoutPixels` method to make this list; in principle we could get rid of `VisitAllReadoutPixels`, but it may be useful for something else someday, so let's just keep it.

You could make such a list in `AlgorithmInit`, and think about what data structure would be fastest to draw from (e.g., `std::set` would be better than `std::vector`, since order doesn't matter). @wdconinc any suggestions?

Alternatively you could randomly generate the sensor ID number, pixel x, and pixel y, but you'd have to make sure to choose valid numbers. Would it be faster to generate 3 random numbers from small lists vs. 1 random number from a large list?

src/algorithms/digi/PhotoMultiplierHitDigi.cc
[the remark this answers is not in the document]
wdconinc: Yeah, this is a bug in DD4hep and EDM4hep. And I fear neither of them will fix this. In any case, https://github.com/key4hep/EDM4hep/issues/202 now points it out.

src/services/geometry/richgeo/ReadoutGeo.cc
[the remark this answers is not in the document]
wdconinc: I wouldn't create a list to pick from. Just go with the alternative since in this case your structure is just a uniform cube anyway, without any holes, right? If I recall (from @S-tuberosum's work), there are O(1000^2) pixels, so your list will be O(1M) in memory for a very regular grid. Lookup would be fast, but still it's something to avoid.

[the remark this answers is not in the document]
c-dilks: Yeah it's a 4D cube with no holes. So we have:
- 6 sectors, from DD4hep constant `DRICH_num_sectors`
- almost 1400 sensor modules
  - more than we need, for reasons as discussed in dRICH meetings
  - this number is trickier to get, but you can count them taking inspiration from the `ReadoutGeo::VisitAllReadoutPixels` loop
  - should be the same for each sector, but it would be safe to double check
- 8x8 pixels in x and y, where the number 8 comes from DD4hep constant `DRICH_num_px`
- and finally the fixed dRICH subsystem ID

Once you have those numbers, encode with https://github.com/eic/EICrecon/blob/08b4fa77027eedb2bde1b9404e5f08784de7bc00/src/services/geometry/richgeo/ReadoutGeo.cc#L47-L55
(don't use addition and bitshifting I mentioned earlier, since `x` and `y` are signed (but in practice they only range 0-7))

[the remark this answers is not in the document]
c-dilks: In fact, let's make it easy to get the number of sensors per sector: https://github.com/eic/epic/pull/410
So then you can just use `DRICH_num_sensors` instead of counting.

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
```diff
@@ -17,6 +20,11 @@ namespace eicrecon {
       double pedMean       = 200.0;
       double pedError      = 3.0;
 
+      // noise
+      bool enableNoise       = false;
```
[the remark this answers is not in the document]
c-dilks: `rate==0` will certainly disable the noise injection, but I kind of like having this boolean so that we can more easily turn it on or off without adjusting the rate. We can certainly move to an `enum` if we have more noise models, but for now I vote we keep it simple.

src/services/geometry/richgeo/ReadoutGeo.cc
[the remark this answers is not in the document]
c-dilks: @DelloStritto's latest changes implement this with `RichGeo::VisitAllRngPixels`, but I have some further comments which will appear in new conversations below.

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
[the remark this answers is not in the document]
wdconinc: Sounds good to me.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -137,13 +140,41 @@ eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::Si
...
+          float p = m_cfg.noiseRate*m_cfg.noiseTimeWindow*dd4hep::ns;
```
c-dilks, line 153: We eventually want to put the units in `DRICH.cc` instead, that is,
```cpp
digi_cfg.noiseTimeWindow = 20.0 * dd4hep::ns; // [ns]  // DRICH.cc
```
and
```cpp
float p = m_cfg.noiseRate*m_cfg.noiseTimeWindow / dd4hep::ns;  // here
```
This is a more general issue though, and I plan to take care of all these config parameter units in a separate PR.

But my main question here is do you really need the `*dd4hep::ns` here?

src/services/geometry/richgeo/ReadoutGeo.h
```diff
@@ -0,0 +1,61 @@
...
+namespace richgeo {
+  class ReadoutGeo {
+    public:
+
+      // constructor
+      ReadoutGeo(std::string detName_, dd4hep::Detector *det_, bool verbose_=false);
+      ~ReadoutGeo() {}
+
+      // random number generators
+      TRandomMixMax m_random;
```
c-dilks, line 31: - Can this be moved to `private`?
- Can this take a `seed`, perhaps the one from `PhotoMultiplierHitDigiConfig`?
- Thread unsafe, but that's a [separate known issue](https://github.com/eic/EICrecon/issues/539)

src/services/geometry/richgeo/ReadoutGeo.cc
```diff
@@ -0,0 +1,108 @@
...
+          } // end xy-segmentation loop
+        }
+      } // end sensor loop (for all sectors)
+    }; // end definition of m_loopCellIDs
+
+    // define k random cell IDs generator
+    m_rngCellIDs = [this] (std::function<void(uint64_t)> lambda, float p) {
+      m_log.PrintLog("call RngReadoutPixels for systemID = {} = {}", m_systemID, m_detName);
+
+      int k = p*m_num_sec*m_num_mod*m_num_px*m_num_px;
```
c-dilks, line 77: Is this really the number of noise hits we want? Do we assume `p << 1`?

src/services/geometry/richgeo/ReadoutGeo.cc
```diff
@@ -0,0 +1,108 @@
...
+	int y = m_random.Uniform(0., m_num_px);
+
+	// encode cellID
+	dd4hep::long64 cellID_dd4hep;
+	m_readoutCoder->set(cellID_dd4hep, "system", m_systemID);
+	m_readoutCoder->set(cellID_dd4hep, "sector", isec);
+	m_readoutCoder->set(cellID_dd4hep, "module", imod);
+	m_readoutCoder->set(cellID_dd4hep, "x",      x);
+	m_readoutCoder->set(cellID_dd4hep, "y",      y);
+	uint64_t cellID(cellID_dd4hep); // DD4hep uses `dd4hep::long64`, but EDM4hep uses `uint64_t`
```
c-dilks, line 92: This block of code now appears twice, so it would be better to move it (and the other block from `m_loopCellIDs`) to its own function.

src/algorithms/digi/PhotoMultiplierHitDigi.h
```diff
@@ -70,5 +76,6 @@ class PhotoMultiplierHitDigi : public WithPodConfig<PhotoMultiplierHitDigiConfig
     void qe_init();
     template<class RndmIter, typename T, class Compare> RndmIter interval_search(RndmIter beg, RndmIter end, const T &val, Compare comp) const;
     bool qe_pass(double ev, double rand) const;
+    bool has_noise_digits(float noiseRate, int timeWindows) const;
```
c-dilks, line 79: No longer used and can be deleted
```suggestion
```

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -277,3 +308,8 @@ dd4hep::Position eicrecon::PhotoMultiplierHitDigi::get_sensor_local_position(uin
 
   return pos_transformed;
 }
+
+bool  eicrecon::PhotoMultiplierHitDigi::has_noise_digits(float noiseRate, int timeWindow) const
+{
+  return (m_rngUni() < (noiseRate*timeWindow*dd4hep::ns));
+}
```
c-dilks, line 315: No longer used and can be deleted
```suggestion
```

src/services/geometry/richgeo/ReadoutGeo.cc
DelloStritto: It seems to be reasonable considering the results at the test beam. Also, as already mentioned by you and Wouter, if needed, we can add different noise types later

src/algorithms/digi/PhotoMultiplierHitDigi.cc
DelloStritto: I moved it to DRICH.cc. It is needed to convert the time windows in seconds (the rate in Hz). But I don't think I need it when I estimate the probability.

src/services/geometry/richgeo/ReadoutGeo.cc
```diff
@@ -0,0 +1,94 @@
...
+
+// constructor
+richgeo::ReadoutGeo::ReadoutGeo(std::string detName_, dd4hep::Detector *det_, bool verbose_)
+  : m_detName(detName_), m_det(det_), m_log(Logger::Instance(verbose_))
+{
+  // capitalize m_detName
+  std::transform(m_detName.begin(), m_detName.end(), m_detName.begin(), ::toupper);
+
+  // random number generators
+  m_random.SetSeed(37);
```
c-dilks, line 16:
```suggestion
  m_random.SetSeed(1); // default seed
```

src/services/geometry/richgeo/ReadoutGeo.h
```diff
@@ -0,0 +1,76 @@
...
+	return cellID;
+	// m_log.PrintLog("    x={:<2} y={:<2} => cellID={:#018X}", x, y, cellID);
+      }
+
+      // loop over readout pixels, executing `lambda(cellID)` on each
+      void VisitAllReadoutPixels(std::function<void(uint64_t)> lambda) { m_loopCellIDs(lambda); }
+
+      // generated k rng cell IDs, executing `lambda(cellID)` on each
+      void VisitAllRngPixels(std::function<void(uint64_t)> lambda, float p) { m_rngCellIDs(lambda, p); }
+
```
c-dilks, line 50:
```suggestion

// set RNG seed
void SetSeed(unsigned long seed) { m_random.SetSeed(seed); }
```

src/algorithms/digi/PhotoMultiplierHitDigi.h
```diff
@@ -63,9 +76,26 @@ class PhotoMultiplierHitDigi : public WithPodConfig<PhotoMultiplierHitDigiConfig
       return dd4hep::Position( v.x*dd4hep::mm, v.y*dd4hep::mm, v.z*dd4hep::mm );
     }
 
+    // set readout info
+    void SetReadoutGeo(richgeo::ReadoutGeo *readout) { m_readoutGeo = readout; };
```
c-dilks, line 80:
```suggestion
    void SetReadoutGeo(richgeo::ReadoutGeo *readout) {
      m_readoutGeo = readout;
      m_readoutGeo.SetSeed(m_cfg.seed);
    }
```

c-dilks (commented):
Some small changes to propagate `seed` to `ReadoutGeo` RNG. (should be tested)

c-dilks (dismissed):

c-dilks:
I approve this PR and have tested it, at least showing that it does the job (next we should see how this impacts IRT PID).

Here are 100 pions aimed with the same momentum, without noise:


And with noise:


I have, however, pushed a bunch of commits to this branch and therefore would prefer a second PR approval. @wdconinc has already looked through this, could you take another look when you get a chance? Or @veprbl if you have some time?

wdconinc:
> could you take another look when you get a chance?

On vacation. Deferring to @veprbl

c-dilks:
There's no rush; there was a lot of discussion and changes on this PR. I will go ahead and merge this branch into our dRICH development branch, `irt-algo` (this PR targets `main`), so that we can begin testing it with IRT.

src/algorithms/digi/PhotoMultiplierHitDigi.h
```diff
@@ -23,17 +23,30 @@
...
+#include <functional>
 
 #include "PhotoMultiplierHitDigiConfig.h"
 #include <algorithms/interfaces/WithPodConfig.h>
 
 namespace eicrecon {
 
+using CellIDType = decltype(edm4hep::SimTrackerHitData::cellID);
+using TimeType   = decltype(edm4hep::SimTrackerHitData::time);
+
```
veprbl, line 35: Move from `eicrecon::` to `eicrecon:: PhotoMultiplierHitDigi::`

src/services/geometry/richgeo/ReadoutGeo.cc
```diff
@@ -0,0 +1,94 @@
...
+
+    // get number of sectors
+    m_num_sec = m_det->constant<int>("DRICH_num_sectors");
+    // get number of sensors
+    m_num_mod = m_det->constant<int>("DRICH_num_sensors");
+    // get number of pixels along one side of a SiPM
+    m_num_px = m_det->constant<int>("DRICH_num_px");
+
+    // define cellID looper
+    m_loopCellIDs = [this] (std::function<void(uint64_t)> lambda) {
```
veprbl, line 40: Not clear why this can't be a method...

veprbl: 1. You don't explain what kind of flexibility
2. Could it be that the code simply doesn't belong in the service?

c-dilks: > Could it be that the code simply doesn't belong in the service?

Where do you think it should go?

c-dilks: > 1. You don't explain what kind of flexibility

It will be more clear what we need when the pfRICH (or others) starts using these algorithms.

veprbl: > > Could it be that the code simply doesn't belong in the service?
> 
> Where do you think it should go?

In the algorithm itself.

c-dilks: This function is specific to the dRICH, and no part of `PhotoMultiplierHitDigi` should be subsystem-specific. `richgeo` is designed to keep PID-subsystem-dependent code out of all algorithms.

c-dilks:
N.B. this algorithm is not (yet) used in production.

veprbl (approved):

PR #561, opened 2023-03-21
CalorimeterIslandCluster: autotune transverse shower profile parameter

veprbl:
The current implementation uses calorimeter cell size for value of shower width parameter. This approach probably doesn't work for calorimeters other than the Imaging. The width should be related to the Moli`ere radius for the material, but possibly also the geometry (even location of the cluster in the calorimeter). This commit implements a parameter scan that should help users to get get cluster splitting working, while we are trying to figure out a better way to handle this.

Example reconstruction for epic_sciglass_only:
eicrecon_clustering_autotune

Can be tested using an option like: `-PEcalBarrelSciGlassProtoClusters:splitCluster=1` (`-PEcalBarrelSciGlassProtoClusters:LogLevel=debug` will show the values of "lambda")

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Only for users had splitting manually enabled

No

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
+++ b/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
@@ -65,6 +65,7 @@ void CalorimeterIslandCluster::AlgorithmInit(std::shared_ptr<spdlog::logger>& lo
         {"globalDistRPhi", {globalDistRPhi, {dd4hep::mm, dd4hep::rad}}}, {"globalDistEtaPhi", {globalDistEtaPhi, {1., dd4hep::rad}}}
     };
 
+        hitsDist = globalDistEtaPhi; // FIXME
 
     // set coordinate system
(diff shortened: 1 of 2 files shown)


src/algorithms/calorimetry/CalorimeterIslandCluster.cc
```diff
@@ -65,6 +65,7 @@ void CalorimeterIslandCluster::AlgorithmInit(std::shared_ptr<spdlog::logger>& lo
         {"globalDistRPhi", {globalDistRPhi, {dd4hep::mm, dd4hep::rad}}}, {"globalDistEtaPhi", {globalDistEtaPhi, {1., dd4hep::rad}}}
     };
 
+        hitsDist = globalDistEtaPhi; // FIXME
```
veprbl, line 68: To be addressed in #638

Issue #527, opened 2023-03-07
Fix the silicon tracker digitization threshold


ShujieL:
### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. the default threshold is 0 in SiliconTrackerDigiConfig.h
2.The unit conversion (GeV to keV) in line 55, 56 of SiliconTrackerDigi.cc is wrong

### Expected Result: (what do you expect when you execute the steps above)
We fixed the reported bugs above and updated the default threshold to 1.2 keV which is about 3 times the mean noise of 110 electrons hitting silicon sensors.

ShujieL:
Along with this bug fix, we also hope to have the threshold variable (see https://github.com/eic/EICrecon/blob/f537c7644b6f8a808ec3687a00b01b635ae93ae7/src/algorithms/digi/SiliconTrackerDigiConfig.h#L6) as a comment line option / flag. @DraTeots can you help us with this?

PR #570, opened 2023-03-24
fixed the silicon tracker digitization threshold

blianggilman:
I fixed the bug that had incorrect units in the silicon energy thresholds. I also changed the silicon energy threshold to be 5 keV.

- [x] Bug fix (issue #527 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: Raised the silicon energy threshold from 0 keV.

See talk from: https://indico.bnl.gov/event/18890/

No, this PR does not introduce any breaking changes.

This will change the default behavior.


DraTeots (dismissed):

src/algorithms/digi/SiliconTrackerDigiConfig.h
```diff
@@ -3,7 +3,7 @@
 namespace eicrecon {
 
     struct SiliconTrackerDigiConfig {
-        double threshold  = 0;
+        double threshold  = 5; //keV
```
veprbl, line 6:
```suggestion
        double threshold  = 5 * units::keV;
```

src/algorithms/digi/SiliconTrackerDigi.cc
```diff
@@ -52,8 +52,8 @@ eicrecon::SiliconTrackerDigi::produce(const std::vector<const edm4hep::SimTracke
 
 
         double edep = sim_hit->getEDep();
-        if (edep * units::keV < m_cfg.threshold) {
-            m_log->debug("  edep is below threshold of {:.2f} [keV]", m_cfg.threshold / units::keV);
+        if (edep / units::keV < m_cfg.threshold) {
+            m_log->debug("  edep is below threshold of {:.2f} [keV]", m_cfg.threshold);
```
veprbl, line 56:
```suggestion
        if (edep < m_cfg.threshold) {
            m_log->debug("  edep is below threshold of {:.2f} [keV]", m_cfg.threshold / units::keV);
```

veprbl (commented):
Why introduce different unit systems for different variables

src/algorithms/digi/SiliconTrackerDigi.cc
veprbl: Also remove `namespace units = TGeoUnit;` above

veprbl: cc #577

bschmookler: Hi. I think the changes @veprbl are not correct if we put the number "5" in the SiliconTrackerDigiConfig.h file for the threshold. In that case, we need to convert edep from GeV to keV (or m_cfg.threshold from keV to GeV) before doing the comparison.

Can we also set the threshold in SiliconTrackerDigiConfig.h to 5*units::keV? Then it would be consistent with the changes Dmitry made.

veprbl: Yes, see the suggestion above. @DraTeots marked it as resolved for unknown reason.

bschmookler: Ok. I see it now. Thanks. 

@blianggilman if you think that all looks reasonable, you can select the "commit suggestion" for both the SiliconTrackerDigiConfig.h and the SiliconTrackerDigi.cc files.

veprbl (approved):

veprbl:
Look fine now

PR #578, opened 2023-03-28
Use JANA's new first-class PODIO integration

wdconinc:

This PR allows eicrecon to use PODIO associations correctly, resolving a major source of confusion, entropy, and bugs.
As of v2.1.0, the JANA2 framework has first-class support for PODIO. This means that users can directly read and 
write PODIO collections. Even if users continue to use the classic JANA vector-of-pointers style, under the hood the PODIO objects are always registered to a collection and that collection is registered to a frame. JANA understands and abides by PODIO's memory ownership semantics, including supporting subset collections. 

This PR also converts three problematic factories into multifactories, which support multiple outputs. These factories used direct object ID rewriting in order to set up valid associations, and/or had particularly thorny PODIO ownership issues. They are now models for how to write factories in the future. These factories are: MatchClusters_factory, ParticlesWithTruthPID_factory, and TrackingResult_factory. 

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- The way that users should write future factories has changed. Users should look to TrackingResult_factory as an example of how to write factories with multiple outputs. 

- Output files are now written using Podio's Frame writer, which is incompatible with Podio's older EventStore writer.

- JEventProcessorPODIO used to maintain a list of factories that have thrown an exception. After it crashed once, it would never be called again. Now, it will be called again on every event. The reason for this change in behavior is that PODIO v0.16.3 has a bug where collection IDs depend on insertion order, which causes the PODIO frame writer to crash. Excepting factories still produce an empty collection (thus preserving the insertion order), which the JEventProcessorPODIO then persists.


nathanwbrei (approved):

veprbl:
Fails clang-tidy check:
```
/home/runner/work/EICrecon/EICrecon/src/services/io/podio/JFactoryPodioT.h:11:10: error: 'datamodel_glue.h' file not found [clang-diagnostic-error]
#include "datamodel_glue.h"
         ^~~~~~~~~~~~~~~~~~
```

nathanwbrei:
I'm guessing clang-tidy is complaining about the "" instead of doing the full #include <services/io/podio/datamodel_glue.h>

src/services/io/podio/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ plugin_glob_all(${PLUGIN_NAME})
 # but that never really worked quite right. Thus, we just force-run it
 # whenever cmake is run.
 execute_process (
-        COMMAND python3 ${PROJECT_SOURCE_DIR}/make_datamodel_glue.py WORKING_DIR=${PROJECT_BINARY_DIR} EDM4HEP_INCLUDE_DIR=${EDM4HEP_INCLUDE_DIR} EDM4EIC_INCLUDE_DIR=${EDM4EIC_INCLUDE_DIR}
+        COMMAND python3 ${PROJECT_SOURCE_DIR}/make_datamodel_glue.py WORKING_DIR=${PROJECT_SOURCE_DIR} EDM4HEP_INCLUDE_DIR=${EDM4HEP_INCLUDE_DIR} EDM4EIC_INCLUDE_DIR=${EDM4EIC_INCLUDE_DIR}
```
c-dilks, line 37: What's the reason for this change? I'm wondering why re-building is now very slow, but reverting this change seems to help speed things up (only tested on one of my development branches).

wdconinc: @nbrei in case he didn't get this (I created the PR so I may be the only one getting the notifications)

nathanwbrei: It's because every factory now needs datamodel_glue.h for the PodioTypeMap that defines the relationship between podio types and their collections. The build dir for the podio plugin isn't on the include path for the other plugins, and the source dir is. I understand that using datamodel_glue all over the place is undesirable because it makes compile times longer and is just generally ugly. Luckily this is temporary. By adding one `using MyPodioObject::collection_t = MyPodioObjectCollection` to the generated EDM headers we can get rid of this.

veprbl: Contrary to this report, the compilation times don't change for me when I try to compile revisions before this PR.

Issue #573, opened 2023-03-27
Inconsistent definitions of energy resolutions


veprbl:
```
src/detectors/B0ECAL/RawCalorimeterHit_factory_B0ECalRawHits.h
39:        u_eRes = {0.0,0.02,0.0};
57:        app->SetDefaultParameter("B0ECAL:B0ECalRawHits:energyResolutions",u_eRes);

src/detectors/BEMC/RawCalorimeterHit_factory_EcalBarrelScFiRawHits.h
39:        u_eRes = {0.0 * dd4hep::MeV};
56:        app->SetDefaultParameter("BEMC:EcalBarrelScFiRawHits:energyResolutions",u_eRes);

src/detectors/BEMC/RawCalorimeterHit_factory_EcalBarrelImagingRawHits.h
39:        u_eRes = {0.0, 0.02, 0.0};
53:        app->SetDefaultParameter("BEMC:EcalBarrelImagingRawHits:energyResolutions",u_eRes);

src/detectors/BEMC/RawCalorimeterHit_factory_EcalBarrelSciGlassRawHits.h
39:        u_eRes =  {0.0, 0.0, 0.0};
58:        app->SetDefaultParameter("BEMC:EcalBarrelSciGlassRawHits:energyResolutions",u_eRes);

src/detectors/EEMC/RawCalorimeterHit_factory_EcalEndcapPRawHits.h
39:        u_eRes = {3.16 * dd4hep::MeV, 1.5 * dd4hep::MeV, 0.0 * dd4hep::MeV};
56:        app->SetDefaultParameter("EEMC:EcalEndcapPRawHits:energyResolutions",u_eRes);

src/detectors/EEMC/RawCalorimeterHit_factory_EcalEndcapPInsertRawHits.h
39:        u_eRes = {3.16 * dd4hep::MeV, 1.5 * dd4hep::MeV, 0.0 * dd4hep::MeV};
56:        app->SetDefaultParameter("EEMC:EcalEndcapPInsertRawHits:energyResolutions",u_eRes);

src/detectors/EEMC/RawCalorimeterHit_factory_EcalEndcapNRawHits.h
39:        u_eRes = {0.0 * dd4hep::MeV, 20 * dd4hep::MeV, 0.0 * dd4hep::MeV};
58:        app->SetDefaultParameter("EEMC:EcalEndcapNRawHits:energyResolutions",u_eRes);

[...]

PR #408, opened 2022-12-28
fix: propagate parameters from reco_flags to cpp plugins

wdconinc:
### Briefly, what does this PR introduce?
This addresses a lot of the discrepancies in https://eic.github.io/EICrecon/#/table_flags/flags_view

### What kind of change does this PR introduce?
- [x] Bug fix (issue: reco_flags duplicates parameters)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #582, opened 2023-04-04
EcalEndcapNRawHits: revert back to unitless definitions

veprbl:
The MeV multipliers were introduced in #408, but they don't make sense from the dimensional analysis point (terms are divided by sqrt(E), 1 and E respectively). Each element of the array would need to have its own unit, but the commonly accepted convention is to use fractions and assume GeV. This commit restores that convention. The actual behavior in reconstruction is not changed by this PR, because the values stay the same.

cc #573

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


wdconinc:
I think we should be explicit in units. Why not `1/sqrt(dd4hep::GeV)` or something like that?

johnny8266 (approved):

PR #583, opened 2023-04-05
EcalBarrelScFiRawHits: revert back to unitless definitions

veprbl:
The current value was introduced in 13b9c92188d9cacedc5c4b9ca0c80404c21f1c3e. There should have been three values, and using units for those doesn't make sense from the dimensional analysis point (terms are divided by sqrt(E), 1 and E respectively). Each element of the array would need to have its own unit, but the commonly accepted convention is to use fractions and assume GeV. The actual behavior in reconstruction is not changed by this PR, because the values stay the same.

See also https://github.com/eic/EICrecon/issues/573

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


wdconinc (approved):
Looks good to me here, but even with zero I think there's a benefit to being explicit with the (correct) units.

PR #584, opened 2023-04-05
EcalEndcapP{,Insert}RawHits: revert back to unitless definitions

veprbl:
The MeV multipliers were introduced in #408, but they don't make sense from the dimensional analysis point (terms are divided by sqrt(E), 1 and E respectively). Each element of the array would need to have its own unit, but the commonly accepted convention is to use fractions and assume GeV. This commit restores that convention. The actual behavior in reconstruction is not changed by this PR, because the values stay the same.

See also https://github.com/eic/EICrecon/issues/573

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/detectors/EEMC/RawCalorimeterHit_factory_EcalEndcapPInsertRawHits.h
+++ b/src/detectors/EEMC/RawCalorimeterHit_factory_EcalEndcapPInsertRawHits.h
@@ -36,7 +36,7 @@ public:
 
         // Set default values for all config. parameters in CalorimeterHitDigi algorithm
         m_input_tag = "EcalEndcapPInsertHits";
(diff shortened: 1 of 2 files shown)


rymilton (commented):
Weird that they were multiplied by MeV. Thanks for changing this!

Will wait to merge until @jizhongling also approves since he's in charge of the ECal.

jizhongling:
I also wanted to make this change before. So it is good to change to unitless values.

veprbl:
@rymilton @jizhongling Can you, please, approve this PR then.

rymilton (approved):

PR #594, opened 2023-04-08
Tune the single photon energy resolution to 11%/sqrt(E [GeV]) \oplus 2%

jizhongling:
This PR changes the default energy resolution of pECal to 11%/sqrt(E [GeV]) \oplus 2% for single photon input.
This is suggested by Oleg.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: Resolution tuning

No breaking changes and users do not need to change their code.

Yes, it changes the default energy resolution of pECal.


veprbl:
Could you explain how values that are a fraction of a percent can make a difference for you?

jizhongling:
> Could you explain how values that are a fraction of a percent can make a difference for you?

The smearing is done for each tower. There are tens of towers for each cluster. As a result, a fraction of the percent of the tower smearing can give percent level clustering smearing.

veprbl:
> > Could you explain how values that are a fraction of a percent can make a difference for you?
> 
> The smearing is done for each tower. There are tens of towers for each cluster. As a result, a fraction of the percent of the tower smearing can give percent level clustering smearing.

From the point of statistics, what you say doesn't make sense:

$\delta \\{ \sum E_i \\} = \sqrt{ \sum (\delta E_i)^2 } = \sqrt{\sum (\frac{\delta E_i}{E_i})^2 E_i^2} \sim \frac{\delta \\{E_{tower}\\}}{E_{tower}} \times \sqrt{N_{eff}}$

The relative uncertainty on $E_{tot.} = \sum E_i$ is then $~\frac{1}{N_{eff}}$.

Is the effect, instead, coming from the $1/\sqrt{E}$ term?

How did you obtain the values for those sub-percent numbers?

jizhongling:
> > > Could you explain how values that are a fraction of a percent can make a difference for you?
> > 
> > 
> > The smearing is done for each tower. There are tens of towers for each cluster. As a result, a fraction of the percent of the tower smearing can give percent level clustering smearing.
> 
> From the point of statistics, what you say doesn't make sense:
> 
> δ{∑Ei}=∑(δEi)2=∑(δEiEi)2Ei2∼δ{Etower}Etower×Neff
> 
> The relative uncertainty on Etot.=∑Ei is then  1Neff.
> 
> Is the effect, instead, coming from the 1/E term?
> 
> How did you obtain the values for those sub-percent numbers?

I agree with you that if each tower has the same energy then $\delta E_{total} = \frac{\delta E_{tower}}{E_{tower}} / \sqrt{N_{eff}}$.

However, if the majority of energy is deposited in one tower and the cluster contains tens of towers with small energy, then each tower can contribute a constant $\delta E_{tower} = E_{tower} * \frac{a^2}{E_{tower}}$, where $\frac{a}{\sqrt{E}}$ is the E dependent resolution.

I tried to build a simple toy model to simulate the relation between tower smearing and cluster resolution. I got the same order of magnitude but I didn't get the exact results. I obtained the sub-percent numbers by running and tuning the EICRecon simulations.

FriederikeBock:
@jizhongling are you tuning such that you meet your test beam results? Could you post the obtained E-resolution as a function of E for the clusters? I think that's the only one we can judge it on right now.

jizhongling:
> @jizhongling are you tuning such that you meet your test beam results? Could you post the obtained E-resolution as a function of E for the clusters? I think that's the only one we can judge it on right now.

[Slide 2](https://indico.bnl.gov/event/19173/contributions/75657/attachments/47094/79873/main.pdf) is the tuning plot. I will discuss this in tomorrow's Calo meeting.

veprbl (dismissed):
Approve for now, assuming further studies can shed some light on this.

veprbl (approved):

PR #599, opened 2023-04-11
feat: associate MC photons in `PhotoMultiplierHitDigi` output

c-dilks:
Use `edm4eic::MCRecoTrackerHitAssociation` along with `edm4eic::RawTrackerHit`, to include the MC photon information in the output of `PhotoMultiplierHitDigi`

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

In general no breaking changes, but:
- until a new version of `EDM4eic` is released, which includes datatype `MCRecoTrackerHitAssociation`, this will not build in `eic-shell`
- may introduce conflicts with #555, but likely straightforward to resolve
only for the dRICH (the only user of `PhotoMultiplierHitDigi`)


c-dilks:
Ready when new version of `EDM4eic` is released.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Chao Peng
+// Copyright (C) 2023 Chao Peng, Thomas Britton, Christopher Dilks
```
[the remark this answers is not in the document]
wdconinc: Also elsewhere.

src/detectors/DRICH/DRICH.cc
```diff
@@ -24,9 +24,10 @@ extern "C" {
 
     // Digitization
     PhotoMultiplierHitDigiConfig digi_cfg;
-    digi_cfg.seed            = 0;
+    digi_cfg.seed            = 5; // FIXME: set to 0 for a 'unique' seed, but
+                                  // that seems to delay the RNG from actually randomizing
     digi_cfg.hitTimeWindow   = 20.0; // [ns]
-    digi_cfg.timeStep        = 0.0625; // [ns]
+    digi_cfg.timeStep        = 16.0; // [ns^-1]
```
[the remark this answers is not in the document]
wdconinc: I just looked at some TDC spec sheets and I think this quantity is usually expressed as time, not reciprocal. Instead of `step`, maybe use `LSB` or `resolution`, which is how it is described in spec sheets (resolution is also how it is done for silicon tracker digitization).

src/algorithms/digi/PhotoMultiplierHitDigi.cc
c-dilks, line 2:
```suggestion
// Copyright (C) 2022, 2023, Chao Peng, Thomas Britton, Christopher Dilks
```

src/algorithms/digi/PhotoMultiplierHitDigi.h
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Chao Peng
+// Copyright (C) 2023 Chao Peng, Thomas Britton, Christopher Dilks
```
c-dilks, line 2:
```suggestion
// Copyright (C) 2022, 2023, Chao Peng, Thomas Britton, Christopher Dilks
```

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
```diff
@@ -7,19 +10,20 @@ namespace eicrecon {
...
+      double timeStep      = 16.0;  // number of TDC counts per unit time // [ns^-1]
+      double speMean       = 80.0;  // mean ADC counts for a single photon
+      double speError      = 16.0;  // sigma of ADC counts for a single photon
+      double pedMean       = 200.0; // mean ADC counts for the pedestal
+      double pedError      = 3.0;   // sigma of ADC counts for the pedestal
 
       // SiPM pixels
       bool   enablePixelGaps = false; // enable/disable removal of hits in gaps between pixels
-      double pixelSize       = 3.0;   // [mm] // pixel (active) size
+      double pixelSize       = 3.0;   // pixel (active) size // [mm]
```
wdconinc, line 26: Multiply with units, e.g.
```suggestion
      double pixelSize       = 3.0 * dd4hep::mm;   // pixel (active) size
```

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
```diff
@@ -1,3 +1,6 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023 Christopher Dilks
```
c-dilks, line 2:
```suggestion
// Copyright (C) 2022, 2023, Christopher Dilks
```

src/detectors/DRICH/DRICH.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022, Christopher Dilks
+// Copyright 2023, Christopher Dilks
```
c-dilks, line 1:
```suggestion
// Copyright (C) 2022, 2023, Christopher Dilks
```

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
```diff
@@ -7,19 +10,20 @@ namespace eicrecon {
     public:
 
       // random number generator seed
-      unsigned long seed = 0;
+      // FIXME: don't use 0 if `TRandomMixMax` is the RNG, it can get "stuck"
```
wdconinc, line 13: How about we fail out when `seed == 0` and TRandomMixMax is used? Can we implement that? (another PR probably)

src/global/digi/PhotoMultiplierHitDigi_factory.cc
```diff
@@ -1,23 +1,20 @@
-// Created by Christopher Dilks
-// Based on SiliconTrackerDigi_factory
-// Subject to the terms in the LICENSE file found in the top-level directory.
-//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023 Christopher Dilks
```
c-dilks, line 2:
```suggestion
// Copyright (C) 2022, 2023, Christopher Dilks
```

src/global/digi/PhotoMultiplierHitDigi_factory.h
```diff
@@ -1,7 +1,5 @@
-// Created by Christopher Dilks
-// Based on SiliconTrackerDigi_factory
-// Subject to the terms in the LICENSE file found in the top-level directory.
-//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023 Christopher Dilks
```
c-dilks, line 2:
```suggestion
// Copyright (C) 2022, 2023, Christopher Dilks
```

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -63,23 +63,25 @@ void eicrecon::PhotoMultiplierHitDigi::AlgorithmChangeRun() {
...
-          double time;
+          decltype(edm4hep::SimTrackerHitData::time) time;
           dd4hep::Position pos;
+          std::vector<const edm4hep::SimTrackerHit*> mc_hits;
         };
         std::unordered_map<decltype(edm4eic::RawTrackerHitData::cellID), std::vector<HitData>> hit_groups;
         // collect the photon hit in the same cell
         // calculate signal
-        for(const auto& ahit : sim_hits) {
+        for(auto &ahit : sim_hits) {
```
wdconinc, line 81: Why do we lose `const` here? Well, probably we don't lose it and it's inside the `auto`, but manifest `const auto` is preferred .Also, `AlgorithmProcess` probably needs to take a `const std::vector<>& sim_hits`.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -63,23 +63,25 @@ void eicrecon::PhotoMultiplierHitDigi::AlgorithmChangeRun() {
 //------------------------
 // AlgorithmProcess
 //------------------------
-std::vector<edm4eic::RawTrackerHit*>
-eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(std::vector<const edm4hep::SimTrackerHit*>& sim_hits) {
-
+std::vector<edm4eic::MCRecoTrackerHitAssociation*> eicrecon::PhotoMultiplierHitDigi::AlgorithmProcess(
```
wdconinc, line 66: Does it make sense to wait a few days until we can return both the RawTrackerHit collection AND the (optional) MCRecoTrackerHitAssocation collection? #578

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
c-dilks: Added a warning and changed the default to `1`.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
c-dilks: Sure. The noise hits from #555 won't have associated `SimTrackerHit` collections, so in that case the output `RawTrackerHit` collection would have more hits than the output association collection. Or do you think we should fill the association collection with noise hits too, but with empty `SimTrackerHit`s? I can make either case work in the IRT wrapper algorithm (which has "cheat modes" to access the true photon information, if desired).

wdconinc: No, I think it's fine (correct, even) to not have associations for noise hits. Associations don't need to provide an association for every entry, but every association must point to a valid entry in each collection.

src/detectors/DRICH/DRICH.cc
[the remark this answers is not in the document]
c-dilks: resolved by 5abba9e37f9e56751d69a5c60871e4a67152a4f2

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
c-dilks: I thought we agreed to convert numbers to a standard system [mm,GeV,ns,radians] as soon as possible, so putting a configuration parameter in the DD4hep unit system is not necessary; instead any numbers read from the DD4hep service should also be converted to the standard system by division. Are we changing the standard again?

src/algorithms/digi/PhotoMultiplierHitDigi.cc
c-dilks: No idea, reverted, and now using `const std::vector<>& sim_hits` as suggested.

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
wdconinc: Yes, but we put them into this standard unit system by entering them with those units when we provide a numerical value. Starting from then on they are in our unit system. We should not just enter their value without a unit, because then it is not obvious which units they are in. See also EICrecon factory parameters.

c-dilks: So are we now re-allowing config params to be in DD4hep units? Should this be enforced as a standard? That is fine with me, but should be done in a separate PR.
In that case, https://github.com/eic/EICrecon/pull/390 could be useful.

c-dilks: (when/if we move to some config param `yaml` file or something)

c-dilks:
Next commits will use the new MultiFactory

src/algorithms/digi/PhotoMultiplierHitDigiConfig.h
wdconinc: We do want units on parameters, and they should be dd4hep units. That's what all algorithms and factories are doing, e.g. https://github.com/eic/EICrecon/blob/main/src/algorithms/calorimetry/CalorimeterHitDigi.h#L47 and https://github.com/eic/EICrecon/blob/main/src/detectors/BEMC/CalorimeterHit_factory_EcalBarrelImagingRecHits.h#L37. Only time doesn't always have units because dd4hep doesn't have a consistent unit for times.

c-dilks:
> Next commits will use the new MultiFactory

Working on https://github.com/eic/EICrecon/pull/612 to keep this cleaner. Doesn't work yet...

src/algorithms/digi/PhotoMultiplierHitDigi.cc
c-dilks: I've now done this, but in a separate PR which targets merging into this one, so that the additional changes are clear. See https://github.com/eic/EICrecon/pull/612, which has been tested locally and works fine.

wdconinc:
> Let's wait til EDM4eic 2.0.0 makes it to the CVMFS image

It builds... but poor scaling on a new build mechanism is causing the eic_container to eats its own tail in the scheduled nightly jobs... Working on that.

wdconinc:
`cvmfs` sync fixed; restarted check.

c-dilks:
CI passed, updating w.r.t. `main`

c-dilks:
And confirmed these files are all identical to the ones on my `irt-algo` dev branch, where I do the real testing with the dRICH.

wdconinc (commented):
Just minor comments.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -30,6 +30,7 @@ void eicrecon::PhotoMultiplierHitDigi::AlgorithmInit(dd4hep::Detector *detector,
     m_cfg.Print(m_log, spdlog::level::debug);
 
     // random number generators
+    if(m_cfg.seed==0) m_log->warn("using seed=0 may cause thread-unsafe behavior of TRandom"); // FIXME: remove when resolved
```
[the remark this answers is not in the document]
c-dilks: Done, and added checkboxes to https://github.com/eic/EICrecon/issues/539 description.

src/global/digi/PhotoMultiplierHitDigi_factory.cc
```diff
@@ -49,24 +36,21 @@ void eicrecon::PhotoMultiplierHitDigi_factory::Init() {
...
-  std::vector<const edm4hep::SimTrackerHit*> sim_hits;
-  for(const auto &input_tag: GetInputTags()) {
-    try {
-      for (const auto hit : event->Get<edm4hep::SimTrackerHit>(input_tag))
-        sim_hits.push_back(hit);
-    } catch(std::exception &e) {
-      m_log->critical(e.what());
-      throw JException(e.what());
-    }
+  auto sim_hits = static_cast<const edm4hep::SimTrackerHitCollection*>(event->GetCollectionBase(GetInputTags()[0]));
```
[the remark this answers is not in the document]
wdconinc: This is something @nathanwbrei is aware of and will be fixed in a newer JANA2 version.

PR #601, opened 2023-04-12
cmake: don't run ls -R on configuration

veprbl:
This avoids polluted logs. Users can be directed to use `ls -R`, `find .` or `cmake --trace` instead.


wdconinc (approved):
Yes, so much text! Thanks for removing.

wdconinc (approved):

PR #608, opened 2023-04-13
feat: various features and fixes in `richgeo` for IRT PID

c-dilks:
Changes to the `richgeo` geometry service:
- `ActsGeo` (not to be confused with the ACTS service)
  - add function to quickly determine if a point is within a radiator
- `IrtGeo`
  - add `cellID` -> pixel position converter (which corrects the position to the sensor surface), and should be faster than `dd4hep::rec::CellIDPositionConverter`
  - add refractive index table (used by IRT)
- `RichGeo`
  - new class `Sensor` to hold some sensor information
- multiple units fixes: strip DD4hep units as soon as possible, preferring defaults such as `mm`; this has been tested thoroughly with https://github.com/eic/EICrecon/pull/393
- replace local singleton logger by EICrecon-standard `spdlog`

- [x] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Tests for the changes have been added (see #393)

no (except for a small issue elsewhere, resolved by https://github.com/eic/drich-dev/pull/63)
no


src/services/geometry/richgeo/ActsGeo.cc
```diff
@@ -1,10 +1,38 @@
...
+{
+  // capitalize m_detName
+  std::transform(m_detName.begin(), m_detName.end(), m_detName.begin(), ::toupper);
+
+  // set `WithinRadiator`
+  if(m_detName=="DRICH") {
+    auto zmax             = m_det->constant<double>("DRICH_zmax")  / dd4hep::mm;
+    auto aerogelZpos      = m_det->constant<double>("DRICH_aerogel_zpos")      / dd4hep::mm;
+    auto aerogelThickness = m_det->constant<double>("DRICH_aerogel_thickness") / dd4hep::mm;
+    WithinRadiator[richgeo::kAerogel] = [this,aerogelZpos,aerogelThickness] (double x, double y, double z) {
```
wdconinc, line 18: Does it make sense to change the capture of `this` to `const` here? Even make the lambda itself const?

src/services/geometry/richgeo/IrtGeo.cc
```diff
@@ -1,51 +1,83 @@
...
   // IRT geometry handles
   m_irtDetectorCollection = new CherenkovDetectorCollection();
   m_irtDetector = m_irtDetectorCollection->AddNewDetector(m_detName.c_str());
+
+  // cellID conversion
+  m_cellid_converter = std::make_shared<const dd4hep::rec::CellIDPositionConverter>(*m_det);
+
+  // define the `cell ID -> pixel position` converter, correcting the returned
+  // `Position` to be at the sensor surface
+  m_irtDetector->m_ReadoutIDToPosition = [this] (auto cell_id) {
```
wdconinc, line 41: Same here: `this` as `const` capture? I'm worried about potential for unintentional thread unsafe write when using non const operations on some things that come in through the captures.

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -75,58 +73,58 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
     m_irtDetector->AddOpticalBoundary(isec, mirrorOpticalBoundary);
-    m_log.PrintLog("");
-    m_log.PrintLog("  SECTOR {:d} MIRROR:", isec);
-    m_log.PrintLog("    mirror x = {:f} cm", mirrorCenter.x());
-    m_log.PrintLog("    mirror y = {:f} cm", mirrorCenter.y());
-    m_log.PrintLog("    mirror z = {:f} cm", mirrorCenter.z());
-    m_log.PrintLog("    mirror R = {:f} cm", mirrorRadius);
+    m_log->debug("");
+    m_log->debug("  SECTOR {:d} MIRROR:", isec);
+    m_log->debug("    mirror x = {:f} mm", mirrorCenter.x());
```
wdconinc, line 106: Canonical for unit quantities:
```suggestion
    m_log->debug("    mirror x = {:f} mm", mirrorCenter.x() / dd4hep::mm);
```
Even if dd4hep changes internally to cm, it would still print in mm here.

wdconinc (commented):
Looks good. I'm a bit confused by the dividing out of units now. E.g.
```
  auto sensorSize       = m_det->constant<double>("PFRICH_sensor_size") / dd4hep::mm;
```
Why divide out `mm` when you still have a variable that should be in some unit system. Don't you want to keep `richgeo::Sensor::size` in dd4hep units too?

src/services/geometry/richgeo/IrtGeoDRICH.cc
c-dilks: Dividing by `dd4hep::mm` already happened a few lines above:
https://github.com/eic/EICrecon/blob/433a93afc02377b90e79e7f33fcb71a146c2c39e/src/services/geometry/richgeo/IrtGeoDRICH.cc#L92-L96

src/services/geometry/richgeo/ActsGeo.cc
c-dilks: `this` was only captured here to access `m_log`, and actually I only ever used `WithinRadiator` in #619, and no longer need it, so it can be removed entirely.

c-dilks:
> Looks good. I'm a bit confused by the dividing out of units now. E.g.
> 
> ```
>   auto sensorSize       = m_det->constant<double>("PFRICH_sensor_size") / dd4hep::mm;
> ```
> 
> Why divide out `mm` when you still have a variable that should be in some unit system. Don't you want to keep `richgeo::Sensor::size` in dd4hep units too?

This approach comes from back when we decided to use EDM4hep units here as the standard unit system:
```
1 = millimeter = GeV = nanosecond = radian
```
 Since this `PFRICH_sensor_size` constant is originally in DD4hep units, and the EDM4hep length unit is `mm`, I divide the `dd4hep::mm` so `sensorSize` is definitely in `mm`; this was the suggested approach by @sly2j around that time: convert any quantity in a non-EDM4hep unit system to the EDM4hep unit system ASAP, and only put it in some other system when strictly necessary.

Have we changed this approach or am I misunderstanding what we all prefer? Are we now preferring DD4hep units as the standard internal unit system (since in another PR you mentioned I should be putting JANA configuration parameters in DD4hep units too)?

src/services/geometry/richgeo/IrtGeo.cc
c-dilks: I couldn't find an example of `const` capture, could you clarify what you mean?

I'm wondering if instead I could capture `*this` (a copy of `this`), but the way it's currently written I don't think that will work since this lambda is defined before members such as `m_irtDetector` are completely built.

I can move the definition of this lambda to somewhere after the `DD4hep_to_IRT` (which sets everything that's needed in the capture), and then capturing `*this` might be okay, since that's the version of `this` I want to copy. My thought is that if `[*this]` indeed makes a copy, then it doesn't matter what happens to the original `this`. What do you think?

An alternative approach is a giant `std::unordered_map`... but I think that would be slower since this lambda already uses the CellID-to-position converter, and the number of hits is << the number of pixels.

c-dilks: Capturing `*this` leads to a double free error, and requires `IrtGeo` to have no pure virtual functions.

Instead, I propose capturing only the members I need by value (except for the logger, captured by reference). I also moved the definition of the lambda to _after_ `DD4hep_to_IRT`, so that everything it captures is fully defined.

```cpp
  m_irtDetector->m_ReadoutIDToPosition = [
    &m_log = this->m_log, // capture logger by reference
    // capture instance members by value, so those owned by `this` are not mutable here
    cell_mask        = this->m_irtDetector->GetReadoutCellMask(),
    cellid_converter = this->m_cellid_converter,
    sensor_info      = this->m_sensor_info
  ] (auto cell_id) {
    // ...
  };
```

c-dilks: changes: 446c18df83163df237e7aae15202cc8a67e476ff

c-dilks: tested in https://github.com/eic/EICrecon/pull/393 and works.

c-dilks:
> Have we changed this approach or am I misunderstanding what we all prefer? Are we now preferring DD4hep units as the standard internal unit system (since in another PR you mentioned I should be putting JANA configuration parameters in DD4hep units too)?

If we have changed this approach, I would prefer to comply in a separate PR, where I can fix `PhotoMultiplierHitDigi`, and any others.

c-dilks:
Ready for re-review; new commits since previous review start from commit [refactor: remove RichGeo::WithinRadiator lambda](https://github.com/eic/EICrecon/pull/608/commits/31b4939b66d2fb40ca2fd0c362171963a00df54e) onward

src/services/geometry/richgeo/RichGeo_service.h
```diff
@@ -30,6 +28,9 @@ class RichGeo_service : public JService {
     RichGeo_service(JApplication *app) : m_app(app) {}
     virtual ~RichGeo_service();
 
+    // return pointer to the main DD4hep Detector
+    dd4hep::Detector *GetDD4hepGeo() { return m_dd4hepGeo; };
```
c-dilks, line 32: @veprbl should `GetDD4hepGeo` be virtual too?

veprbl: Yes

c-dilks:
Can we merge this soon? I'm trying to clean up some things in [eic/drich-dev](https://github.com/eic/drich-dev) in preparation for dRICH software tutorials, and this PR is blocking some PRs there. This PR does not impact production running, since the dRICH IRT is not in production.

veprbl (approved):

PR #609, opened 2023-04-13
Updated a calculation in seeder configuration to ACTS units

bschmookler:
This PR modifies the calculation of the transverse momentum per unit radius so that it is now in ACTS units. The original equation seems to have been copied from here:

https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/SeedFinderConfig.hpp#L258

It is mentioned there that the calculation should be updated to ACTS units.

In any event, this change seems to have a negligible effect on the output of the seed finder.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc (approved):
I guess it's a good thing that being 300 times off is still resulting in a negligible effect...

PR #610, opened 2023-04-13
BUGFIX: CalorimeterIslandCluster: compare absolute values for distance-based adjacency settings

veprbl:
Currently, if the adjacency is defined using distance-based options like u_localDistXY, there will be cases when distance vector reported by hitsDist will have negative components. When those components are compared to positive numbers, this leads to an always-true neigbourhood condition. The clusters may be wrongly merged because of that.


Chao1009 (approved):
Thanks for the fix!

PR #611, opened 2023-04-14
EcalEndcapNIslandProtoClusters: switch to using adjacencyMatrix

veprbl:
Switching to adjacencyMatrix, assuming we are going to migrate other detectors to it as well.


wdconinc:
Rebased to pick up PODIO frames.

PR #612, opened 2023-04-14
feat: output raw hits and associations from `PhotomultiplierHitDigi`

c-dilks:

**NOTE**: This PR targets the branch associated with https://github.com/eic/EICrecon/pull/599, and therefore adds on to it.

- output both the raw hits and associations, using the new `JMultifactory` feature
  - new `struct PhotoMultiplierHitDigiResult` is used as the return value of `AlgorithmProcess` (rather than a `std::pair`, for better flexibility and readability)
- switch from `std::vector`s of raw pointers to PODIO collections
- unrelated: style changes to `DRICH.cc` to be more consistent with #393 (we are factorizing that giant PR into smaller PRs, and these changes will make this process a bit easier)

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no

no

### DIFF
--- a/src/algorithms/digi/PhotoMultiplierHitDigi.cc
+++ b/src/algorithms/digi/PhotoMultiplierHitDigi.cc
@@ -79,10 +79,11 @@ eicrecon::PhotoMultiplierHitDigiResult eicrecon::PhotoMultiplierHitDigi::Algorit
         std::unordered_map<decltype(edm4eic::RawTrackerHitData::cellID), std::vector<HitData>> hit_groups;
         // collect the photon hit in the same cell
         // calculate signal
+        m_log->trace("{:-<70}","Loop over simulated hits ");
         for(const auto& sim_hit : *sim_hits) {
             auto edep_eV = sim_hit.getEDep() * 1e9; // [GeV] -> [eV] // FIXME: use common unit converters, when available
             auto id      = sim_hit.getCellID();
-            m_log->trace("hit: pixel id={:#X}  edep = {} eV", id, edep_eV);
+            m_log->trace("hit: pixel id={:#018X}  edep = {} eV", id, edep_eV);
 
             // overall safety factor
             if (m_rngUni() > m_cfg.safetyFactor) continue;
@@ -102,6 +103,7 @@ eicrecon::PhotoMultiplierHitDigiResult eicrecon::PhotoMultiplierHitDigi::Algorit
             }
 
             // cell time, signal amplitude, truth photon
+            m_log->trace(" -> hit accepted");
             auto   time = sim_hit.getTime();
             double amp  = m_cfg.speMean + m_rngNorm() * m_cfg.speError;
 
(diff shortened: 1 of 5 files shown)


c-dilks:
Ready for review, but CI checks won't run yet until EDM4eic v2.0.0 is in the image.

c-dilks:
Hold up, found a bug

c-dilks:
> Hold up, found a bug

Fixed by the removal of another pesky raw pointer. Downstream IRT code is confirmed to be fully functional with this PR.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -111,66 +115,69 @@ std::vector<edm4eic::MCRecoTrackerHitAssociation*> eicrecon::PhotoMultiplierHitD
...
                 hit_assoc.setRawHit(raw_hit);
-                for(auto &mc_hit : data.mc_hits)
-                  hit_assoc.addToSimHits(*mc_hit);
-                hit_assocs.push_back(new edm4eic::MCRecoTrackerHitAssociation(hit_assoc)); // force immutable
+                for(auto i : data.sim_hit_indices)
+                  hit_assoc.addToSimHits(sim_hits->at(i));
             }
         }
-        return hit_assocs;
+        return std::move(result);
```
wdconinc, line 180: https://www.youtube.com/watch?t=2284

src/algorithms/digi/PhotoMultiplierHitDigi.h
```diff
@@ -28,15 +28,20 @@
 
 namespace eicrecon {
 
+struct PhotoMultiplierHitDigiResult {
+  std::unique_ptr<edm4eic::RawTrackerHitCollection> raw_hits;
+  std::unique_ptr<edm4eic::MCRecoTrackerHitAssociationCollection> hit_assocs;
+};
```
wdconinc, line 34: This is now not necessary anymore thanks to multifactories which allow a simple return value of `std::tuple<edm4eic::RawTrackerHitCollection, edm4eic::MCRecoTrackerHitAssociationCollection>`, and you can `return {raw_hits, hit_assocs}`. As applied here: https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/ParticlesFromTrackFit.cc#L140 (with probably unnecessary move semantics).

wdconinc: One could of course argue whether a struct or pair are really all that different.

wdconinc (approved):

src/algorithms/digi/PhotoMultiplierHitDigi.h
c-dilks: I'm aware of this trend, but I prefer a struct or class, since:
- easier to find what is included in the output when reading through `AlgorithmProcess`: just search for the returned struct instead of each of the tuple members
- easier to use from the caller, e.g., to get the raw hits, say `result.raw_hits` instead `result->first`; however, it is easier to loop over a tuple or pair, which can be convenient in factories

My preference isn't too strong, so if you'd rather this algorithm be more consistent with others, I'll switch it to a tuple.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
c-dilks: Good prank... my music player was having problems, then I clicked this link and thought I was having issues with Youtube too... anyway I'll drop the unnecessary `std::move`.

PR #616, opened 2023-04-18
Add 3D cluster shape analysis

dhevang:

Cluster shape properties are calculated from the 3D covariance matrix of hits within a cluster.
2 coordinate system options are provided: eta-phi-z and x-y-z, which can be toggled with the m_xyClusterProfiling variable.
This adds to the existing routine, which has been moved out of the header file and into the CalorimeterClusterRecoCoG.cc.

- [ ] Bug fix (issue #__)
- [ x ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -246,14 +246,16 @@ edm4eic::Cluster* CalorimeterClusterRecoCoG::reconstruct(const edm4eic::ProtoClu
   // Calculate cluster profile:
   //    radius,
   //   	dispersion (energy weighted radius),
-  //   	sigma_long
-  //   	sigma_short
-  //   	sigma_z
-  double radius = 0, dispersion = 0, lambda_1 = 0, lambda_2 = 0, lambda_3 = 0;
-  double w_sum = 0;
+  //    eta-phi cluster widths (2D)
+  //   	x-y-z cluster widths (3D)
+  float radius = 0, dispersion = 0, w_sum = 0;
 
-  Eigen::Matrix3f sum2 = Eigen::Matrix3f::Zero();
-  Eigen::Vector3f sum1 = Eigen::Vector3f::Zero();
+  Eigen::Matrix2f sum2_2D = Eigen::Matrix2f::Zero();
+  Eigen::Matrix3f sum2_3D = Eigen::Matrix3f::Zero();
+  Eigen::Vector2f sum1_2D = Eigen::Vector2f::Zero();
+  Eigen::Vector3f sum1_3D = Eigen::Vector3f::Zero();
+  Eigen::Vector2cf eigenValues_2D = Eigen::Vector2cf::Zero();
+  Eigen::Vector3cf eigenValues_3D = Eigen::Vector3cf::Zero();
 
   if (cl.getNhits() > 1) {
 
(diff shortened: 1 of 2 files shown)


src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -6,6 +6,7 @@
  *  Logarithmic weighting is used for mimicing energy deposit in transverse direction
  *
  *  Author: Chao Peng (ANL), 09/27/2020
+            Dhevan Gangadharan (UH): cluster profiling from Eigenvalues
```
veprbl, line 9: Please don't keep revision story in comments. Add yourself to the SPDX copyright section above instead. By doing that you confirm that you are the sole author of the changes and have rights to issue your code under LGPL3 or newer.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h
```diff
@@ -13,6 +13,8 @@
 
 #include <random>
 
+#include <Eigen/Dense>
```
veprbl, line 16: This header is an implementation detail, should go to the .cc file.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,3 +162,162 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
+      const auto delta = cl.getPosition() - hit.getPosition();
+      radius += delta * delta;
+
+      double w = weightFunc(hit.getEnergy(), cl.getEnergy(), m_logWeightBase, 0);
+      w_sum += w;
+      dispersion += delta * delta * w;
+
+      double pos_1 = edm4eic::anglePolar( hit.getPosition() );
+      double pos_2 = edm4eic::angleAzimuthal( hit.getPosition() );
+      double pos_3 = hit.getPosition().z;
```
veprbl, line 272: I don't understand how this works to put these into a matrix. From the point of dimensional analysis it would make sense to do:
```suggestion
      double pos_1 = edm4eic::anglePolar( hit.getPosition() ) * hit.getPosition().mag();
      double pos_2 = edm4eic::angleAzimuthal( hit.getPosition() ) * hit.getPosition().mag();
      double pos_3 = hit.getPosition().z;
```
But then, we might as well
```suggestion
      double pos_1 = hit.getPosition().x;
      double pos_2 = hit.getPosition().y;
      double pos_3 = hit.getPosition().z;
```
which is what `m_xyClusterProfiling` does. Perhaps, we only need one calculation variant always?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,3 +162,162 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
+      double sigma_33 = sum_33 / w_sum - (sum_3/w_sum) * (sum_3/w_sum);
+      double sigma_12 = sum_12 / w_sum - (sum_1/w_sum) * (sum_2/w_sum);
+      double sigma_13 = sum_13 / w_sum - (sum_1/w_sum) * (sum_3/w_sum);
+      double sigma_23 = sum_23 / w_sum - (sum_2/w_sum) * (sum_3/w_sum);
+
+      // covariance matrix
+      Eigen::MatrixXd cov3(3,3);
+      cov3(0,0) = sigma_11;  cov3(1,1) = sigma_22;  cov3(2,2) = sigma_33;
+      cov3(0,1) = sigma_12;  cov3(0,2) = sigma_13;  cov3(1,2) = sigma_23;
+      cov3(1,0) = cov3(0,1); cov3(2,0) = cov3(0,2); cov3(2,1) = cov3(1,2);
```
veprbl, line 305:
```suggestion
      Eigen::Matrix3f cov3;
      cov3 << sigma_11, sigma_12, sigma_13,
              sigma_12, sigma_22, sigma_23,
              sigma_13, sigma_23, sigma_33;
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,3 +162,162 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
+      }
+      sum_11 += w * pos_1 * pos_1;
+      sum_22 += w * pos_2 * pos_2;
+      sum_33 += w * pos_3 * pos_3;
+      sum_12 += w * pos_1 * pos_2;
+      sum_13 += w * pos_1 * pos_3;
+      sum_23 += w * pos_2 * pos_3;
+      sum_1  += w * pos_1;
+      sum_2  += w * pos_2;
+      sum_3  += w * pos_3;
```
veprbl, line 286: If `Vector3f pos`, then
```
sum2 += w * pos * pos.transpose();
sum += w * pos;
sumw += w;
```
...
```
sigma = sum2 / w_sum - sum * sum.transpose() / (sumw * sumw)
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,3 +162,162 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
+      lambda_2 = eigenValues[1].real();
+      lambda_3 = eigenValues[2].real();
+    }
+  }
+
+  cl.addToShapeParameters( radius );
+  cl.addToShapeParameters( dispersion );
+  cl.addToShapeParameters( lambda_1 );
+  cl.addToShapeParameters( lambda_2 );
+  cl.addToShapeParameters( lambda_3 );
```
veprbl, line 320: We need to document the meaning and units in
https://github.com/eic/EDM4eic/blob/c8ffba77708dfd01a9f1ee2134001d59e47920cb/edm4eic.yaml#L249

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,3 +162,162 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
+      // covariance matrix
+      Eigen::MatrixXd cov3(3,3);
+      cov3(0,0) = sigma_11;  cov3(1,1) = sigma_22;  cov3(2,2) = sigma_33;
+      cov3(0,1) = sigma_12;  cov3(0,2) = sigma_13;  cov3(1,2) = sigma_23;
+      cov3(1,0) = cov3(0,1); cov3(2,0) = cov3(0,2); cov3(2,1) = cov3(1,2);
+
+      // Eigenvalues correspond to cluster's 2nd moments (sigma_long, sigma_short, sigma_z)
+      Eigen::EigenSolver<Eigen::MatrixXd> es(cov3, false); // set to true for eigenvector calculation
+      auto eigenValues = es.eigenvalues();
+      lambda_1 = eigenValues[0].real(); // imaginary parts correspond to unphysical roots
```
veprbl, line 310:
```suggestion
      lambda_1 = eigenValues[0].real();
```
I believe, a symmetric matrix always has real eigenvalues.

veprbl (commented):
This is a nice change. I have few suggestions.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: Done.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: Yes, it's a bit odd to mix eta-phi with z, but I think the scale (and units) of each dimension may not really matter.  x-y-z dimensions is much more intuitive, I agree.  I included the eta-phi-z because other collaborations like ALICE have used eta-phi in their 2D analyses.  We can exclude the eta-phi-z for now if needed.  What do you think?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: Good suggestion!

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: Indeed.  I added this explanation to the comment

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: It's mentioned that one should alter the "metadata".  How do I do that?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h
dhevang: Done.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
veprbl: x-y-z makes sense and theta-phi or eta-phi makes sense, but not theta-phi-z. The dimension matters. For an argument, imagine, we present your algorithm not with mm (default length unit in EDM4hep), but in, say, meters or in micrometers. That will change directions of your eigenvectors. Also note that the z and the polar basis vectors are collinear at eta=0.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
veprbl: I don't think it's possible to do that now: AIDASoft/podio#407
For now, just updating the comment in EDM4eic repo would suffice.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: I now have both 2D (theta-phi) and 3D (x-y-z) shape analyses included separately.  All 5=2+3 eigenvalues are included in the shape parameter list.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,3 +161,159 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
+  // for simple 2D CoG clustering
+  cl.setIntrinsicTheta(edm4eic::anglePolar(cl.getPosition()));
+  cl.setIntrinsicPhi(edm4eic::angleAzimuthal(cl.getPosition()));
+  // TODO errors
+
+  //_______________________________________
+  // Calculate cluster profile:
+  //    radius,
+  //   	dispersion (energy weighted radius),
+  //    eta-phi cluster widths (2D)
```
veprbl, line 249: I would like to point out that eta is not given by `edm4eic::anglePolar`. A possible resolution could be:
```suggestion
  //    theta-phi cluster widths (2D)
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,3 +161,159 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
+      // eigenvalues of symmetric real matrix are always real
+      eigenValues_2D = es_2D.eigenvalues();
+      eigenValues_3D = es_3D.eigenvalues();
+    }
+  }
+
+  cl.addToShapeParameters( radius );
+  cl.addToShapeParameters( dispersion );
+  cl.addToShapeParameters( eigenValues_2D[0].real() ); // 2D eta-phi cluster width 1
+  cl.addToShapeParameters( eigenValues_2D[1].real() ); // 2D eta-phi cluster width 2
```
veprbl, line 313: Here as well

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: Yes, simple misprint in the comments.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
dhevang: Yes, same here.

veprbl (approved):
Looks good to me

PR #626, opened 2023-04-21
ImagingPixelReco: throw on errors in initialize()

veprbl:
Makes the algorithm throw an exception in `initialize` instead of dereferencing a null pointer in `process()`


wdconinc (approved):
Looks good.

PR #629, opened 2023-04-23
cmake: add BUILD_TESTING option, disable on CI for clang++

veprbl:
Fixes current CI

#### Context

Container was updated to add Catch2, but that package is somehow broken for clang++. Let's, for now, disable it.


wdconinc:
Still needed? Catch2 doesn't use LTO in latest container, 4pm.

Issue #625, opened 2023-04-21
An exception in factory `Init()` should cause termination with an error exit code


veprbl:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released):
- Which version (or `HEAD` for the most recent on git):
- Any specific OS or system where the issue occurs?
- Any special versions of ROOT or Geant4?

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1.
2.

### Expected Result: (what do you expect when you execute the steps above)
Reconstruction terminates

### Actual Result: (what do you get when you execute the steps above)
```
[2023-04-21 18:24:45.464] [JEventProcessorPODIO] [error] Omitting PODIO collection 'FooBarFactory' due to exception: Unknown exception in JFactoryT::Init().
```
and reconstruction continues

nathanwbrei:
I fully agree that we should remove the try/catch block inside JEventProcessorPODIO and let crashing factories crash rather than attempting to keep going. Most of the factories are crashing because there is a JEvent::Get<T>() for some T that is not present in the input file (due to geometry, etc). We can use the non-excepting forms of Get/GetSingle, or we can wrap them in the try/catch block instead.

PR #637, opened 2023-04-28
Fixes for Dmitry R.'s platform

nathanwbrei:
@DraTeots helped me quite a bit with PR #578. He made a lot of small fixes in order to get EICrecon to build on his system. For the sake of keeping the huge PR as manageable as possible, I didn't include these in the final PR. Rather than throw this work away, however, I cherry-picked the changes that *didn't* make it into main so that we can have a look at them and decide what to keep. 

Probably the most important fix in here is an issue with Eigen alignments. 

The changes related to exception handling are going to inform my response to issue #625. 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


veprbl:
Can we review these changes in separate PRs? Some are not uncontroversial.

nathanwbrei:
> Can we review these changes in separate PRs? Some are not uncontroversial.

Absolutely! I fully expect this us to cherry pick and/or drop some of these commits. The important thing is that we all see what the changes are, and more or less agree on which should make it into master. 

When it comes to specifics:

* Some of the changes relating to Eigen alignment look uncontroversial to me (e.g. the missing pragma once's). Although I'm pretty sure there was also some code that got commented out in there, and I don't know if that code was important.

* Removing the redundant `ParticlesWithAssociation` class (and some of the many many unused headers) seems like a good idea because it would reduce code entropy and might even improve compilation times slightly.

* The exception handling changes definitely need to be their own PR, and I am experimenting with doing it a little bit differently than what is here. 

* I have no idea what the changes to the CMake project structure entail or why we might want them. Are these the changes that you flagged as being possibly controversial?

wdconinc:
@nathanwbrei @veprbl @DraTeots Can this be closed?

veprbl:
I didn't see useful changes that weren't yet ported.

PR #605, opened 2023-04-13
algorithm_test: Implement a test for CalorimeterIslandCluster

veprbl:
Add first unit test for an algorithm.

PR #640, opened 2023-04-30
services: use virtual methods, like in Juggler

veprbl:

This includes a change needed for #605, separated for purposes of focused review.
Unit testing could benefit from a proper isolation of the plugins and services. Services implemented in Juggler define interfaces with virtual method dispatching. This PR does a small step towards that scheme.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


faustus123:
I'm not as familiar with juggler testing internals and don't quite understand this so I may need a little more explanation. 

I take it the main goal is to provide isolation for unit testing so one can say, create a test using code that needs the `JDD4hep_service`, but without actually needing to load the plugin that provides it. i.e. the test can use a dummy class that supports the interface without actually needing to link to `DD4hep`. 

I see how using an interface class would allow this, assuming the `JDD4hep_service` class inherits from that interface. In that case though, the interface class would define the virtual methods which confuses me as to why virtual methods are needed in `JDD4hep_service` itself. Does the scheme you're setting up include subclassing `JDD4hep_service`?

veprbl:
> I'm not as familiar with juggler testing internals and don't quite understand this so I may need a little more explanation.

I don't think there is testing in Juggler, so we are all mostly on the same page here. However I did find that Juggler already had decent services implemented, including some reasonable design for interfaces:
https://eicweb.phy.anl.gov/EIC/juggler/-/blob/master/JugBase/src/components/GeoSvc.h
It is really surprising to me that this was not simply ported to EICrecon. Maybe, I'm missing some context here.

> I take it the main goal is to provide isolation for unit testing so one can say, create a test using code that needs the `JDD4hep_service`, but without actually needing to load the plugin that provides it. i.e. the test can use a dummy class that supports the interface without actually needing to link to `DD4hep`.

Correct, we may want to implement mock services for our unit tests.

> I see how using an interface class would allow this, assuming the `JDD4hep_service` class inherits from that interface. In that case though, the interface class would define the virtual methods which confuses me as to why virtual methods are needed in `JDD4hep_service` itself. Does the scheme you're setting up include subclassing `JDD4hep_service`?

At this point, I'm not even suggesting adding any class hierarchy (like introducing an interface superclass like `IGeoSvc`). This will only make the calls virtual. That way the current tests won't require loading of service plugins for dynamic linker to resolve all the text symbols (virtual calls rely only on the vtable layout). This PR alone won't yet allow to test the functionality that requires those services, more work will be needed after that.

faustus123:
OK. I think I get it now. I'll put in a review to approve it.

faustus123 (approved):
This seems reasonable. It shouldn't change any of the existing functionality, but will do what Dmitry is wanting it to do: allow a more independent level of unit testing.

Issue #621, opened 2023-04-20
Improve ACTS seed phi angle calculation


bschmookler:
The ACTS seed phi is currently calculated at the first space point:

https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackSeeding.cc#L120

We can calculate it at the seed DCA to the origin using the approach described on slides 7-11 here:

https://indico.bnl.gov/event/18894/contributions/75735/attachments/47132/79949/ACTS%20seed%20reconstruction.pdf

PR #641, opened 2023-05-01
Updated seed phi calculation and seed finder configuration

bschmookler:
This PR updates the calculation of the azimuthal angles for the ACTS seeds as discussed on slides 7-11 here:

https://indico.bnl.gov/event/18894/contributions/75735/attachments/47132/79949/ACTS%20seed%20reconstruction.pdf

Also, this PR updates the seed finder configuration parameters to the most up-to-date values based Emma's and Rey's studies:

https://indico.bnl.gov/event/18894/contributions/75735/attachments/47132/79952/param%20exploration%20%282%29.pdf

Some default values for the seed filter are also set, but do not seem to be used in the track seeding algorithm right now.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #621)
- [ ] Documentation update
- [ ] Other: __

No
No


src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
```diff
@@ -1,102 +1,168 @@
...
+    float m_zMin = -1500. * Acts::UnitConstants::mm; // min z to look for hits to compose seeds
+    float m_deltaRMinTopSP = 10. * Acts::UnitConstants::mm; // Min distance in r between middle and top SP in one seed
+    float m_deltaRMaxTopSP = 200. * Acts::UnitConstants::mm; // Max distance in r between middle and top SP in one seed
+    float m_deltaRMinBottomSP = 10. * Acts::UnitConstants::mm; // Min distance in r between middle and bottom SP in one seed
+    float m_deltaRMaxBottomSP = 200. * Acts::UnitConstants::mm; // Max distance in r between middle and bottom SP in one seed
+    float m_collisionRegionMin = -250 * Acts::UnitConstants::mm; // Min z for primary vertex
+    float m_collisionRegionMax = 250 * Acts::UnitConstants::mm; // Max z for primary vertex
+
+    float m_maxSeedsPerSpM = 0; // max number of seeds a single middle sp can belong to - 1
+    float m_cotThetaMax = 27.29; // Cotangent of max theta angle (27.29 corresponds to eta = 4)
```
wdconinc, line 33:
```suggestion
    float m_cotThetaMax = 1.0 / tan(2. * atan(exp(-4.0)); // Cotangent of max theta angle (based on eta)
```

wdconinc (approved):
This looks good to me. I verified that all constants have units etc.

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
```diff
@@ -1,102 +1,168 @@
...
+    float m_sigmaScattering = 5; // How many standard devs of scattering angles to consider
+    float m_radLengthPerSeed = 0.1; // Average radiation lengths of material on the length of a seed
+    float m_minPt = (100. * Acts::UnitConstants::MeV) / m_cotThetaMax; // MeV (in Acts units of GeV) - minimum transverse momentum
+    float m_bFieldInZ = 1.7 * Acts::UnitConstants::T; // T (in Acts units of GeV/[e*mm]) - Magnetic field strength
+    float m_beamPosX = 0; // x offset for beam position
+    float m_beamPosY = 0; // y offset for beam position
+    float m_impactMax = 3. * Acts::UnitConstants::mm; // Maximum transverse PCA allowed
```
osbornjd, line 40: One thing to consider for the future (definitely not for this PR, as this is already a needed update) - this will (obviously) limit the seeds that are found to be located within 3 mm of the nominal x-y origin. For the vast majority of tracks that is fine, but long lived decays and/or secondaries will never be found. There are two options for future development:
1. deploy iterative tracking which runs an additional seeding pass, either with a re-tuned version of this seeder or some other seeder, to find the missing seeds
2. Continue tuning this set of parameters to allow for larger `m_impactMax`.

I would suggest option (1), but again this is definitely not for this PR and requires additional future development. I just thought I'd mention that the Acts seeder takes this value very seriously

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
```diff
@@ -1,102 +1,168 @@
...
+    float m_zMax = 1700. * Acts::UnitConstants::mm; // max z to look for hits to compose seeds
+    float m_zMin = -1500. * Acts::UnitConstants::mm; // min z to look for hits to compose seeds
+    float m_deltaRMinTopSP = 10. * Acts::UnitConstants::mm; // Min distance in r between middle and top SP in one seed
+    float m_deltaRMaxTopSP = 200. * Acts::UnitConstants::mm; // Max distance in r between middle and top SP in one seed
+    float m_deltaRMinBottomSP = 10. * Acts::UnitConstants::mm; // Min distance in r between middle and bottom SP in one seed
+    float m_deltaRMaxBottomSP = 200. * Acts::UnitConstants::mm; // Max distance in r between middle and bottom SP in one seed
+    float m_collisionRegionMin = -250 * Acts::UnitConstants::mm; // Min z for primary vertex
+    float m_collisionRegionMax = 250 * Acts::UnitConstants::mm; // Max z for primary vertex
+
+    float m_maxSeedsPerSpM = 0; // max number of seeds a single middle sp can belong to - 1
```
osbornjd, line 32: Is this a new definition in a newer version of Acts? As far as I'm aware this is just the maximum number of seeds a single middle spacepoint could belong to (not -1) - if this is true, this configuration should in principle return 0 seeds (I would naively think...) if the middle spacepoint can not belong to any triplet. Am I missing something or is this a typo?

osbornjd (dismissed):
Looks good to me other than my question about the `maxSeedsPerSpM` - thanks for updating all of these parameters @bschmookler . In any case I don't think it is a problem to merge this as it isn't explicitly being used in production, so these parameters can always continue to be updated further as more tuning is performed.

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
bschmookler: Thanks. Let me double check this with Emma. We definitely see seeds being generated when this parameter is set to 0, but maybe it defaults to some other value when set <1.

wdconinc: Is there a chance that Emma gets back quickly on this? It would be good to include this in the release today.

bschmookler: Emma is away and probably won't be able to reply until tomorrow. 

You can find the study she did for this parameter on slide 20 here:

https://indico.bnl.gov/event/18894/contributions/75735/attachments/47132/79952/param%20exploration%20%282%29.pdf

osbornjd: I think it's fine to merge for now, as in any case we know from the previous study that these parameters are better than the current (wrong) ones. I think we should double check this in the future though, I'll look through the seeding code tomorrow morning to double check whether or not this is either somehow just skipped over, not representative of what I think it is, or is somehow defaulting to some other value (or none of these).

osbornjd (dismissed):

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
```diff
@@ -1,102 +1,168 @@
...
+    float m_cotThetaMax = 1.0 / tan(2. * atan(exp(-4.0)); // Cotangent of max theta angle (based on eta)
```
wdconinc, line 33:
```suggestion
    float m_cotThetaMax = 1.0 / tan(2. * atan(exp(-4.0))); // Cotangent of max theta angle (based on eta)

```

wdconinc (changes requested):

wdconinc (approved):

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
wdconinc: Deferred to #648

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
osbornjd: Just to follow up on this, [here](https://github.com/acts-project/acts/blob/3b4b5c741c8541491d496a36b917b00b344d52d1/Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp#L49) is the definition of this parameter in Actsv21.1.0 orthogonal seeder config.

Doing a grep in my own repo for v21.1.0 I find that `maxSeedsPerSpM` is used only in `SeedFilter.ipp` [here](https://github.com/acts-project/acts/blob/3b4b5c741c8541491d496a36b917b00b344d52d1/Core/include/Acts/Seeding/SeedFilter.ipp#L270-L272). As you suggested, the algorithm makes a cut on the total number of allowed seeds+1, so the current value of 0 basically allows only 1 seed per middle spacepoint. This is probably (partly) why there are so many duplicates, especially at central rapidities, as a single space point can belong to 1 seed but the seed volume to search within encompasses >3 layers. At some point down the road it may be worth trying to optimize this value in conjunction with the seed search volume to limit the number of duplicates (while obviously also maintaining high efficiency). But in any case, my apologies for the confusion - evidently the documentation associated with the parameter itself should be updated in Acts.

bschmookler: Hi Joe. Thanks for confirming this.

Issue #628, opened 2023-04-22
EICrecon takes 20 minutes to build


veprbl:
```
time cmake --build build -- -j1
```
```output
real    22m48.523s
user    21m30.882s
sys     1m0.390s
```

veprbl:
An output of compiling with `-DCMAKE_CXX_COMPILER_LAUNCHER=time -DCMAKE_CXX_LINKER_LAUNCHER=time` suggests that the time spent linking is negligible.

veprbl:
Also, apparently, it's not related to datamodel_glue.h or datamodel_includes.h.

nathanwbrei:
I've heard that clang has some nifty compile-time profiling tools, though I haven't tried them out yet. I'd be interested in how you profile this. Or maybe I should figure it out.

veprbl:
> I've heard that clang has some nifty compile-time profiling tools, though I haven't tried them out yet.

Glad that you've posted this sentence. I wasn't able to figure out how to ask google about this. `-ftime-trace` looks promising.

veprbl:
Looking at few files, Eigen is a big offender, but that's what it does. There are some "spdlog" and "JANA" headers that don't take much time (~1s?), but they should add up over the codebase. It came to me as a surprise, but a lot of time is spent in the "backend" (even for a `-DCMAKE_BUILD_TYPE=` build).

<img width="526" alt="image" src="https://user-images.githubusercontent.com/245573/234119559-c286d440-a2bc-4380-8c9b-6318841edb67.png">

veprbl:
I've tried using include-what-you-use (IWYU), and it gave me about 1 minute speedup (5%). Not much, given that it requires quite a bit of tuning to get a nice and working diff.

veprbl:
Sorry, not sure what happened. I didn't mean to forcefully assign anyone.

nathanwbrei:
No worries! I just assumed you gave up in frustration and were hoping I'd think of something

wdconinc:
Yeah, I noticed that GitHub had attached this to the #784 but I couldn't figure out how to detach it again. That PR explicitly doesn't address this and likely makes it worse :-D

veprbl:
Robots don't get irony

DraTeots:
Interesting insights using this tool: 
https://github.com/aras-p/ClangBuildAnalyzer

[...]

PR #642, opened 2023-05-01
datamodel_glue.h: don't include datamodel_includes.h

veprbl:

The `datamodel_glue.h` is included from all PODIO factories and it further includes `datamodel_includes.h`, that then includes header files of all EDM4eic collections. In principle, we are looking to get rid of `datamodel_glue.h`. We might as well start making the code to work, at least without the `datamodel_includes.h` by placing explicit includes where needed.

This provides a very minor improvement for the compilation speed, cc #628.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


src/services/io/podio/make_datamodel_glue.py
```diff
@@ -73,13 +73,18 @@ def AddCollections(datamodelName, collectionfiles):
         header = '#include <'+ datamodelName + header_fname + '>'
         header_lines.append(header)
 
+        type_map.append('namespace ' + datamodelName + ' {')
```
wdconinc, line 76:
```suggestion
        type_map.append('namespace ' + datamodelName.lower() + ' {')
```
Is this needed? I thought the entry point headers for the data models are EDM4hep.h etc right? On my phone, hard to check.

wdconinc (approved):

src/services/io/podio/make_datamodel_glue.py
veprbl: It's already lowecase, see:
https://github.com/eic/EICrecon/pull/642/files#diff-a4f8457644f0e66b97c328c3b74947e07aa878793c0bbed9cdf84a596188a597L91-L92

PR #647, opened 2023-05-01
CalorimeterIslandCluster: fix evaluator error format string

veprbl:
Prevents users from seeing confusing error messages.


wdconinc (approved):
Looks good.

PR #651, opened 2023-05-05
linux-eic-shell.yml: fix llvm-cov when running on a branch

veprbl:
A minor fix for a failure seen on main.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -146,7 +146,7 @@ jobs:
           -instr-profile=src/tests/algorithms_test/algorithms_test.profdata \
           "${COV_OPTIONS[@]}" | jq '.data | map(.totals.regions.percent) | .[]' | xargs printf "%.1f\n")
         curl --request POST \
-          --url https://api.github.com/repos/${{ github.repository }}/statuses/${{ github.event.pull_request.head.sha }} \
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -146,7 +146,7 @@ jobs:
           -instr-profile=src/tests/algorithms_test/algorithms_test.profdata \
           "${COV_OPTIONS[@]}" | jq '.data | map(.totals.regions.percent) | .[]' | xargs printf "%.1f\n")
         curl --request POST \
-          --url https://api.github.com/repos/${{ github.repository }}/statuses/${{ github.event.pull_request.head.sha }} \
+          --url https://api.github.com/repos/${{ github.repository }}/statuses/${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }} \
```
wdconinc, line 149: I think this also works:
```suggestion
          --url https://api.github.com/repos/${{ github.repository }}/statuses/${{ github.event.pull_request.head.sha || github.sha }} \
```
If the first doesn't exist, the second is used. I think we use this pattern elsewhere. It is annoying indeed that there's no single way to get a sha.

wdconinc (approved):

wdconinc (approved):

Issue #652, opened 2023-05-05
eicrecon-gcc-gun-EcalLumiSpec test is broken on main


veprbl:
https://github.com/eic/EICrecon/actions/runs/4888399445/jobs/8726255033#step:5:496
```
[JEventProcessorPODIO] [error] Omitting PODIO collection 'EcalLumiSpecClusters' due to exception: JFactoryPodioT: Unable to add collection to frame as frame is missing!.
```

nathanwbrei:
This looks a lot like issue #631. I can help you with this if you need it ASAP

nathanwbrei:
There's a "good" fix in PR #636, but that's held up by other crashes I haven't had the chance to debug. There are several levels of quick fixes available though

dhevang:
Indeed, looks like the same thing as #631. If a good fix could be available in a week time scale, I would wait for that.
Otherwise, would you just suggest to include EcalLumiSpecClusterAssociations in the podio:output_include_collections as per #631 ?

nathanwbrei:
> 

Go ahead and see if adding it fixes the problem. If not, I'll jump in

veprbl:
Looks like #653 fixes it. I'm so confused right now.

nathanwbrei:
I can explain, but it's going to a long unpleasant explanation and the only meaningful takeaway is that using event->Insert() in this context is bad. If this makes the bug go away for now, just merge it, and I'll make a note to circle back and fix the collection name as part of PR #636.

nathanwbrei:
Another, slightly better fix is to go to `github/workflows/linux-eic-shell.yml`, and change line 319 so that `-Ppodio:output_include_collections=EcalLumiSpecRawHits,EcalLumiSpecRecHits,EcalLumiSpecClusters` becomes -Ppodio:output_include_collections=EcalLumiSpecRawHits,EcalLumiSpecRecHits,EcalLumiSpecClusters,EcalLumiSpecClusterAssociations`.

veprbl:
> I can explain, but it's going to a long unpleasant explanation and the only meaningful takeaway is that using event->Insert() in this context is bad. If this makes the bug go away for now, just merge it, and I'll make a note to circle back and fix the collection name as part of PR #636.

What is so special about EcalLumiSpecClusterAssociations compared to EcalLumiSpecCluster***s***Associations?

veprbl:
And why does it not matter for other factories that use "...ClusterAssociations"?

[...]

PR #653, opened 2023-05-05
Fix EcalLumiSpecClusters collections

veprbl:
Let's try to revert my changes and see if it helps.

Resolves: #652


dhevang (approved):
Error of missing collection was removed with this change.

Issue #564, opened 2023-03-22
`richgeo`: add ACTS surface downstream to help filter tracks which will not pass through the RICH


c-dilks:
**Is your feature request related to a problem? Please describe.**
Reduce the pollution of errors from track propagation, where the trajectory does not go through the RICH volume:
```
ProjectTrack   ERROR     Propagation reached the step count limit of 1000 (did 1000 steps)
```

**Describe the solution you'd like**
Before propagating tracks in RICH volumes, check if the track is likely to actually pass through the RICH volume. This can be done by checking if the trajectory can be propagated to:
```
for aerogel radiator: a disc at the aerogel entrance window
for gas radiator:     a disc at DRICH_zmin + DRICH_snout_length (the plane where the snout touches the cylindrical tank)
```

If the test propagation is successful, proceed with the full propagation to discs in the radiator volume.

Additionally, for gas, if the radial value of the test propagated point is larger than the maximum snout radius (`DRICH_rmax1`), then this track only passes through the gas volume; do not allow propagation to discs in the gas radiator which are in the snout (otherwise those will also cause error pollution).

Caveat: for tracks which miss the dRICH, we will still get such an error from the test propagation, one for each radiator. To suppress these, consider asking the ACTS developers to reduce the corresponding `ACTS_ERROR` to `ACTS_VERBOSE`

**Describe alternatives you've considered**
Live with the pollution and annoy everyone.

PR #655, opened 2023-05-07
feat(DRICH)!: enable raw hits and projected tracks by default

c-dilks:
The dRICH IRT-PID implementation has not been approved in time for production 23.5. In order to at least have _something_ useful from the dRICH, let's store the digitized raw hits, `DRICHRawHits`, and the projected track points in the radiators, `DRICH{Aerogel,Gas}Tracks`. These will enable the usage of these production data as input to the IRT algorithm.

For each charged particle which passes through both radiators, `DRICH{Aerogel,Gas}Tracks` store 15 `TrackPoints` and the dRICH typically has less than 30 `DRICHRawHits`. We are not storing the associated `SimTrackerHits` at this time.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @chchatte92

no

yes


c-dilks:
https://github.com/eic/EICrecon/issues/564 should be tested for and/or resolved before merging.

c-dilks:
Much pollution... 
https://github.com/eic/EICrecon/actions/runs/4905199544/jobs/8758965025?pr=655

wdconinc:
Why does the valgrind job take so much longer than without the RICH reconstruction? Timeout after 6 hours compared to 1h23 in a [recent run](https://github.com/eic/EICrecon/actions/runs/4980543761/jobs/8913897110) is quite a difference. How does this compare in regular runtime?

c-dilks:
> Why does the valgrind job take so much longer than without the RICH reconstruction? Timeout after 6 hours compared to 1h23 in a [recent run](https://github.com/eic/EICrecon/actions/runs/4980543761/jobs/8913897110) is quite a difference. How does this compare in regular runtime?

I haven't checked, but I suspect this is because of https://github.com/eic/EICrecon/issues/564, where tracks which don't pass through the dRICH radiators are still being run through this propagator algorithm (for a 1,000 attempts).

c-dilks:
If we are already running this month's production, we might as well just close this PR.

wdconinc:
If the IRT takes so long, then we need to figure out whether and how this impacts overall production for the next production campaigns too. @rahmans1 @T-Britton 

We are already concerned about the slowdown by a factor 3 compared to two months ago, and this seems like it is at least another factor 5 for reconstruction before the valgrind job times out. Part of the assessment here is to look at the call graphs from jana which the reconstruction produces (or can produce if we don't have that in the workflows here) and then looking at top consumers for common workflows, but we need to do that for an optimized build and not in valgrind's VM. This can also fall under benchmarking. @torrijeske @veprbl

c-dilks:
This is not yet IRT code, this is input to it, in particular the track propagator. Its memory leaks should be fixed by https://github.com/eic/EICrecon/pull/656 but I need advice how to deal with #564.

c-dilks:
Actually, let's remove the `DRICH*Tracks` collections from this, leaving only the digitizer to produce `DRICHRawHit`; valgrind should not slow down in this case, or we have issues in the digitizer...

c-dilks:
CI valgrind check took the normal amount of time, and no issues are detected in the digitizer (`PhotoMultiplierHitDigi`).

PR #656, opened 2023-05-07
refactor+fix: use collections and multi-factory in `RichTrack_factory`; fix potential leaks in `TrackPropagation`

c-dilks:
- return `Collection` of `TrackSegments` in `TrackPropagation::propagateToSurfaceList`, rather than a (potentially leaky) new `TrackSegment` pointer
  - the loop over input trajectories is moved from `RichTrack_factory` to `TrackPropagation::propagateToSurfaceList`, which makes the PR diff appear to be more complex than it actually is
  - fix additional possible memory leaks by replacing any `return new raw_pointer` calls with `return std::make_unique<...>(...)`, and update any code which depends on this change
- improve the configuration with `RichTrackConfig`, which configures the _factory_, not the algorithm, since the factory is used to connect to the `richgeo` service (and not the more general `TrackPropagation` algorithm)

- [x] Bug fix (memory leaks)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Tests for the changes have been added: tested in `irt-algo` branch

no, unless the `TrackPropagation` algorithm is called by anything external of EICrecon

no

### DIFF
--- a/src/algorithms/tracking/TrackPropagation.cc
+++ b/src/algorithms/tracking/TrackPropagation.cc
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Wenqing Fan, Barak Schmookler, Whitney Armstrong, Sylvester Joosten, Dmitry Romanov
+// Copyright (C) 2022, 2023 Wenqing Fan, Barak Schmookler, Whitney Armstrong, Sylvester Joosten, Dmitry Romanov, Christopher Dilks
 
 #include <cmath>
 #include <algorithm>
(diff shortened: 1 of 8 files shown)


src/global/pid/RichTrack_factory.cc
```diff
@@ -1,53 +1,53 @@
...
-    else if(param_prefix.rfind("Gas")!=std::string::npos) m_numPlanes = 10;
+  // get list of radiators
+  std::map<int,std::string> radiator_list;
+  for(auto& output_tag : GetOutputTags()) {
+    auto radiator_id = richgeo::ParseRadiatorName(output_tag/*, m_log*/); // FIXME: m_log argument requires PR #608
+    radiator_list.insert({
+        radiator_id,
+        richgeo::RadiatorName(radiator_id/*, m_log*/) // FIXME: m_log argument requires PR #608
+        });
   }
```
c-dilks, line 29: It would be better to merge #608 first, so we don't have to do this workaround
```suggestion
  for(auto& output_tag : GetOutputTags()) {
    auto radiator_id = richgeo::ParseRadiatorName(output_tag, m_log);
    radiator_list.insert({
        radiator_id,
        richgeo::RadiatorName(radiator_id, m_log)
        });
  }
```

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -67,66 +67,82 @@ namespace eicrecon {
...
+    {
+      // logging
+      m_log->trace("Propagate trajectories: --------------------");
+      m_log->trace("number of trajectories: {}",trajectories.size());
+
+      // start output collection
+      auto track_segments = std::make_unique<edm4eic::TrackSegmentCollection>();
+
+      // loop over input trajectories
+      for(const auto& traj : trajectories) {
```
veprbl, line 91: I don't understand why do you prefer to move this loop inside the function. That restricts flexibility without any (performance?) gains.

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -67,66 +67,82 @@ namespace eicrecon {
...
-        // add the `TrackPoint` to the `TrackSegment`
-        track_segment.addToPoints(*point);
-
-      } // end `targetSurfaces` loop
-
-      // output
-      track_segment.setLength(length);
-      track_segment.setLengthError(length_error);
-      return new edm4eic::TrackSegment(track_segment);
+    std::unique_ptr<edm4eic::TrackSegmentCollection> TrackPropagation::propagateToSurfaceList(
```
veprbl, line 78: If you need to return a collection, it probably should be one of the STL ones, not PODIO.

veprbl (commented):
It looks like there was a memory leak to fix here. Nice find.

src/algorithms/tracking/TrackPropagation.cc
c-dilks: This is for modularity: less code in the factory, more code in the algorithm. Ideally the factories should just be boilerplate at this point, but in many cases this is still not 100% possible. This refactor does not restrict flexibility since input `trajectories` can have `size==1`.

src/algorithms/tracking/TrackPropagation.cc
c-dilks: We should have been working with PODIO collections from the start. The `std::vector<raw pointer to PODIO object>` object is not easy to work with in practice, because the PODIO collection which owns them is never clear (ownership is needed to make relations); this initial aversion of PODIO collections has been the cause of multiple memory issues downstream.

c-dilks:
> It looks like there was a memory leak to fix here. Nice find.

Possibly more at https://github.com/eic/EICrecon/issues/661.

src/algorithms/tracking/TrackPropagation.cc
veprbl: Algorithm returns STL collections of owned smart pointers, factory packages it for a specific way to output information by moving pointers out into the PODIO collection.

veprbl: I think PODIO design is in wrong here, if they followed single responsibility rules, the implementation of collections would be an overstep. If using PODIO collections for the return types aligns with our factory pattern, why can't we use  them in the method arguments as well?

My understanding was that the goal for algorithms was to encapsulate the logic of the algorithms, not the logic of the ePIC reconstruction. They would not be able to absorb all of it anyways, because the arrangement of factories into a pipeline is made by EICrecon.

c-dilks: > why can't we use them in the method arguments as well?

We normally do; at least I do for all dRICH algorithms. This algorithm cannot, however, since its input `eicrecon::TrackingResultTrajectory` datatype is not an `EDM4*` datatype (which in my opinion should never have been allowed, since it can't be related to any other EDM objects).

> My understanding was that the goal for algorithms was to encapsulate the logic of the algorithms, not the logic of the ePIC reconstruction.

Correct, but any part of the ePIC software stack needs some "module" (model) to hold the data. This is one of the most important modules since it is the most difficult to change. Last summer, in the critical software decision process, the consensus for this model was PODIO+EDM4hep+EDM4eic and therefore we should expect to be able to use it and all of its features everywhere, including a "decoupled" algorithms framework.

This is a very good discussion, especially as input/feedback to the eventual decoupling of algorithms from the reconstruction framework; it's very non-trivial to draw the line between the algorithms and framework. However, I'm afraid this discussion has veered off topic from this PR review. Perhaps we can continue this in https://github.com/orgs/eic/discussions where there can be more visibility (even though it's been quiet nowadays...).

veprbl: > We normally do; at least I do for all dRICH algorithms.

I am satisfied with this answer. If you are saying that we can and should do this, we need to make a motion adopt this convention everywhere uniformly.

> This algorithm cannot, however, since its input eicrecon::TrackingResultTrajectory datatype is not an EDM4* datatype (which in my opinion should never have been allowed, since it can't be related to any other EDM objects).

Could you file an issue, so that we can transition it to EDM4eic?

---

> Correct, but any part of the ePIC software stack needs some "module" (model) to hold the data. This is one of the most important modules since it is the most difficult to change. Last summer, in the critical software decision process, the consensus for this model was PODIO+EDM4hep+EDM4eic and therefore we should expect to be able to use it and all of its features everywhere, including a "decoupled" algorithms framework.

I didn't think that this was the point of making that decision. The primary goal was setting the on-disk format PODIO, and picking a specific set of data structures (list structures; names and encodings of the fields) for algorithms was secondary. But what we are discussing is just about using collections API from PODIO in the algorithms.

I think, to have discussion here is very appropriate, as we have a concrete code to discuss. Hopefully, it doesn't disrupt your work or anything like that.

c-dilks: > motion adopt this convention everywhere uniformly.

Yes, this should be brought up with those working on the `algorithms` API (@sly2j, whether algorithm input/output should be PODIO Collections or STL collections of (pointers to) PODIO objects). Fortunately EICrecon can handle both Collections and `std::vector<raw pointer>` (and probably other structures / smart pointers), and it's not too painful to refactor, should we force a convention on everyone. I like having conventions and consistency, but another solution would be to let algorithm developers choose.

However, I vote for PODIO collections, because `std::vector<raw pointer>` has wasted a lot of my time, especially causing difficulty when _reading_ EICrecon output with something else (successful writing does not guarantee successful reading downstream, so having CI benchmarks which _read_ the data is critical for stability).

> Could you file an issue, so that we can transition it to EDM4eic?

https://github.com/eic/EICrecon/issues/681

veprbl (approved):
No additional comments

veprbl (approved):
No additional comments

src/algorithms/tracking/TrackPropagation.cc
sly2j: In all cases podio collections: this is required to properly deal with associations and to allow for optimized memory layouts down the road (the collection and, by extension the frame owns the memory). There is much more that podio does behind the screens that is critical for good operations, especially in a multi-threaded context. We should under no circumstance work with raw pointers to Podio objects in our algorithms. `operator new` is banned in normal user code, and shared pointers/unique pointers should be used only for managing algorithm resources.  Note that many of the problems we have encountered in the last year were due to nonstandard implementation/usage of podio objects that could have been avoided by sticking to the API.

PR #660, opened 2023-05-08
CalorimeterHitDigi: remove u_ref

veprbl:
In the current code, `u_ref` is used to set value for bits of the cell ID that were removed by `id_mask`. The resulting value `hid` is printed out to logs and used as unique key, but never propagates anywhere. Removing support to u_ref will have no effect on the reconstruction, and will be equivalent to setting `u_ref` to `{0}`.

See discussion in https://eic.cloud.mattermost.com/main/pl/d1c8cm1rabnrjnxqumbyioua7c on how u_ref is part of a legacy logic that had to be disabled.


FriederikeBock:
I think this change will break something for the LFHCal, as I am deliberately masking here cellID of the individual layers in Z and only keeping the readout-layer z. Please don't remove this option.

veprbl:
I'm removing `u_ref`, not `u_fields` that does the masking.

veprbl:
Please be more specific about what is wrong, so that we can accept or reject this PR and not just hang.

FriederikeBock:
So going more in detail, if I understand it correctly. With the code as is in principle it would have been possible to set the ID you are masking in `ufield` to any value you set in the correspoding `uref`. With this change this would automatically set the ID for the `ufield` to 0?

veprbl:
In this particular instance, the `u_ref` allows to set some bits, but they are not used in a meaningful way (any value works equivalently for a uniqueness key) and not propagated anywhere beyond printed as log messages (cellID value of the peak energy hit used instead).
https://github.com/eic/EICrecon/blob/3ca05a0a560de426be80edc9fe470952cf80a1da/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L207
https://github.com/eic/EICrecon/blob/3ca05a0a560de426be80edc9fe470952cf80a1da/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L223-L225
https://github.com/eic/EICrecon/blob/3ca05a0a560de426be80edc9fe470952cf80a1da/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L254-L255

PR #636, opened 2023-04-28
WIP: Calo clustering factories now use JMultifactory

nathanwbrei:
### Briefly, what does this PR introduce?

The calorimeter clustering factories need to produce associations alongside clusters, and the current approach is hacky and bug-prone. JANA 2.1.0 introduces multifactories to solve this problem cleanly. This PR migrates all of the calorimeter clustering factories that produce associations to use multifactories. 

### What kind of change does this PR introduce?
- [x] Bug fix (issue #631)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

PR #666, opened 2023-05-15
fix: Cluster factory factory so cluster matching has associations 

wdconinc:
In 23.05.1 we realized that cluster matching (though now implementated) doesn't actually work yet due to the absence of cluster associations being produced in clustering. One could fix this in 17 factories (#636) but I really didn't want to do that so here's an approach that reduces duplicate code (but otherwise is similar enough, with the exception of adding configuration structures).

Todo:
- [x] depends on #664 

Closes: #636

- [x] Bug fix (issue: no cluster associations)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @veprbl @nathanwbrei 

No.

Now associations will be written too.


wdconinc:
Net lines of code: -1500...

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -108,49 +88,51 @@ void CalorimeterClusterRecoCoG::AlgorithmProcess() {
...
           for (const auto& pclhit1: pclhits) {
             m_log->trace("{}: {}", pclhit1.getCellID(), pclhit1.getEnergy());
           }
           m_log->trace("MC hits: ");
-          for (const auto& mchit1: mchits) {
-            m_log->trace("{}: {}", mchit1->getCellID(), mchit1->getEnergy());
-          }
+          //for (const auto& mchit1: mchits) {
+          //  m_log->trace("{}: {}", mchit1->getCellID(), mchit1->getEnergy());
+          //}
```
wdconinc, line 114: Re-enable. Not sure why I disabled this.

nathanwbrei:
It's so satisfying when you can remove huge pieces of redundant code like that!

wdconinc:
And it will also work for all other calorimetry factories... In a different PR.

veprbl:
I also appreciate the effort. A bit concerned about the size of the diff to review, though :)

src/algorithms/calorimetry/CalorimeterClusterRecoCoGConfig.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+namespace eicrecon {
+
+    struct CalorimeterClusterRecoCoGConfig {
+
+        std::string energyWeight;
+        std::string moduleDimZName;
+
+        double sampFrac = 1.;
```
FriederikeBock, line 13: Would it be possible to rename this variable, as is not really the sampling fraction which we necessarily scale with here but rather a more general calibration factor? I think that would improve the readability of the code even further.

wdconinc: Not in this PR

FriederikeBock:
Thanks for this massive effort! From reading the code it seems to be ok and certainly removes duplication. Have you run any explicit tests for any of the calorimeters to test that you get what you expect on the cluster level, if so could you list them in the PR please.

wdconinc:
Ran out of time to do a full test. The test I did do was the proof of concept implementation on scfi clusters only, and that resulted in the correct cluster distributions and non-empty associations. Extending this to all cluster factories resulted in failing tests, which makes a pre/post comparison premature.

First this needs some debugging of the segfaults (i.e. why does podio try to write collections to the frame from multi-factories that should have been disabled due to lack of necessary hit collections, even when activating the successful_collections in the podio processor @nathanwbrei). Someone with a debug build of podio outside of eic-shell will have an easier time identifying which collection ends up with a nullptr.

If we want this by June (and I think we do) someone would need to take this over from here since I won't be able to complete it until in June.

A next step, as soon as this is demonstrated to work or even in parallel, may be a refactor for similar code reduction on the other calorimetry factories which are all also just duplicated code. What's needed there is conversion to using the config mixin on the algorithm side, and removing the concrete factories in separate translation units in favor of a templated solution. Finally, I don't think there is a way around the using varargs constructor to pick up the required constructors, by maybe someone can find a way around it by using a different construct.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -234,6 +249,20 @@ void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent> &event) {
...
+    //            all). See also below, at "TODO: NWB:".
+    for (const auto& coll_name : m_collections_to_write) {
+        try {
+            const auto* coll_ptr = event->GetCollectionBase(coll_name);
+        }
+        catch(std::exception &e) {
+            // chomp
+        }
+    }
+
```
wdconinc, line 265: FYI @nathanwbrei, this fixes collection IDs in the order of `m_collections_to_write` before letting algorithms pick them in random order.

wdconinc:
And now on ReconstructedParticles after matching. This is looping over ReconstructedParticleAssociations and plotting reconstructed vs generated total momentum.

```
events->Draw("std::hypot(ReconstructedParticles.momentum.x[ReconstructedParticleAssociations.recID],ReconstructedParticles.momentum.y[ReconstructedParticleAssociations.recID],ReconstructedParticles.momentum.z[ReconstructedParticleAssociations.recID]):std::hypot(MCParticles.momentum.x[ReconstructedParticleAssociations.simID],MCParticles.momentum.y[ReconstructedParticleAssociations.simID],MCParticles.momentum.z[ReconstructedParticleAssociations.simID])")
```
image

wdconinc:
This work is now complete:
- clusters finding produces associations,
- cluster matching with tracks produces associations,
- remaining unmatched clusters are turned into neutrals.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -0,0 +1,91 @@
...
+        // This prefix will be used for parameters
+        std::string plugin_name  = eicrecon::str::ReplaceAll(GetPluginName(), ".so", "");
+        std::string param_prefix = plugin_name + ":" + GetTag();
+
+        // Use JDD4hep_service to get dd4hep::Detector
+        auto geoSvc = app->template GetService<JDD4hep_service>();
+        m_detector = geoSvc->detector();
+
+        // SpdlogMixin logger initialization, sets m_log
+        SpdlogMixin<T>::InitLogger(JChainMultifactoryT<CalorimeterClusterRecoCoGConfig>::GetPrefix(), "info");
```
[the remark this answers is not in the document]
wdconinc: It just isn't initialized in https://github.com/eic/EICrecon/blob/main/src/extensions/jana/JChainMultifactoryT.h#L35. We can set it by calling `SetPrefix` explicitly.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -0,0 +1,91 @@
...
+
+    //------------------------------------------
+    // Init
+    void Init() override {
+
+        auto app = GetApplication();
+
+        // This prefix will be used for parameters
+        std::string plugin_name  = eicrecon::str::ReplaceAll(GetPluginName(), ".so", "");
+        std::string param_prefix = plugin_name + ":" + GetTag();
```
wdconinc, line 43: This would set the prefix, I think.
```suggestion
        std::string param_prefix = plugin_name + ":" + GetTag();
        JChainMultifactoryT<CalorimeterClusterRecoCoGConfig>::SetPrefix(param_prefix);
```

wdconinc: Nevermind, probably irrelevant.

src/extensions/jana/JChainMultifactoryGeneratorT.h
```diff
@@ -37,10 +36,14 @@ class JChainMultifactoryGeneratorT : public JFactoryGenerator {
...
     {
-        Init();
     };
 
     void GenerateFactories(JFactorySet *factory_set) override {
+        // initialization is delayed to let caller set plugin name first
+        if (!m_init_done) {
+            Init();
+            m_init_done = true;
+        }
```
veprbl, line 46: cc @nathanwbrei

veprbl:
I see some odd behavior. When I run brycecanyon configuration with main and with this PR, new configuration options *appear* for:
 - `FHCAL:HcalEndcapPIslandProtoClusters:*`
 - `FHCAL:HcalEndcapPTruthClusters:*`
 - `FHCAL:LFHCALClusters:*`
 - `FHCAL:LFHCALIslandProtoClusters:*`
 - `BEMC:EcalBarrelSciGlassProtoClusters:*`

But some collections *disappear* from the output file:

 - `EcalBarrelSciGlassClusters`
 - `EcalBarrelSciGlassTruthClusters`
 - `HcalEndcapPClusterAssociations`
 - `HcalEndcapPClusters`
 - `HcalEndcapPTruthClusterAssociations`
 
I don't understand how to reconcile this.

wdconinc:
There is no SciGlass and HcalEndcapP in brycecanyon. These collections should not have been present before, but a workaround was to write them anyway (see "TODO: NWB: Re-enable me" at https://github.com/eic/EICrecon/pull/666/files#diff-08c369ce77f61e0738a1b6bea175ebb10a4df985a97ced435296883bcca0992cL271). So, that is intended behavior as part of fixing the treatment of collections with missing dependencies.

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -3,60 +3,208 @@
...
+              .enableEtaBounds = false
+            },
+            app   // TODO: Remove me once fixed
+          )
+        );
+
+        app->Add(
+          new JChainMultifactoryGeneratorT<Cluster_factory_LFHCALClusters>(
+             "LFHCALClusters",
+            {"LFHCALIslandProtoClusters",  // edm4eic::ProtoClusterCollection
```
wdconinc, line 194:
```suggestion
            {"LFHCALProtoClusters",  // edm4eic::ProtoClusterCollection
```
as in https://github.com/eic/EICrecon/pull/666/files#diff-65cecdd85beb3ba502721f7ec59a332712e214e8c85a8b4804af074edfabe929L55

wdconinc: I mean https://github.com/eic/EICrecon/pull/666/files#diff-803292d5047f9f39733fc1e4c25e513ffd878b83d7bd3419daaf5c50f2c3896eL56

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -3,60 +3,208 @@
...
+              .enableEtaBounds = true
+            },
+            app   // TODO: Remove me once fixed
+          )
+        );
+
+        app->Add(
+          new JChainMultifactoryGeneratorT<Cluster_factory_HcalEndcapPInsertClusters>(
+             "HcalEndcapPInsertClusters",
+            {"HcalEndcapPInsertIslandProtoClusters",  // edm4eic::ProtoClusterCollection
```
wdconinc, line 151:
```suggestion
            {"HcalEndcapPInsertProtoClusters",  // edm4eic::ProtoClusterCollection
```
as in https://github.com/eic/EICrecon/pull/666/files#diff-105ac13e3fd40754f5377b992a6630adef70b348cf900618fb4477c8a89fdf51L46

veprbl: This doesn't seem right because (on `main`):
https://github.com/eic/EICrecon/blob/dedbad46a5fbe4d043af2b9f0e0e13b4a84ab95d/src/detectors/FHCAL/ProtoCluster_factory_HcalEndcapPInsertIslandProtoClusters.h#L29

src/detectors/FHCAL/FHCAL.cc
veprbl: This doesn't seem right because (on `main`):
https://github.com/eic/EICrecon/blob/main/src/detectors/FHCAL/ProtoCluster_factory_LFHCALIslandProtoClusters.h

veprbl: The actual inconsistency seems to be in the name `EcalBarrelSciGlassProtoClusters`

veprbl:
Okay. I agree, collections should not be there in the first place. I'm not able to see why old version doesn't produce parameters, but, at least, it's good that it works now.

veprbl (approved):

PR #591, opened 2023-04-06
Introduction of LFHCal & adjustments for processing of fECal

FriederikeBock:
### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

PR #671, opened 2023-05-16
Introduce EDM4hep unit system

veprbl:
### Rationale

Most of the number manipulation is performed on the values from the EDM4hep input files or EDM4eic output files. Unfortunately unit system of those differs from the one adopted in Geant or TGeo. To make matters more complicated, DD4hep units can switch between Geant and TGeo (with later being the default).

| Geant      | TGeo                | EDM4hep/EDM4eic |
| ------- | ------------- | --------- |
| milimeter  | centimeter          | milimeter       |
| nanosecond | second              | nanosecond      |
| MeV        | GeV                 | GeV             |

We would like to adopt a single unit system to be the default in EICrecon, and the goal of this PR is to set it to the one of the EDM. The current codebase uses units from the `dd4hep::` namespace, for which we have the following common cases:

 - Energy units like keV/MeV/GeV are compatible with EDM. Energy (and time) values are not expected to occur in geometry data structures.
 - Distance units are incompatible with values in EDM, but may be compatible with the geometry. There are likely some bugs there, those cases need to be analyzed manually.
 - Timing units are often not specified in the code, as unit system in EDM is different by several orders of magnitude and it's really hard to make the mistake of using `dd4hep::`.

### Related discussions

 - https://github.com/eic/EICrecon/pull/390
 - https://github.com/eic/EDM4eic/pull/35


src/common/unit_system.h
```diff
@@ -0,0 +1,50 @@
...
+    // | milimeter  | centimeter          | milimeter       |
+    // | nanosecond | second              | nanosecond      |
+    // | MeV        | GeV                 | GeV             |
+    // +------------+---------------------|-----------------+
+    //
+
+    // distance
+    static constexpr double mm = 1.0; // millimeter
+    static constexpr double cm = 1e1 * mm; // centimeter
+    static constexpr double micrometer = 1e-3 * mm; // micrometer
```
c-dilks, line 31: Good to have both, because some write "u" to mean Greek letter "mu" (`dd4hep::micrometer == dd4hep::um`).
```suggestion
    static constexpr double um = 1e-3 * mm; // micrometer
    static constexpr double micrometer = um;
```

src/common/unit_system.h
```diff
@@ -0,0 +1,50 @@
...
+    static constexpr double nm = 1e-6 * mm; // micrometer
+    static constexpr double m = 1e3 * mm; // meter
+
+    // time
+    static constexpr double ns = 1.0; // nano second
+    static constexpr double second = 1e9 * ns; // second
+    static constexpr double milisecond = 1e6 * ns; // milisecond
+    static constexpr double microsecond = 1e3 * ns; // milisecond
+    static constexpr double picosecond = 1e-3 * ns; // pico second
```
c-dilks, line 40: Let's just use the short names:
```suggestion
    static constexpr double ns = 1.0; // nanosecond
    static constexpr double s = 1e9 * ns; // second
    static constexpr double ms = 1e6 * ns; // millisecond
    static constexpr double us = 1e3 * ns; // microsecond
    static constexpr double microsecond = us;
    static constexpr double ps = 1e-3 * ns; // picosecond
```

src/algorithms/calorimetry/CalorimeterHitDigi.h
```diff
@@ -48,7 +52,7 @@ class CalorimeterHitDigi {
 
     // digitization settings
     unsigned int             m_capADC;
-    double                   m_capTime = 1000.;
+    double                   m_capTime = 1000. * unit::ns;
```
c-dilks, line 55: Is whatever uses this aware that units are now included?

src/detectors/FHCAL/CalorimeterHit_factory_LFHCALRecHits.h
```diff
@@ -30,7 +34,7 @@ class CalorimeterHit_factory_LFHCALRecHits : public eicrecon::JFactoryPodioT<edm
 
         // digitization settings, must be consistent with digi class
         m_capADC=65536;//2^16
-        m_dyRangeADC=1 * dd4hep::GeV;
+        m_dyRangeADC=1 * unit::GeV;
         m_pedMeanADC=20;
         m_pedSigmaADC=0.8;
         m_resolutionTDC=10 * dd4hep::picosecond;
```
c-dilks, line 40:
```suggestion
        m_resolutionTDC=10 * unit::ps;
```

c-dilks: and many others, I only glanced through and caught a few of these, and `grep` can find the rest

src/detectors/FHCAL/ProtoCluster_factory_LFHCALIslandProtoClusters.h
```diff
@@ -30,8 +34,8 @@ class ProtoCluster_factory_LFHCALIslandProtoClusters : public eicrecon::JFactory
         m_input_tag = "LFHCALRecHits";
 
         m_splitCluster=false;
-        m_minClusterHitEdep=1 * dd4hep::MeV;
-        m_minClusterCenterEdep=100.0 * dd4hep::MeV;
+        m_minClusterHitEdep=1 * unit::MeV;
+        m_minClusterCenterEdep=100.0 * unit::MeV;
 
         // neighbour checking distances
         m_sectorDist=0 * dd4hep::cm;
```
c-dilks, line 41:
```suggestion
        m_sectorDist=0 * unit::cm;
```

src/detectors/FHCAL/CalorimeterHit_factory_LFHCALRecHits.h
veprbl: I have not touched times and lengths (at least not in the public PR). The `dd4hep::picosecond` is likely a blatant bug here, but we would need to check those cases more carefully.

src/algorithms/calorimetry/CalorimeterHitDigi.h
veprbl: This was originally introduced in https://github.com/eic/EICrecon/pull/591/files#r1163225587 by @FriederikeBock where we didn't understand why dd4hep units didn't work.

We are not really using this cutoff. I don't think that the timing information is propagated correctly, and we don't simulate any signal shaping, so the exact value is a moot. I don't change the number, just declare the unit value.

src/common/unit_system.h
veprbl: I like the idea for mu -> u. Implemented.

src/common/unit_system.h
veprbl: Done.

src/detectors/FHCAL/ProtoCluster_factory_LFHCALIslandProtoClusters.h
veprbl: Same situation as with timing. The original values with units may be subtly broken (unit systems differ by a factor of 10, not 10^9 like for timing) for non-zero. Putting the EDM4hep units should fix that, but I'd like to review that more carefully, and, perhaps, separately.

src/detectors/FHCAL/CalorimeterHit_factory_LFHCALRecHits.h
veprbl: No bug was there. Now updated to use `units::`.

src/detectors/FHCAL/ProtoCluster_factory_LFHCALIslandProtoClusters.h
c-dilks: Good point.. probably not a good idea to update _everything_ in this PR.

veprbl:
Will need to be redone.

Issue #673, opened 2023-05-17
Changing the value of threshold factor for ScFi


AkshayaVijay:
changing the value of the threshold factor from 5.0 to 36.0488 to have a threshold value of 0.5MeV for ScFi

PR #674, opened 2023-05-17
EcalBarrelScFiRecHits: update value of thresholdFactor

AkshayaVijay:
Updated the value of the threshold factor from 5 to 36.00488 to have a threshold of 0.5MeV

- [X] Other feature (issue #673 )

### DIFF
--- a/src/detectors/BEMC/CalorimeterHit_factory_EcalBarrelScFiRecHits.h
+++ b/src/detectors/BEMC/CalorimeterHit_factory_EcalBarrelScFiRecHits.h
@@ -32,7 +32,7 @@ public:
         m_resolutionTDC=10 * dd4hep::picosecond;//{this, "resolutionTDC", 10 * ps};
 
         // zero suppression values
-        m_thresholdFactor=5.0;// from ATHENA's reconstruction.py
+        m_thresholdFactor=36.0488;// from ATHENA's reconstruction.py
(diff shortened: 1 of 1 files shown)


src/detectors/BEMC/CalorimeterHit_factory_EcalBarrelScFiRecHits.h
```diff
@@ -32,7 +32,7 @@ class CalorimeterHit_factory_EcalBarrelScFiRecHits : public eicrecon::JFactoryPo
         m_resolutionTDC=10 * dd4hep::picosecond;//{this, "resolutionTDC", 10 * ps};
 
         // zero suppression values
-        m_thresholdFactor=5.0;// from ATHENA's reconstruction.py
+        m_thresholdFactor=36.0488;// from ATHENA's reconstruction.py
```
wdconinc, line 35: For future reference, we probably don't need this many significant figures. No need to change it here though.

wdconinc (dismissed):
You should probably also ask @mariakzurek for a review.

src/detectors/BEMC/CalorimeterHit_factory_EcalBarrelScFiRecHits.h
AkshayaVijay: Okay.

mariakzurek (approved):

PR #675, opened 2023-05-18
266 integrate lowq2 tagger reconstruction

simonge:
Adds reconstructed particles from the LowQ2 tagger and includes them in the track matching for comparing the truth/reconstructed inclusive kinematics.
Almost everything has been implemented with stand alone factories using relatively simple methods. I will add comments with know issues which I'd like some feedback on.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Currently will need to source eicrecon-this.sh before running.

Adds some new factories to the default reconstruction.


simonge:
Ideally most of this could have been implemented with acts or using some of the available functions without full integration, I didn't work out how to approach this so have simple implementations with some obvious limits currently. Additionally I am unsure if I am using the data model correctly or if the data model is even designed to support this.

Tracking clusters - I have included a very simple cluster constructor which just links any pixels which see a hit and finds a central point based on weighting the positions based on energy. It is likely in this detector we will have non-negligible numbers of overlapping hits once all of the backgrounds have been included.

Tracking - Simple brute force loop over all combinations of hits, fitting with a straight line and cutting on the fit chi2. Combinatorics get impossible very quickly when adding background hits, a Kalman filter or neural network will need to be used at this stage to quickly eliminate options.

Reconstruction - ROOT TMVA approach to reconstructing the primary track parameters. Unsure of the best way of linking to an xml location which stores the weights. This also needs to address the problem of passing the beam energy through the simulation and reconstruction, the neural network weights currently included have been trained on a normalized energy but they still need to know how to scale the momentum back up once the vector has been made.

simonge:
This is currently passing all of the checks but shouldn't be for a couple of reasons I'll need some help fixing.

1. The BookMVA function from ROOT is returning a <FATAL> error when the file or method it's looking for aren't found. How do I pass this to the JANA logger to exit the run?
2. A new environment path will need to be introduced to store the weight file and needs to be set before running, where/how should this be stored? My current implementation used $JANA_PLUGIN_PATH and works ok in the container but can contain a many paths as it does in the checks here so isn't useable more generally.

DraTeots:
> 1. The BookMVA function from ROOT is returning a  error when the file or method it's looking for aren't found. How do I pass this to the JANA logger to exit the run?

You should throw JException with the explanation if such thing occurs

> 2. A new environment path will need to be introduced to store the weight file and needs to be set before running, where/how should this be stored? My current implementation used $JANA_PLUGIN_PATH and works ok in the container but can contain a many paths as it does in the checks here so isn't useable more generally.

There is a mechanism for such files called [LargeCalibrations](https://github.com/JeffersonLab/JANA2/blob/7a1c2a9961e154983356ed9e2e8e33e6ce594e77/src/libraries/JANA/Calibrations/JLargeCalibration.h#L99) 

The JLargeCalibration class is used to manage local resource files.
These would typically be larger files that are costly to download
every time the program is run. (i.e. if they come from the
JCalibration system.) Files are kept in a central location
on the local filesystem so only one copy needs to exist.

Full documentation lives in the link above. In short, you specify file name and resource directory. Files could be automatically downloaded from network storage the first time if they are not yet in resource directory. Thus we may put them at the same place as magnetic field maps and then specify in LargeCalibration class.

wdconinc:
Before you jump in and start using `JLargeCalibration`, we need to discuss this. We already have solutions that are integrated with the production campaigns, and I would rather not add another solution to this that we don't currently use.

DraTeots:
IMO it works the same as DD4Hep pulls magnetic field maps. The difference is that it is done via JANA code because this piece of code has nothing to do with DD4Hep. Thus my naïve approach would be to use it absolutely the same way through the same infrastructure

simonge:
Is there a JANA:RESOURCE_DIR which is included in the source/build automatically? Not knowing all of the solutions currently being deployed, my preferred approach would be to have a separate environment variable which could just be the library path for the main EICrecon JANA plugins. @veprbl suggested this (or similar) rather than my request in #703

simonge:
Alternatively iterate over the ":" delimited contents of JANA_PLUGIN_PATH checking if the path name exists.

simonge:
The plugin is now at least loading correctly looking at the CI output. I'm not sure any of the event samples used in the tests will see events in the Low Q2 tagger? Please can someone trawl through and check some of the changes I've made.

src/global/tracking/tracking.cc
```diff
@@ -66,9 +67,17 @@ void InitPlugin(JApplication *app) {
             "CentralTrackingParticles",                       // Tag name for multifactory
             {"CentralCKFTrajectories"},                       // eicrecon::TrackingResultTrajectory
             {"outputParticles",                               // edm4eic::ReconstructedParticle
-             "outputTrackParameters"},                        // edm4eic::TrackParameters
+             "outputTrackParametersT"},                        // edm4eic::TrackParameters
```
veprbl, line 70: Needs better naming.

src/detectors/LOWQ2/LowQ2ProtoCluster_factory.cc
```diff
@@ -0,0 +1,132 @@
...
+    m_log->info("LowQ2 Tagger Decoding complete...");
+
+  }
+
+
+  void LowQ2ProtoCluster_factory::ChangeRun(const std::shared_ptr<const JEvent> &event) {
+    // Nothing to do here
+  }
+
+  void LowQ2ProtoCluster_factory::Process(const std::shared_ptr<const JEvent> &event) {
```
veprbl, line 60: Algorithms like this and the others are supposed to be added to src/algorithms under LGPL3. Unit tests are also much welcome.

src/global/tracking/TrackerParticleCollector_factory.h
```diff
@@ -0,0 +1,42 @@
...
+
+
+namespace eicrecon {
+
+    /// This factory just collects reconstructed hits edm4eic::TrackerHit from different sources
+    /// And makes a single array out of them
+    class TrackerParticleCollector_factory : public JChainFactoryT<edm4eic::TrackParameters, TrackerHitReconstructionConfig> {
+
+    public:
+        TrackerParticleCollector_factory( std::vector<std::string> default_input_tags, TrackerHitReconstructionConfig cfg ):
```
veprbl, line 20: Would be nice to have consistent naming with respect to other implementations like https://github.com/eic/EICrecon/blob/main/src/global/pid/MergeTrack_factory.cc
Any ideas how to achieve that?

src/algorithms/reco/InclusiveKinematicsElectron.cc
```diff
@@ -99,19 +99,23 @@ namespace eicrecon {
...
     // Get incoming hadron beam
     const auto pi_coll = find_first_beam_hadron(mcparts);
-    if (pi_coll.size() == 0) {
-      m_log->debug("No beam hadron found");
-      return kinematics;
+//     if (pi_coll.size() == 0) {
+//       m_log->debug("No beam hadron found");
+//       return kinematics;
+//     }
+    if (pi_coll.size() > 0) {
```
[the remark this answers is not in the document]
simonge: Currently some files don't contain some beam particles but probably should be stored in meta data in the longer term e.g. beamgas backgrounds don't care what the other beam is. This just sets a default if that's the case but isn't a good long term solution.

src/detectors/LOWQ2/LowQ2Reconstruction_factory.cc
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  void LowQ2Reconstruction_factory::Process(const std::shared_ptr<const JEvent> &event) {
+
+    auto inputtracks =  event->Get<edm4eic::TrackParameters>(m_input_tag);
+
+    std::vector<edm4eic::TrackParameters*> outputLowQ2Particles(inputtracks.size());
+
+    // Reconstructed particle members which don't change
+    std::int32_t type   = 0; // Check?
```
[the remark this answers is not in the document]
simonge: This was a note to myself to check what this type referred to and if it should be 0?

src/detectors/LOWQ2/LowQ2ProtoCluster_factory.cc
```diff
@@ -0,0 +1,132 @@
...
+
+      // Filter to make sure everything is on the same detector layer
+      auto layerFilter = (module==module[maxIndex])*(layer==layer[maxIndex]);
+
+      // Loop over hits, adding neighbouring hits as relevant
+      while(indexList.size()){
+
+	auto index = indexList[0];
+	auto filter = avaliable*layerFilter*(abs(x-x[index])<=1)*(abs(y-y[index])<=1)*(abs(t-t[index])<1);
+	avaliable = avaliable*(!filter);
```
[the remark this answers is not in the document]
simonge: The logic works fine as far as I've tested, there might be more efficient and clear ways to code it.

src/detectors/LOWQ2/LowQ2ProtoCluster_factory.cc
simonge: I'm hoping to split bits of this PR into algorithms to be more in line with the philosophy of EICrecon in a future PR.

src/global/tracking/TrackerParticleCollector_factory.h
simonge: This was just a copy of 
[https://github.com/eic/EICrecon/blob/main/src/global/tracking/TrackerHitCollector_factory.cc
](https://github.com/eic/EICrecon/blob/main/src/global/tracking/TrackerHitCollector_factory.cc
)
Would it make sense to have a template collector/merge factory?

src/algorithms/reco/InclusiveKinematicsElectron.cc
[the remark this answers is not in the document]
simonge: For this and other beam configuration related parameters I have created some parameters in the JEventProcessorPODIO in anticipation of them being read from metadata in the future. Let me know if there's a better location for this.

src/detectors/LOWQ2/LowQ2Reconstruction_factory.cc
```diff
@@ -0,0 +1,138 @@
...
+    const char* env_p = getenv(m_environment_path.c_str());
+    if (env_p) {
+
+	std::string dir_path;
+        std::stringstream envvar_ss(env_p);
+        while (getline(envvar_ss, dir_path, ':')) {
+	    std::string weightName = dir_path +"/"+ m_file_path;
+	    if (std::filesystem::exists(weightName)){
+   		try{
+      		    m_method = dynamic_cast<TMVA::MethodBase*>(m_reader.BookMVA( m_method_name, weightName ));
```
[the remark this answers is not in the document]
simonge: The electron passes through the beam line dipole and quadropole magnets, a linear model would provide a very poor reconstruction.

Where would be best to put the training scripts? I have versions on my github but will tidy them up a bit to be a bit more legible

wdconinc:
I have concerns with this PR. It is turning into an everything and the kitchen sink PR. For example, all changes to inclusive kinematics classes are not related to the low Q2 tagger, and if you are going to change an algorithm interface to InclusiveKinematicsElectron, then you need to be consistent and do all similar algorithms. Please split this in several smaller PRs and please open a wider discussion before unilaterally making substantial changes like changes to defaults or configuration interfaces.

simonge:
Point taken, I've been jumbling up the integration of the detector with trying to analysis of some of the simulated backgrounds. I'll roll back the InclusiveKinematicsElectron changes.

I was more hoping for some advise as to where to locate the beam config parameters as they are quite important for the PR but don't belong entirely within the scope of the LowQ2

wdconinc:
> I was more hoping for some advise as to where to locate the beam config parameters

I agree that this needs discussion and a solution. But it should be in a separate PR, and probably even a dedicated discussion in a meeting. It also affects how to pass this in the geometry. Can you open a separate PR with your changes to InclusiveKinematicsElectron?

simonge:
> I agree that this needs discussion and a solution. But it should be in a separate PR, and probably even a dedicated discussion in a meeting.

Ok I will take this out too and just add it as a config for LowQ2Reconstruction_factory as that's all that uses it at the moment. Then in the future PR (after discussion) try and propagate the beam parameters through the code usefully

simonge:
I could split this further into track reconstruction from the detector and leave the ML particle reconstruction for a following PR. This would probably be of rather limited use for physics studies.

veprbl:
> The electron passes through the beam line dipole and quadropole magnets, a linear model would provide a very poor reconstruction.

I also heard @ajentsch saying that he plans to move to ML reconstruction too, as, apparently, the linear model doesn't work too, in some extreme cases, in FF. However, we still would prefer a simple model. I see that the weight file is quite big (will take a while to parse on each EICrecon invocation, which will add up...) and contains 4 hidden layers. I would not want to take away too much of your time, but I think we could use an optimization here. Would it be possible to do the same fit with a simpler model?

> Where would be best to put the training scripts? I have versions on my github but will tidy them up a bit to be a bit more legible

Ideally those go into detector benchmarks [1] so that we always have the latest artifacts produced for every version of the epic geometry. Of course, it would really help to be able to see your code to have a concrete discussion.

[1] https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks

src/detectors/LOWQ2/LowQ2ProtoCluster_factory.cc
veprbl: A reason to split the logic into algorithms ASAP is because we would also like to have someone contribute basic unit tests for this code.

src/global/tracking/TrackerParticleCollector_factory.h
veprbl: I think, it's fine to keep it as a copy for this PR. We would want to file an issue to track possible merging of those factories.

src/algorithms/reco/InclusiveKinematicsElectron.cc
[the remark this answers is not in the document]
veprbl: Perhaps, a way to address @wdconinc's suggestion would be to take this change out from this PR and address this properly, or start a discussion on how to do it.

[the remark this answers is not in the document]
veprbl: Oh, I didn't notice that you've already done it.

simonge:
> Ideally those go into detector benchmarks [1] so that we always have the latest artifacts produced for every version of the epic geometry. Of course, it would really help to be able to see your code to have a concrete discussion.
> 
> [1] https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks

Is this still the correct location for benchmarks and training scripts? The same repository is on this github organisation.

It would be nice to get this merged in some form. I keep on needing to repeat the same changes when merging of main doesn't track the files quite right and duplicating global changes on the local files to keep it up to date.

wdconinc:
> It would be nice to get this merged in some form.

I completely agree. This is just a large PR which hampers getting it reviewed efficiently. It also appears that changes are still being added, which makes it seem like the changes are not final yet.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,177 @@
+// Created by Simon Gardner to do LowQ2 tracking
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
```
wdconinc, line 3: Use SPDX license header throughout for all src/algorithms files.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,177 @@
...
+                  edm4hep::Vector3f pos(exitPos.x()*10,exitPos.y()*10,exitPos.z()*10); //Temp unit transform
+                  edm4eic::Cov3f    posError;
+
+
+                  edm4hep::Vector3f mom(outVec.x(),outVec.y(),outVec.z()); //Temp unit transform
+                  edm4eic::Cov3f    momError;
+
+
+//                edm4eic::TrackPoint outTrack(pos,posError,mom,momError,time,timeError,theta,phi,dirError,path,pathError);
+//                outputTracks->push_back(outTrack);
```
wdconinc, line 163: See https://github.com/eic/EICrecon/pull/862/commits/f10c14fdab71cffefbfea5d28cfd98cf2dd79b7f for a robust solution to EDM4eic versioning.

wdconinc: Or just remove the TrackPoint if you don't want to write it.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,177 @@
...
+                  if(outVec.Z()>0) outVec*=-1;
+
+                  //position vector crosses x axis, swap to z exit later
+                  auto exitPos = outPos-(outPos.x()/outVec.x())*outVec;
+
+                  // Create track parameters edm4eic structure
+                  // TODO - populate more of the fields
+                  int type = 0;
+                  // Plane Point
+                  edm4hep::Vector2f loc(exitPos.y()*10,exitPos.z()*10); //Temp unit transform
```
wdconinc, line 136: The `loc` is position on a cylinder, i.e. `(std::hypot(x,y), z)`.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,177 @@
...
+                  outputTracks->push_back(outTrack);
+
+                  // Plane Point
+                  edm4hep::Vector3f pos(exitPos.x()*10,exitPos.y()*10,exitPos.z()*10); //Temp unit transform
+                  edm4eic::Cov3f    posError;
+
+
+                  edm4hep::Vector3f mom(outVec.x(),outVec.y(),outVec.z()); //Temp unit transform
+                  edm4eic::Cov3f    momError;
+
```
wdconinc, line 160: This may fail compilation unless you add `[[maybe_unused]]`.

src/algorithms/fardetectors/FarDetectorLinearTracking.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <spdlog/spdlog.h>
+
+// Event Model related classes
+#include <edm4eic/TrackParametersCollection.h>
+
+#include <extensions/jana/JChainFactoryT.h>
+#include <extensions/spdlog/SpdlogMixin.h>
+#include <spdlog/logger.h>
+#include <services/geometry/dd4hep/JDD4hep_service.h>
```
wdconinc, line 15: Use `"` for all headers inside EICrecon (tree-wide).

This matters since we build EICrecon in CI while a system-wide EICrecon is already available, so using `<services/geometry/dd4hep/JDD4hep_service.h>` will find the system-wide version, while `"services/geometry/dd4hep/JDD4hep_service.h"` will find the local version which might have local changes.

We don't have a CI test for this, but really should have one.

src/algorithms/fardetectors/FarDetectorLinearTrackingConfig.h
```diff
@@ -0,0 +1,19 @@
...
+#include "FarDetectorTrackerClusterConfig.h"
+
+namespace eicrecon {
+  struct FarDetectorLinearTrackingConfig {
+
+    FarDetectorTrackerClusterConfig detconf;
+
+    int   layer_hits_max{4};
+
+    float chi2_max{4};
```
wdconinc, line 16: Add comment with info.

src/algorithms/fardetectors/FarDetectorLinearTrackingConfig.h
```diff
@@ -0,0 +1,19 @@
...
+
+#pragma once
+
+#include <spdlog/spdlog.h>
+#include "FarDetectorTrackerClusterConfig.h"
+
+namespace eicrecon {
+  struct FarDetectorLinearTrackingConfig {
+
+    FarDetectorTrackerClusterConfig detconf;
```
wdconinc, line 12: Nesting config is not allowed. That prevents specifying configuration in a simple format where we don't need to know the exact structure underlying this. Each config should be simple types only.

src/algorithms/fardetectors/FarDetectorLinearTrackingConfig.h
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023, Simon Gardner
+
+#pragma once
+
+#include <spdlog/spdlog.h>
```
wdconinc, line 6: Config should not include anything beyond e.g. `<string>`.
```suggestion
```

src/algorithms/fardetectors/FarDetectorMLReconstructionConfig.h
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2022, 2023, Christopher Dilks, Luigi Dello Stritto
+
```
wdconinc, line 3: Make sure to update all copyright headers to be correct.

src/algorithms/fardetectors/FarDetectorTrackerCluster.cc
```diff
@@ -0,0 +1,143 @@
...
+
+    // Set up clustering variables
+    ROOT::VecOps::RVec<bool> available(module.size(), 1);
+    ROOT::VecOps::RVec<int>  indices  (module.size());
+
+//     std::vector<eicrecon::TrackerCluster*> outputClusters;
+    auto outputClusters = std::make_unique<edm4hep::TrackerHitCollection>();
+
+    for(ulong i = 0; i<indices.size(); i++)
+      indices[i] = i;
```
wdconinc, line 55: Add braces for readability.

src/algorithms/fardetectors/FarDetectorTrackerCluster.cc
```diff
@@ -0,0 +1,143 @@
...
+
+      double xPos      = 0;
+      double yPos      = 0;
+      double zPos      = 0;
+      float  weightSum = 0;
+
+      float esum   = 0;
+      float t0     = 0;
+      float tError = 0;
+      //pCluster->associatedHits = std::vector<edm4eic::RawTrackerHit>;
```
wdconinc, line 71: Remove all commented lines, or add another comment explaining why commented code should be maintained.

src/algorithms/fardetectors/FarDetectorTrackerCluster.cc
```diff
@@ -0,0 +1,143 @@
...
+
+      // Filter to make sure everything is on the same detector layer
+      auto layerFilter = (module==module[maxIndex])*(layer==layer[maxIndex]);
+
+      // Loop over hits, adding neighbouring hits as relevant
+      while(indexList.size()){
+
+        auto index = indexList[0];
+        auto filter = available*layerFilter*(abs(x-x[index])<=1)*(abs(y-y[index])<=1)*(abs(t-t[index])<1);
+        available = available*(!filter);
```
wdconinc, line 91: This seems clunky.

src/algorithms/pid/ParticlesWithPID.cc
```diff
@@ -79,9 +78,9 @@ namespace eicrecon {
                     continue;
                 }
 
-                const auto p_mag = std::hypot(p.x, p.y, p.z);
-                const auto p_phi = std::atan2(p.y, p.x);
-                const auto p_eta = std::atanh(p.z / p_mag);
+                const double p_mag = std::hypot((double)p.x, (double)p.y, (double)p.z);
```
wdconinc, line 81: Use `static_cast`.

wdconinc: and you can just do a `static_cast` on the result of `hypot`.

src/algorithms/pid/ParticlesWithPID.cc
```diff
@@ -92,6 +91,12 @@ namespace eicrecon {
                                    deta < m_cfg.etaTolerance &&
                                    dsphi < sinPhiOver2Tolerance;
 
+               // Matching kinematics with the static variables doesn't work at low angles and within beam divergence
+               // Maybe reconsider variables used or divide into regions
+               if ((p_eta < -5) || (edm4eic::eta(mom) < -5)) {
```
wdconinc, line 96: Can't we use `edm4eic::eta(p)` here?

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -0,0 +1,66 @@
+// Copyright 2023, Simon Gardner
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
+//
+
+#include <Acts/Propagator/Navigator.hpp>
```
wdconinc, line 6: Shouldn't need an Acts header here.

src/detectors/LOWQ2/LowQ2_DNN_CPU.weights.xml
wdconinc, line 1: This cannot be part of the repository.

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -0,0 +1,66 @@
...
+    // -------------------------------
+
+    // Digitization
+    SiliconTrackerDigiConfig digi_cfg;
+    //digi_cfg.timeResolution = 2.5; // Change timing resolution.
+
+    //Clustering config
+    FarDetectorTrackerClusterConfig cluster_cfg;
+    //Ensure same detector is passed to digi and clustering
+    cluster_cfg.readout = "TaggerTrackerHits";
```
wdconinc, line 37: Preferred is:
```suggestion
    FarDetectorTrackerClusterConfig cluster_cfg{
      //Ensure same detector is passed to digi and clustering
      .readout = "TaggerTrackerHits",
    };
```
or even directly into the lines below without naming a variable.

src/factories/fardetectors/FarDetectorLinearTracking_factoryT.h
```diff
@@ -0,0 +1,100 @@
...
+              logger()->debug("Find layer field {}, index = {}",  cfg.detconf.xField, cfg.detconf.x_idx);
+            }
+            if (!cfg.detconf.yField.empty()) {
+              cfg.detconf.y_idx = id_dec->index(cfg.detconf.yField);
+              logger()->debug("Find layer field {}, index = {}", cfg.detconf.yField, cfg.detconf.y_idx);
+            }
+          } catch (...) {
+            logger()->error("Failed to load ID decoder for {}", cfg.detconf.readout);
+            throw JException("Failed to load ID decoder");
+          }
```
wdconinc, line 69: We should not modify `cfg` in the factory. Assume it will become `const` at some point, at least within the scope of this class. This must be done in the algorithm.

src/factories/fardetectors/FarDetectorLinearTracking_factoryT.h
```diff
@@ -0,0 +1,100 @@
...
+
+        }
+
+    private:
+
+        eicrecon::FarDetectorLinearTracking m_reco_algo;        // Actual digitisation algorithm
+
+        dd4hep::BitFieldCoder *id_dec{nullptr};
+        std::shared_ptr<JDD4hep_service> m_geoSvc;
```
wdconinc, line 96: Factory doesn't need these as member fields.

src/factories/fardetectors/FarDetectorMLReconstruction_factoryT.h
```diff
@@ -0,0 +1,91 @@
...
+      eicrecon::FarDetectorMLReconstruction m_reco_algo;        // ML reconstruction fitting algorithm
+
+      float m_electron_beamE = 10*dd4hep::GeV;
+
+      float m_electron{0.000510998928}; //TODO: Link to constant elsewhere?
+
+      // Stuff to add to config
+      std::string m_method_name{"DNN_CPU"};
+      std::string m_file_path{"LowQ2_DNN_CPU.weights.xml"};
+      std::string m_environment_path{"JANA_PLUGIN_PATH"};
```
wdconinc, line 87: Unused in factory.

src/factories/fardetectors/FarDetectorTrackerCluster_factoryT.h
```diff
@@ -0,0 +1,108 @@
...
+            m_log->debug("Find layer field {}, index = {}",  cfg.xField, cfg.x_idx);
+          }
+          if (!cfg.yField.empty()) {
+            cfg.y_idx = id_dec->index(cfg.yField);
+            m_log->debug("Find layer field {}, index = {}", cfg.yField, cfg.y_idx);
+          }
+        } catch (...) {
+          m_log->error("Failed to load ID decoder for {}", cfg.readout);
+          throw JException("Failed to load ID decoder");
+        }
```
wdconinc, line 75: Into algorithm.

src/factories/fardetectors/FarDetectorTrackerCluster_factoryT.h
```diff
@@ -0,0 +1,108 @@
...
+
+      }
+
+
+  private:
+      eicrecon::FarDetectorTrackerCluster m_reco_algo;        // Actual digitisation algorithm
+
+      dd4hep::BitFieldCoder *id_dec{nullptr};
+      std::shared_ptr<JDD4hep_service> m_geoSvc;
```
wdconinc, line 104: Remove.

src/global/tracking/tracking.cc
```diff
@@ -90,5 +91,21 @@ void InitPlugin(JApplication *app) {
...
+      {"CombinedTrajectories"},
+        app));    // Output collection name
+
+    // Tracker hits collector from ACTS and other factories
+    app->Add(new JChainMultifactoryGeneratorT<TrackerTrajectoryCollector_factory>(
+                                                                                  "CombinedSeededTrajectories",
+      {"CentralCKFSeededTrajectories",  // ACTS output
+          "LowQ2Trajectories"},            // Low Q2 output
+      {"CombinedSeededTrajectories"},
+        app));  // Output collection name
```
wdconinc, line 108: Fix alignment.

wdconinc:
The more we can split this up in little pieces and submit as separate PRs, the sooner this can get merged in pieces too. Take what's stable and split it off, but make sure to address the comments here since they would not propagate to a new PR.

michael-pitt:
Hi @simonge 
Does the beam energy is hardcoded `float m_electron_beamE = 10*dd4hep::GeV;` or it can be defined by the user?

veprbl:
> > Ideally those go into detector benchmarks [1] so that we always have the latest artifacts produced for every version of the epic geometry. Of course, it would really help to be able to see your code to have a concrete discussion.
> > [1] https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks
> 
> Is this still the correct location for benchmarks and training scripts? The same repository is on this github organisation.

Yes, the eicweb is primary and GitHub is currently only a mirror. Do you have access to eicweb?

simonge:
> > > Ideally those go into detector benchmarks [1] so that we always have the latest artifacts produced for every version of the epic geometry. Of course, it would really help to be able to see your code to have a concrete discussion.
> > > [1] https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks
> > 
> > 
> > Is this still the correct location for benchmarks and training scripts? The same repository is on this github organisation.
> 
> Yes, the eicweb is primary and GitHub is currently only a mirror. Do you have access to eicweb?

I believe I have access, yes thanks. Can I check if my understanding of how this might end up working is reasonably accurate, how much of this should/could be automated.

A large source of specific training events are stored somewhere.
If the detector geometry is updated (ideally only triggered on specific features changing) the simulation and reconstruction are run, only needing to output the relevant collections.
The network is retrained on the output and tagged and saved (somewhere) with benchmarks that check it's not messed everything up.

veprbl:
> > > > Ideally those go into detector benchmarks [1] so that we always have the latest artifacts produced for every version of the epic geometry. Of course, it would really help to be able to see your code to have a concrete discussion.
> > > > [1] https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks
> > > 
> > > 
> > > Is this still the correct location for benchmarks and training scripts? The same repository is on this github organisation.
> > 
> > 
> > Yes, the eicweb is primary and GitHub is currently only a mirror. Do you have access to eicweb?
> 
> I believe I have access, yes thanks. Can I check if my understanding of how this might end up working is reasonably accurate, how much of this should/could be automated.
> 
> A large source of specific training events are stored somewhere. If the detector geometry is updated (ideally only triggered on specific features changing) the simulation and reconstruction are run, only needing to output the relevant collections. The network is retrained on the output and tagged and saved (somewhere) with benchmarks that check it's not messed everything up.

This all sounds correct. I think, it would be okay to run training on every detector_benchmarks run using only few events. For producing the calibrations for production we'd manually trigger benchmarks to simulate many more events. My understanding is that @sly2j can pin arbitrary artifacts so that they don't get removed from eicweb and we can access them from there.

PR #679, opened 2023-05-24
feat(dRICH): link unique `Track` to `TrackSegment`s

c-dilks:

This is more of a workaround for an uncomfortable design issue:
- the factory `RichTrack_factory` creates multiple track projections (datatype `edm4eic::TrackSegment`) for a single input `eicrecon::TrackingResultTrajectory`, but downstream code needs a way to know which of these `edm4eic::TrackSegments` came from the same input `eicrecon::TrackingResultTrajectory`
- `eicrecon::TrackingResultTrajectory` is not an `EDM4*` datatype, and therefore the `edm4eic::TrackSegment` objects this factory creates cannot link to them
- in the data model `edm4eic::TrackSegment` can have a 1-1 relation to an `edm4eic::Track`
- therefore, as a workaround, we generate a unique, empty `edm4eic::Track` for each input `eicrecon::TrackingResultTrajectory`, which can be used in this 1-1 relation from `edm4eic::TrackSegment` as a unique identifier to encode which `edm4eic::TrackSegment`s came from the same input track
- finally, an additional 'TrackID' output tag is needed, to to guarantee these related `edm4eic::Track` objects have an owner and are accessible by downstream readers

Additional thoughts
1. if `eicrecon::TrackingResultTrajectory` were an `EDM4*` datatype, than that is what `edm4eic::TrackSegment` should link to; alternatively we should be able to run the `TrackPropagator` for `edm4eic::Track` objects
2. if two different PID subsystems use this factory, this workaround will not work, since the created unique `edm4eic::Track` objects will be different; but the hope is that thought (1) above can be addressed, then we can get rid of this workaround and the two PID subsystems will have a way of identifying a unique charged particle track which passed through both of them
3. alternative solution: `RichTrack_factory` for the dRICH creates 2 `edm4eic::TrackSegmentCollection`s: one for aerogel and another for gas. We could combine their `edm4eic::TrackSegment`s to 1 (combining gas and aerogel), but this solution seems to have less flexibility whenever we want to implement another PID algorithm other than IRT, and does not resolve thought (2) above

- [x] Bug fix (a downstream bug on the `irt-algo` branch)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Tests for the changes have been added (running benchmarks locally)

Yes, already addressed on the `irt-algo` branch; no breaking changes for `main`

No

### DIFF
--- a/src/detectors/DRICH/DRICH.cc
+++ b/src/detectors/DRICH/DRICH.cc
@@ -77,7 +77,7 @@ extern "C" {
     app->Add(new JChainMultifactoryGeneratorT<RichTrack_factory>(
           "DRICHTracks",
           {"CentralCKFTrajectories"},
-          {"DRICHAerogelTracks", "DRICHGasTracks"},
+          {"DRICHAerogelTracks", "DRICHGasTracks", "DRICHTrackIDs"},
           track_cfg,
           app
           ));
--- a/src/detectors/DRICH/DRICH.cc
+++ b/src/detectors/DRICH/DRICH.cc
@@ -77,7 +77,7 @@ extern "C" {
     app->Add(new JChainMultifactoryGeneratorT<RichTrack_factory>(
           "DRICHTracks",
           {"CentralCKFTrajectories"},
-          {"DRICHAerogelTracks", "DRICHGasTracks", "DRICHTrackIDs"},
+          {"DRICHAerogelTracks", "DRICHGasTracks"},
           track_cfg,
           app
           ));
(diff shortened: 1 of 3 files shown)


c-dilks:
The more I work with this implementation the less happy I am about it. I now prefer to have `RichTrack_factory` output 3 track propagations:
- aerogel track points in a TrackSegment
- gas track points in a TrackSegment
- the above track points all merged to one TrackSegment

This will maximize flexibility downstream, and still provide single merged TrackSegments. Hopefully someday we can relate these TrackSegments to the input Trajectory (or Track).

PR #684, opened 2023-05-30
CalorimeterHitReco: proper handling for segmentation for dimension

veprbl:
Not all segmentation types will be ordered as XYZ. One notable example: CartesianStripZ gives only one dimension, which triggers ASAN error.

This implements only support for CartesianGridXY. It is assumed that we should be looking to deprecate dimension field from EDM4eic. Hence, this implementation only fixes the case which have worked before. The case where wrong coordinates were set before will now report warnings and use the cartesian bounding box.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitReco.cc
@@ -208,10 +208,18 @@ void CalorimeterHitReco::AlgorithmProcess() {
 //                dd4hep::Position(gpos.x(), gpos.y(), gpos.z()));//dd4hep::Position(gpos.x, gpos.y, gpos.z)
         std::vector<double> cdim;
         // get segmentation dimensions
-        if (converter->findReadout(local).segmentation().type() != "NoSegmentation") {
-            cdim = converter->cellDimensions(cellID);
+        auto segmentation_type = converter->findReadout(local).segmentation().type();
+        auto cell_dim = converter->cellDimensions(cellID);
+        if (segmentation_type == "CartesianGridXY") {
+            cdim.resize(3);
+            cdim[0] = cell_dim[0];
+            cdim[1] = cell_dim[1];
             m_log->error("Using segmentation for cell dimensions: {}", fmt::join(cdim, ", "));
         } else {
+            if (segmentation_type != "NoSegmentation") {
+                m_log->warn("Usupported segmentation type \"{}\"", segmentation_type);
+            }
+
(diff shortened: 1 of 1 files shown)


veprbl:
This has been branched off from #2010, where it fixes reconstruction under ASAN.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -208,21 +208,31 @@ void CalorimeterHitReco::AlgorithmProcess() {
...
         std::vector<double> cdim;
         // get segmentation dimensions
-        if (converter->findReadout(local).segmentation().type() != "NoSegmentation") {
-            cdim = converter->cellDimensions(cellID);
+        auto segmentation_type = converter->findReadout(local).segmentation().type();
+        auto cell_dim = converter->cellDimensions(cellID);
+        if (segmentation_type == "CartesianGridXY") {
+            cdim.resize(3);
+            cdim[0] = cell_dim[0];
+            cdim[1] = cell_dim[1];
```
wdconinc, line 216: Just confirming, this doesn't leave undefined `cdim[2]`?

veprbl: Yes, resize requires `T` to be `DefaultInsertable`. In order to not zero the appended items, one would need to provide a custom `Allocator`, which we, obviously, don't do.

wdconinc (approved):
Looks good.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -208,21 +208,31 @@ void CalorimeterHitReco::AlgorithmProcess() {
...
         } else {
+            if (segmentation_type != "NoSegmentation") {
+                m_log->warn("Usupported segmentation type \"{}\"", segmentation_type);
+            }
+
             // Using bounding box instead of actual solid so the dimensions are always in dim_x, dim_y, dim_z
             cdim = converter->findContext(cellID)->volumePlacement().volume().boundingBox().dimensions();
             std::transform(cdim.begin(), cdim.end(), cdim.begin(),
                            std::bind(std::multiplies<double>(), std::placeholders::_1, 2));
+            m_log->error("Using bounding box for cell dimensions: {}", fmt::join(cdim, ", "));
```
wdconinc, line 227: Were these `error`s intended to be `debug`s? Physics benchmark now looks like https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/1659794 ...

PR #688, opened 2023-06-01
fix: throw exceptions for algorithm failures, instead of warning printouts

c-dilks:
There are several factories that `catch` exceptions in underlying algorithms and just print a warning. We should try to `throw` these exceptions instead, however:
- there may still be some algorithms where this is not a good idea, in other words, there may be a reason just printing a warning message is preferred; IMO this should be fixed in the algorithm and we shouldn't just be warning on exceptions
- this PR likely does not address all of the cases

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no, but there may be "regular" warning messages that are now thrown exceptions

maybe (see bullet points above)

### DIFF
--- a/src/global/digi/PhotoMultiplierHitDigi_factory.cc
+++ b/src/global/digi/PhotoMultiplierHitDigi_factory.cc
@@ -50,7 +50,7 @@ void eicrecon::PhotoMultiplierHitDigi_factory::Process(const std::shared_ptr<con
     SetCollection<edm4eic::MCRecoTrackerHitAssociation>(GetOutputTags()[1], std::move(result.hit_assocs));
   }
   catch(std::exception &e) {
-    m_log->warn("Exception in underlying algorithm: {}. Event data will be skipped", e.what());
+    throw JException(e.what());
   }
 
 }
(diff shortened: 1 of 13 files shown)


wdconinc (approved):
This looks good to me.

Part of the 'problem' here stems from the arches/brycecanyon split and the fact that we have currently no way to define a different algorithm graph for each. That means that we must now define a single algorithm graph and rely on the behavior that missing collections in the input file are indicators of which branch in the graph to simply ignore. That's not ideal, but it is what we have.

So, the question becomes, is throwing exceptions and handling them downstream in a consistent way going to allow us to do the above? To me, it seems like it will (since it doesn't change the calorimeter factories at all).

veprbl:
Regarding arches/brycecanyon configurations support, it must be that there are bugs with running algorithms for non-existing collections. My reasoning for assuming a bug is following: Setting `input_tag=""` is a suggested way to disable a factory. I assumed that worked because of the unavailable names, not because the `""` had a special meaning. If that is true, the factories are supposed to disable themselves and not even let algorithms fail.

Issue #700, opened 2023-06-05
HcalEndcapNRecHits segmentation PolarGridRPhi2 unsupported in CalorimeterHitReco


wdconinc:
The HcalEndcapN uses a [PolarGridRPhi2](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/PolarGridRPhi2.cpp) segmentation. This is [unsupported in CalorimeterHitReco](https://github.com/eic/EICrecon/blob/main/src/algorithms/calorimetry/CalorimeterHitReco.cc#LL220C37-L220C37), resulting in a warning for each event.

PR #701, opened 2023-06-05
feat: support PolarGridRPhi2 segmentation in CalorimeterHitReco

wdconinc:
This supports PolarGridRPhi2 segmentations in CalorimeterHitReco, similar to the CartesianGridXY. [As far as I know](https://github.com/search?q=repo%3Aeic%2Fepic%20PolarGridRPhi2&type=code), only used by HcalEndcapN so impact needs to be assessed by that group.

- [x] Bug fix (issue #700)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Maybe, cell dimension not written as bounding box but as actual cell dimensions, may affect clustering downstream.

Yes, cell dimension not written as bounding box but as actual cell dimensions, may affect clustering downstream.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitReco.cc
@@ -215,9 +215,14 @@ void CalorimeterHitReco::AlgorithmProcess() {
             cdim[0] = cell_dim[0];
             cdim[1] = cell_dim[1];
             m_log->error("Using segmentation for cell dimensions: {}", fmt::join(cdim, ", "));
+        } else if (segmentation_type == "PolarGridRPhi2") {
+            cdim.resize(3);
+            cdim[0] = cell_dim[0];
+            cdim[1] = cell_dim[1];
+            m_log->error("Using segmentation for cell dimensions: {}", fmt::join(cdim, ", "));
         } else {
             if (segmentation_type != "NoSegmentation") {
-                m_log->warn("Usupported segmentation type \"{}\"", segmentation_type);
+                m_log->warn("Unsupported segmentation type \"{}\"", segmentation_type);
             }
 
             // Using bounding box instead of actual solid so the dimensions are always in dim_x, dim_y, dim_z


src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -215,9 +215,14 @@ void CalorimeterHitReco::AlgorithmProcess() {
...
             m_log->debug("Using segmentation for cell dimensions: {}", fmt::join(cdim, ", "));
+        } else if (segmentation_type == "PolarGridRPhi2") {
+            cdim.resize(3);
+            cdim[0] = cell_dim[0];
+            cdim[1] = cell_dim[1];
+            m_log->debug("Using segmentation for cell dimensions: {}", fmt::join(cdim, ", "));
         } else {
             if (segmentation_type != "NoSegmentation") {
-                m_log->warn("Usupported segmentation type \"{}\"", segmentation_type);
+                m_log->warn("Unsupported segmentation type \"{}\"", segmentation_type);
```
wdconinc, line 225: Gratuitous typo correction.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -215,9 +215,14 @@ void CalorimeterHitReco::AlgorithmProcess() {
             cdim[0] = cell_dim[0];
             cdim[1] = cell_dim[1];
             m_log->debug("Using segmentation for cell dimensions: {}", fmt::join(cdim, ", "));
+        } else if (segmentation_type == "PolarGridRPhi2") {
+            cdim.resize(3);
+            cdim[0] = cell_dim[0];
+            cdim[1] = cell_dim[1];
```
veprbl, line 221: Makes no sense to implement this one. Downstream code expects things to be comparable in XYZ with cm as units.

wdconinc: So is bounding box what we want? We should at least be explicit about it.

veprbl: No, we don't want it either, and likely we simply don't use the `dimensions` field for EndcapNHcal, at least in clusterer. This design is broken, I will be pushing to drop this eventually.

wdconinc: But we should do something that doesn't cause an error message for every event. It's affecting ops.

veprbl: Didn't we hide the message into debug?

wdconinc: Not the message for when the segmentation isn't recognized. That's still a warning, as it probably should be.

PR #702, opened 2023-06-05
CalorimeterHitReco: warn about unsupported segementation only once

veprbl:
There doesn't seem to be a warn_once facility in spdlog and algorithms are not supposed to rely on common services, so the flag is implemented inline.

Resolves: #700
Closes: #701


wdconinc:
Not sure how a `warn_once` would be implemented in a service. It would seem to need to keep track of all callers, which would impose a different calling syntax...

That said, is `call_once` potentially appropriate here, https://en.cppreference.com/w/cpp/thread/call_once ?

wdconinc (approved):
In any case, this is fine and ready for merging. We can think about how to implement warn_once consistently in a different PR.

veprbl:
> Possibly we can collapse duplicates instead, https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/sinks/dup_filter_sink.h .

Saw that. It looks like it won't help with the interleaved messages.

PR #713, opened 2023-06-09
Add GitHub action for PR backporting

veprbl:
This allows one to create labels with special name like `backport v1.2` that would automatically create a PR to duplicate change into a maintenance branch.

### DIFF
--- a/.github/workflows/pr-backport.yml
+++ b/.github/workflows/pr-backport.yml
@@ -1,0 +1,17 @@
+name: Backport merged pull request
+on:
+  pull_request_target:
+    types: [closed]
+permissions:
+  contents: write # so it can comment
+  pull-requests: write # so it can create pull requests
+jobs:
+  backport:
+    name: Backport pull request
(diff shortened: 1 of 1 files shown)


wdconinc:
Yes, thanks!!! We have needed this not just here but in eic-spack too!

wdconinc (approved):
Security-related questions (since `pull_request_target` with `write/write` permissions):
- who can label?
- can we block labels that refer to non-`v[0-9].[0-9]` branches?

I am trying to ensure that no external fork can create a hostile PR, label it, close it and still get stuff merged.

veprbl:
Only org-members can apply label. The action only creates a PR after the original is merged. The backport PR is not a threat – it's not approved or merged automatically.

veprbl:
I don't think blocking of labels is needed, it's not like incorrect labels will appear or that users can't make a PR targeting arbitrary branch. It can be done, though:
https://github.com/korthout/backport-action/blob/main/README.md#label_pattern

PR #716, opened 2023-06-14
fix: CKF makes tracks instead of particles

osbornjd:
This PR changes the default behavior of the CKF to make `edm4eic::Tracks` instead of `edm4eic::ReconstructedParticles`, where a track would be preferable since we won't necessarily have calo/PID information yet. This prefaces changes to the vertexing EDM which will better accommodate vertex to track relations.

To keep the default behavior the same, there is a new factory added that just converts the tracks to reconstructed particles in the same way that is currently done. This way, there should be no effect to the downstream physics output. At some point down the road this could be modified to e.g. add more detailed information (e.g. calo/PID info), or just removed completely by another module that actually combines all of the subsystem information.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No


osbornjd:
Marked as draft as I'm interested to see if this breaks the CI at all, I can't tell any changes from my local tests.

PR #720, opened 2023-06-29
feat: more modern CMake by using targets instead of include_directories

wdconinc:
This aims to modernize, simplify, and in some cases make more robust the CMake configuration. Primarily this involves removing constructs like `set(Acts_INCLUDE_DIR ${Acts_DIR}/../../../include)` which are prone to change, and are not necessary when using targets such as `ActsCore`, etc. This PR also removes a number of unused blocks of code.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: cmake modernization)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Not so surprised that we don't find the data model glue.
```
/home/runner/work/EICrecon/EICrecon/src/services/io/podio/JEventProcessorPODIO.cc:7:10: fatal error: 'datamodel_glue.h' file not found
#include <datamodel_glue.h>
         ^~~~~~~~~~~~~~~~~~
1 error generated.
```
since that is one header we do need to pass an explicit include directory for.

veprbl:
The issue with datamodel_glue is that it needs to be includable from any factory. We need to get rid of the hack, but I'm not sure if facilities in PODIO exist or even were reported. cc @nathanwbrei

src/services/io/podio/CMakeLists.txt
```diff
@@ -30,18 +30,17 @@ plugin_add(${PLUGIN_NAME} )
...
 add_custom_command(
         OUTPUT ${podio_BINARY_DIR}/datamodel_glue.h ${podio_BINARY_DIR}/datamodel_includes.h
         COMMAND python3 ${PROJECT_SOURCE_DIR}/make_datamodel_glue.py WORKING_DIR=${podio_BINARY_DIR} EDM4HEP_INCLUDE_DIR=${EDM4HEP_INCLUDE_DIR} EDM4EIC_INCLUDE_DIR=${EDM4EIC_INCLUDE_DIR}
         MAIN_DEPENDENCY ${PROJECT_SOURCE_DIR}/make_datamodel_glue.py
+	DEPENDS ${EDM4HEP_DEPENDS} ${EDM4EIC_DEPENDS}
 )
 
-# Add include directories (works same as target_include_directories)
-message(STATUS "fmt_INCLUDE_DIR=${fmt_INCLUDE_DIR}")
-plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_BINARY_DIR}/src ${CMAKE_CURRENT_SOURCE_DIR} ${EICRECON_SOURCE_DIR}/src)
```
wdconinc, line 41: This is the one that should stay here to find the `datamodel_glue` header.

wdconinc:
Removing the need for the data model glue itself is out of the scope of this PR. It'd be nice to have something that doesn't require it but this is just about CMake cleanup.

nathanwbrei:
The PODIO people definitely know that we need the `using` statement. However I'm not aware of anyone having ever implemented it, probably because I volunteered myself to PR it in a moment of hopefulness. I agree with Wouter that we should keep it out of the scope for this particular PR but remember to remove the corresponding `target_include_directories` when the time comes.

nathanwbrei:
I still owe you guys a JANA 'cleanup' release that removes the need for a bunch of other ugly hacks in EICrecon. Let me know if this affects or is affected by anything else on the roadmap. I've been really busy with a different project but I haven't forgotten you!

wdconinc:
Clean up release is out of scope for the July campaign, but it'd be great if we can get that tagged by mid July or so.

wdconinc:
After this last commit modifying the `make_datamodel_glue.py`, all explicit `../../` constructs are now gone.

When using a generator expression to get `INTERFACE_INCLUDE_DIRECTORIES` we cannot use the `file(GLOB)` to `DEPENDS` on the list of all headers in EDM4hep and EDM4eic, but that was only an addition in this PR and intended to resolve a corner case of installing different EDM4hep/EDM4eic headers into the same `/usr/local/include` path. I think that the generator expression will force a re-execution of `make_datamodel_glue.py` anyway when either the EDM4hep or EDM4eic targets are updated.

src/services/io/podio/CMakeLists.txt
```diff
@@ -31,17 +27,19 @@ plugin_glob_all(${PLUGIN_NAME})
...
+                WORKING_DIR=${CMAKE_CURRENT_BINARY_DIR}
+                EDM4HEP_INCLUDE_DIR=$<REMOVE_DUPLICATES:$<TARGET_PROPERTY:EDM4HEP::edm4hep,INTERFACE_INCLUDE_DIRECTORIES>>
+                EDM4EIC_INCLUDE_DIR=$<REMOVE_DUPLICATES:$<TARGET_PROPERTY:EDM4EIC::edm4eic,INTERFACE_INCLUDE_DIRECTORIES>>
         MAIN_DEPENDENCY ${PROJECT_SOURCE_DIR}/make_datamodel_glue.py
 )
 
 # Add include directories (works same as target_include_directories)
-message(STATUS "fmt_INCLUDE_DIR=${fmt_INCLUDE_DIR}")
-plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_BINARY_DIR}/src ${CMAKE_CURRENT_SOURCE_DIR} ${EICRECON_SOURCE_DIR}/src)
+plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
```
veprbl, line 39:
```suggestion
plugin_include_directories(${PLUGIN_NAME} PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
```

cmake/jana_plugin.cmake
```diff
@@ -223,37 +219,36 @@ macro(plugin_add_event_model _name)
...
     if(NOT EDM4EIC_FOUND)
         find_package(EDM4EIC REQUIRED)
-        set(EDM4EIC_INCLUDE_DIR ${EDM4EIC_DIR}/../../include)
     endif()
 
     # Add include directories
-    # (same as target_include_directories but for both plugin and library)
     # ${podio_BINARY_DIR} is an include path to datamodel_glue.h
-    plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${podio_INCLUDE_DIR} ${EDM4EIC_INCLUDE_DIR} ${EDM4HEP_INCLUDE_DIR} ${podio_BINARY_DIR})
+    plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${podio_BINARY_DIR})
```
veprbl, line 226:
```suggestion
    plugin_include_directories(${PLUGIN_NAME} PUBLIC ${podio_BINARY_DIR})
```

veprbl (approved):

veprbl (dismissed):

cmake/jana_plugin.cmake
```diff
@@ -223,37 +219,36 @@ macro(plugin_add_event_model _name)
...
     if(NOT EDM4EIC_FOUND)
         find_package(EDM4EIC REQUIRED)
-        set(EDM4EIC_INCLUDE_DIR ${EDM4EIC_DIR}/../../include)
     endif()
 
     # Add include directories
-    # (same as target_include_directories but for both plugin and library)
     # ${podio_BINARY_DIR} is an include path to datamodel_glue.h
-    plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${podio_INCLUDE_DIR} ${EDM4EIC_INCLUDE_DIR} ${EDM4HEP_INCLUDE_DIR} ${podio_BINARY_DIR})
+    plugin_include_directories(${PLUGIN_NAME} PUBLIC ${podio_BINARY_DIR})
```
wdconinc, line 226: Actually, this doesn't do anything. `podio_BINARY_DIR` isn't a thing.

veprbl (dismissed):

veprbl:
```
In file included from /home/runner/work/EICrecon/EICrecon/src/detectors/EEMC/EEMC.cc:6:
In file included from /home/runner/work/EICrecon/EICrecon/src/extensions/jana/JChainMultifactoryGeneratorT.h:12:
/home/runner/work/EICrecon/EICrecon/src/extensions/jana/JChainMultifactoryT.h:22:10: fatal error: 'datamodel_glue.h' file not found
#include <datamodel_glue.h>
         ^~~~~~~~~~~~~~~~~~
1 error generated.
```

cmake/jana_plugin.cmake
veprbl: It's a thing. CMake defines <projectname>_BINARY_DIR. This allows to reference podio plugin independently of what is the top level project.

wdconinc: Right. I meant that we're not writing anything to `${podio_BINARY_DIR}` anymore in `datamodel_glue`.

wdconinc: But we should. Let me redo this.

src/services/io/podio/CMakeLists.txt
```diff
@@ -31,17 +27,19 @@ plugin_glob_all(${PLUGIN_NAME})
...
+                WORKING_DIR=${CMAKE_CURRENT_BINARY_DIR}
+                EDM4HEP_INCLUDE_DIR=$<REMOVE_DUPLICATES:$<TARGET_PROPERTY:EDM4HEP::edm4hep,INTERFACE_INCLUDE_DIRECTORIES>>
+                EDM4EIC_INCLUDE_DIR=$<REMOVE_DUPLICATES:$<TARGET_PROPERTY:EDM4EIC::edm4eic,INTERFACE_INCLUDE_DIRECTORIES>>
         MAIN_DEPENDENCY ${PROJECT_SOURCE_DIR}/make_datamodel_glue.py
 )
 
 # Add include directories (works same as target_include_directories)
-message(STATUS "fmt_INCLUDE_DIR=${fmt_INCLUDE_DIR}")
-plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_BINARY_DIR}/src ${CMAKE_CURRENT_SOURCE_DIR} ${EICRECON_SOURCE_DIR}/src)
+plugin_include_directories(${PLUGIN_NAME} PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
```
veprbl, line 39: Not sure that this is needed. We are not supposed to link from plugin to plugin.
```suggestion
```

cmake/jana_plugin.cmake
```diff
@@ -223,37 +219,32 @@ macro(plugin_add_event_model _name)
...
     if(NOT EDM4EIC_FOUND)
         find_package(EDM4EIC REQUIRED)
-        set(EDM4EIC_INCLUDE_DIR ${EDM4EIC_DIR}/../../include)
     endif()
 
-    # Add include directories
-    # (same as target_include_directories but for both plugin and library)
-    # ${podio_BINARY_DIR} is an include path to datamodel_glue.h
-    plugin_include_directories(${PLUGIN_NAME} SYSTEM PUBLIC ${podio_INCLUDE_DIR} ${EDM4EIC_INCLUDE_DIR} ${EDM4HEP_INCLUDE_DIR} ${podio_BINARY_DIR})
-
```
veprbl, line 233: The `${podio_BINARY_DIR}` was intended for `#include <datamodel_glue.h>` to work.

wdconinc: Yeah, I see that now, but it's confusing. We should not redefine the `podio` project in `src/services/io/podio` when we also include a target `podio` that sets its own variables `podio_*`. I'd rather we called this internal project something else that is more explicit, like `podio_datamodel_glue` or so. But then it differs from the plugin name, so let's just keep this as is.

src/services/io/podio/CMakeLists.txt
wdconinc: Confirmed that this is needed. Without this, `datamodel_glue.h` is not found by `src/services/io/podio/JEventProcessorPODIO.cc`.

cmake/jana_plugin.cmake
wdconinc: Maybe `project(datamodel)` is sufficient to make this separation clearer. The difference in name with the plugin does not cause problems.

wdconinc: https://github.com/eic/EICrecon/pull/720/commits/cd869e67d502f340a37f58c09e7f3e34c439abba

src/services/io/podio/CMakeLists.txt
wdconinc: Reverted in https://github.com/eic/EICrecon/pull/720/commits/acb8e746f5b4744e9bc4871ba7eaa1e756835eb3

src/services/io/podio/CMakeLists.txt
```diff
@@ -31,21 +27,23 @@ plugin_glob_all(${PLUGIN_NAME})
...
 
 # Add libraries (works same as target_include_directories)
-plugin_link_libraries(${PLUGIN_NAME} EDM4HEP::edm4hep EDM4HEP::edm4hepDict EDM4EIC::edm4eic EDM4EIC::edm4eic_utils podio::podioRootIO)
+plugin_link_libraries(${PLUGIN_NAME} fmt::fmt EDM4HEP::edm4hep EDM4HEP::edm4hepDict EDM4EIC::edm4eic EDM4EIC::edm4eic_utils podio::podioRootIO)
 
 # Create a ROOT dictionary with the vector<edm4hep::XXXData> types defined. Without
 # this, root will complain about not having a compiled CollectionProxy.
-root_generate_dictionary(G__datamodel_vectors ${podio_BINARY_DIR}/datamodel_includes.h MODULE ${PLUGIN_NAME}_plugin LINKDEF datamodel_LinkDef.h)
+root_generate_dictionary(G__datamodel_vectors ${CMAKE_CURRENT_BINARY_DIR}/datamodel_includes.h MODULE ${PLUGIN_NAME}_plugin LINKDEF datamodel_LinkDef.h)
```
veprbl, line 46:
```suggestion
root_generate_dictionary(G__datamodel_vectors ${datamodel_BINARY_DIR}/datamodel_includes.h MODULE ${PLUGIN_NAME}_plugin LINKDEF datamodel_LinkDef.h)
```
Make things searchable.

src/services/io/podio/CMakeLists.txt
veprbl:
```suggestion
plugin_include_directories(${PLUGIN_NAME} PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
```

src/services/io/podio/CMakeLists.txt
wdconinc: I'm not sure about this one. It also means that the internals of the project would depend on the name of the project. If someone changes the name from `datamodel` to something else, wouldn't they reasonably expect that only external users would need updating?

veprbl (approved):

PR #722, opened 2023-06-29
ci: asan and lsan suppressions

wdconinc:
This adds asan and lsan suppressions files for use in CI.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/asan.supp
+++ b/.github/asan.supp
@@ -1,2 +1,1 @@
-# Ref. https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-reports-in-external-libraries
 interceptor_via_lib:libCore.so
--- a/.github/asan.supp
+++ b/.github/asan.supp
@@ -1,2 +1,2 @@
 # Ref. https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-reports-in-external-libraries
-interceptor_via_lib:/usr/local/lib/root/libCore.so
+interceptor_via_lib:libCore.so
--- a/.github/asan.supp
+++ b/.github/asan.supp
@@ -1,0 +1,2 @@
+# Ref. https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-reports-in-external-libraries
+interceptor_via_lib:/usr/local/lib/root/libCore.so
--- a/.github/lsan.supp
+++ b/.github/lsan.supp
@@ -1,1 +1,2 @@
 leak:libCore.so
+leak:edm4hep::*::createBuffers
--- a/.github/lsan.supp
+++ b/.github/lsan.supp
@@ -1,2 +1,1 @@
-# Ref. https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer
-/usr/local/lib/root/libCore.so
+leak:libCore.so
--- a/.github/lsan.supp
+++ b/.github/lsan.supp
@@ -1,0 +1,2 @@
+# Ref. https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer
+/usr/local/lib/root/libCore.so
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -33,8 +33,8 @@ jobs:
             CMAKE_BUILD_TYPE: Debug
             CXXFLAGS: -fprofile-instr-generate -fcoverage-mapping
     env:
-      ASAN_OPTIONS: malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
-      LSAN_OPTIONS: exitcode=0
+      ASAN_OPTIONS: suppressions=.github/asan.supp:malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
+      LSAN_OPTIONS: suppressions=.github/lsan.supp:exitcode=0
     steps:
     - uses: actions/checkout@v3
     - name: Prepare ccache timestamp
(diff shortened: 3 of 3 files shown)


.github/asan.supp
```diff
@@ -0,0 +1,2 @@
+# Ref. https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-reports-in-external-libraries
+interceptor_via_lib:/usr/local/lib/root/libCore.so
```
wdconinc, line 2:
```console
AddressSanitizer: failed to parse suppressions
```

wdconinc:
It's possible this asan suppressions file is only supported in llvm...

wdconinc:
I was using CI to test here. I didn't test this yet on my own system.

veprbl:
Backtrace leaks are to be fixed by https://github.com/JeffersonLab/JANA2/pull/224, I suppose.
I'm investigating leaks in calo algorithms #731

veprbl:
> [FATAL] Detected timeout in worker! Stopping.

wdconinc:
I hate those timeouts.

veprbl:
Should we disable them in EICrecon globally? I don't understand how multithreadedness could make a difference.

wdconinc:
> Should we disable them in EICrecon globally?

Yes, but I don't know how easy it would be to do that, while still allowing people to override the default. It's a JANA2 setting so changing the default should happen there, but it's considered a desirable feature (not sure what's desirable about continuing when something has gone wrong, but that's a matter of philosophy I guess).

I was kinda waiting waiting until we have proper configuration management and I would have added it there as one of the first settings.

veprbl:
Maybe timeouts exist because people expect to encounter deadlocks.

wdconinc:
That is the argument: you don't want your production jobs to hang indefinitely just because they cannot connect to a database server for calibration info. To which I answered: you don't want your production jobs to succeed when the calibrations are not those from your database server. I'm all for timeouts on database connections, but when it fails to connect within a timeout, it should just fail the job.

.github/lsan.supp
```diff
@@ -0,0 +1 @@
+leak:libCore.so
```
wdconinc, line 1:
```suggestion
leak:libCore.so
leak:libedm4hep.so
```

.github/asan.supp
```diff
@@ -0,0 +1 @@
+interceptor_via_lib:libCore.so
```
wdconinc, line 1:
```suggestion
interceptor_via_lib:libCore.so
interceptor_via_lib:libedm4hep.so
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -329,8 +329,8 @@ jobs:
         particle: [e]
         detector_config: [ip6_extended]
     env:
-      ASAN_OPTIONS: malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
-      LSAN_OPTIONS: exitcode=0
+      ASAN_OPTIONS: suppressions=.github/asan.supp:malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
+      LSAN_OPTIONS: suppressions=.github/lsan.supp:exitcode=0
```
wdconinc, line 333:
```suggestion
      ASAN_OPTIONS: suppressions=${{ github.workspace }}/.github/asan.supp:malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
      LSAN_OPTIONS: suppressions=${{ github.workspace }}/.github/lsan.supp:exitcode=0
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -380,8 +380,8 @@ jobs:
         - beam: 5x41
           minq2: 1000
     env:
-      ASAN_OPTIONS: malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
-      LSAN_OPTIONS: exitcode=0
+      ASAN_OPTIONS: suppressions=.github/asan.supp:malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
+      LSAN_OPTIONS: suppressions=.github/lsan.supp:exitcode=0
```
wdconinc, line 384:
```suggestion
      ASAN_OPTIONS: suppressions=${{ github.workspace }}/.github/asan.supp:malloc_context_size=20:detect_leaks=1:verify_asan_link_order=0:detect_stack_use_after_return=1:detect_odr_violation=0:new_delete_type_mismatch=0
      LSAN_OPTIONS: suppressions=${{ github.workspace }}/.github/lsan.supp:exitcode=0
```

.github/lsan.supp
veprbl: Would that not hide EDM4hep objects allocated by EICrecon?

wdconinc: That's what I was wondering (and reason why I hadn't committed it after making the comment). Needs testing but my local EICrecon copy is tied up...

veprbl: Which leak are you looking to suppress by filtering out EDM4hep?

wdconinc: The numerous occurrences of e.g. https://github.com/eic/EICrecon/actions/runs/5448228500/jobs/9911415555?pr=722#step:5:4037:
```
Indirect leak of 336 byte(s) in 14 object(s) allocated from:
    #0 0x7fa81439d4c8 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:95
    #1 0x7fa80f567c2b in edm4hep::(anonymous namespace)::createBuffers(bool) [clone .lto_priv.2] (/usr/local/lib/libedm4hep.so+0x65c2b)
```

But I was going to test what else it hides.

veprbl: How about 
```suggestion
leak:libCore.so
leak:createBuffers
```
then?

wdconinc: Or `leak:edm4hep::*::createBuffers` to make sure it's still in EDM4hep. I'll give that a try.

wdconinc:
```suggestion
leak:libCore.so
leak:edm4hep::*::createBuffers
```

wdconinc:
FYI, I backported https://github.com/JeffersonLab/JANA2/pull/224 into the JANA2 we are using, so the exception backtrace leaks should be gone.

veprbl (approved):

PR #731, opened 2023-07-02
[ci-asan-lsan-suppressions] treewide: avoid leaking memory in users of CalorimeterHitDigi

veprbl:
Fixes a memory leak as seen in https://github.com/eic/EICrecon/actions/runs/5438209432/jobs/9889373595?pr=722

Clearing the array prevents memory from being released in
https://github.com/eic/EICrecon/blob/6f6abf218abb3ece03b60a4875ba5996777b2274/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L123-L124
This assumes that JANA doesn't release objects pointed to the `rawhits` array.

### DIFF
--- a/src/detectors/B0ECAL/RawCalorimeterHit_factory_B0ECalRawHits.h
+++ b/src/detectors/B0ECAL/RawCalorimeterHit_factory_B0ECalRawHits.h
@@ -88,7 +88,6 @@ public:
 
         // Hand owner of algorithm objects over to JANA
         Set(rawhits);
-        rawhits.clear(); // not really needed, but better to not leave dangling pointers around
     }
 
 };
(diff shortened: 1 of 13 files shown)


veprbl:
It seems like not clearing the array prevents the memory leak.

JANA2 should call delete for us
https://github.com/JeffersonLab/JANA2/blob/4cbf895c2757c5d39facb9d0c61f6a42e5d6d0b9/src/libraries/JANA/JFactoryT.h#L167C48-L169
unless `JFactory::NOT_OBJECT_OWNER` is set. @nathanwbrei Any idea what might be happening here?

wdconinc:
Ultimately this needs to get fixed by converting the algorithm and factory to use podio collections and letting the collections do the memory management. Having to do this ourselves, and making mistakes doing so, has always been the argument in favor of collections.

veprbl:
Yes, this was mostly intended as a test to push our understanding of JANA. This issue was not supposed to happen, as JANA is supposed to take ownership of pointers in the vector, which apparently doesn't happen.

I will submit a proper fix with use of the allocator from PODIO collections.

PR #734, opened 2023-07-04
feat!: enable dRICH PID in production

c-dilks:
Include dRICH output collections in default `eicrecon`.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no

yes, new output collections

### DIFF
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -51,6 +51,16 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "TOFBarrelRecHit",
             "TOFEndcapRecHits",
 
+            // DRICH
+            "DRICHRawHits",
+            "DRICHRawHitsAssociations",
+            "DRICHAerogelTracks",
+            "DRICHGasTracks",
+            "DRICHMergedTracks",
+            "DRICHAerogelIrtCherenkovParticleID",
+            "DRICHGasIrtCherenkovParticleID",
+            // "DRICHMergedIrtCherenkovParticleID",
+
             // MPGD
             "MPGDBarrelRecHits",
             "MPGDDIRCRecHits",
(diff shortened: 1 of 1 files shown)


c-dilks:
First enabling all collections, as a test. We will likely propose including only a subset of them, to avoid storing duplicate information.

wdconinc:
@chchatte92 @c-dilks This would enable the other dRICH collections by default as well. Rebased. I think this will be useful to merge too. Thoughts?

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -51,6 +51,16 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
...
 
+            // DRICH
+            "DRICHRawHits",
+            "DRICHRawHitsAssociations",
+            "DRICHAerogelTracks",
+            "DRICHGasTracks",
+            "DRICHMergedTracks",
+            "DRICHAerogelIrtCherenkovParticleID",
+            "DRICHGasIrtCherenkovParticleID",
+            // "DRICHMergedIrtCherenkovParticleID",
```
c-dilks, line 62: Let's drop the redundant (merged) ones.
```suggestion
            "DRICHAerogelIrtCherenkovParticleID",
            "DRICHGasIrtCherenkovParticleID",
```

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -75,6 +83,7 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "ReconstructedChargedParticleAssociations",
             "ReconstructedSeededChargedParticles",
             "ReconstructedSeededChargedParticleAssociations",
+            // "ReconstructedChargedParticleIDs",
```
c-dilks, line 86: Do we need to include these? The one-to-many relation from `ReconstructedChargedParticles` points to them already, but I can't remember if this collection needs to be explicitly included in the output.
```suggestion
            "ReconstructedChargedParticleIDs",
```

c-dilks:
> This would enable the other dRICH collections by default as well.

This will be very useful for high-stats studies from the September production. At minimum we need hits and track projections.

wdconinc:
Cool, can someone turn this from draft into full PR? I can approve and we can be done with this quickly.

wdconinc (dismissed):

wdconinc (approved):

PR #735, opened 2023-07-04
fix(RICH): minimize number of failed track propagation attempts

### LINKED ISSUE #564: `richgeo`: add ACTS surface downstream to help filter tracks which will not pass through the RICH

c-dilks:
Close #564

Adds a "filter surface" to test track propagation; if successful, allow propagation to a list of surfaces.

The filter surface for a RICH radiator is set to the xy-plane within the radiator with highest |z|, since if a track can be propagated to that plane, it likely can be propagated to most of the other radiator's planes.

This change minimizes the number of track propagation attempts and should minimize the number of errors seen in #564 down to 2 such errors per charged particle that misses the RICH (one for each RICH radiator; will be 4 such errors when we start using the pfRICH too).

- [x] Bug fix (issue #564)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
no


c-dilks:
Actually, now I see none of these errors, even before this PR.
```
ProjectTrack   ERROR     Propagation reached the step count limit of 1000 (did 1000 steps)
```

wdconinc:
> Actually, now I see none of these errors, even before this PR.

Could you confirm why you don't see the errors anymore, even without this PR, and close if the reason why you don't see the errors is sufficient resolution?

c-dilks:
> Could you confirm why you don't see the errors anymore, even without this PR, and close if the reason why you don't see the errors is sufficient resolution?

I confirm they don't appear even with the EICrecon `acts` service log level set to `trace` (which seems to set ACTS log level to max verbosity). 

I checked the [history of Propagator.ipp](https://github.com/acts-project/acts/commits/main/Core/include/Acts/Propagator/Propagator.ipp), where these messages are printed, and it's not clear to me if any recent changes helped us here.

I have not read enough ACTS code or documentation to understand, but do you know if there is now some recent feature which aborts a track propagation attempt if the trajectory is known not to pass through the target surface?

At any rate, for any trajectory which does not pass through the dRICH vessel, this PR will reduce the calls to the ACTS propagator from 35 down to a maximum of 2 (one for each radiator).... on 2nd thought let me bring that down to 1, by just checking the last plane in gas.

wdconinc:
The relevant history is https://github.com/acts-project/acts/commits/v21.1.1/Core/include/Acts/Propagator/Propagator.ipp because our ACTS version is pinned by the track reconstruction WG. Any recent changes would have to have been inside EICrecon or possibly the geometry (e.g. TOF conversion into ACTS surfaces may have improved with https://github.com/eic/epic/pull/466).

c-dilks:
> on 2nd thought let me bring that down to 1, by just checking the last plane in gas.

Never mind, propagation is done per radiator, so we'd have to be sure to check gas projections first, then skip aerogel projection if gas projection has zero points.

I don't think the negligible speed up from 2 propagation attempts to 1 is worth the additional complexity in `RichTrack_factory` and the consequent decrease in generality.

I'm already unhappy with the current level of generality here, since small issues in track projections can cause hidden problems in IRT PID, but at least this PR is an improvement. I also dislike having so much setup code in the factory, but I'm awaiting the full decoupling of `algorithms` to better address this concern.

wdconinc (commented):
I can't see any issue, but this isn't a part of the code base I am familiar with. I don't even know what's in most of the structures or what the memory implications are.

c-dilks:
> I can't see any issue, but this isn't a part of the code base I am familiar with. I don't even know what's in most of the structures or what the memory implications are.

@veprbl may have some familiarity with this, given https://github.com/eic/EICrecon/pull/726 review; would you mind reviewing this too?

I'd ask one of the original authors in the copyright notice, but I don't think any of them are familiar with this particular `TrackPropagation::propagateToSurfaceList` method.

wdconinc (approved):
I had a chance to dig into this more and it looks fine. Let's merge it.

veprbl (approved):
No change in the output when reconstructing rec_dis_18x275_minQ2=1000_brycecanyon.edm4eic.root

wdconinc:
The backport error is because of deleting the branch. I've seen it before. But we don't need the backport, as discussed, since this isn't active in v1.4.

PR #743, opened 2023-07-05
Fix a warning in JDD4hep_service.h

veprbl:
```
src/services/geometry/dd4hep/JDD4hep_service.h:40:13: warning: '~JDD4hep_service' overrides a destructor but is not marked 'override' [-Wsuggest-destructor-override]
    virtual ~JDD4hep_service();
            ^
prefix/include/JANA/Services/JServiceLocator.h:38:13: note: overridden virtual function is here
    virtual ~JService() = default;
            ^
```

### DIFF
--- a/src/services/geometry/dd4hep/JDD4hep_service.h
+++ b/src/services/geometry/dd4hep/JDD4hep_service.h
@@ -37,7 +37,7 @@ class JDD4hep_service : public JService
 {
 public:
     JDD4hep_service( JApplication *app ) : app(app) {}
-    virtual ~JDD4hep_service();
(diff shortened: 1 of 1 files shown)


wdconinc:
Mainly wondering if by overriding explicitly we will still call `~JService()`... This seems like saying: `~JDD4hep_service` overrides `~JService`, so don't call the virtual base class destructor, we'll take care of it in ` ~JDD4hep_service()`

veprbl:
I got this warning from clang. I now see that the keyword implies a confusing statement. I doubt it actually disables the destructor chaining, but we should be good without it.

PR #745, opened 2023-07-05
MPGD Hits for new Tracking Config

mposik1983:
Add MPGD Hits to new MPGD layers (outer barrel, forward and backward endcaps)

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: Adds hits for new tracking geometry

No.

No


wdconinc:
We should use RawHits instead of DigiHits. I hadn't noticed that tracking uses something against our naming conventions (which are based on the type name in EDM4hep and EDM4eic).

mposik1983:
> We should use RawHits instead of DigiHits. I hadn't noticed that tracking uses something against our naming conventions (which are based on the type name in EDM4hep and EDM4eic).

I see. I was just updating the existing code. I will change "Digi" to "Raw" then in all tracking cases

src/detectors/MPGD/MPGD.cc
```diff
@@ -23,30 +23,72 @@ void InitPlugin(JApplication *app) {
...
             "MPGDBarrelRecHits",    // Output data tag
              tracker_hit_reco_cfg));           // Hit reco default config for factories
 
     // Digitization
     SiliconTrackerDigiConfig dirc_digi_default_cfg;
-    barrel_digi_default_cfg.threshold = 0.25 * dd4hep::keV;
-    barrel_digi_default_cfg.timeResolution = 8;
-    app->Add(new JChainFactoryGeneratorT<SiliconTrackerDigi_factory>({"MPGDDIRCHits"}, "MPGDDIRCDigiHits", dirc_digi_default_cfg));
+    dirc_digi_default_cfg.threshold = 0.25 * dd4hep::keV;
+    dirc_digi_default_cfg.timeResolution = 10;
```
wdconinc, line 40: This stupid typo mistake could/should have been avoided with a structure like this:
```
    SiliconTrackerDigiConfig dirc_digi_default_cfg {
      .threshold = 0.25 * dd4hep::keV,
      .timeResolution = 10,
    };
```

wdconinc: (not your typo, but the pre-existing typo you fixed)

src/detectors/MPGD/MPGD.cc
```diff
@@ -23,30 +23,72 @@ void InitPlugin(JApplication *app) {
...
     app->Add(new JChainFactoryGeneratorT<TrackerHitReconstruction_factory>(
-            {"MPGDDIRCDigiHits"},     // Input data collection tags
+            {"MPGDDIRCRawHits"},     // Input data collection tags
             "MPGDDIRCRecHits",   // Output data tag
             dirc_hit_reco_cfg));          // Hit reco default config for factories
 
+    // Digitization
+    SiliconTrackerDigiConfig outerMPGDbarrel_digi_default_cfg;
+   outerMPGDbarrel_digi_default_cfg.threshold = 0.25 * dd4hep::keV;
+   outerMPGDbarrel_digi_default_cfg.timeResolution = 10;
```
wdconinc, line 54:
```suggestion
    outerMPGDbarrel_digi_default_cfg.threshold = 0.25 * dd4hep::keV;
    outerMPGDbarrel_digi_default_cfg.timeResolution = 10;
```

src/detectors/MPGD/MPGD.cc
wdconinc: Let's postpone this to a different PR.

wdconinc (dismissed):

src/detectors/MPGD/MPGD.cc
mposik1983: I do like that structure, it is much easier to read!

wdconinc: See https://github.com/eic/EICrecon/issues/746

Also, I think the pythonic trailing comma on the `timeResolution` line may not work (or maybe it does... I think this is the one place where C(++) may actually allow trailing commas.)

wdconinc (approved):

veprbl:
This broke the tracking (messages `CKFTracking    WARNING   No tracks found`)

PR #748, opened 2023-07-07
fix: missing comma in tracking.cc

wdconinc:
This fixes the tracking for me locally. Presumably this failed before (likely not silently but without an error return code because EICrecon depends on failing without error codes).

### DIFF
--- a/src/global/tracking/tracking.cc
+++ b/src/global/tracking/tracking.cc
@@ -40,7 +40,7 @@ void InitPlugin(JApplication *app) {
                          "TOFBarrelRecHit",             // TOF hits
                          "TOFEndcapRecHits",
 			 "MPGDBarrelRecHits",           // MPGD
-                         "MPGDDIRCRecHits"
+                         "MPGDDIRCRecHits",
                          "OuterMPGDBarrelRecHits",
                          "BackwardMPGDEndcapRecHits",
                          "ForwardMPGDEndcapRecHits",


wdconinc:
I don't think there was any error or warning in the log before about this (https://github.com/eic/EICrecon/actions/runs/5481863750/jobs/9986831374#step:5:383), and IT SHOULD NOT EVEN HAVE RUN! Sorry, I'm feeling frustrated about this.

c-dilks:
You're not alone, this exact sort of error frustrated me a couple months ago (when I was adding a new feature). I really don't like that this is valid C++:
```cpp
std::string a = "b" "c";  //  sets a = "bc"
```

mposik1983 (approved):

veprbl:
We could add `-Werror=string-concatenation`, at least to clang builds (https://clang.llvm.org/docs/DiagnosticsReference.html#id826)

c-dilks:
> We could add `-Werror=string-concatenation`, at least to clang builds (https://clang.llvm.org/docs/DiagnosticsReference.html#id826)

Yes! Then at least the CI build matrix will catch it (should enable `-Werror` on CI too)

veprbl:
TBH, I would enable it in CMake, it's not ambiguous and would save individuals some time during development.

veprbl:
I've tried
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0ae56fc2..da75e2dd 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,12 @@ if(NOT "${CMAKE_CXX_STANDARD}")
     set(CMAKE_CXX_STANDARD 17)
 endif()
 
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag(-Wstring-concatenation HAS_W_ERROR_STRING_CONCATENATION)
+if(HAS_W_ERROR_STRING_CONCATENATION)
+    add_compile_options(-Wstring-concatenation -Werror=string-concatenation)
+endif()
+
 # Export compile commands as json for run-clang-tidy
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
```
But it doesn't catch the error. Apparently it doesn't care about list initializers:
https://godbolt.org/z/W1a5zs4vd
https://stackoverflow.com/questions/76288726/c-c-warn-or-prohibit-literal-string-concatenation

veprbl:
It is, however, caught by clang-tidy:
```
src/global/tracking/tracking.cc:43:26: warning: suspicious string literal, probably missing a comma [bugprone-suspicious-missing-comma]
                         "MPGDDIRCRecHits"
                         ^
```
I see @wdconinc is already on it in the recent PRs.

Issue #738, opened 2023-07-04
Doesn't compile with DD4hep from master


veprbl:
```
In file included from src/services/geometry/richgeo/ReadoutGeo.cc:6:
src/services/geometry/richgeo/ReadoutGeo.h:35:10: error: no type named 'long64' in namespace 'dd4hep'
        dd4hep::long64 cellID_dd4hep;
        ~~~~~~~~^
```

wdconinc:
FYI https://github.com/AIDASoft/DD4hep/pull/1125 is in master, not in 1.25.1.

wdconinc:
I don't like all these interface-breaking changes in DD4hep either (and I assume the next version will be 1.26). But in this case at least an `auto` may work in both 1.25.1 and master.

c-dilks:
When do we plan to include https://github.com/AIDASoft/DD4hep/pull/1125 in our images?

`auto` won't help this `ReadoutGeo.h` case, but with 1125 we have [the new `CellID` type](https://github.com/MarkusFrankATcernch/DD4hep/blob/35dc496e2be17a55b3df7f1cc2ac6f821ca764fe/DDCore/include/DDSegmentation/BitFieldCoder.h#L26).

c-dilks:
> Can you use `dd4hep::rec::DDSegmentation::CellID`? https://github.com/AIDASoft/DD4hep/blob/master/DDRec/include/DDRec/CellIDPositionConverter.h#L29

Yeah, looks like it; blame indicates this line is 6 years old.

c-dilks:
> I'm hesitant to try to patch in a PR that touches so many source files since it inevitably requires a few other things to be patch in too. If we can avoid it, I'd prefer treating it without the patch.

@veprbl if `richgeo::ReadoutGeo` is not the only issue, can you post the other relevant errors?

veprbl:
> > I'm hesitant to try to patch in a PR that touches so many source files since it inevitably requires a few other things to be patch in too. If we can avoid it, I'd prefer treating it without the patch.
> 
> @veprbl if `richgeo::ReadoutGeo` is not the only issue, can you post the other relevant errors?

Yes, at least according to `make --keep-going`.

veprbl:
If common solution is not possible, we can always use `#if DD4HEP_VERSION_GE(1,26)`

c-dilks:
> Yes, at least according to `make --keep-going`.

Any PID or RICH specific ones?

[...]

PR #749, opened 2023-07-07
fix(richgeo): DD4hep CellID type

c-dilks:
fix https://github.com/eic/EICrecon/issues/738

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no

no

### DIFF
--- a/src/services/geometry/richgeo/RichGeo.h
+++ b/src/services/geometry/richgeo/RichGeo.h
@@ -11,7 +11,7 @@
 
 namespace richgeo {
 
-  typedef decltype(edm4hep::SimTrackerHitData::cellID) CellIDType;
+  using CellIDType = decltype(edm4hep::SimTrackerHitData::cellID);
 
   // sensors
   // -----------------------------------------------------------------------
--- a/src/services/geometry/richgeo/RichGeo.h
+++ b/src/services/geometry/richgeo/RichGeo.h
@@ -7,8 +7,11 @@
 #include <iostream>
 #include <fmt/format.h>
 #include <spdlog/spdlog.h>
+#include <edm4hep/SimTrackerHitData.h>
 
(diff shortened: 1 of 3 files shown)


c-dilks:
Confirmed EICrecon builds with no errors with DD4hep `master`; did not check if it runs.

src/services/geometry/richgeo/RichGeo.h
```diff
@@ -7,9 +7,12 @@
 #include <iostream>
 #include <fmt/format.h>
 #include <spdlog/spdlog.h>
+#include <edm4hep/SimTrackerHitData.h>
 
 namespace richgeo {
 
+  typedef decltype(edm4hep::SimTrackerHitData::cellID) CellIDType;
```
wdconinc, line 14: `using` is preferred in modern C++.
```suggestion
  using CellIDType = decltype(edm4hep::SimTrackerHitData::cellID);
```
Ref. https://www.nextptr.com/tutorial/ta1193988140/how-cplusplus-using-or-aliasdeclaration-is-better-than-typedef (just a first hit on google)

wdconinc: I also think `*_t` as a type alias is a more typical naming convention.

wdconinc: So maybe `cellID_t` or something?

c-dilks: It doesn't seem like we have any convention for alias names here in EICrecon.

wdconinc: Yeah, I don't feel strongly about this. Absent formatting guidance, I'll resolve this.

wdconinc (approved):

PR #753, opened 2023-07-07
ci: pre-commit clang-format (but not CI)

wdconinc:
Run clang-format on pre-commit. Hopefully leads to fewer missing spaces and incorrect indentation.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @veprbl FYI

No.

No.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,5 @@
+ci:
+    skip: [clang-format]
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.4.0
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1,3 +1,5 @@
+ci:
+    skip: [clang-format]
```
veprbl, line 2: What does this do?

veprbl (approved):

.pre-commit-config.yaml
wdconinc: Prevents pre-commit from doing its thing like here: https://github.com/eic/EICrecon/pull/848/commits/0914efe0268a6a0f04bd960c78f642a6fe49eb09. It makes sure that clang-format is only going in after review by a human and won't automatically mess up someone's carefully crafted alignment.

PR #755, opened 2023-07-08
ci: include-what-you-use, clang-include-fixer

wdconinc:
Include only what you use, CI check and fixes upload. See https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md for motivation (and caveats).

Also, `clang-include-fixer`, see https://clang.llvm.org/extra/clang-include-fixer.html.

TODO:
- [x] Requires support in nightly container, being added in https://eicweb.phy.anl.gov/containers/eic_container/-/merge_requests/641/

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


.github/workflows/linux-eic-shell.yml
```diff
@@ -119,6 +119,15 @@ jobs:
       with:
         name: clang-tidy-fixes.yml
         path: clang_tidy_fixes.yml
+    - uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        run: |
+          iwyu_tool -p build -o clang > iwyu_fixes.yml
```
wdconinc, line 126:
```suggestion
          iwyu_tool -p build $(git diff --name-only ${{github.event.pull_request.base.sha}}) -- -Xiwyu --no_comments -Xiwyu --cxx17ns > iwyu_fixes.log
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -119,6 +119,15 @@ jobs:
...
+    - uses: actions/upload-artifact@v3
+      with:
+        name: iwyu_fixes.yml
+        path: iwyu_fixes.yml
```
wdconinc, line 130:
```suggestion
        name: iwyu_fixes.log
        path: iwyu_fixes.log
```

wdconinc:
Out of the box, without any tweaking, this set of iwyu mappings now results in the changeset at https://github.com/eic/EICrecon/pull/994/. That doesn't compile (in some cases due to stuff that iwyu isn't going to catch, like implicit conversions) but it passes a minimum threshold of reasonableness (for me).

wdconinc:
@veprbl This PR for iwyu in CI is now ready for review.

This runs iwyu on changed files for pull requests, and full file set for pushes to main. It uploads a patch of files that need to be changed, and for pull requests it generates a job failure (but is not a required job. Automatically applying changes to the PR is not possible within this approach (nor, I think, desired because iwyu only generates changes that still need human review).

An example of diffs generated is at https://github.com/eic/EICrecon/actions/runs/6510717157/job/17686511123?pr=997#step:14:11 based on a change in #997.

veprbl (approved):
Diff LGTM

PR #751, opened 2023-07-07
Add reconstructed electron factory, algorithm utilizing E/p cut 

jdbrice:
### Briefly, what does this PR introduce?
Provides electron candidates as a first step towards a DIS electron finder.
This PR uses the recently provided MC particle to Reco cluster associations to produce a `ReconstructedElectrons` collection. Electrons candidates are identified with an E/p cut using Reco particle momentum and Redo cluster energy (only from ECAL detectors for now).

Couple of comments/potential improvements for reviewers:
- I am not sure how best to use the Config objects to allow parameters to be modified (e.g. min/max E/p in this case)
- AFAIK this creates a copy of the reconstructed particles - since this (ReconstructedElectrons) is a strict subset of the ReconstructedParticles collection, is there a way to store as such (without duplicating)
- I am currently pulling from (by name) several locations (see `reco.cc`) to get all possible source of `edm4eic::MCRecoClusterParticleAssociation` - is there a way to instead ask for all sources of a given type?

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added: 
  - Not sure if new tests needed, 
  - "ReconstructedElectrons" added to polio default output (include list) - simply running on a recent sim file should suffice as a test - expected behavior is to produce the "ReconstructedElectrpons" collection. A second level of test is to ensure that the collection is not empty (assuming at least some electrons within acceptance are in the simulated input)
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators: discussion in S&C meeting on June 5

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No

### Does this PR change default behavior?
No

PR #756, opened 2023-07-08
linux-eic-shell.yml: fix llvm-cov for PRs from fork

veprbl:
Fixes an issue reported in https://github.com/eic/EICrecon/pull/751#issuecomment-1627448876

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -155,16 +155,20 @@ jobs:
           COV_PERCENT=$(llvm-cov-15 export $LIB_PATH \
             -instr-profile=src/tests/algorithms_test/algorithms_test.profdata \
             "${COV_OPTIONS[@]}" | jq '.data | map(.totals.regions.percent) | .[]' | xargs printf "%.1f\n")
-          curl --request POST \
-            --url https://api.github.com/repos/${{ github.repository }}/statuses/${{ github.event.pull_request.head.sha || github.sha }} \
-            --header "Accept: application/vnd.github+json" \
-            --header "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
Looks good. Possibly also an opportunity to highlight as job summary instead of simple echo to log, https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/.

veprbl:
This is an interesting feature. Would be nice to be able to convert html reports to markdown.

wdconinc:
Isn't html a subset of gfmd?

veprbl:
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN"
        "http://www.w3.org/TR/html4/strict.dtd">
<HTML>
  <HEAD>
    <TITLE>The document title</TITLE>
  </HEAD>
  <BODY>
    <H1>Main heading</H1>
    <P>A paragraph.</P>
    <P>Another paragraph.</P>
    <UL>
      <LI>A list item.</LI>
      <LI>Another list item.</LI>
    </UL>
  </BODY>
</HTML>

wdconinc:
https://github.github.com/gfm/#raw-html

veprbl:
Close enough. Though, the links will be dead, since there is no place to host other pages.

wdconinc:
Oh what I would give for gh-pages deployments from pull requests...

veprbl:
You can commit to a repo, so something is possible, but it's not as straightforward as publishing artifact.

wdconinc:
The most promising approach I see currently is to setup a staging environment in a namespace separate from the public deployment environment, where we can then publish under e.g. `pull/id/` or so. Ref https://github.com/actions/deploy-pages

Having this work would make pr reviews of all these gh-pages repos so much easier too.

wdconinc:
We've gotten off topic on this PR though ;-)

PR #741, opened 2023-07-05
algorithms/calorimetry: pass PODIO collections instead of std::vector

veprbl:
This fixes memory leaks like:

```
Direct leak of 407056 byte(s) in 50882 object(s) allocated from:
    #0 0x55e48bee007d in operator new(unsigned long) (/home/runner/work/EICrecon/EICrecon/bin/eicrecon+0xeb07d) (BuildId: 5069f7d2185cef71541ac1e93e2f6643618de2c4)
    #1 0x7fa316881fc5 in CalorimeterHitDigi::signal_sum_digi() /home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterHitDigi.cc:254:23
    #2 0x7fa3168807eb in CalorimeterHitDigi::AlgorithmProcess() /home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterHitDigi.cc:127:9
    #3 0x7fa3150b2e0c in RawCalorimeterHit_factory_LFHCALRawHits::Process(std::shared_ptr<JEvent const> const&) /home/runner/work/EICrecon/EICrecon/src/detectors/FHCAL/RawCalorimeterHit_factory_LFHCALRawHits.h:86:9
    #4 0x7fa315a0fcc1 in eicrecon::JFactoryPodioT<edm4hep::RawCalorimeterHit>::Create(std::shared_ptr<JEvent const> const&) /home/runner/work/EICrecon/EICrecon/src/services/io/podio/JFactoryPodioT.h:171:13
    #5 0x7fa315a4e8a8 in JFactoryT<edm4hep::RawCalorimeterHit>::GetOrCreate(std::shared_ptr<JEvent const> const&) /opt/software/linux-debian12-x86_64_v2/gcc-12.2.0/jana2-2.1.0-hj25lfdff2mcbllcvr5lojwa4sqpnpb5/include/JANA/JFactoryT.h:83:13
    #6 0x7fa315a4d87b in std::vector<edm4hep::RawCalorimeterHit const*, std::allocator<edm4hep::RawCalorimeterHit const*> > JEvent::Get<edm4hep::RawCalorimeterHit>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const /opt/software/linux-debian12-x86_64_v2/gcc-12.2.0/jana2-2.1.0-hj25lfdff2mcbllcvr5lojwa4sqpnpb5/include/JANA/JEvent.h:325:27
```

[...]

PR #762, opened 2023-07-13
fix: clang-tidy readability-qualified-auto

wdconinc:
`auto` type deduction can be used to deduce `const` and pointer nature of the type, but that obscures this nature in the code, making it less readable.

This modifies incorrect looking statements like
```cpp
    for( auto hit : ForwardRomanPotRecHits() ){
        hist1D["ForwardRomanPotRecHits_time"]->Fill( hit->getTime());
    }
```
with more readable statements like
```cpp
    for( const auto *hit : ForwardRomanPotRecHits() ){
        hist1D["ForwardRomanPotRecHits_time"]->Fill( hit->getTime());
    }
```
where `hit` is now manifestly a pointer.

This does not introduce any `auto` where none was present, and only adds `const`, `*`, and `&` where appropriate. Also, it doesn't modify `const auto hit` into e.g. `const auto *const hit`.

See https://clang.llvm.org/extra/clang-tidy/checks/readability/qualified-auto.html for details.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: code style changes only

No.

No.


veprbl:
I think, `auto` is commonly overused. Explicit types are great for readability. Now, another argument for `auto`, is that it allows to modify types without touching any code. By adding `const` and `*` that convenience is somewhat diminished. Not sure what happens to `&` and how `std::unique_ptr` fits in, but okay. I would really like to have the `auto` keyword banned except for the cases when the variable type is already spelled on the same line of code.

src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -136,7 +137,7 @@ void CalorimeterHitDigi::AlgorithmProcess()  {
 void CalorimeterHitDigi::single_hits_digi(){
 
      // Create output collections
-    for ( auto ahit : simhits ) {
+    for ( const auto *ahit : simhits ) {
```
veprbl, line 140: I expect a conflict with #741. Collections return objects (with pointers inside), not references or pointers.

wdconinc:
My approach on the matter of anything that `clang-tidy` suggests is that it makes more sense to defer to the collective compiler developer community who comes up with these checks, than to open up the can of worms of selective adherence based on the individual opinion of some physicists (my opinions included). This approach also extends to `clang-format`: more important to follow something standard than everyone using it as an opportunity to air their opinions about various parts of the conventions... That's not to say I agree or disagree with your statement; I'm saying that whether I agree or disagree is not really important.

> By adding `const` and `*` that convenience is somewhat diminished.

Maybe, maybe not. `auto` is rather forgiving, at least about `const` modifiers. Look at https://github.com/eic/EICrecon/blob/main/src/services/geometry/richgeo/IrtGeo.cc#L78, for example, where the `const auto` really means `CherenkovRadiator *const`. When `clang-tidy` turns this into `auto *const`, I'm more satisfied with what that implies (see https://github.com/eic/EICrecon/pull/762/files#diff-2b184d8a273a37ab7c03736321faf5b1cfe412ce4e4c048638c3aa7dd0b257a8).

veprbl (approved):

wdconinc:
Passes all checks without diffs now.

veprbl (approved):

PR #766, opened 2023-07-14
fix: rewrite TrackParamTruthInit to create edm4eic::TrackParameters

wdconinc:
This reworks `TrackParamTruthInit` to create `edm4eic::TrackParametersCollection`s. That then means that JANA2 doesn't need to clean up the `SinglyChargedBoundParameters` it seems to be having trouble with. It also means we can actually store the truth track seeds in the edm4eic output trees.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @veprbl @bschmookler 

No.

Yes, stores truth track seeds in output ROOT tree.


wdconinc:
Reverted the gdb prefix that was used for debuggin the issue here, since it prevents the ASAN operation. Prefixed line was
```console
gdb -return-child-result -ex run -ex "thread apply all bt" -ex "quit" --args \
```

veprbl:
I think we should also drop the `src/algorithms/tracking/JugTrack/Track.hpp` and be explicit about using the ACTS types.

veprbl:
> I think we should also drop the `src/algorithms/tracking/JugTrack/Track.hpp` and be explicit about using the ACTS types.

Actually bad idea, we should stick to what is done in `ActsExamples`

veprbl (approved):
Everything looks fine to me and it produces identical output to main.

wdconinc:
> Everything looks fine to me and it produces identical output to main.

I was still having some issues confirming this on my local system, but I think I may have some other stuff going on (maybe still some local files that didn't get recompiled, or I'm doing something stupid: the differences I was seeing were much worse than when this was still on top of `ckf_realseed` so it doesn't make sense). If you compared the outputs before and after, then I trust this is fine and we should merge.

veprbl:
I've revisited the check. Turns out, there was a bug in the script that I've used to test this. There are some differences in CI outputs for DIS. I've tested locally with single particle simulations and the files are identical except for some collection indices. Still need have to look into DIS event issue to be sure.

veprbl:
I've followed up on this in #779.

PR #767, opened 2023-07-14
Upgrade Jet Factories to Multifactories

ruse-traveler:

This PR upgrades the existing jet factories to multifactories and also modifies the jet reconstruction algorithm such that it returns a PODIO collection of reconstructed particles rather than a standard vector.

- [ ] Bug fix (issue #__)
- [ X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/algorithms/reco/JetReconstruction.cc
```diff
@@ -23,31 +23,32 @@ namespace eicrecon {
     m_log->trace("Initialized");
   }
 
-  std::vector<edm4eic::ReconstructedParticle*> JetReconstruction::execute(
+  edm4eic::ReconstructedParticleCollection* JetReconstruction::execute(
```
veprbl, line 26:
```suggestion
  std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::execute(
```

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -23,31 +23,32 @@ namespace eicrecon {
...
     m_log->trace("  Number of particles: {}", momenta.size());
 
     // Particles for jet reconstrution
     std::vector<PseudoJet> particles;
     for (const auto &mom : momenta) {
+
       // Only cluster particles within the given pt Range
       if ((mom->pt() > m_minCstPt) && (mom->pt() < m_maxCstPt)) {
-        particles.emplace_back(mom->px(), mom->py(), mom->pz(), mom->e() );
+        particles.push_back( PseudoJet(mom->px(), mom->py(), mom->pz(), mom->e()) );
```
wdconinc, line 46: You are reverting changes in the target branch.

wdconinc (changes requested):

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -57,41 +58,48 @@ namespace eicrecon {
...
-        cst_edm.setEnergy(csts[j].e());
-        cst_edm.setMass(csts[j].m());
-        //jet_edm.addToParticles(cst_edm);  // FIXME: global issue with podio reference
+        edm4eic::MutableReconstructedParticle cst_output;
+        cst_output.setType(1);
+        cst_output.setPDG(i);
+        cst_output.setMomentum(edm4hep::Vector3f(csts[j].px(), csts[j].py(), csts[j].pz()));
+        cst_output.setEnergy(csts[j].e());
+        cst_output.setMass(csts[j].m());
+        //jet_output.addToParticles(cst_output);  // FIXME: global issue with podio reference
```
veprbl, line 90: This one would need to be routed properly, something along the lines of
```c++
add_to_particle(jet_output, input_collection[csts[j].user_index()])
```
where
```c++
template <typename T>
add_to_particle(edm4eic::MutableReconstructedParticle &p, T &other);

template <>
add_to_particle(edm4eic::MutableReconstructedParticle &p, edm4eic::ReconstructedParticle &other) {
  p.addToParticles(other);
}

template <>
add_to_particle(edm4eic::MutableReconstructedParticle &p, edm4eic::MCParticle &other) {
  p.addToMCParticles(other);
}
```

src/algorithms/reco/JetReconstruction.h
```diff
@@ -6,11 +6,13 @@
 
 // standard c includes
 #include <vector>
-// DD4hep includes
+// jana includes
```
veprbl, line 9:
```suggestion
```

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -23,31 +23,32 @@ namespace eicrecon {
     m_log->trace("Initialized");
   }
 
-  std::vector<edm4eic::ReconstructedParticle*> JetReconstruction::execute(
+  edm4eic::ReconstructedParticleCollection* JetReconstruction::execute(
     const std::vector<const edm4hep::LorentzVectorE*> momenta) {
```
veprbl, line 27: Would need to use templated type here
```suggestion
    const T &input_collection) {
```

veprbl (commented):
Looks good. Mostly rambling about unusual formatting.
Please use smart pointers. We also would like to get proper associations, I'm putting down some ideas on how to integrate that.

src/global/reco/reco.cc
```diff
@@ -37,7 +36,6 @@ void InitPlugin(JApplication *app) {
     app->Add(new JChainMultifactoryGeneratorT<MatchClusters_factory>(
         "ReconstructedParticlesWithAssoc",
         { "EcalEndcapNClusters",
-          "EcalBarrelScFiClusters",
```
veprbl, line 40: This doesn't seem right...

src/global/reco/reco.cc
```diff
@@ -65,22 +63,17 @@ void InitPlugin(JApplication *app) {
...
-        "EcalEndcapNClusterAssociations",
-        "EcalEndcapPClusterAssociations",
-        "EcalEndcapPInsertClusterAssociations",
-        "EcalLumiSpecClusterAssociations",
-        },
-        "ReconstructedElectrons"
+    app->Add(new JChainMultifactoryGeneratorT<GeneratedJets_factory>(
+             "GeneratedJets",
+            {"MCParticles"},
+            {"GeneratedJets"},
```
veprbl, line 69: We don't need to change these factories to mutifactories at all. They need to output jets (currently modeled as `edm4hep::ReconstructedParticles`, we then refer to constituents that are already written in the MCParticles and ReconstructedParticles collections.

src/global/reco/reco.cc
```diff
@@ -65,22 +63,17 @@ void InitPlugin(JApplication *app) {
...
+            app
+    ));
+    app->Add(new JChainMultifactoryGeneratorT<ReconstructedJets_factory>(
+            "ReconstrucctedJets",
+            {"ReconstructedParticles"},
```
bspage912, line 74: Change input collection to ReconstructedChargedParticles?

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -23,31 +23,32 @@ namespace eicrecon {
...
     }
 
     m_log->trace("  Number of particles: {}", momenta.size());
 
     // Particles for jet reconstrution
     std::vector<PseudoJet> particles;
     for (const auto &mom : momenta) {
+
       // Only cluster particles within the given pt Range
       if ((mom->pt() > m_minCstPt) && (mom->pt() < m_maxCstPt)) {
```
bspage912, line 45: May be for another PR to address, but logically, I'm not sure we want to impose the particle / track pT cuts here as we likely want them to be different depending on the input. We usually place some threshold on tracks (and calorimeter clusters) based on the tracker performance, but place a much lower pT threshold on stable MC particles.

bspage912 (commented):
I have no comment on the code changes themselves as I am still trying to understand all this myself :-). 

I added comments on using ReconstructedChargedParticles vs ReconstructedParticles as the input in the ReconstructedJets_factory, on possibly restricting the input particle species in GeneratedJets_factory, and on where we logically want to place input track/particle pT thresholds.

src/algorithms/reco/JetReconstruction.cc
ruse-traveler: Whoops! Thank you for catching that: I deleted the wrong line while cleaning up.

src/algorithms/reco/JetReconstruction.cc
ruse-traveler: Ah! Thank you for catching that!

src/global/reco/reco.cc
ruse-traveler: 😩 Aggh. I deleted that by accident. Thank you for catching that!

src/global/reco/reco.cc
ruse-traveler: Out of curiosity: if we wanted to have multiple jet collections (e.g. a collection of track-only jets and then a collection of particle flow jets), would it be ideal to use multifactories?

src/global/reco/ReconstructedJets_factory.cc
```diff
@@ -1,49 +1,74 @@
...
         // Nothing to do here
-    }
+        return;
+
+    }  // end 'BeginRun(std::shared_ptr<JEvent&>)'
+
+
 
     void ReconstructedJets_factory::Process(const std::shared_ptr<const JEvent> &event) {
-        auto rc_particles = event->Get<edm4eic::ReconstructedParticle>("ReconstructedParticles");
```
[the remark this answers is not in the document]
ruse-traveler: Whoops! Good catch: that should be ReconstructedChargedParticles. But that's correct! For multifactories, the input and output tags are specified in the reco.cc file.

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -1,52 +1,77 @@
...
+          std::vector<const edm4hep::LorentzVectorE*> momenta;
+          for (const auto& particle : input) {
+
+            // select only final state particles
+            const bool is_final_state = (particle -> getGeneratorStatus() == 1);
+            if (!is_final_state) continue;
+
+            const auto& momentum = particle -> getMomentum();
+            const auto& energy   = particle -> getEnergy();
+            momenta.push_back(new edm4hep::LorentzVectorE(momentum.x, momentum.y, momentum.z, energy));
```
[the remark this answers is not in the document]
ruse-traveler: That's a good point... And I would say we should. But what if we add an option to the algorithm to pick either charged or all generated particles?

src/algorithms/reco/JetReconstruction.cc
ruse-traveler: Definitely! I was thinking on addressing this in another PR, but in making the options for both factories configurable, we'll also be able to set the kinematic thresholds differently between the reconstructed and generated jets.

src/global/reco/reco.cc
ruse-traveler: Ahhhh I see. I guess for my edification: are multifactories exclusively for collections that require multiple inputs (e.g. reconstructed particles)?

veprbl: Multifactories are needed for factories that want to produce multiple outputs.

Just to clarify (cc @wdconinc), we are likely to convert all factories (even those producing a single output) to multfactories eventually. Presumably that would not impact the performance, but a consistent interface would simplify implementation of Gaudi-style steering via configuration in Python. Latter is the current informal consensus regarding the direction to go forward. Having said that, producing multiple kinds of jets from the same factories doesn't seem like a sound choice: there is no shared computation between the two procedures, but the shared state opens them up to introducing bugs. We would be much better off making algorithms sufficiently configurable and running those from separately defined factories. In that sense, you might be doing the right thing for the wrong reason.

src/global/reco/reco.cc
```diff
@@ -65,22 +63,17 @@ void InitPlugin(JApplication *app) {
...
-        },
-        "ReconstructedElectrons"
+    app->Add(new JChainMultifactoryGeneratorT<GeneratedJets_factory>(
+             "GeneratedJets",
+            {"MCParticles"},
+            {"GeneratedJets"},
+            app
+    ));
+    app->Add(new JChainMultifactoryGeneratorT<ReconstructedJets_factory>(
+            "ReconstrucctedJets",
```
veprbl, line 73:
```suggestion
            "ReconstructedJets",
```

src/global/reco/ReconstructedJets_factory.h
```diff
@@ -3,37 +3,49 @@
 
 #pragma once
 
-#include <edm4eic/ReconstructedParticleCollection.h>
+// jana includes
```
veprbl, line 6:
```suggestion
```

src/global/reco/GeneratedJets_factory.cc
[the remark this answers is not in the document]
ruse-traveler: Forgot to comment here: but to implement this, I added an additional condition to the generated jets factory to pick out only charged particles!

ruse-traveler:
Hi all, just wanted to touch base on this PR to make sure it gets in before the September campaign. If everyone's okay with implementing PODIO references in a separate PR, then I think I've addressed all comments so far. (If no one objects, then I'll mark all the above conversations as resolved.)

Let me know if I missed anything or if there's anything else I need to do to make sure this gets merged!

wdconinc:
> Let me know if I missed anything or if there's anything else I need to do to make sure this gets merged!

There are conflicts...

ruse-traveler:
> > Let me know if I missed anything or if there's anything else I need to do to make sure this gets merged!
> 
> There are conflicts...

Whoops! I should've specified: aside from the conflicts! 😓 

(I'll take care of those, though!)

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -1,52 +1,72 @@
...
-        std::vector<const edm4hep::LorentzVectorE*> momenta;
-        for (const auto& p : mc_particles) {
-          if (p -> getGeneratorStatus() == 1) {
-            const auto& mom    = p -> getMomentum();
-            const auto& energy = p -> getEnergy();
-            momenta.push_back(new edm4hep::LorentzVectorE(mom.x, mom.y, mom.z, energy));
+
+        // loop over input tags
+        size_t iInput = 0;
+        for (const std::string& input_tag : GetInputTags()) {
```
veprbl, line 41: Why are you iterating over input collections? This can't work because down below you overwrite previous result with the latest collection.

I would stick to just one input particle collection `GetInputTags()[0]` and leave the other inputs reserved for possible future use. If we need to merge multiple particle collections, let's make that code into a separate factory.

ruse-traveler: Ah! This is a holdover from my misunderstanding of what a multifactory was for: my thinking was that I would loop over specified input collections and produce an output for each one. I didn't remove it in the last batch of commits just because I thought it wouldn't make a difference.

But I agree with your suggestion: I'll stick to `GetTags()[0]` and reserve the others for future use!

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -23,26 +23,27 @@ namespace eicrecon {
...
-        particles.emplace_back(mom->px(), mom->py(), mom->pz(), mom->e() );
+        particles.emplace_back( PseudoJet(mom->px(), mom->py(), mom->pz(), mom->e()) );
```
veprbl, line 46:
```suggestion
        particles.emplace_back(mom->px(), mom->py(), mom->pz(), mom->e());
```

src/global/reco/reco.cc
```diff
@@ -34,7 +33,7 @@ void InitPlugin(JApplication *app) {
     app->Add(new JChainMultifactoryGeneratorT<MatchClusters_factory>(
         "ReconstructedParticlesWithAssoc",
         { "EcalEndcapNClusters",
-          "EcalBarrelScFiClusters",
+          "EcalBarrelSciFiClusters",
```
veprbl, line 36:
```suggestion
          "EcalBarrelScFiClusters",
```

src/global/reco/reco.cc
```diff
@@ -62,22 +61,17 @@ void InitPlugin(JApplication *app) {
...
-    app->Add(new JChainFactoryGeneratorT<ReconstructedElectrons_factory>(
-        {"MCParticles", "ReconstructedChargedParticles", "ReconstructedChargedParticleAssociations",
-        "EcalBarrelScFiClusterAssociations",
-        "EcalEndcapNClusterAssociations",
-        "EcalEndcapPClusterAssociations",
-        "EcalEndcapPInsertClusterAssociations",
-        "EcalLumiSpecClusterAssociations",
-        },
-        "ReconstructedElectrons"
+    app->Add(new JChainMultifactoryGeneratorT<GeneratedJets_factory>(
```
veprbl, line 64: Submit a separate PR for this
```suggestion
    app->Add(new JChainFactoryGeneratorT<ReconstructedElectrons_factory>(
         {"MCParticles", "ReconstructedChargedParticles", "ReconstructedChargedParticleAssociations",
         "EcalBarrelScFiClusterAssociations",
         "EcalEndcapNClusterAssociations",
         "EcalEndcapPClusterAssociations",
         "EcalEndcapPInsertClusterAssociations",
         "EcalLumiSpecClusterAssociations",
         },
         "ReconstructedElectrons"

    app->Add(new JChainMultifactoryGeneratorT<GeneratedJets_factory>(
```

src/global/reco/reco.cc
```diff
@@ -19,7 +19,6 @@
 #include "InclusiveKinematicsSigma_factory.h"
 #include "GeneratedJets_factory.h"
 #include "ReconstructedJets_factory.h"
-#include "ReconstructedElectrons_factory.h"
```
veprbl, line 22:
```suggestion
#include "ReconstructedElectrons_factory.h"

```

src/global/reco/reco.cc
```diff
@@ -62,22 +62,27 @@ void InitPlugin(JApplication *app) {
...
-        "ReconstructedElectrons"
+         {"MCParticles", "ReconstructedChargedParticles", "ReconstructedChargedParticleAssociations",
+         "EcalBarrelScFiClusterAssociations",
+         "EcalEndcapNClusterAssociations",
+         "EcalEndcapPClusterAssociations",
+         "EcalEndcapPInsertClusterAssociations",
+         "EcalLumiSpecClusterAssociations",
+         },
+         "ReconstructedElectrons"
+
```
veprbl, line 74:
```suggestion
    ));

```

src/global/reco/GeneratedJets_factory.h
```diff
@@ -3,37 +3,47 @@
...
             public SpdlogMixin {
 
     public:
-        explicit GeneratedJets_factory(std::vector<std::string> default_input_tags):
-            JChainFactoryT<edm4eic::ReconstructedParticle>(std::move(default_input_tags)) {
-        }
+
+        // ctor
+        explicit GeneratedJets_factory(std::string tag,
+                                       const std::vector<std::string>& input_tags,
```
veprbl, line 26:
```suggestion
        explicit GeneratedJets_factory(const std::vector<std::string>& input_tags,
```

src/global/reco/ReconstructedJets_factory.h
```diff
@@ -3,37 +3,47 @@
...
             public SpdlogMixin {
 
     public:
-        explicit ReconstructedJets_factory(std::vector<std::string> default_input_tags):
-            JChainFactoryT<edm4eic::ReconstructedParticle>(std::move(default_input_tags)) {
-        }
+
+        // ctor
+        explicit ReconstructedJets_factory(std::string tag,
+                                           const std::vector<std::string>& input_tags,
```
veprbl, line 26:
```suggestion
        explicit ReconstructedJets_factory(const std::vector<std::string>& input_tags,
```

src/global/reco/reco.cc
```diff
@@ -62,22 +62,28 @@ void InitPlugin(JApplication *app) {
...
+         "EcalEndcapNClusterAssociations",
+         "EcalEndcapPClusterAssociations",
+         "EcalEndcapPInsertClusterAssociations",
+         "EcalLumiSpecClusterAssociations",
+         },
+         "ReconstructedElectrons"
     ));
 
+    app->Add(new JChainMultifactoryGeneratorT<GeneratedJets_factory>(
+             "GeneratedJets",
```
veprbl, line 77:
```suggestion
```

src/global/reco/reco.cc
```diff
@@ -62,22 +62,28 @@ void InitPlugin(JApplication *app) {
...
+            {"MCParticles"},
+            {"GeneratedJets"},
+            app
+    ));
+    app->Add(new JChainMultifactoryGeneratorT<ReconstructedJets_factory>(
+            "ReconstructedJets",
```
veprbl, line 83:
```suggestion
```

src/global/reco/ReconstructedJets_factory.cc
```diff
@@ -1,49 +1,63 @@
...
 
+        // grab input collection
+        auto input = event->Get<edm4eic::ReconstructedParticle>(GetInputTags()[0]);
```
veprbl, line 40: Avoid conversion to `std::vector`
```suggestion
        auto input = static_cast<const edm4eic::ReconstructedParticleCollection*>(event->GetCollectionBase(GetInputTags()[0]));
```

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -1,52 +1,67 @@
...
+
+    }  // end 'BeginRun(std::shared_ptr<JEvent&>)'
+
+
 
     void GeneratedJets_factory::Process(const std::shared_ptr<const JEvent> &event) {
-        auto mc_particles = event->Get<edm4hep::MCParticle>("MCParticles");
 
+        // grab input collection
+        auto input = event->Get<edm4hep::MCParticle>(GetInputTags()[0]);
```
veprbl, line 40: Avoid conversion to `std::vector`
```suggestion
        auto input = static_cast<const edm4hep::MCParticleCollection*>(event->GetCollectionBase(GetInputTags()[0]));
```

src/global/reco/GeneratedJets_factory.h
```diff
@@ -3,37 +3,47 @@
...
+        void BeginRun(const std::shared_ptr<const JEvent> &event) override;
 
         /** Event by event processing **/
         void Process(const std::shared_ptr<const JEvent> &event) override;
 
     protected:
-        JetReconstruction m_jet_algo;
 
-    };
+        std::vector<std::string> m_input_tags;
```
veprbl, line 44: Avoid adding dead code
```suggestion
```

src/global/reco/ReconstructedJets_factory.h
```diff
@@ -3,37 +3,47 @@
...
+        void BeginRun(const std::shared_ptr<const JEvent> &event) override;
 
         /** Event by event processing **/
         void Process(const std::shared_ptr<const JEvent> &event) override;
 
     protected:
-        JetReconstruction m_jet_algo;
 
-    };
+        std::vector<std::string> m_input_tags;
```
veprbl, line 44:
```suggestion
```

src/global/reco/reco.cc
```diff
@@ -62,22 +62,28 @@ void InitPlugin(JApplication *app) {
...
-        },
-        "ReconstructedElectrons"
+         {"MCParticles", "ReconstructedChargedParticles", "ReconstructedChargedParticleAssociations",
+         "EcalBarrelScFiClusterAssociations",
+         "EcalEndcapNClusterAssociations",
+         "EcalEndcapPClusterAssociations",
+         "EcalEndcapPInsertClusterAssociations",
+         "EcalLumiSpecClusterAssociations",
+         },
+         "ReconstructedElectrons"
```
veprbl, line 73:
```suggestion
        {"MCParticles", "ReconstructedChargedParticles", "ReconstructedChargedParticleAssociations",
        "EcalBarrelScFiClusterAssociations",
        "EcalEndcapNClusterAssociations",
        "EcalEndcapPClusterAssociations",
        "EcalEndcapPInsertClusterAssociations",
        "EcalLumiSpecClusterAssociations",
        },
        "ReconstructedElectrons"
```

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -1,52 +1,67 @@
...
-        }
+        for (const auto& particle : input) {
 
-        auto jets = m_jet_algo.execute(momenta);
-        for (const auto &mom : momenta) {
-          delete mom;
+            // select only final state charged particles
+            const bool is_final_state = (particle->getGeneratorStatus() == 1);
+            const bool is_charged = (particle->getCharge() != 0.);
+            if (!is_final_state || !is_charged) continue;
```
veprbl, line 49: Doesn't fit the description - please submit separately
```suggestion
            if (!is_final_state) continue;
```

src/global/reco/reco.cc
```diff
@@ -62,22 +62,28 @@ void InitPlugin(JApplication *app) {
...
 
+    app->Add(new JChainMultifactoryGeneratorT<GeneratedJets_factory>(
+             "GeneratedJets",
+            {"MCParticles"},
+            {"GeneratedJets"},
+            app
+    ));
+    app->Add(new JChainMultifactoryGeneratorT<ReconstructedJets_factory>(
+            "ReconstructedJets",
+            {"ReconstructedChargedParticles"},
```
veprbl, line 84: Please submit separately
```suggestion
            {"ReconstructedParticles"},
```

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -1,52 +1,66 @@
...
-
+// standard c includes
 #include <memory>
-
 #include <JANA/JEvent.h>
-#include "services/log/Log_service.h"
-
+#include <services/log/Log_service.h>
+// event data model definitions
 #include <edm4hep/MCParticle.h>
```
veprbl, line 9:
```suggestion
#include <edm4hep/MCParticleCollection.h>
```

src/global/reco/ReconstructedJets_factory.cc
```diff
@@ -1,49 +1,63 @@
...
-#include "ReconstructedJets_factory.h"
-
+// standard c includes
 #include <memory>
-
 #include <JANA/JEvent.h>
-#include "services/log/Log_service.h"
+#include <services/log/Log_service.h>
+// event data model definitions
+#include <edm4eic/ReconstructedParticle.h>
```
veprbl, line 9:
```suggestion
#include <edm4eic/ReconstructedParticleCollection.h>
```

src/global/reco/ReconstructedJets_factory.cc
```diff
@@ -1,49 +1,63 @@
...
-        for (const auto& p : rc_particles) {
-          // TODO: Need to exclude the scattered electron
-          const auto& mom    = p -> getMomentum();
-          const auto& energy = p -> getEnergy();
-          momenta.push_back(new edm4hep::LorentzVectorE(mom.x, mom.y, mom.z, energy));
+        for (const auto& particle : input) {
+
+            // TODO: Need to exclude the scattered electron
+            const auto& momentum = particle->getMomentum();
+            const auto& energy = particle->getEnergy();
```
veprbl, line 48:
```suggestion
            const auto& momentum = particle.getMomentum();
            const auto& energy = particle.getEnergy();
```

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -1,52 +1,66 @@
...
 
-        auto jets = m_jet_algo.execute(momenta);
-        for (const auto &mom : momenta) {
-          delete mom;
+            // select only final state charged particles
+            const bool is_final_state = (particle->getGeneratorStatus() == 1);
+            if (!is_final_state) continue;
+
+            const auto& momentum = particle->getMomentum();
+            const auto& energy = particle->getEnergy();
```
veprbl, line 51:
```suggestion
            const auto& momentum = particle.getMomentum();
            const auto& energy = particle.getEnergy();
```

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -1,52 +1,66 @@
...
-            momenta.push_back(new edm4hep::LorentzVectorE(mom.x, mom.y, mom.z, energy));
-          }
-        }
+        for (const auto& particle : input) {
 
-        auto jets = m_jet_algo.execute(momenta);
-        for (const auto &mom : momenta) {
-          delete mom;
+            // select only final state charged particles
+            const bool is_final_state = (particle->getGeneratorStatus() == 1);
```
veprbl, line 47:
```suggestion
            const bool is_final_state = (particle.getGeneratorStatus() == 1);
```

veprbl:
I took liberty to finish the transitions from vectors to collections, fixed some issues introduced while rebasing and reverted charged particle cut.

veprbl (approved):
Checked outputs before/after. This PR introduces only minimal changes: jets are placed into collections before the constituents.

ruse-traveler:
@veprbl Thank you for taking care of those! Out of curiosity: is there a reason to revert the charged particle cut?

Either way, it's something we can discuss more when we revisit the constituent PODIO relations (i.e. for the next campaign). So this all looks good to me, and I'm okay with merging if everyone else is!

veprbl:
Bundling changes places unnecessary burden on reviewers. Switch to multifactories should not cause any changes in the output, and it should be easy to demonstrate that. Thus, changes to the cuts don't belong in this change.

veprbl:
@ruse-traveler Can you resubmit charge cut for review, or do you want me to do it? Please, create new branch in this repo this time. I also suggest renaming the collection to {Reconstructed,Generated}ChargedJets to communicate the difference in the result.

ruse-traveler:
> @ruse-traveler Can you resubmit charge cut for review, or do you want me to do it? Please, create new branch in this repo this time. I also suggest renaming the collection to {Reconstructed,Generated}ChargedJets to communicate the difference in the result.

Can do! I'll put together a PR for that using a branch from this repo. And makes sense (re: reverting the cut)! I'll make sure to not sneak in extra changes in future PRs...

PR #769, opened 2023-07-15
linux-eic-shell.yml: fix the apt-get failures

veprbl:
Fix for failures like https://github.com/eic/EICrecon/actions/runs/5556691154/jobs/10150278283
https://github.com/actions/runner-images/issues/7452


wdconinc:
We could cache the /var/cache/apt directory for both lists and archives.

wdconinc (approved):

PR #772, opened 2023-07-15
fix: GeneratedParticles: ignore MCParticles with generatorStatus != 1

wdconinc:
The GeneratedParticles branch all MCParticles which are thrown into the detector, in the ReconstructedParticles format (for easy comparison). This means we should not include `generatorStatus == 0` as was the case until now.

- [x] Bug fix (issue: GeneratedParticles includes more particles than intended)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, excludes secondary particles (which should be determined from MCParticles based on the bit pattern in simulatorStatus).


bspage912 (approved):
Seems pretty straightforward.

PR #774, opened 2023-07-16
ci: ccache max size increase and show stats; key name changes

wdconinc:
For debug builds the previous 500MB doesn't fit the entire build so we end up rebuilding all the time. This bumps it up to 1500MB and adds a stats summary at the end which should allow us to assess how many misses we have.

For more efficient lookups in branches off branches, this explicitly adds the PR head and base branch to the lookup list, and changes the ccache key to the head branch for PRs. This means the temporary merge ref "774/merge" won't be used as the key, since that cannot be determined by dependent branches.

- [x] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -51,6 +51,8 @@ jobs:
         key: ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.ref_name }}-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
         restore-keys: |
              ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.ref_name }}-
+             ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.event.pull_request.head.ref }}-
+             ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.event.pull_request.base.ref }}-
              ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-
              ccache-${{ matrix.CC }}-
              ccache-
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -48,7 +48,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: .ccache
-        key: ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.ref_name }}-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
+        key: ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.event.pull_request.head.ref || github.ref_name }}-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
         restore-keys: |
              ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.ref_name }}-
              ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.event.pull_request.head.ref }}-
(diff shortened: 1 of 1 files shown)


wdconinc:
The cache key and lookup is now as follows.
```console
Run actions/cache@v3
  with:
    path: .ccache
    key: ccache-gcc-Debug-ci-ccache-max-dir-2023-07-16T14:40+00:00
    restore-keys: ccache-gcc-Debug-774/merge-
  ccache-gcc-Debug-ci-ccache-max-dir-
  ccache-gcc-Debug-main-
  ccache-gcc-Debug-
  ccache-gcc-
  ccache-
```
A new PR will find the base branch to restore. An update PR will find the head branch with higher priority. A push to main will find the ref_name with highest priority, and all PR head and base entry will be skipped due to empty entries there (`ccache-gcc-Debug--` with double dash).

wdconinc:
Gcc debug build:
```
Stats updated:      Sun Jul 16 15:08:04 2023
Cacheable calls:     800 /  800 (100.0%)
  Hits:              423 /  800 (52.88%)
    Direct:          391 /  423 (92.43%)
    Preprocessed:     32 /  423 ( 7.57%)
  Misses:            377 /  800 (47.12%)
Successful lookups:
  Direct:            391 /  800 (48.88%)
  Preprocessed:       32 /  409 ( 7.82%)
Local storage:
  Cache size (GB):  0.51 / 1.50 (33.89%)
  Files:             782
  Hits:              423 /  800 (52.88%)
  Misses:            377 /  800 (47.12%)
  Reads:            1600
  Writes:            782
```
500MB was too small but maybe 1.5GB is too much... We can evaluate in a few days.

.github/workflows/linux-eic-shell.yml
```diff
@@ -48,17 +48,19 @@ jobs:
       uses: actions/cache@v3
       with:
         path: .ccache
-        key: ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.ref_name }}-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
+        key: ccache-${{ matrix.CC }}-${{ matrix.CMAKE_BUILD_TYPE }}-${{ github.event.pull_request.head.ref || github.ref_name }}-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
```
veprbl, line 51: I'm not able to understand this. Could you comment on ref/ref_name use here?

wdconinc: The idea here was to use the branch name in the key, not whatever github decides to call its temporary merge branches. E.g. for this PR, the cache key should be `ci-ccache-max-dir`, not `774/merge` or `gh-readonly-queue/main/pr-774`.

If we (as I'm prone to do) want to branch off from another branch that is not `main`, then adding the base branch ref as a lookup key only works if that has been used as a key for writing.

In other words, currently we write caches with key `main` or `774/mere` (pull request) or `gh-readonly-queue/main/pr-774` (merge queue), that's what's in `github.ref_name`. If we branch off of `ci-ccache-max-dir`, then the new branch has no way of knowing it should look for a cache with key `774/merge`. This changes the keys to use `main` or `ci-ccache-max-dir`.

wdconinc: See https://github.com/eic/EICrecon/actions/caches?query=ccache for names of cached objects:
- ccache-gcc-Debug-gh-readonly-queue/main/pr-1006-7b1e880bc061f94a676ab46159a8cc1d2f3c11a2-2023-09-23T01:08+00:00
- ccache-gcc-Release-995/merge-2023-09-23T00:58+00:00
- ccache-gcc-Debug-main-2023-09-22T23:33+00:00

It's only for `main` that we actually use the branch name.

veprbl: I understand better now. Plus, I now see that only `base.ref` and `head.ref` are available for the pull_request event: https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request. From what I understand they are not fully qualified ref like `github.ref`, but more like `github.ref_name`.

There are also `github.base_ref` and `github.head_ref`, which should be synonyms to `github.even.pull_request.*.ref`. We don't use `pull_request_target`, so we don't really prefer either.

veprbl (approved):

wdconinc:
Rebased to resolve minor conflict. Needs new review.

veprbl (approved):

PR #775, opened 2023-07-16
ci: compress debug symbols

wdconinc:
This compresses the debug symbol section in both clang and gcc compiled debug builds. Smaller builds mean faster cache retrievals and faster pipelines.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -25,13 +25,14 @@ jobs:
           - CC: gcc
             CXX: g++
             CMAKE_BUILD_TYPE: Debug
+            CXXFLAGS: -gz
           - CC: clang
             CXX: clang++
             CMAKE_BUILD_TYPE: Release
           - CC: clang
             CXX: clang++
             CMAKE_BUILD_TYPE: Debug
-            CXXFLAGS: -fprofile-instr-generate -fcoverage-mapping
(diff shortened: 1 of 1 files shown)


wdconinc:
New:
```
Total size of all the files uploaded is 423576670 bytes
File upload process has finished. Finalizing the artifact upload
Artifact has been finalized. All files have been successfully uploaded!
The raw size of all the files that were specified for upload is 677966964 bytes
The size of all the files that were uploaded is 423576670 bytes. This takes into account any gzip compression used to reduce the upload size, time and storage
```

Previous:
```
Total size of all the files uploaded is 426393028 bytes
File upload process has finished. Finalizing the artifact upload
Artifact has been finalized. All files have been successfully uploaded!

The raw size of all the files that were specified for upload is 1718165558 bytes
The size of all the files that were uploaded is 426393028 bytes. This takes into account any gzip compression used to reduce the upload size, time and storage
```

I guess there's not really much difference in the artifacts since those are also zipped before upload. An install size of 677MB versus 1.7GB is not nothing. Maybe this should be enabled in CMake also/instead?

wdconinc:
Closing in favor of #777, which does this by default without the need to pass flags in CI.

PR #777, opened 2023-07-16
feat: use -gz if supported in Debug and RelWithDebInfo

wdconinc:
Like #775 this compresses debug symbols when supported. I think this is a useful default.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, by default compresses debug symbols.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,9 @@ cmake_minimum_required(VERSION 3.16)
 
 project(EICRecon)
 
+# CMake includes
+include(CheckCXXCompilerFlag)
+
 # Set version based on git
 include(cmake/git_version.cmake)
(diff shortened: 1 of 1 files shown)


wdconinc:
Compared to https://github.com/eic/EICrecon/pull/775#issuecomment-1637127249, the uploaded build artifacts are indeed compressed here too.
```
Total size of all the files uploaded is 423578035 bytes
File upload process has finished. Finalizing the artifact upload
Artifact has been finalized. All files have been successfully uploaded!

The raw size of all the files that were specified for upload is 677967261 bytes
The size of all the files that were uploaded is 423578035 bytes. This takes into account any gzip compression used to reduce the upload size, time and storage
```

veprbl (approved):

PR #778, opened 2023-07-16
feat: rewrite MC2SmearedParticle to operate on collections

wdconinc:
This reworks the MC2SmearedParticle algorithm and factory to operate on podio collections, which are passed to the algorithm. It moves the loop over MCParticles into the algorithm (maybe at some point the factories can parallelize this, but we're not there, and consistency in how algorithms interact with collections is more important right now). It also removes the (defunct) smearing configuration.

This also removes the `generatorStatus == 0` entries as done in #772.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: use collections in MC2SmearedParticles)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Include #772 too.

src/algorithms/reco/MC2SmearedParticle.cc
```diff
@@ -2,61 +2,46 @@
...
-    if (mc_particle->getGeneratorStatus() > 1) {
-        m_log->debug("ignoring particle with generatorStatus = {}", mc_particle->getGeneratorStatus());
-        return nullptr;
+std::unique_ptr<edm4eic::ReconstructedParticleCollection> eicrecon::MC2SmearedParticle::produce(const edm4hep::MCParticleCollection* mc_particles) {
+    auto rec_particles = std::make_unique<edm4eic::ReconstructedParticleCollection>();
+    for (const auto& mc_particle: *mc_particles) {
+
+        if (mc_particle.getGeneratorStatus() != 1) {
+            m_log->debug("ignoring particle with generatorStatus = {}", mc_particle.getGeneratorStatus());
+            return nullptr;
```
wdconinc, line 21:
```suggestion
            continue;
```

wdconinc:
This differs from the current `main` branch in the removal of non-generated `GeneratedParticles` only (sometimes O(1000) per event for large s, large Q<sup>2</sup> events), and this does not differ at all from the artifacts in #772 (in particular https://github.com/eic/EICrecon/actions/runs/5613960299).

veprbl (dismissed):
LGTM

veprbl (approved):

PR #780, opened 2023-07-17
feat: rewrite InclusiveKinematics algorithms to operate on podio collections

wdconinc:
This converts the InclusiveKinematics algorithms and factories (back) to using podio collections (from `std::vector<T*>`).

Because JANA2 expects `std::unique_ptr`, the algorithms are creating `std::unique_ptr`s of the collections. Ultimately it may be nice to just create collection and `std::move` it...

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/reco/Beam.h
+++ b/src/algorithms/reco/Beam.h
@@ -13,11 +13,11 @@ namespace eicrecon {
 
   template<class T>
   auto find_first_with_pdg(
-      const std::vector<T*>& parts,
+      const T& parts,
       const std::set<int32_t>& pdg) {
-    std::vector<T*> c;
+    T c;
     for (const auto& p: parts) {
-      if (pdg.count(p->getPDG()) > 0) {
+      if (pdg.count(p.getPDG()) > 0) {
         c.push_back(p);
         break;
       }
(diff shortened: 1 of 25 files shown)


wdconinc:
Huh, just like #778 this seems to crash in the same place... https://github.com/eic/EICrecon/actions/runs/5578084822/jobs/10192400500#step:5:847, but only for gun simulations, not for DIS.

wdconinc:
@nathanwbrei In jana2, who is supposed to catch exceptions where before they are ignored? This was a frustrating debugging exercise (resolved by 034f7a0) because again a class of factories got completely disabled by an exception, and EICrecon just goes on without notifying anything.
- Inside the [Beam.h](https://github.com/eic/EICrecon/blob/inclusive-kinematics-podio-collections/src/algorithms/reco/Beam.h#L18) podio throws an exception when you try to insert an entry from one collection into another collection without using a subset collection.
- The [algorithm](https://github.com/eic/EICrecon/blob/inclusive-kinematics-podio-collections/src/algorithms/reco/InclusiveKinematicsDA.cc) doesn't catch the exception
- The [factory](https://github.com/eic/EICrecon/blob/inclusive-kinematics-podio-collections/src/global/reco/InclusiveKinematicsDA_factory.cc) doesn't catch the exception
- The [JChainFactoryT](https://github.com/eic/EICrecon/blob/inclusive-kinematics-podio-collections/src/extensions/jana/JChainFactoryT.h) doesn't catch the exception (no Process override anyway)
- The [JFactoryPodioT](https://github.com/eic/EICrecon/blob/inclusive-kinematics-podio-collections/src/services/io/podio/JFactoryPodioT.h#L171) catches the except but throws it again
- ... (now we're in JANA2 land)
- ultimately nothing is printed about this exception, and it's just silently eaten somewhere.

What is the design intent here?

wdconinc:
This now has no differences anymore with the target `main` branch, e.g. https://github.com/eic/EICrecon/actions/runs/5649213606/job/15303675618?pr=780#step:10:72.

Ready for review and merging.

ruse-traveler (approved):
Looks great! Good to see the PODIO integration progressing!

PR #783, opened 2023-07-18
fix: place truth vertex on perigee surface in truth seeding

wdconinc:
This fixes the truth vertex issue introduced in #766 and discussed in #779 by placing the truth vertex on the perigee surface with its two-vector. This 'mostly' recovers the original behavior, but it necessarily has differences because the truth vertex propagation is modeled differently (but now identical to how this is done in realistic seeding).

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @veprbl @bschmookler 

No.

Yes, some changes in truth seeded tracks to be expected.


src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -92,7 +93,7 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
         // Insert into edm4eic::TrackParameters, which uses numerical values in its specified units
         auto track_parameter = track_parameters->create();
         track_parameter.setType(-1); // type --> seed(-1)
-        track_parameter.setLoc({0.0, 0.0}); // 2d location on surface [mm]
+        track_parameter.setLoc({std::hypot(v.x, v.y), v.z}); // 2d location on surface [mm]
```
wdconinc, line 96: Not sure why MCParticle has Vector3d for vertices, but whatever:
```suggestion
        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y)), static_cast<float>(v.z)}); // 2d location on surface [mm]
```
A choice between two casts and a double hypot or three casts and a float hypot... probably doesn't matter.

bschmookler (approved):
This is very close to what we have for the realistic seeding. 

The only difference is that here, we use the generated particle's (x,y,z) vertex. For the real seeds, we fit 3 points and find the point of closest approach in the (x,y) plane, and extract the z value when r=0 from a linear fit. 

The generated particle's vertex may not be the closest point to the perigee point at (x,y,z)=(0,0,0). The CKF may assume that we calculated that point of closest approach to the perigee point when we pass in the seed parameters. So, if the particle's generation point is very different from the point of closest approach to (0,0,0), this can have some effect on the CKF results. I'm not entirely sure on this.

Since we know the particle's momentum vector and generation vertex and the magnetic field, we can calculate the point of closest approach to (0,0,0) point. That can be a future improvement.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -92,7 +93,7 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
         // Insert into edm4eic::TrackParameters, which uses numerical values in its specified units
         auto track_parameter = track_parameters->create();
         track_parameter.setType(-1); // type --> seed(-1)
-        track_parameter.setLoc({0.0, 0.0}); // 2d location on surface [mm]
+        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y)), static_cast<float>(v.z)}); // 2d location on surface [mm]
```
veprbl, line 96: So coordinates in
https://github.com/eic/EICrecon/blob/3a54c0125862d80326f1b6b5b273b793acbbe3b6/src/global/tracking/CKFTracking_factory.cc#L83
can be effectively provided with
https://github.com/eic/EICrecon/blob/3a54c0125862d80326f1b6b5b273b793acbbe3b6/src/global/tracking/CKFTracking_factory.cc#L65-L66
?

wdconinc: Not quite. The non-zero three vector position with zero two-vector on the surface specifies a unique point. The zero three vector with non-zero two-vector on the surface specifies a circle (of closest approach).

wdconinc:
Merging with approval from @bschmookler after discussion at this morning's [trk recon meeting](https://indico.bnl.gov/event/20074/).

PR #784, opened 2023-07-18
feat: make the factory-factory a variadic template (more evil things)

### LINKED ISSUE #628: EICrecon takes 20 minutes to build

wdconinc:
The wordiness of the current factory-factory,
```cpp
class Cluster_factory_EcalBarrelSciGlassTruthClusters :
  public CalorimeterClusterRecoCoG_factoryT<Cluster_factory_EcalBarrelSciGlassTruthClusters>
{
  public:
    template <typename... Args>
    Cluster_factory_EcalBarrelSciGlassTruthClusters(Args&&... args)
    :  CalorimeterClusterRecoCoG_factoryT<Cluster_factory_EcalBarrelSciGlassTruthClusters>(std::forward<Args>(args)...)
      { }
};
```
has been bothering me since #666. This always seemed like it should be possible to express it more elegantly, with a `using` or so.

However, the naive approach
```cpp
using Cluster_factory_EcalBarrelSciGlassTruthClusters =
CalorimeterClusterRecoCoG_factoryT<Cluster_factory_EcalBarrelSciGlassTruthClusters>;
```
fails out of the gate since you can't use the lhs in the rhs.

This PR introduces an approach that does allow the `using` approach to work, and it actually makes it even snappier:
```cpp
using Cluster_factory_EcalBarrelSciGlassTruthClusters = CalorimeterClusterRecoCoG_factoryT<>;
```
(because why repeat yourself).

This doesn't fix #628...

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @veprbl @nathanwbrei 

No.

No.

### DIFF
--- a/src/detectors/B0ECAL/B0ECAL.cc
+++ b/src/detectors/B0ECAL/B0ECAL.cc
@@ -14,19 +14,8 @@
 
 
 namespace eicrecon {
-    class Cluster_factory_B0ECalTruthClusters: public CalorimeterClusterRecoCoG_factoryT<Cluster_factory_B0ECalTruthClusters> {
-    public:
-        template <typename... Args>
-        Cluster_factory_B0ECalTruthClusters(Args&&... args)
-        : CalorimeterClusterRecoCoG_factoryT<Cluster_factory_B0ECalTruthClusters>(std::forward<Args>(args)...) { }
-    };
-
-    class Cluster_factory_B0ECalClusters: public CalorimeterClusterRecoCoG_factoryT<Cluster_factory_B0ECalClusters> {
-    public:
-        template <typename... Args>
-        Cluster_factory_B0ECalClusters(Args&&... args)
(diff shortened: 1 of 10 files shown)


wdconinc:
I think this is now 1 step away from a generic factory-factory that can generate any factory. All this only configurable at compile time and with impact on compile time and object size... but it's what we can do now to stop the expansion of factory classes.

wdconinc:
My understanding (not having looked into this) is that Jana dispatches data flow to factories based on the types that those factories produce. But a good question for @nathanwbrei.

veprbl:
If I had to guess, the reason why JFactoryT needs to be provided a type is because you want to be calling a proper destructor, unless you are able to inherit from JObject and have it go through a virtual dispatch. JANA has no idea what types a factory can consume. In our case, all calorimetry algorithms produce the same type for each detector. We shouldn't need multiple classes per algorithm, just multiple instances of the same class.

PR #785, opened 2023-07-19
feat: convert calorimetry JFactoryT to JChainFactoryT

wdconinc:
All of the calorimetry factories are currently JFactoryT instead of JChainFactoryT, which means that they don't allow for providing a configuration object in the plugin (think: BEMC.cc).

This converts all calorimetry JFactoryT factories to JChainFactoryT, which requires that we specify also the set of input tags, and the output tag in the JChainFactoryGeneratorT constructor in the plugin. These are therefore removed from the factories since they are set in the plugin.

The third argument of the JChainFactoryGeneratorT can become the config object, which can now be introduced in a separate PR, algorithm by algorithm.

All this fits in a three-step process towards simplification:
- make calorimetry algorithms accept external config objects
- collect all configuration in plugin
- remove or reduce dedicate factory code

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators:
  - @veprbl I think this should be largely independent of the conversion to collections, but some diffs will touch close to the context of those changes. I'm going to rebase this after the conversion to collections is done.

No.

No.

### DIFF
--- a/src/detectors/EHCAL/EHCAL.cc
+++ b/src/detectors/EHCAL/EHCAL.cc
@@ -34,7 +34,7 @@ extern "C" {
           {"HcalEndcapNRawHits"}, "HcalEndcapNRecHits"
         ));
         app->Add(new JChainFactoryGeneratorT<CalorimeterHit_factory_HcalEndcapNMergedHits>(
-          {"HcalEndcapNRawHits"}, "HcalEndcapNMergedRecHits"
+          {"HcalEndcapNRawHits"}, "HcalEndcapNMergedHits"
         ));
         app->Add(new JChainFactoryGeneratorT<ProtoCluster_factory_HcalEndcapNTruthProtoClusters>(
           {"HcalEndcapNRecHits", "HcalEndcapNHits"}, "HcalEndcapNTruthProtoClusters"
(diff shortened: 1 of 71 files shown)


src/detectors/EHCAL/EHCAL.cc
```diff
@@ -26,11 +27,21 @@ extern "C" {
...
-        app->Add(new JFactoryGeneratorT<ProtoCluster_factory_HcalEndcapNTruthProtoClusters>());
-        app->Add(new JFactoryGeneratorT<ProtoCluster_factory_HcalEndcapNIslandProtoClusters>());
+        app->Add(new JChainFactoryGeneratorT<RawCalorimeterHit_factory_HcalEndcapNRawHits>(
+          {"HcalEndcapNHits"}, "HcalEndcapNRawHits"
+        ));
+        app->Add(new JChainFactoryGeneratorT<CalorimeterHit_factory_HcalEndcapNRecHits>(
+          {"HcalEndcapNRawHits"}, "HcalEndcapNRecHits"
+        ));
+        app->Add(new JChainFactoryGeneratorT<CalorimeterHit_factory_HcalEndcapNMergedHits>(
+          {"HcalEndcapNRawHits"}, "HcalEndcapNMergedRecHits"
```
wdconinc, line 37:
```suggestion
          {"HcalEndcapNRawHits"}, "HcalEndcapNMergedHits"
```

wdconinc (commented):
Re-add the `HcalEndcapNMergedHits` collection

veprbl (approved):
No output changes when compared to main branch

PR #786, opened 2023-07-19
fix: prefer pragma once in factories

wdconinc:
`#pragma once` has [caveats](https://en.wikipedia.org/wiki/Pragma_once#Caveats), but should be avoided in factories where frequent copy-paste workflows present risks if the header guards are not updated.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

  - clang-tidy doesn't support this as a test

No.

No.


ruse-traveler (commented):
This is quite reasonable to me. (Apologies for the delayed feedback!)

wdconinc:
One more factory conversion and this PR has become moot :-)

PR #792, opened 2023-07-24
fix: rm failing MC2SmearedParticleConfig include

wdconinc:
This removes an include of a removed header, removed in #778.

When we run CI tests, we use an eic-shell container that has EICrecon already installed. That means that removing a header file in the test build can still cause that header file to be found in the system directories. This then only shows up in an actual clean build later on.

Short of revamping our CI, I don't really know of a way to prevent this from happening in the short term.

- [x] Bug fix (issue: failing nightly at https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/1799855#L1974)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/global/reco/reco.cc
+++ b/src/global/reco/reco.cc
@@ -8,7 +8,6 @@
 
 #include <extensions/jana/JChainFactoryGeneratorT.h>
 #include <extensions/jana/JChainMultifactoryGeneratorT.h>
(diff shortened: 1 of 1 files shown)


ruse-traveler (dismissed):
Makes sense to me given PR #778!

sly2j (approved):

PR #804, opened 2023-07-27
fix: Simplified track seeder findPCA (was: findROOT) function and guard against returning NaN values

bschmookler:
This PR fixes addition NAN values which can occasionally get returned by the track seeder `findRoot` function. The NAN values arise from negative `sqrt` arguments. Prior to this fix, this issue occurred in about 1 out of 5000 single muon events.

These events with the negative `sqrt` arguments can occur when the `circleFit` routine returns a large Y value for the circle center and a small X value.

An example seed where this occurs is shown below:

`Circle Parameters: R = 4557.015625, X0 = 0.979015, Y0 = -4557.020996`
`Output from findRoot: miny = -0.005411, minx = 1.958031, miny2 = -9114.037159, minx2 = -nan`
`Point-of-closest approach returned in Global coordinates (x,y,z): -nan -0.005411 0.000969`

Final seed parameters before fix:
```
localpos(0) = -nan, localpos(1) = -nan
theta = 2.166248, phi = -nan
qOverP = -0.356472, charge = -1
```

With this PR, this seed now correctly returns `(minx,miny)`.

Note that the `findRoot` function could return a 'mixed' pair (e.g. `minx2` and `miny`). This is also fixed in this PR.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, it writes out the seeded tracking results to the EICRecon output ROOT file.


src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -157,12 +157,21 @@ std::pair<float, float> eicrecon::TrackSeeding::findRoot(std::tuple<float,float,
 		      * pow(Y0,4)) + square(X0) * Y0 + pow(Y0, 3))
     / (square(X0) + square(Y0));
 
+  //If one of the calculated y values is farther than R from the circle center, return the other pair
+  if( square(R) - square(miny - Y0) < 0 )
+	return std::make_pair(-std::sqrt(square(R) - square(miny2 - Y0)) + X0 , miny2);
+  if( square(R) - square(miny2 - Y0) < 0 )
+	return std::make_pair(std::sqrt(square(R) - square(miny - Y0)) + X0 , miny);
```
wdconinc, line 164: Why is `square(R) - square(miny2 - Y0)` positive when `square(R) - square(miny - Y0)` is negative? That doesn't seem obvious.

wdconinc (commented):
I'm surprised by the changes this introduces in e.g. https://github.com/eic/EICrecon/actions/runs/5676333499/job/15383179977?pr=804#step:10:769: many changes in loc.a and some sign changes to phi.

src/algorithms/tracking/TrackSeeding.cc
bschmookler: I think you are right that it is not guaranteed that one of the sqrts has to have a positive argument. But in that case, both `minx` and `minx2` would have to be on the order of 1e-6, which is unlikely.

In the example event shown above, if I use the values of R, X0, Y0 and then calculate `miny, minx, miny2, and minx2` by hand with higher precision, I get `-0.0054761640, -9114.0365, 1.9580288, 1.1812951e-06`, respectively. The `minx2` value is close to zero, which causes issues with the float precision when subtracting 2 large numbers.

Maybe there is a better way to write this part of the code?

wdconinc: In essence
```cpp
  const double R0 = std::hypot(X0, Y0); // ovflw safety
  const double miny_alt = Y0 * (1 + R/R0); // assumes R0 > 0
  const double miny2_alt = Y0 * (1 - R/R0); // assumes R0 > 0
  const double minx_alt = miny_alt * X0/Y0; // assumes Y0 != 0
  const double minx2_alt = miny2_alt * X0/Y0; // assumes Y0 != 0
```
and then pick the one with lowest $r$.

wdconinc: Part of me doesn't believe this could really be this simple. But then again, another part of me didn't believe it could be that complicated at the start... So, let me know if I'm overlooking something.

bschmookler: Hi Wouter. Thanks for all these suggestions.

I will add a guard for the `R0 = 0 (< epsilon)` case.

Otherwise, we may not need to check both possible points. I think the point

`x = X0 * (1 - R/R0) and y = Y0 * (1- R/R0)`

should always be on the circle and on the line connecting the origin to the circle center. And it will always be closer to the origin than the opposite point

`x = X0 * (1 + R/R0) and y = Y0 * (1+ R/R0)`

bschmookler: This should hold whether R0 is larger or smaller than R.

bschmookler: I made the changes I suggested. I was able to run 10k muons without any issue. At the seed level, the ACTS loc-a and phi reconstruction get modified only very slightly. The other seed parameters don't change at all, as expected.

image1

image2

wdconinc:
I got too involved in this one, so marking @veprbl for review.

wdconinc:
But could you update the PR description since that is what becomes the squash merge commit message?

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -86,8 +86,10 @@ std::vector<edm4eic::TrackParameters*> eicrecon::TrackSeeding::makeTrackParams(S
       float R = std::get<0>(RX0Y0);
       float X0 = std::get<1>(RX0Y0);
       float Y0 = std::get<2>(RX0Y0);
-      if (R > std::sqrt(std::cbrt(std::numeric_limits<float>::max()))) {
-        // avoid future float overflow for hits on a line
+      if (R > std::numeric_limits<float>::max() ||
+	std::abs(X0) > std::numeric_limits<float>::max() ||
+	std::abs(Y0) > std::numeric_limits<float>::max()) {
+        // avoid float overflow for hits on a line
```
wdconinc, line 92: I think the pathological cases (straight line, circle around origin) need to be caught before entering this function, not in the function. It will be easier to handle them in the caller than here. If you have a straight line, the perigee is different but still determined. if you have a center point at the origin, you don't need to assign an invalid point but discard the seed.

wdconinc (commented):
In addition to the comment on the code, I think `findRoot` is not a good name anymore.

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -86,8 +86,10 @@ std::vector<edm4eic::TrackParameters*> eicrecon::TrackSeeding::makeTrackParams(S
       float R = std::get<0>(RX0Y0);
       float X0 = std::get<1>(RX0Y0);
       float Y0 = std::get<2>(RX0Y0);
-      if (R > std::sqrt(std::cbrt(std::numeric_limits<float>::max()))) {
-        // avoid future float overflow for hits on a line
+      if (R > std::numeric_limits<float>::max() ||
```
veprbl, line 89: Are you trying to express `std::isfinite` https://en.cppreference.com/w/cpp/numeric/math/isfinite ?

veprbl:
This looks much better now. I don't understand the original expression with square roots and 6-th power polynomials to judge the source of NaNs.

In my personal opinion, it would be really nice to just reuse something with Acts. I've started a thread on Mattermost, maybe we can discuss.

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -149,21 +151,19 @@ std::pair<float, float> eicrecon::TrackSeeding::findRoot(std::tuple<float,float,
...
-    / (square(X0) + square(Y0));
 
-  const double miny2 = (-std::sqrt(square(X0) * square(R) * square(Y0) + square(R)
-		      * pow(Y0,4)) + square(X0) * Y0 + pow(Y0, 3))
-    / (square(X0) + square(Y0));
+  const double R0 = std::hypot(X0, Y0);
 
-  const double minx = std::sqrt(square(R) - square(miny - Y0)) + X0;
-  const double minx2 = -std::sqrt(square(R) - square(miny2 - Y0)) + X0;
+  //If center of circle is at origin, there is no point-of-closest approch
```
veprbl, line 157:
```suggestion
  //If center of circle is at origin, there is no point-of-closest approach
```

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -86,8 +86,10 @@ std::vector<edm4eic::TrackParameters*> eicrecon::TrackSeeding::makeTrackParams(S
       float R = std::get<0>(RX0Y0);
       float X0 = std::get<1>(RX0Y0);
       float Y0 = std::get<2>(RX0Y0);
-      if (R > std::sqrt(std::cbrt(std::numeric_limits<float>::max()))) {
-        // avoid future float overflow for hits on a line
+      if (R > std::numeric_limits<float>::max() ||
+	std::abs(X0) > std::numeric_limits<float>::max() ||
+	std::abs(Y0) > std::numeric_limits<float>::max()) {
```
veprbl, line 91:
```suggestion
      if (!(std::isfinite(R) &&
	std::isfinite(std::abs(X0)) &&
	std::isfinite(std::abs(Y0))) {
```

veprbl (dismissed):

src/algorithms/tracking/TrackSeeding.cc
bschmookler: I renamed `findRoot `to `findPCA` (for point-of-closest approach). I also moved the check on the circle center point to before the calling of this function.

Following the discussion at the track reconstruction meeting on Thursday, we can add additional functionality to handle the linear case later. We will also consider a modification to using Acts functions to find the seed parameters in the future.

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -92,6 +92,13 @@ std::vector<edm4eic::TrackParameters*> eicrecon::TrackSeeding::makeTrackParams(S
         // avoid float overflow for hits on a line
         continue;
       }
+      if ( std::hypot(X0,Y0) < std::numeric_limits<double>::epsilon() ||
+	!std::isfinite(std::hypot(X0,Y0)) ) {
```
veprbl, line 96: `!std::isfinite(std::hypot(X0,Y0))` - this condition is redundant given we already checked `X0` and `Y0` by this point.

veprbl: On a second thought, this is not redundant, there are cases that overflow (`std::hypot(std::numeric_limits<float>::max(), 1.0e35f)`)

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -86,10 +86,19 @@ std::vector<edm4eic::TrackParameters*> eicrecon::TrackSeeding::makeTrackParams(S
...
       float Y0 = std::get<2>(RX0Y0);
-      if (R > std::sqrt(std::cbrt(std::numeric_limits<float>::max()))) {
-        // avoid future float overflow for hits on a line
+      if (!(std::isfinite(R) &&
+	std::isfinite(std::abs(X0)) &&
+	std::isfinite(std::abs(Y0)))) {
+        // avoid float overflow for hits on a line
         continue;
       }
+      if ( std::hypot(X0,Y0) < std::numeric_limits<double>::epsilon() ||
```
veprbl, line 95: Would like to point out that `X0` and `Y0` are `float`s. Perhaps
```suggestion
      if ( std::hypot(X0,Y0) < std::numeric_limits<decltype(std::hypot(X0,Y0))>::epsilon() ||
```

veprbl (approved):

veprbl (approved):

PR #809, opened 2023-07-27
drop SciGlass

veprbl:
SciGlass reconstruction was never used for serious studies. It won't be needed by ePIC in the future.

This allows us to drop a whole algorithm.


wdconinc:
Any chance the ecal endcap n might need it?

veprbl:
Probably not. And I have no idea if this code even works.

wdconinc (approved):
Looks good. I'm in a cleanup mood.

PR #815, opened 2023-07-28
feat: enable ccache by default if found, unless ENABLE_CCACHE=OFF

wdconinc:
Local development with repeated compile cycles is greatly improved by ccache compiler caching. This enables ccache use by default if `ccache` is available, unless the user disables it. If a user explicitly passes a specific `ccache` as `CMAKE_CXX_COMPILER_LAUNCHER`, it will use that one, though.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, will compile with `ccache` if available.


veprbl:
`echo 'export CXX=ccache' >> ~/.profile`?

wdconinc:
> `echo 'export CXX=ccache' >> ~/.profile?`

Yeah, but I don't want to enable it for ALL my compiles (otherwise it really becomes useless because with spack dev I probably rotate through a few GB of cache every day or two).

This is just a convenience feature (which doesn't even work, https://github.com/eic/EICrecon/actions/runs/5694084760/job/15434537689?pr=815), so comments like this are welcome.

veprbl:
Maybe in combination with https://github.com/direnv/direnv ?

wdconinc:
Mostly addressed by #940.

PR #817, opened 2023-07-30
JOmniFactory as potential successor to JChainFactory 

nathanwbrei:

This PR introduces JOmniFactory, which is designed to have the same functionality as JChainMultifactory with a notably cleaned-up interface. This will eventually support instantiating and wiring factories via an external configuration file and/or script, and may form the basis of a 'generic algorithms' JANA adapter. The hope is that we can migrate all EICrecon factories to use this as their base class and thereby substantially reduce user confusion and boilerplate code.
 
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/extensions/jana/JOmniFactoryGeneratorT.h
+++ b/src/extensions/jana/JOmniFactoryGeneratorT.h
@@ -79,7 +79,7 @@ public:
         // Create throwaway factory so we can populate its config using our map<string,string>.
         FactoryT factory;
         factory.ConfigureAllParameters(config_params);
-        auto config = factory.GetConfig();
+        auto config = factory.config();
 
         m_wirings.push_back({.m_tag=tag,
                              .m_default_input_tags=default_input_tags,
@@ -101,7 +101,7 @@ public:
             // factory->SetTag(wiring.m_tag);
             // We do NOT want to do this because JMF will use the tag to suffix the collection names
             // TODO: NWB: Change this in JANA
-            factory->GetConfig() = wiring.m_default_cfg;
+            factory->config() = wiring.m_default_cfg;
 
             // Set up all of the wiring prereqs so that Init() can do its thing
             // Specically, it needs valid input/output tags, a valid logger, and
--- a/src/extensions/jana/JOmniFactoryGeneratorT.h
+++ b/src/extensions/jana/JOmniFactoryGeneratorT.h
@@ -1,0 +1,120 @@
+// Copyright 2023, Jefferson Science Associates, LLC.
+// Subject to the terms in the LICENSE file found in the top-level directory.
+// Created by Nathan Brei
+
+#pragma once
+
+#include <JANA/JFactorySet.h>
+#include <JANA/JFactoryGenerator.h>
+#include <vector>
+
+template<class FactoryT>
+class JOmniFactoryGeneratorT : public JFactoryGenerator {
+public:
+    using FactoryConfigType = typename FactoryT::ConfigType;
+
+private:
+    struct TypedWiring {
+        std::string m_tag;
+        std::vector<std::string> m_default_input_tags;
+        std::vector<std::string> m_default_output_tags;
+        FactoryConfigType m_default_cfg; /// Must be properly copyable!
+    };
+
+    struct UntypedWiring {
+        std::string m_tag;
+        std::vector<std::string> m_default_input_tags;
+        std::vector<std::string> m_default_output_tags;
+        std::map<std::string, std::string> m_config_params;
+    };
+
+public:
+
+
+    explicit JOmniFactoryGeneratorT(std::string tag,
+                                    std::vector<std::string> default_input_tags,
+                                    std::vector<std::string> default_output_tags,
+                                    FactoryConfigType cfg,
+                                    JApplication* app) {
+        m_app = app;
+        m_wirings.push_back({.m_tag=tag,
+                             .m_default_input_tags=default_input_tags,
+                             .m_default_output_tags=default_output_tags,
+                             .m_default_cfg=cfg
+                            });
+    };
+
+    explicit JOmniFactoryGeneratorT(std::string tag,
+                                    std::vector<std::string> default_input_tags,
+                                    std::vector<std::string> default_output_tags,
+                                    JApplication* app) {
+        m_app = app;
+        m_wirings.push_back({.m_tag=tag,
+                                .m_default_input_tags=default_input_tags,
+                                .m_default_output_tags=default_output_tags
+                                });
+
+    }
+
+    explicit JOmniFactoryGeneratorT(JApplication* app) : m_app(app) {
+    }
+
+    void AddWiring(std::string tag,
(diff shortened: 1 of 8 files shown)


src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
```diff
@@ -0,0 +1,62 @@
...
+    // Process
+    void Process(const std::shared_ptr<const JEvent> &event) override {
+
+        // TODO: NWB: We are using GetCollectionBase because GetCollection is temporarily out of commission due to JFactoryPodioTFixed
+        auto proto = static_cast<const edm4eic::ProtoClusterCollection*>(event->GetCollectionBase(GetInputTags()[0]));
+        auto mchits = static_cast<const edm4hep::SimCalorimeterHitCollection*>(event->GetCollectionBase(GetInputTags()[1]));
+
+        try {
+            auto clusters_with_assocs = m_algo.process(proto, mchits);
+            SetCollection<edm4eic::Cluster>(GetOutputTags()[0], std::move(clusters_with_assocs.first));
```
wdconinc, line 52: At some point a generic algorithms approach will need to get collections and set collections for an arbitrary number of inputs and outputs, still by string tag. I was working on modifying `DeclarePodioOutputs` and `SetCollection` that allows a syntax like:
```suggestion
            SetCollections<std::tuple<edm4eic::Cluster, std::emd4eicMCRecoClusterParticleAssociation>>(GetOutputTags(), std::move(clusters_with_assocs));
```
But I got somewhat lost in `std::indexed_sequence`s and fold expressions needed for this :-)

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
```diff
@@ -0,0 +1,62 @@
...
+    const dd4hep::Detector* m_detector;
+    eicrecon::CalorimeterClusterRecoCoG m_algo;
+
+public:
+    void Init() override {
+
+        DeclarePodioOutput<edm4eic::Cluster>(GetOutputTags()[0]);
+        DeclarePodioOutput<edm4eic::MCRecoClusterParticleAssociation>(GetOutputTags()[1]);
+
+        DeclareParameter("samplingFraction", GetConfig().sampFrac);
```
wdconinc, line 25: One thing I haven't figured out yet is how we aim to allow algorithms to define their config options so we don't have to duplicate them in the factory. How does this approach get us closer to that? A generic factory will likely have to take the algorithm class as a template argument, and we'll have to use that algorithm's information to build a list of `DeclareParameters` with another fold expression? It won't work with the current struct approach in the algorithms, that's for sure (it doesn't have a way to get 
names, for starters). Maybe we should just propagate the algorithms property interface already. Those are mature enough.

src/extensions/jana/JOmniFactoryGeneratorT.h
```diff
@@ -0,0 +1,65 @@
...
+        // Factory is ready
+        factory_set->Add(factory);
+    }
+
+private:
+    std::string m_tag;
+    std::vector<std::string> m_default_input_tags;
+    std::vector<std::string> m_default_output_tags;
+    FactoryConfigType m_default_cfg;                   /// Default config for a factories. (!) Must be properly copyable
+    JApplication* m_app; // TODO: NWB: Remove me
```
wdconinc, line 64: We should have jana2 2.1.1 in CI now. I think that has the GetApplication in JMultiFactory which was the only thing that's missing for this, right?

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
nathanwbrei: I know what you mean. I think that the variadic SetCollections() should live at the JOmniFactory level and not the JMultifactory level though. Once we do this, the only logic left in the factory is associating ConfigT fields with parameter names, and acquiring Services. Services are easy to handle via another parameter pack, and maybe the name associations should be attached to the ConfigT struct? However I want to get the story straight on how we wire everything from a config file before we go down another rabbit hole.

wdconinc: I agree, this wasn't going into JMultiFactory but indeed needs the Omni factory (I had a different name in my experimentation).

src/extensions/jana/JOmniFactoryGeneratorT.h
nathanwbrei: Sadly, it looks like we will continue to need to pass m_app to the JOmniFactoryGenerator manually. Why, you ask? For the OmniFactory we need to call Init() before we call JFactorySet::Add, and we can't safely call Init() until the OmniFactory has been given a JApplication. Note this is NOT the case for JChainFactory and JMultifactory, which should be just fine.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
nathanwbrei: Just brainstorming here: Probably by using a templated visitor pattern. Consider: 
```c++

class ConfigT {
    int bucket_count = 2;
    float threshold = 6.7;
}

template <>
void Configure<ConfigT>(ParameterRegistrar& visitor, ConfigT& config) {
    visitor.RegisterParameter<int>("buckets", &config.bucket_count);
    visitor.RegisterParameter<float>("threshold", &config.threshold);
}
```

We then have separate subclasses of `ParameterRegistrar` for JANA, Gaudi, Art, etc.

wdconinc: The other approach, in algorithms ([`rroperty.h`](https://github.com/eic/algorithms/blob/main/core/include/algorithms/property.h#L73), [`ClusterRecoCoG`](https://github.com/eic/algorithms/blob/main/calorimetry/include/algorithms/calorimetry/ClusterRecoCoG.h#L54)), is to have algorithms define properties which get registered internally. This can then be queried by the framework.

nathanwbrei: Yeah, we are also using that pattern in JANA in `JEventProcessorSequential.h`. I have mixed feelings about it. The major pluses I see (compared to the visitor approach) are:
1. We can change the implementation at a very deep level without interfering with user's code
2. There is very little surface area for people to get confused
 
The major minuses I see are:
1. We no longer have a POD
2. If something goes wrong we are going to have to reason very carefully about C++ member initialization order

I would usually reach for a visitor pattern first just based off personal preference, but the more I think this one through the better it looks. I can imagine a JOmniFactory that represents collection types and Services the same way, and this could be a good first step to making JOmniFactory fully generic. I'll give it a whirl.

nathanwbrei:
Closing this PR because it is redundant with #865. (This one is based on my personal fork, which obviously cannot publish to github pages, so the tests can't pass)

PR #820, opened 2023-07-30
fix: restore LGPL license for algorithms

wdconinc:
This clarifies that the algorithms (and arguably everything else here) was always LGPL as per the upstream license conditions of juggler.

- [x] Bug fix (issue #2, in part)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/LICENSE
+++ b/src/algorithms/LICENSE
@@ -1,0 +1,166 @@
+
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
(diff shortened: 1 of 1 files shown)


wdconinc:
This doesn't require positive approval from all contributors since it merely clarifies the reality that this code is LGPL licensed.

src/algorithms/LICENSE
```diff
@@ -0,0 +1,166 @@
+
```
veprbl, line 1: move to root with name `LICENSE.LGPL3`

wdconinc: That would imply that everything is LGPL, which is not what I'm trying to merge here (though I think it is). This location makes it clear that algorithms is LGPL, which no one argues against.

veprbl (approved):

PR #807, opened 2023-07-27
feat: migrate the imaging calorimeter algorithms to collections and modern factories

wdconinc:
### Briefly, what does this PR introduce?
This modifies the ImagingPixelReco, ImagingTopoCluster, ImagingClusterReco, and TruthEnergyPositionClusterMerger algorithms to use podio collections and modern (multi)factory approaches.

~~Someone might quibble about the `std::vector<std::vector<std::pair<uint32_t, const edm4eic::CalorimeterHit&>>> groups`, but I think it's stable here: it is a finite lifetime object with references to within a const collection.~~ No, this wasn't stable.

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators @sly2j @Chao1009 @mariakzurek 

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #821, opened 2023-07-31
fix: convert ImagingPixelReco to use CalorimeterHitReco

wdconinc:
This replaces the ImagingPixelReco with the more broadly applicable CalorimeterHitReco (and removes ImagingPixelReco), after discussion with @Chao1009. That fixes the time resolution mismatches discussed in #807.

At some point CalorimeterHitDigi and CalorimeterHitReco will learn ToT and other features of modern digitization (including AstroPix) and then we can update the parameters here. For now, this uses a 'simple' TDC assumption with bit resolution of 3.25 ns, after discussion with @mariakzurek.

After this PR is applied to `imaging-cluster-reco-collections` in #807, the differences for a single event (first array is main vs second array is this branch) are
```
EcalBarrelImagingClusters.time
NOT EQUAL
[[0.00033, 0.00034, 0.000649]]
[[3.25, 3.25, 6.5]]
EcalBarrelImagingMergedClusters.time
NOT EQUAL
[[3.87, 6.79, 3.64]]
[[3.25, 6.5, 3.25]]
```
i.e the timing for EcalBarrelImagingClusters is now sensible instead of incorrectly of picosecond scale, and it is indeed digitized with 3.25 bit resolution (as it happens, tracks take about 3.5 ns to reach the detector, or at least 2.5 ns for photons at eta = 0). That time is now used for the merged clusters.

- [x] Bug fix (issue: incorrect hit time reconstruction in imaging pixels)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @Chao1009 @mariakzurek @sly2j 

No.

No.


veprbl (approved):
The collections that changed are:

* EcalBarrelImagingClusters.time
* EcalBarrelImagingMergedClusters.time
* EcalBarrelImagingRawHits.timeStamp
* EcalBarrelImagingRecHits.dimension.x
* EcalBarrelImagingRecHits.dimension.y
* EcalBarrelImagingRecHits.time

All values of dimension.{x,y} have changed from 0. to 0.5. Timing of the hits have bumped from various fractional values < 1ns to a fixed value of 3.25 ns (occasionally 6.5).

PR #822, opened 2023-07-31
Realistic scfi/imaging energy/position cluster merger

wdconinc:
This adds the realistic scfi/imaging energy/position cluster merger to the default reconstruction. It also renames the truth energy/position cluster merger collections and associations to end with `..TruthClusters`.

There is a bit of a design decision here that was made: if the matched energy and position cluster disagree on the MCParticle entry, then two associations are written, one for each possible entry, and at half weight.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @Chao1009 @mariakzurek @sly2j 

No.

No.


wdconinc:
Rebased, conflicts resolved, ready for review again.

wdconinc:
@sly2j @mariakzurek @Chao1009 FYI This is a new feature and the only way this gets into the August campaign is with a review/approval by tomorrow cob.

veprbl:
It would seem like this finds 272 clusters in rec_dis_18x275_minQ2=1000_brycecanyon.edm4eic.root instead of 1984 before. Is that a known effect?

wdconinc:
Matching parameters are absolutely not optimized yet. That's a study for when this is in eic-shell. Let me check again if there something else going on here.

wdconinc:
We had some discussion about this PR in the barrel ecal simulation meeting (cc @mariakzurek). The reason for poor matching in the absence of better parameters is likely the tight eta matching requirement of 0.2 (ScFi has no eta resolution at all so this likely just restricts found matches to the center of the detector) and the energy matching of 50% which may still be too small for the poor imaging layer energy resolution. This really needs a more extensive study to determine the best parameters. But getting this into EICrecon would allow that study to happen more easily (e.g. by @AkshayaVijay at some point later).

veprbl (approved):
Merged clusters don't seem to be used in reco yet
https://github.com/eic/EICrecon/blob/ed64e33f043b3e62b5d2c5319a700773adcd10a6/src/global/reco/reco.cc#L34-L44
should be fine to get this in to get studies going

Issue #782, opened 2023-07-17
LFHCal between 80% and 90% of reconstruction time consumption


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): `main`
- Which version (or `HEAD` for the most recent on git): `HEAD`
- Any specific OS or system where the issue occurs? `eic-shell`
- Any special versions of ROOT or Geant4? no

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Pick latest `main` branch at https://github.com/eic/EICrecon/actions?query=branch%3Amain+ (currently https://github.com/eic/EICrecon/actions/runs/5575953696)
2. Download `rec_dis_18x275_minQ2=1000_brycecanyon.dot` (currently https://github.com/eic/EICrecon/suites/14350109589/artifacts/808204419)
3. Unzip artifact
4. Filter zeros with `sed -i 's|0.0%"]|0.0%", penwidth=0]|g' jana.dot`
5. Convert to svg with `dot -Tsvg jana.dot > jana.svg`
6. Find the LFHCAL branch (which should be easily visible

See here:
![jana](https://github.com/eic/EICrecon/assets/4656391/5a840ceb-134a-4557-a2aa-884505e53994)

### Expected Result: (what do you expect when you execute the steps above)
Pareto principle: 80% of time spent in 20% of algorithms.

### Actual Result: (what do you get when you execute the steps above)
Over 80% of time is spent in LFHCAL alone.

wdconinc:
This is all in the factory that creates `edm4eic::ProtoCluster:LFHCALIslandProtoClusters`.

veprbl:
I suspect the implementation of the adjacencyMatrix is the culprit. I was originally investigating a possibility to use cling instead of CLHEP one from DD4hep. The main issue that there doesn't seem to be a way to instantiate an additional TInterpreter's to have them thread-local.

wdconinc:
https://github.com/eic/EICrecon/blob/main/src/detectors/FHCAL/ProtoCluster_factory_LFHCALIslandProtoClusters.h#L52-L66 yeah :-D I didn't know that had grown so much.

veprbl:
It could be just the O(N^2) due to combinatorics in the number of hits.

I bet we could cut the time in half by replacing this line with `break;`:
[...]

PR #824, opened 2023-08-01
CalorimeterIslandCluster: compile adjacencyMatrix expressions using cling

veprbl:
This allows to significantly speed up evaluation of the adjacency checks. The price is an additional dependency on ROOT.

Fixes: #782


wdconinc:
Is this possible to be thread safe without mutexes? Can we have multiple algorithms potentially running `AlgorithmInit` all at the same time?

wdconinc:
> Why would we run AlgorithmInit in parallel?

Because JANA2, as far as I understand, doesn't just initialize all algorithms at the start, or even one at a time. It initializes on demand, when the first event happens to hit a factory. There can be many factories in flight at that time if we have threads going on.

wdconinc (dismissed):
Looks great!

wdconinc (approved):

wdconinc (approved):

PR #825, opened 2023-08-03
Reverted default Silicon Tracker Digi threshold to 0

kkauder:
SiliconTrackerDigiConfig has a threshold. This should be overwritten by sub-systems and thus by default be 0.
NB: be aware of thresholds in npsim! E.g. https://github.com/eic/npsim/pull/9/files

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] _Comments_ have been added

No

Yes

### DIFF
--- a/src/algorithms/digi/SiliconTrackerDigiConfig.h
+++ b/src/algorithms/digi/SiliconTrackerDigiConfig.h
@@ -6,7 +6,7 @@ namespace eicrecon {
 
   struct SiliconTrackerDigiConfig {
     // sub-systems should overwrite their own
-    // NB: be aware of thresholds in npsim!
+    // NB: be aware of thresholds in npsim! E.g. https://github.com/eic/npsim/pull/9/files
     double threshold  = 0 * dd4hep::keV;
     double timeResolution = 8;   /// TODO 8 of what units??? Same TODO in juggler. Probably [ns]
   };


wdconinc:
Since this effectively changes all tracking digitization thresholds from 5 keV to zero, can we be explicit in e.g. https://github.com/kkauder/EICrecon/blob/digithreshfix/src/detectors/BTRK/BTRK.cc#L26 about what value should be used, even if it is supposed to be zero?

kkauder:
> Since this effectively changes all tracking digitization thresholds from 5 keV to zero, can we be explicit in e.g. https://github.com/kkauder/EICrecon/blob/digithreshfix/src/detectors/BTRK/BTRK.cc#L26 about what value should be used, even if it is supposed to be zero?

In principle yes, _but_ only the following change the value from default:
```
./src/detectors/BTOF/BTOF.cc
./src/detectors/MPGD/MPGD.cc
./src/detectors/FOFFMTRK/FOFFMTRK.cc
./src/detectors/ECTOF/ECTOF.cc
./src/detectors/B0TRK/B0TRK.cc
```
So to do this well, some boilerplate like
```
SiliconTrackerDigiConfig barrel_digi_default_cfg;
    barrel_digi_default_cfg.threshold = 0.25 * dd4hep::keV;
    barrel_digi_default_cfg.timeResolution = 10;
    app->Add(new JChainFactoryGeneratorT<SiliconTrackerDigi_factory>({"MPGDBarrelHits"}, "MPGDBarrelRawHits", barrel_digi_default_cfg));
```
should be added to all appropriate subsystems. I'll review the existing ones for consistency but for now won't add this to others

kkauder:
Only MPGD sets it to the (correct) value of 0.25 keV. All others set it to 0, which is also okay in our current setup.

wdconinc (approved):
Ok. Looks good then.

wdconinc:
Only differences introduced (since we don't save `RawHits`) are in `RecHits`, e.g. https://github.com/eic/EICrecon/actions/runs/5752312080/job/15594048110?pr=825#step:10:2877 addition of `RecHits`.

PR #835, opened 2023-08-07
rename EcalBarrelImagingMergedClusters -> EcalBarrelClusters

veprbl:
A shorter name is more convenient. The previous longer name is, perhaps, slightly confusing to newcomers in that it's Imaging and ScFi clusters that are combined, not Imaging and Imaging.


wdconinc (approved):

src/detectors/BEMC/BEMC.cc
```diff
@@ -155,7 +155,7 @@ extern "C" {
           app   // TODO: Remove me once fixed
         ));
         app->Add(new JChainMultifactoryGeneratorT<TruthEnergyPositionClusterMerger_factoryT>(
-          "EcalBarrelImagingMergedClusters",
+          "EcalBarrelClusters",
```
wdconinc, line 158:
```suggestion
          "EcalBarrelTruthClusters",
```

src/detectors/BEMC/BEMC.cc
```diff
@@ -164,8 +164,8 @@ extern "C" {
             "EcalBarrelImagingClusterAssociations"
           },
           {
-            "EcalBarrelImagingMergedClusters",
-            "EcalBarrelImagingMergedClusterAssociations"
+            "EcalBarrelClusters",
+            "EcalBarrelClusterAssociations"
```
wdconinc, line 168:
```suggestion
            "EcalBarrelTruthClusters",
            "EcalBarrelTruthClusterAssociations"
```

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -109,12 +109,12 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "EcalEndcapPInsertTruthClusterAssociations",
             "EcalEndcapPInsertClusters",
             "EcalEndcapPInsertClusterAssociations",
+            "EcalBarrelClusters",
+            "EcalBarrelClusterAssociations",
```
wdconinc, line 113:
```suggestion
            "EcalBarrelTruthClusters",
            "EcalBarrelTruthClusterAssociations",
```

wdconinc (commented):
Since these are truth-reconstructed quantities, we can also use this as an opportunity to clarify that.

src/detectors/BEMC/BEMC.cc
wdconinc: We can postpone the renaming until #822.

wdconinc (approved):

PR #836, opened 2023-08-08
fix: rename imaging/scfi module to sector

wdconinc:
As of https://github.com/eic/epic/pull/490, the naming of the sectors of the imaging and scfi is now correct in the readout field. This is reqiured because there are now actual modules implemented (i.e. chip modules placed on staves). The hierarchy/nomenclature defined by the DSC is system > sector > layer > stave > module > pixel (x,y).

- [x] Bug fix (issue: the usual race condition when we update things...)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/detectors/BEMC/BEMC.cc
+++ b/src/detectors/BEMC/BEMC.cc
@@ -55,7 +55,7 @@ extern "C" {
             .sampFrac = 0.10200085,
             .readout = "EcalBarrelScFiHits",
             .layerField = "layer",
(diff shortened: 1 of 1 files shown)


sly2j (approved):

veprbl:
Changes like this could use some framework to implement a backwards compatibility, right? I assume, inconsistent readouts should cause incorrect behavior in clustering.

PR #839, opened 2023-08-10
ImagingTopoCluster: avoid storing copies of the hits

veprbl:
A similar change was done to CalorimeterIslandCluster::dfs_search when that was converted to using Collections.


wdconinc (approved):
Nice, thanks! We should check the diffs again after a rebase; right now there are diffs in non-imaging collections.

PR #840, opened 2023-08-10
feat: use podio collections in tracking digi/reco; rm unused interfaces

wdconinc:
This converts SiliconTrackerDigi (used for MPGDs too) and TrackerHitReconstruction to processing podio collections instead of individual hits. It adapts their factories to multifactories, and updates the plugins. It removes now unused algorithm interface definitions (IObjectProducer, etc) and the remaining duplicated and unused TruthTrackSeeding algorithm.

- [x] Bug fix (issue: remove unused algorithms, interfaces, and factories)
- [x] New feature (issue: move tracking digitization and reconstruction to podio collections)
- [ ] Documentation update
- [ ] Other: __

No. This should lead to identical output as before..

No.


wdconinc:
Zero differences on local comparison starting from the same simulation output  `sim_dis_18x275_minQ2=1000_brycecanyon.edm4hep.root`.

I think what's going on in the CI here (which shows diffs) is that the most recent `main` run was on older geometry, before there were changes that affect the simulations in this branch. We may need to run a `main` simulation once a day on a schedule.

bschmookler (approved):

wdconinc:
I didn't introduce any digi thresholds that weren't already applied. Do you see a digi threshold somewhere where there wasn't one before?

kkauder:
I just ran `grep -rl threshold src/detectors` and it has a lot more results then I remember (but maybe I grepped for something more specific before)

wdconinc:
This PR should not have had any effect. In reality it had two minor and one major effect:
- the old TrackerHitReconstruction factory just plainly ignore the specified timeResolution (no threshold on reconstruction): this now causes the RecHits timeError to have the intended value,
- due to a change in how the std::unordered_map is written in digi, we have a different order in which hits are stored,
- this unintentionally killed the TrackerHitCollector that puts all of the hits together in a single collection.

Issue #838, opened 2023-08-09
`{CalorimeterIslandCluster,ImagingTopoCluster}::dfs_group` should not use recursion calls


veprbl:
I'm seeing some crashes `Illegal instruction: 4` on macOS with following dumping in Console -> Crash Reports:
```output
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGILL)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   eicrecon [41046]

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff2059aba2 __semwait_signal + 10
1   libsystem_c.dylib             	0x00007fff2051ac1a nanosleep + 196
2   libc++.1.0.dylib              	0x000000010c77b1ab std::__1::this_thread::sleep_for(std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l> > const&) + 91
3   libJANA.dylib                 	0x000000010a7bf01e void std::__1::this_thread::sleep_for<long long, std::__1::ratio<1l, 1000l> >(std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000l> > const&) + 430
4   libJANA.dylib                 	0x000000010a7bd967 JApplication::Run(bool) + 2231
5   eicrecon                      	0x000000010a702dee jana::Execute(JApplication*, jana::UserOptions&) + 1070 (eicrecon_cli.cpp:389)
6   eicrecon                      	0x000000010a6f63be main + 382 (eicrecon.cc:61)
7   libdyld.dylib                 	0x00007fff205e8f3d start + 1

Thread 1:
0   libclang_rt.asan_osx_dynamic.dylib	0x000000010c833eb0 __sanitizer_annotate_contiguous_container + 544
1   BEMC.so                       	0x000000011737984a std::__1::vector<std::__1::pair<unsigned int, edm4eic::CalorimeterHit const*>, std::__1::allocator<std::__1::pair<unsigned int, edm4eic::CalorimeterHit const*> > >::__annotate_contiguous_container(void const*, void const*, void const*, void const*) const + 26 (vector:864)
[...]

PR #841, opened 2023-08-10
CalorimeterIslandCluster,ImagingTopoCluster: use breadth first search

veprbl:
We are not looking to discover closest or furthest hits, so DFS and BFS are
interchangeable. Using BFS allows to get rid of recursion.

Fixes: #838


veprbl:
I've tested, this produces identical output on CI.

wdconinc (approved):
I am assuming there are no pathologies that cause performance degradation in our use case.

PR #847, opened 2023-08-13
fix: remove unused includes of ParticlesFromTrackFitResult.h

wdconinc:
This removes unused helper includes which are only relied upon by the `tracking_efficiency` benchmark that is likely non-operational since there is no "CentralTrackingParticles" collection in https://github.com/eic/EICrecon/blob/2e4ace8c161b98377280eee7d9df1618c7d8562a/src/benchmarks/reconstruction/tracking_efficiency/TrackingEfficiency_processor.cc#L70.

TODO:
- [x] evaluate if `TrackingEfficiency_processor` should just be rewritten to use the `ParticlesFromTrackFitResultNew` in https://github.com/eic/EICrecon/blob/2e4ace8c161b98377280eee7d9df1618c7d8562a/src/algorithms/tracking/ParticlesFromTrackFit.h#L13, rewrite it, then remove `ParticlesFromTrackFitResult` entirely,
- [ ] run all these benchmarks to make sure they actually work...

- [x] Bug fix (issue: unused includes)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (dismissed):

wdconinc:
`TrackingEfficiency_processor` should not have used `CentralTrackingParticles` but instead pull the `outputParticles`.

wdconinc:
Now removed the unused header entirely.

veprbl (approved):

PR #848, opened 2023-08-14
feat(ci): add job to run each benchmark separately

wdconinc:
This runs all benchmarks in `src/benchmarks/` for one particle gun job.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: see #847)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @veprbl this will probably fail several of the benchmarks; probably should parse list of plugins or determine from build system instead of hard-coding the list.

No.

No.


.github/workflows/linux-eic-shell.yml
```diff
@@ -283,6 +283,46 @@ jobs:
...
+        name: install-${{ matrix.CC }}-eic-shell-Release
+    - uses: actions/download-artifact@v3
+      with:
+        name: sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root
+    - uses: cvmfs-contrib/github-action-cvmfs@v3
+    - name: Run EICrecon
+      uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: /opt/detector/setup.sh
```
veprbl, line 320:
```suggestion
        setup: |
          /opt/detector/setup.sh
```

veprbl:
I don't think EICrecon-based benchmarks are  a good idea. If something can't be done from an EDM4eic output, we are probably doing something wrong.

wdconinc:
> I don't think EICrecon-based benchmarks are a good idea.

I tend to agree. But the case for removing these benchmarks is going to be stronger when they don't even work. Right now I don't know which ones are dead code and which ones are dead code that doesn't even work.

I do think these are not quite the "benchmarks" in the way we are talking about them, but more diagnostics modules to run alongside with the reconstruction. I don't know if separate diagnostics plugins make more sense than just instrumenting the reconstruction directly, but this is what we have...

To me, there seems to be no clear distinction between the benchmarks, examples and tests, and all are dead code which may or may not work, which may or may not be in line with current thinking, and which may or may not actually run (even if it compiles).

.github/workflows/linux-eic-shell.yml
```diff
@@ -283,6 +283,47 @@ jobs:
         path: sim_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root
         if-no-files-found: error
 
+  eicrecon-benchmarks-plugins:
+    runs-on: ubuntu-latest
+    needs:
+    - build
+    - ddsim-gun
+    strategy:
```
wdconinc, line 291: Don't fail on first failure, but continue testing all benchmarks. 
```suggestion
    strategy:
      fail-fast: false
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -284,6 +284,45 @@ jobs:
...
+    - uses: actions/download-artifact@v3
+      with:
+        name: sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root
+    - uses: cvmfs-contrib/github-action-cvmfs@v3
+    - name: Run EICrecon
+      uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: /opt/detector/setup.sh
+        run:
```
wdconinc, line 320:
```suggestion
        run: |
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -285,6 +285,44 @@ jobs:
         path: sim_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root
         if-no-files-found: error
 
+  eicrecon-benchmarks-plugins:
+    runs-on: ubuntu-latest
+    needs:
+    - build
+    - ddsim-gun
+    strategy:
+      fail-fast: false
```
wdconinc, line 294: Now that they all work, change benchmarks back to fail-fast:
```suggestion
```

src/benchmarks/detectors/EcalBarrelScFiCheck/EcalBarrelScFiCheckProcessor.h
```diff
@@ -0,0 +1,33 @@
...
+#include <edm4hep/SimCalorimeterHit.h>
+#include <edm4hep/RawCalorimeterHit.h>
+#include <edm4eic/ProtoCluster.h>
+
+class EcalBarrelScFiCheckProcessor: public JEventProcessorSequentialRoot {
+private:
+
+    // Data objects we will need from JANA e.g.
+    PrefetchT<edm4hep::SimCalorimeterHit> EcalBarrelScFiHits           = {this, "EcalBarrelScFiHits"};
+    PrefetchT<edm4hep::RawCalorimeterHit> EcalBarrelScFiRawhits        = {this, "EcalBarrelScFiRawHits"};
```
veprbl, line 19:
```suggestion
    PrefetchT<edm4hep::RawCalorimeterHit> EcalBarrelScFiRawHits        = {this, "EcalBarrelScFiRawHits"};
```
and elsewhere..

.github/workflows/linux-eic-shell.yml
```diff
@@ -285,6 +285,43 @@ jobs:
...
+      uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: /opt/detector/setup.sh
+        run: |
+          export DETECTOR_CONFIG=${DETECTOR}_${{ matrix.detector_config }}
+          export JANA_PLUGIN_PATH=$PWD/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
+          chmod a+x bin/*
+          $PWD/bin/eicrecon -Pplugins=${{ matrix.benchmark_plugins }} -Ppodio:output_file=rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root -Pplugins=dump_flags,janadot -Pdump_flags:json=${{ matrix.particle }}_${{ matrix.detector_config }}_flags.json -Pjana:warmup_timeout=0 -Pjana:timeout=0
+
```
veprbl, line 324: No artifacts produced?

wdconinc: It wasn't really the intention to start looking at these artifacts, but we could store them. I am not sure if they contain anything actionable, but it is a good test to make sure they are actually written.

veprbl (approved):
Looked at ScFi and LFHCAL artifacts (latter is empty).

.github/workflows/linux-eic-shell.yml
```diff
@@ -288,6 +288,48 @@ jobs:
         path: sim_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root
         if-no-files-found: error
 
+  eicrecon-benchmarks-plugins:
+    runs-on: ubuntu-latest
+    needs:
+    - build
+    - ddsim-gun
```
wdconinc, line 295:
```suggestion
    - npsim-gun
```

wdconinc:
@veprbl Rebased after ddsim -> npsim; modified dependency in the benchmarks testing job.

veprbl (approved):

PR #851, opened 2023-08-14
services/log: remove timestamp from log messages, allow configuring format

veprbl:
During development, it is often needed to diff log files, which can be cumbersome if timestamps are printed. It doesn't seem like we have a strong need to get timestamps in the production logs at the time, but if that will be needed, it can now be re-enabled using eicrecon::LogFormat option.

### DIFF
--- a/src/services/log/Log_service.cc
+++ b/src/services/log/Log_service.cc
@@ -16,6 +16,10 @@ Log_service::Log_service(JApplication *app) {
     m_log_level_str = "info";
     m_application->SetDefaultParameter("eicrecon:LogLevel", m_log_level_str, "log_level: trace, debug, info, warn, error, critical, off");
     spdlog::default_logger()->set_level(eicrecon::ParseLogLevel(m_log_level_str));
+
+    m_log_format_str = "[%n] [%^%l%$] %v";
+    m_application->SetDefaultParameter("eicrecon:LogFormat", m_log_level_str, "spdlog pattern string");
+    spdlog::set_pattern(m_log_format_str);
 }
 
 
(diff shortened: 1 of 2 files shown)


src/services/log/Log_service.cc
```diff
@@ -16,6 +16,10 @@ Log_service::Log_service(JApplication *app) {
     m_log_level_str = "info";
     m_application->SetDefaultParameter("eicrecon:LogLevel", m_log_level_str, "log_level: trace, debug, info, warn, error, critical, off");
     spdlog::default_logger()->set_level(eicrecon::ParseLogLevel(m_log_level_str));
+
+    m_log_format_str = "[%n] [%^%l%$] %v";
+    m_application->SetDefaultParameter("eicrecon:LogFormat", m_log_level_str, "spdlog pattern string");
```
wdconinc, line 21: Wasn't there something about spaces that `SetDefaultParameter` couldn't handle in the adjacency matrix?

veprbl: Good point. I don't see an issue in CI, is that because the default value is unaffected?

wdconinc: Yeah, I think the default is not affected. The `SetDefaultParameter` function doesn't set the default parameter that is passed as the argument, but instead it overwrites the variable (passed by reference) with the non-default value passed on the command line... Maybe `GetNonDefaultParameter` would have been clearer.

wdconinc: Who knows what other characters JANA2 might trip up over (I like my logging like valgrind... `eicrecon -Peicrecon:LogFormat="==%n==\\%10!!"`) but I'd say let's get this in, clean up the logs, and we can flag this again upstream.

wdconinc (approved):

Issue #725, opened 2023-06-30
ParticlesFromTrackFit: needs to create a relation from edm4eic::ReconstructedParticle to edm4eic::Track


veprbl:
This is, in principle, supported by our EDM
https://github.com/eic/EDM4eic/blob/4d9b768bf233db092da06da6b4d8cd541abe6404/edm4eic.yaml#L182-L184
but not implemented:
https://github.com/eic/EICrecon/blob/158e3e0364a95965b249b0ffac16534e8123c6ad/src/algorithms/tracking/ParticlesFromTrackFit.cc#L129C11-L136

(Not clear to me how `edm4eic::Track` and `edm4eic::TrackParameters` relate)

veprbl:
#766 is a step towards this

veprbl:
Needs #716

PR #852, opened 2023-08-14
fix: rm unused output{Seeded}Particles from TrackingResult_factory

wdconinc:
This removes the unused output{Seeded}Particles collection from the tracking plugin, and removes the ReconstructedParticle output from the TrackingResult algorithm and factory.

The construction of ReconstructedParticles (which includes PID information) is not something that should be done at this stage of the track reconstruction chain (and indeed the results are ignored).

The `TrackingEfficiency_processor` was updated to process the `ReconstructedChargedParticles` collection which is (later) produced from tracks.

- [x] Bug fix (issue: unused collection, incorrect use of ReconstructedParticles)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/ParticlesFromTrackFit.cc
+++ b/src/algorithms/tracking/ParticlesFromTrackFit.cc
@@ -29,10 +29,9 @@ void eicrecon::Reco::ParticlesFromTrackFit::init(std::shared_ptr<spdlog::logger>
     m_log = log;
 }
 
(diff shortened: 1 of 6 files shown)


veprbl (approved):
No diff except some collectionID changes.

veprbl:
cc #725

PR #853, opened 2023-08-15
fix: readd license and provenance of ActsExamples files

wdconinc:
Several tracking interfaces originate in the ActsExamples namespace under the Examples directory in Acts. Instead of modifying these (changing filename, directory structure, and namespace) and redistributing them under a different license, it is easier for maintainance to keep them as similar as possible. This restores the original directory structure, file names, and adds a clear license. Future Acts upgrades should start with updating these files.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No. Some namespaces changed, some type names changed, untracked local code may need adjustment.

No.


src/algorithms/tracking/ActsExamples/EventData/GeometryContainers.hpp
```diff
@@ -11,8 +11,8 @@
 #include <boost/container/flat_map.hpp>
 #include <boost/container/flat_set.hpp>
 
-#include "Utilities/GroupBy.hpp"
-#include "Utilities/Range.hpp"
+#include "../Utilities/GroupBy.hpp"
```
wdconinc, line 14: First and only file with ugly relative include. Fixing more fundamentally with include directories is likely possible too, but this primarily affects so-called benchmarks which may disappear anyway.

wdconinc:
Empty diffs, as expected.

src/algorithms/tracking/ActsExamples/EventData/GeometryContainers.hpp
wdconinc: In the ActsExamples build, this is resolved by a globally adding the ActsExample directory to target_include_directories. In Juggler and EICrecon previously, this was resolved by placing Utilities under EventData (where it does not belong or is distributed by Acts). This moves these files in the right place, but we don't want to add the ActsExamples directory as a target_include_directories for all targets.

The relative include could be avoided, *in a future PR*, by providing the ActsExamples target_include_directory to the targets that link against libalgoriths_tracking through its public interface.

wdconinc:
Diffing with
```console
find src/algorithms/tracking/ActsExamples -name *.hpp | while read file ; do diff -w -y $(find . ~/git/acts/Examples/ -name $(basename $file)) ; done 
```
indicates some minor changes (even with Acts main right now).

veprbl (approved):
I agree with this approach. Reducing the diff to Acts upstream will help with maintenance, I was thinking about a similar thing myself. Keeping ActsExamples namespace is a bold, but mature move. I've checked, the current diff is indeed still quite large.

PR #856, opened 2023-08-15
CalorimeterHitDigi: don't add a hit when none of the merged hits meet the timing requirement

veprbl:
There was an edge case when all hits could be skipped in the previous loop and then the algorithm would still proceed to add a hit with edep=0. Most common case to trigger that, would be if there is only a single hit (e.g. if we were to use this code path for calorimeters for which the hit merging is not required).

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ ] Changes have been communicated to collaborators @FriederikeBock @steinber @rymilton

No

No

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -190,6 +190,7 @@ std::unique_ptr<edm4hep::RawCalorimeterHitCollection> CalorimeterHitDigi::signal
                 }
             }
         }
+        if (time > m_cfg.capTime) continue;
 
         // safety check
         const double eResRel = (edep > m_cfg.threshold)


wdconinc:
Not just an edge case, but quite common. Here is 5x41 brycecanyon Q2>1 EcalBarrelScFiRawHits.amplitude diff:
```
  [[0, 2281, 4091, 0, 0, 0, 0, ..., 18815, 4014, 7974, 0, 7257, 519, 22256], ...]
  [[2281, 4091, 1949, 1955, 7243, 7250, ..., 4014, 7974, 7257, 519, 22256], ...]
```

wdconinc (approved):

veprbl:
You can't test this on CI right now because npsim is broken in the container: https://chat.epic-eic.org/main/pl/szkezeezwjnb3fiaiihg8ze87o

veprbl:
Nevermind, I looked at the wrong thing.

veprbl:
I don't see statistically significant discrepancies beyond:
image
image

wdconinc:
Yeah, this is consistent. As https://github.com/eic/EICrecon/actions/runs/5980650859/job/16227488637?pr=856#step:10:7516 indicates, only raw hits are removed (and they do have amplitude 0). In island clustering a hit can connect two otherwise disconnected clusters, so we can have more clusters due to this.

wdconinc:
I think we can merge this, if you're satisfied with it.

PR #859, opened 2023-08-16
fix: mv init TrackParameters edm4eic to ActsExamples conversion into algorithm

wdconinc:
### Briefly, what does this PR introduce?
This moves the conversion of the initial track parameters into ActsExamples structures from the CKFTracking factory into the algorithm, for improved modularity and algorithm reuse. No other changes. Diffs should come back clean.

### What kind of change does this PR introduce?
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

Issue #687, opened 2023-06-01
CKF assumes one track per seed


osbornjd:
Currently the CKF takes the output and puts the information directly into an `Acts::MultiTrajectory` type object, which an additional algorithm `ParticlesFromTrackFit` reads from and translates into an `edm4eic` object. As discussed in the track reconstruction meeting today, the interpretation currently assumes that the CKF only finds one branch per input track seed as shown [here](https://github.com/eic/EICrecon/blob/4620036e6b3940563ba0de8ee12ad851d1b9923c/src/algorithms/tracking/ParticlesFromTrackFit.cc#L52) and [here](https://github.com/eic/EICrecon/blob/4620036e6b3940563ba0de8ee12ad851d1b9923c/src/algorithms/tracking/ParticlesFromTrackFit.cc#L61).  Currently we are basically taking the longest track branch per track seed as "the track", which is probably a very good assumption in no background environments. This likely does not matter for tracking without background as there are probably very few cases where multiple branches per input track seed are found. However, this will (my suspicion is) be a very bad assumption once background is implemented. We should consider modifying this to including all branches for a given track seed, which may then require some algorithm to identify which branch is the best and should be taken as "the track fit" associated to a given track seed.

osbornjd:
Related to this, what is [this](https://github.com/eic/EICrecon/blob/4620036e6b3940563ba0de8ee12ad851d1b9923c/src/algorithms/tracking/ParticlesFromTrackFit.cc#L110-L117) piece of code doing? It appears to be taking all of the found branches from the CKF and then iterating over them as if they are track states, then setting the reconstructed particle momentum and charge to be that of some track state?

PR #860, opened 2023-08-17
feat: produce edm4eic::TrajectoryCollection in CKFTracking 

wdconinc:
This moves the creation of podio output types into the CKFTracking algorithms (it should accept and produce only podio collections), with the assumptions from the previous ParticlesFromTrackFit algorithm. No changes expected (with the exception of changed collection IDs due to the addition of a new output collection).

The CKFTracking algorithm now produces both the desired podio output type and the deprecated ActsExamples::Trajectories, which is still used by some downstream algorithms. Next is to migrate those algorithms over.

The TrackingResult factory and ParticlesFromTrackFit algorithms are now obsolete and removed.

TODO:
- [x] merge #859 first, it is included here

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/algorithms/tracking/CKFTracking.cc
```diff
@@ -123,24 +159,92 @@ namespace eicrecon {
...
+
+                // Collect the trajectory summary info
+                auto trajectoryState = Acts::MultiTrajectoryHelpers::trajectoryState(mj, trackTip);
+
+                // Create trajectory
+                auto trajectory = trajectories->create();
+
+                trajectory.setChi2(trajectoryState.chi2Sum / trajectoryState.NDF);
+                trajectory.setNdf(trajectoryState.NDF);
+
```
bschmookler, line 199: We should add 

`trajectory.nMeasurements(trajectoryState.nMeasurements);`

etc...

bschmookler: Based on the description here ( https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L337), we should do
`
trajectory.setChi2(trajectoryState.chi2Sum);`

wdconinc: Yeah, I was following the EDM4hep plugin convention and didn't know we had picked the opposite convention already.

src/global/tracking/tracking.cc
```diff
@@ -76,43 +76,32 @@ void InitPlugin(JApplication *app) {
...
-            {"CentralCKFTrajectories"},                       // ActsExamples::Trajectories
-            {"outputTrackParameters"},                        // edm4eic::TrackParameters
-            app));
+            {"CentralCKFActsTrajectories"}, "CentralTrackVertices"));
 
     app->Add(new JChainMultifactoryGeneratorT<ParticlesWithTruthPID_factory>(
             "ChargedParticlesWithAssociations",                // Tag name for multifactory
             {"MCParticles",                                    // edm4hep::MCParticle
-            "outputTrackParameters"},                          // edm4eic::TrackParameters
+            "CentralCKFActsTrajectories"},                          // edm4eic::TrackParameters
```
bschmookler, line 94: I don't understand this part. `CentralCKFActsTrajectories ` is a vector of `ActsExamples::Trajectories*` (ACTS multitrajectory). So, I assume we need to rewrite to convert it `edm4eic::TrackParameters` type before passing it to this factory.

bschmookler: ?

wdconinc: Just an incorrect comment. It needs Acts trajectories.

wdconinc: Ok, I see what you mean. Let me revisit and clarify. There is indeed a part missing. The track parameters are linked in the trajectories (both acts and podio) and the PID matching should use the podio trajectories as a start.

src/algorithms/tracking/CKFTracking.cc
```diff
@@ -158,17 +166,85 @@ namespace eicrecon {
...
+                const auto& trackTip = trackTips.front();
+
+                // Collect the trajectory summary info
+                auto trajectoryState = Acts::MultiTrajectoryHelpers::trajectoryState(mj, trackTip);
+
+                // Create trajectory
+                auto trajectory = trajectories->create();
+
+                trajectory.setChi2(trajectoryState.chi2Sum / trajectoryState.NDF);
+                trajectory.setNdf(trajectoryState.NDF);
```
wdconinc, line 198:
```suggestion
                trajectory.setChi2(trajectoryState.chi2Sum);
                trajectory.setNdf(trajectoryState.NDF);
                trajectory.setNMeasurements(trajectoryState.nMeasurements);
```

src/global/tracking/tracking.cc
wdconinc: Resolved in https://github.com/eic/EICrecon/pull/860/commits/0406b6523d7f1609894b71611b965b1d6027d6a5.

wdconinc:
No differences in outputs, except for an additional collection which changes 20 collectionIDs:
- brycecanyon: [diff](https://github.com/eic/EICrecon/actions/runs/6003189947/job/16282598058?pr=860#step:11:72)
- craterlake: [diff](https://github.com/eic/EICrecon/actions/runs/6003189947/job/16282598232?pr=860#step:11:72)

You'll want to download the raw logs and search for `NOT EQUAL`.

bschmookler:
I see that the edm4eic::trackparameters get added into the trajectory here: https://github.com/eic/EICrecon/blob/CKFTracking-podio-outputs/src/algorithms/tracking/CKFTracking.cc#L209

And they now get used to make the ReconstructedChargedParticles and associations here: https://github.com/eic/EICrecon/blob/CKFTracking-podio-outputs/src/algorithms/tracking/ParticlesWithTruthPID.cpp#L33

Will these track parameters also be written out to the EICRecon output file with [this](https://github.com/eic/EDM4eic/blob/b46ba9a94d2c1fd4ef722c23348d87a51923e3fe/edm4eic.yaml#L348) data type?

This can be useful for comparison to the seed outputs (which are saved as edm4eic::trackparameters). For example, we can compare the Acts Loca and Locb values at the seed and track output level.

wdconinc:
> Will these track parameters also be written out to the EICRecon output file with [this](https://github.com/eic/EDM4eic/blob/b46ba9a94d2c1fd4ef722c23348d87a51923e3fe/edm4eic.yaml#L348) data type?

Yes, that's exactly the benefit (or one of them) of this approach. It makes the entire data structure much more transparent instead of locking this up in ACTS structures that we cannot save to the ROOT files.

bschmookler:
Sorry, I'm missing one thing. I see that the trajectory written out here:

https://github.com/eic/EICrecon/blob/CKFTracking-podio-outputs/src/global/tracking/CKFTracking_factory.cc#L81

The trajectory has a onetomany association with the track parameters. But I don't see a factory where edm4eic::TrackParameters gets saved.

wdconinc:
> Sorry, I'm missing one thing.

It was indeed missing one thing ;-) Now it writes the track parameters collections too.

wdconinc:
Since you asked for this ;-)

image
image

bschmookler (dismissed):

bschmookler (approved):

src/algorithms/tracking/ParticlesWithTruthPID.cpp
simonge, line 1: Is there a future motivation for changing this to take the TrajectoryCollection rather than TrackParametersCollection directly? Unless different trajectories can link to the same trackparameter, or a trackparameter in the collection isn't associated with a trajectory, the functionality is unchanged.

wdconinc: One (multi)trajectory based on one seed can have multiple track parameters (happens rarely without background). It will be up to the algorithm creating particles from trajectories to treat this appropriately, for example by selecting the best track parameters in a trajectory, using all of them, or something in between. Passing the trajectory instead of a flat list of parameters allows us to make that decision with more information. See https://github.com/eic/EICrecon/issues/687 as well.

simonge: So for a alternative future implementation of this algorithm? This has just made injecting non-acts tracking into the workflow a little more complicated, I'm yet to understand if the trajectory datatype makes sense for this.

wdconinc: That's for future refinements in the algorithms. Ultimately the ParticlesWithTruthPID algorithm should not even take a Trajectory but a Track. But that's also a future refinement.

src/algorithms/tracking/CKFTracking.cc
```diff
@@ -158,17 +169,99 @@ namespace eicrecon {
...
+                            static_cast<float>(covariance(Acts::eBoundTheta, Acts::eBoundPhi)),
+                            static_cast<float>(covariance(Acts::eBoundTheta, Acts::eBoundQOverP)),
+                            static_cast<float>(covariance(Acts::eBoundPhi, Acts::eBoundQOverP))
+                        },
+                        static_cast<float>(parameter[Acts::eBoundTime]),
+                        sqrt(static_cast<float>(covariance(Acts::eBoundTime, Acts::eBoundTime))),
+                        static_cast<float>(boundParam.charge())};
+
+                    track_parameters->push_back(pars);
+                    trajectory.addToTrackParameters(pars);
```
veprbl, line 251: Fun fact: This is not compatible with EDM4eic 2.0.0. I suppose, it's not practical to support it, but want to point it out.

wdconinc: Right, I forgot to update the minimum. The whole reason for 2.1.0 was to allow this...

wdconinc: https://github.com/eic/EICrecon/pull/885

Issue #480, opened 2023-02-13
Writing out track propagation information


bschmookler:
**Is your feature request related to a problem? Please describe.**
The [track propagation algorithm](https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackPropagation.cc) currently writes out the track projection as a **edm4eic::TrackPoint** data type. This cannot be saved to the EICRecon output ROOT tree. Once that is written out, we also need to associate the projection with the correct track in the output file.

**Describe the solution you'd like**
The output of the track propagation algorithm can be saved as a [track segment](https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L371), which can be written out. The track can also be associated in this data type. 

**Describe alternatives you've considered**
Right now, we can access this track propagation information using a Plugin, like [this one](https://github.com/eic/EICrecon/blob/EEMCProj_Proc/src/tests/EEMCProjPlugin/EemcTrkPropagation_processor.cc).

**Additional context**
None.

PR #862, opened 2023-08-17
Create TrackPropagation_factory

tylerkutz:
This PR introduces a TrackPropagation_factory that uses the TrackPropagation algorithm to project tracks to calorimeters.

Propagated track points are stored in a TrackSegmentCollection that is saved in the PODIO output.

Have not yet added association between each edm4eic::TrackSegment and the underlying edm4eic::Track.

- [X] Bug fix (issue #480)
- [X] New feature (issue #480)
- [ ] Documentation update
- [ ] Other: __

Requires new int32_t member to edm4eic::TrackPoint to store detector/surface ID (int32_t) (see https://github.com/eic/EDM4eic/pull/41).


src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,183 @@
...
+	for(auto traj: trajectories) {
+		edm4eic::MutableTrackSegment this_propagated_track;
+		for(unsigned short isurf = 0; auto surf: m_target_surface_list) {
+			try {
+				auto prop_point = m_track_propagation_algo.propagate(traj, surf);
+				if(!prop_point) continue;
+				prop_point->surface = m_target_surface_ID[isurf];
+				this_propagated_track.addToPoints(*prop_point);
+			} catch (std::exception& e) {
+				m_log->warn("Exception in underlying algorithm: {}. Trajectory is skipped", e.what());
```
veprbl, line 56: Is this branch actually exercised? If not, please remove in case we have exceptions that we want to catch. Blanket-ignoring is a practice that occurs in the existing code, but we are looking to move away from it.

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,183 @@
...
+	const double EHCAL_MinR = 0.0;
+	const double EHCAL_MaxR = (m_geoSvc->detector()->constant<double>("HcalEndcapN_rmax") / dd4hep::mm) * extend;
+	auto EHCAL_Bounds       = std::make_shared<Acts::RadialBounds>(EHCAL_MinR, EHCAL_MaxR);
+	auto EHCAL_Trf          = transform * Acts::Translation3(Acts::Vector3(0, 0, EHCAL_Z));
+	m_EHCAL_prop_surface  = Acts::Surface::makeShared<Acts::DiscSurface>(EHCAL_Trf, EHCAL_Bounds);
+	m_target_surface_list.push_back(m_EHCAL_prop_surface);
+	m_target_surface_ID.push_back(m_geoSvc->detector()->constant<int32_t>("HCalEndcapN_ID"));
+
+	/*
+	std::cout << std::endl;
```
veprbl, line 143: Please don't add commented code. Either convert it to logger debug messages or remove entirely.

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,183 @@
...
+
+	auto trajectories = event->Get<eicrecon::TrackingResultTrajectory>(m_input_tag);
+
+	edm4eic::TrackSegmentCollection propagated_tracks;
+
+	for(auto traj: trajectories) {
+		edm4eic::MutableTrackSegment this_propagated_track;
+		for(unsigned short isurf = 0; auto surf: m_target_surface_list) {
+			try {
+				auto prop_point = m_track_propagation_algo.propagate(traj, surf);
```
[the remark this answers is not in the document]
tylerkutz: Yes, my understanding is this is totally harmless but does get cumbersome in the output.  I will check the settings on the EigenStepper and see if I can suppress these messages.

src/global/tracking/TrackPropagation_factory.cc
tylerkutz: Yeah I copied this from another factory but if we're moving away from it I will take it out.

src/global/tracking/TrackPropagation_factory.cc
tylerkutz: Sure, I think this is useful to have in so I will make it a debug message.

src/global/tracking/TrackPropagation_factory.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <edm4eic/ReconstructedParticle.h>
+#include <edm4eic/TrackSegmentCollection.h>
+
+#include <Acts/Surfaces/DiscSurface.hpp>
+#include <Acts/Surfaces/CylinderSurface.hpp>
+
+namespace eicrecon {
+
+	class TrackPropagation_factory : public JFactoryPodioT<edm4eic::TrackSegment>,
```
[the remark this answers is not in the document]
tylerkutz: Your comment and suggested change do not seem consistent, is a `JChainFactoryT` the same as a `JChainMultifactory`?

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -290,8 +290,9 @@ namespace eicrecon {
         m_log->trace("    loc err = {:.4f}", static_cast<float>(covariance(Acts::eBoundLoc1, Acts::eBoundLoc1)));
         m_log->trace("    loc err = {:.4f}", static_cast<float>(covariance(Acts::eBoundLoc0, Acts::eBoundLoc1)));
 
+        int32_t surfaceID = 0; // default value...will be set in TrackPropagation factory
```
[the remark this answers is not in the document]
tylerkutz: I'm not sure this adds anything meaningful.  In the current implementation, `targetSurf` has just been created in the track propagation factory (and has not had a geometry ID set).  Even if a geometry ID is automatically assigned, it's not useful for this application so there's really nothing to be gained by (temporarily) setting the value this way.

tylerkutz:
@veprbl Thanks for all these suggestions, I've just gone through and implemented them in my local repo (except for the two comments that I've responded to above).  Once those are resolved I'll test it and commit the updates.

src/algorithms/tracking/TrackPropagation.cc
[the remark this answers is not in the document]
veprbl: Your factory is not the only user of this algorithm, projectee DRICH surfaces should have the ID assigned. Your factory can be then also changed to assign meaningful ACTS geometry object IDs instead of overwriting them in each EDM object.

src/global/tracking/TrackPropagation_factory.h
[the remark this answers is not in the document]
veprbl: Both are good. The `JChainMultifactory` allows multiple output collections which you may not need. However we are looking to unify around `JChainMultifactory` until/if its successors (like #817) come.

**tl/dr**: please, target `JChainMultiactory`.

tylerkutz:
@veprbl Just pushed some commits that address your comments.

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -208,8 +208,8 @@ namespace eicrecon {
         using Propagator = Acts::Propagator<Stepper>;
         Stepper stepper(magneticField);
         Propagator propagator(stepper);
-        // Acts::Logging::Level logLevel = Acts::Logging::FATAL
-        Acts::Logging::Level logLevel = Acts::Logging::INFO;
+        Acts::Logging::Level logLevel = Acts::Logging::FATAL;
+        //Acts::Logging::Level logLevel = Acts::Logging::INFO;
```
veprbl, line 212:
```suggestion
```

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,194 @@
...
+    double ECAL_avgClusterDepth = 50.;	  // mm
+    double HCAL_avgClusterDepth = 150.;   // mm
+
+    // extend surfaces by ten percent for tracks close to the edge
+    // extend in r for endcaps, extend in z for barrel
+    double extend = 1.1;
+
+    // Create propagation surface for BEMC
+    std::shared_ptr<Acts::CylinderSurface> m_BEMC_prop_surface1;
+    std::shared_ptr<Acts::CylinderSurface> m_BEMC_prop_surface2;
```
veprbl, line 75:
```suggestion
```

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,194 @@
...
+    const double BEMC_R     = m_geoSvc->detector()->constant<double>("EcalBarrel_rmin") / dd4hep::mm;
+    const double BEMC_halfz = (std::max(m_geoSvc->detector()->constant<double>("EcalBarrelBackward_zmax"), m_geoSvc->detector()->constant<double>("EcalBarrelForward_zmax")) / dd4hep::mm) * extend;
+    auto BEMC_Trf         = transform * Acts::Translation3(Acts::Vector3(0, 0, 0));
+    m_BEMC_prop_surface1  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R, BEMC_halfz);
+    m_BEMC_prop_surface2  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R + ECAL_avgClusterDepth, BEMC_halfz);
```
veprbl, line 80: Type is evident from the RHS
```suggestion
    auto m_BEMC_prop_surface1  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R, BEMC_halfz);
    auto m_BEMC_prop_surface2  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R + ECAL_avgClusterDepth, BEMC_halfz);
```

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,194 @@
...
+
+    // extend surfaces by ten percent for tracks close to the edge
+    // extend in r for endcaps, extend in z for barrel
+    double extend = 1.1;
+
+    // Create propagation surface for BEMC
+    std::shared_ptr<Acts::CylinderSurface> m_BEMC_prop_surface1;
+    std::shared_ptr<Acts::CylinderSurface> m_BEMC_prop_surface2;
+    const double BEMC_R     = m_geoSvc->detector()->constant<double>("EcalBarrel_rmin") / dd4hep::mm;
+    const double BEMC_halfz = (std::max(m_geoSvc->detector()->constant<double>("EcalBarrelBackward_zmax"), m_geoSvc->detector()->constant<double>("EcalBarrelForward_zmax")) / dd4hep::mm) * extend;
```
veprbl, line 77:
```suggestion
    const double BEMC_R     = m_geoSvc->detector()->constant<double>("EcalBarrel_rmin") / dd4hep::mm * Acts::UnitConstants::mm;
    const double BEMC_halfz = (std::max(m_geoSvc->detector()->constant<double>("EcalBarrelBackward_zmax"), m_geoSvc->detector()->constant<double>("EcalBarrelForward_zmax")) / dd4hep::mm * Acts::UnitConstants::mm) * extend;
```

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -88,6 +88,9 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "ReconstructedJets",
             "ReconstructedElectrons",
 
+	    // Tracks propagated to calorimeters
+	    "PropagatedTrackPoints",
```
veprbl, line 92: You would not need a comment if the name was self-explanatory. Maybe, "TrackCalorimeterProjections"?

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,194 @@
...
+#include <JANA/JEvent.h>
+#include <algorithms/tracking/JugTrack/TrackingResultTrajectory.hpp>
+#include <services/geometry/acts/ACTSGeo_service.h>
+
+#include <Acts/EventData/MultiTrajectoryHelpers.hpp>
+#include <Acts/Surfaces/RadialBounds.hpp>
+
+#include <extensions/spdlog/SpdlogExtensions.h>
+#include <services/log/Log_service.h>
+#include <spdlog/fmt/ostr.h>
```
veprbl, line 15: Probably not needed
```suggestion
```

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -290,8 +290,10 @@ namespace eicrecon {
         m_log->trace("    loc err = {:.4f}", static_cast<float>(covariance(Acts::eBoundLoc1, Acts::eBoundLoc1)));
         m_log->trace("    loc err = {:.4f}", static_cast<float>(covariance(Acts::eBoundLoc0, Acts::eBoundLoc1)));
 
+	uint16_t detectorID = 0; // default value...will be set in TrackPropagation factory
+        uint16_t surfaceID = targetSurf->geometryId().value();
```
veprbl, line 294: Not clear to me where the uint16_t comes from.
```suggestion
	uint8_t system = 0; // default value...will be set in TrackPropagation factory
        uint64_t surface = targetSurf->geometryId().value();
```

veprbl (commented):
This looks quite nice already

src/algorithms/tracking/TrackPropagation.cc
tylerkutz: This was an option proposed in last week's S&C meeting.  As discussed in the meeting I summarized the proposed solutions in slides and gave Daniel's and my recommendations for how to move forward (see discussion [here](https://github.com/eic/EDM4eic/pull/41).  Nobody gave any comments or input so I proceeded with our recommendation.

veprbl: Alignment doesn't matter: fields can be re-ordered to remove padding, unaligned access has no performance penalty on any applicable CPU.

tylerkutz: I'm not sure why this feedback is coming now on this PR, instead of two weeks ago on the EDM4eic PR?  It was clearly stated in the summary of the S&C meeting that I would proceed with uint16_t unless others preferred something else.

Anyways, do you prefer uint64_t for the surface field?  Please let me know ASAP so that I can change this and the EDM4eic PR.

tylerkutz:
@veprbl @jdbrice 

Just pushed the corresponding changes to this branch.  These commits implement the final changes to the data model, along with some miscellaneous edits based on Dmitry's comments above.

tylerkutz:
Just pushed updates to the track propagation algorithms and factory to reflect the changes requested to `edm4eic::TrackPoint`

This should be ready to approve/merge.

jdbrice:
Thank you @tylerkutz @veprbl @wdconinc 
Can we make sure this is merged in time to be included in addition to the changes to EDM4eic (I see it is now merged, thanks!)

wdconinc:
This hasn't even run a successful pipeline. Let's not get ahead of ourselves.

jdbrice:
@wdconinc Of course, you are right. These need edm4eic to build successfully. If they do not after that we can address it ASAP. I Just meant to remind that these need to be in for the Sept campaign in addition to the edm4eic changes.

wdconinc:
That's what we would all like, yes.

wdconinc:
Here's the first problem already. As we have pointed out, changing the data model is invasive. I just deleted the EDM4eic v2.2.0 version again and released v3.0.0 instead because this really is an interface breaking release with a major version change. That is demonstrated by the fact that neither juggler nor EICrecon compile with the new version before this change.

So, no, we cannot have a version of eic-shell ready by tomorrow to run the CI pipelines for this PR, not until the current main branch of EICrecon (and of juggler) can accommodate both version v2.2.0 and v3.0.0. That's a job for a different PR. I had assumed that this would not be an issue. Right now we can't even build a version of eic-shell that includes EDM4eic v3.0.0.

Cc: @sly2j @mdiefent

tylerkutz:
Was it not expected that EICrecon would also need to be updated?  In order to compile with the change to EDM4eic I had to update all the places where TrackPoint was used (only two places, including where I am using it).  How would EICrecon be able to support multiple definitions of TrackPoint?

src/algorithms/tracking/TrackProjector.cc
```diff
@@ -143,8 +143,13 @@ namespace eicrecon {
...
                 const float pathLengthError = 0;
 
-                // Store track point
+                uint64_t surface = trackstate.referenceSurface().geometryId().value();
+	        uint32_t system = 0;
+
+		// Store track point
                 track_segment.addToPoints({
+				                  surface,
+						  system,
```
wdconinc, line 152: @jdbrice @tylerkutz This is the change you will want to make in main first, through a separate PR, to allow upgrading the data model in a backward and forward compatible way. These two extra fields in the initializer list are the issue. Changing to explicit setters is likely sufficient.

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -305,6 +306,8 @@ namespace eicrecon {
           float pathlengthError{}; ///< Error on the pathlenght
          */
         return std::make_unique<edm4eic::TrackPoint>(edm4eic::TrackPoint{
+                                               surface,
+                                               system,
```
wdconinc, line 310: And the same applies here.

jdbrice: @wdconinc thank you
I will try to setup the PR now. I'll test the changes with an updated (local) EDM4eic and this change against main.

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -0,0 +1,183 @@
...
+
+    m_geoSvc = app->template GetService<JDD4hep_service>();
+
+    SetPropagationSurfaces();
+
+}
+
+void eicrecon::TrackPropagation_factory::Process(const std::shared_ptr<const JEvent> &event) {
+
+    auto trajectories = event->Get<eicrecon::TrackingResultTrajectory>(GetInputTags()[0]);
```
wdconinc, line 37: This is no the right type (anymore, ref #860) for the `CentralCKFTrajectories` passed to this factory, and it will fail compilation.

Either change this here and in the algorithm to use `edm4eic::Trajectory` (preferred) or change the collection to `CentralCKFActsTrajectories` (which will be remove in the future). The `edm4eic::Trajectory` collection is stored in the output ROOT file and can be used for associations, whereas the `eicrecon::TrackingResultTrajectory` collection does not get saved and cannot be used for associations.

tylerkutz: I agree that we should use edm4eic::Trajectory since it allows associations.  I'll change the algorithm and factory to use this.

Just to clarify though, it looks like TrackSegment is still associated with `edm4eic::Track`, not `edm4eic::Trajectory`.  So what would be the appropriate way to add the association to the trajectory?

wdconinc: I don't think all pieces are in place yet to do the associations, so I'm not suggesting that the associations get added now. In particular, yes, we need to insert `edm4eic::Track` here, based on `edm4eic::Trajectory`, and some algorithms will only need to receive the tracks, not the trajectories.

veprbl:
Seems to work:
image
image

wdconinc (approved):

veprbl:
To summarize:
- This now has a temporarily disabled access to ACTS geometry object id. There are some issues with it. I did look at the ACTS code, but could not understand how the issue arises. I will file an issue for that.
- The migration to new tracking collections (https://github.com/eic/EICrecon/pull/862#discussion_r1313180169) is left for a follow up PR.

cc @ShujieL @ruse-traveler

PR #863, opened 2023-08-17
ci: don't fail when if_no_artifact_found: warn

wdconinc:
We set `if_no_artifact_found: warn` on the action-download-artifact (when the default is `fail`), but then we fail anyway because the next step can't find the file. If we don't want to fail, we shouldn't run the next step.

TODO:
- [x] apply to all

- [x] Bug fix (issue: https://github.com/eic/EICrecon/actions/runs/5893882391/job/15987488862)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -341,12 +341,12 @@ jobs:
         name: rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
         if_no_artifact_found: warn
     - uses: actions/checkout@v3
-      if: steps.download_previous_artifact.outputs.found_artifact == true
+      if: steps.download_previous_artifact.outputs.found_artifact == 'true'
       with:
         repository: 'eic/epic-capybara'
         path: epic-capybara
     - uses: eic/run-cvmfs-osg-eic-shell@main
-      if: steps.download_previous_artifact.outputs.found_artifact == true
+      if: steps.download_previous_artifact.outputs.found_artifact == 'true'
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -332,17 +332,21 @@ jobs:
...
+    - name: Download previous artifact
+      id: download_previous_artifact
+      uses: dawidd6/action-download-artifact@v2
       with:
         branch: ${{ github.event.pull_request.base.ref || github.ref_name }}
         path: ref/
         name: rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
         if_no_artifact_found: warn
     - uses: actions/checkout@v3
+      if: steps.download_previous_artifact.outputs.found_artifact == true
```
wdconinc, line 344: The booleans are a lie.
```suggestion
      if: steps.download_previous_artifact.outputs.found_artifact == 'true'
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -332,17 +332,21 @@ jobs:
...
       with:
         repository: 'eic/epic-capybara'
         path: epic-capybara
     - uses: eic/run-cvmfs-osg-eic-shell@main
+      if: steps.download_previous_artifact.outputs.found_artifact == true
```
wdconinc, line 349: This one too.
```suggestion
      if: steps.download_previous_artifact.outputs.found_artifact == 'true'
```

veprbl (approved):

PR #864, opened 2023-08-18
CalorimeterHitDigi: unified code path for digitization with and without the hit merging

veprbl:
This should be easier to maintain. We probably could even refactor merging into
a separate algorithm, if we wanted to.

A side effect of this is that non-merging digitization now reorders hits (the
cell id is used as a key for an std::unordered_map).

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @Chao1009 

No

Yes, the hits will be reordered.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -59,7 +59,7 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
     tRes       = m_cfg.tRes / dd4hep::ns;
     stepTDC    = dd4hep::ns / m_cfg.resolutionTDC;
 
-    id_mask = 0;
+    decltype(id_mask) id_inverse_mask = 0;
     // all these are for signal sum at digitization level
     if (!m_cfg.fields.empty()) {
         // sanity checks
@@ -76,7 +76,7 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
         try {
             auto id_desc = m_detector->readout(m_cfg.readout).idSpec();
             for (auto & field : m_cfg.fields) {
-                id_mask |= id_desc.field(field)->mask();
+                id_inverse_mask |= id_desc.field(field)->mask();
             }
         } catch (...) {
             // a workaround to avoid breaking the whole analysis if a field is not in some configurations
@@ -87,7 +87,7 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
         }
         m_log->info("ID mask in {:s}: {:#064b}", m_cfg.readout, id_mask);
     }
-    id_mask = ~id_mask;
+    id_mask = ~id_inverse_mask;
 }
 
 
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -59,8 +59,8 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
     tRes       = m_cfg.tRes / dd4hep::ns;
     stepTDC    = dd4hep::ns / m_cfg.resolutionTDC;
 
+    id_mask = 0;
     // all these are for signal sum at digitization level
-    merge_hits = false;
     if (!m_cfg.fields.empty()) {
         // sanity checks
         if (!m_detector) {
@@ -75,7 +75,6 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
         // get decoders
         try {
             auto id_desc = m_detector->readout(m_cfg.readout).idSpec();
-            id_mask = 0;
             for (auto & field : m_cfg.fields) {
                 id_mask |= id_desc.field(field)->mask();
             }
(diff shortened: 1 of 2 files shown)


veprbl:
This includes #856 to impose timing cut on hits that were previously routed through (now removed) non-merging path.

Chao1009:
Refactoring a separate merging algorithm should be better, but we need to figure out how to pass the information from merging to digitization.

Current workflow:
`sim_hits (input) -> [merging -> digitization] -> digi_hits (output)`
Make merging a separate algorithm:
`sim_hits (input) > merging -> merged_info (intermediate hits) -> digitization -> digi_hits (output)`

"merged_info" could be the merged simulation hits, but this may rewrite a bunch of redundant information. It can also be some collection like the protocluster, which needs a modification to the data model.

veprbl:
(That discussion beyond this PR) For the outputs of this envisioned factory that could take on hit energy deposition summation, can't use the same `edm4hep::RawCalorimeterHit` as for input?

Chao1009:
Yes, the discussion is not for this PR : )
For the question regarding the `edm4hep::RawCalorimeterHit` (digi_hits), do you mean:
`sim_hits (input) > digitization -> digi_hits (intermediate) -> merging -> digi_hits (output)?`

src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -59,8 +59,8 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
     tRes       = m_cfg.tRes / dd4hep::ns;
     stepTDC    = dd4hep::ns / m_cfg.resolutionTDC;
 
+    id_mask = 0;
```
wdconinc, line 62: Can we at this point call this inverse mask with a docstring as to what the set bits are?

src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -86,64 +85,13 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
             // throw::runtime_error(fmt::format("Failed to load ID decoder for {}", m_cfg.readout));
             return;
         }
-        id_mask = ~id_mask;
         m_log->info("ID mask in {:s}: {:#064b}", m_cfg.readout, id_mask);
-        // all checks passed
-        merge_hits = true;
     }
+    id_mask = ~id_mask;
```
wdconinc, line 90: And then here it actually becomes the mask from the inverted mask.

wdconinc (approved):
This looks great from a code simplification point of view. Let's keep it in that scope.

It's a bit hard to interpret the diffs due to reordering. Can you copy-paste some diffs on the reconstructed clusters into the comments? As far as I can tell there shouldn't be any, right?

src/algorithms/calorimetry/CalorimeterHitDigi.cc
veprbl: Fixed

wdconinc (approved):

veprbl:
Let me comment on how this was tested. I apply a following change on top of this PR preserve the ordering:
```diff
diff --git a/src/algorithms/calorimetry/CalorimeterHitDigi.cc b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
index 74e30f92..924b8432 100644
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -97,6 +97,7 @@ std::unique_ptr<edm4hep::RawCalorimeterHitCollection> CalorimeterHitDigi::proces
     // find the hits that belong to the same group (for merging)
     std::unordered_map<uint64_t, std::vector<std::size_t>> merge_map;
     std::size_t ix = 0;
+    std::vector<uint64_t> hids;
     for (const auto &ahit : simhits) {
         uint64_t hid = ahit.getCellID() & id_mask;
 
@@ -104,13 +105,15 @@ std::unique_ptr<edm4hep::RawCalorimeterHitCollection> CalorimeterHitDigi::proces
         m_log->trace("new cell ID in {:s}: {:#064b}", m_cfg.readout, hid);
 
         merge_map[hid].push_back(ix);
+        hids.push_back(hid);
 
         ix++;
     }
 
     // signal sum
     // NOTE: we take the cellID of the most energetic hit in this group so it is a real cellID from an MC hit
-    for (const auto &[id, ixs] : merge_map) {
+    for (auto id : hids) {
+        auto ixs = merge_map[id];
         double edep     = 0;
         double time     = std::numeric_limits<double>::max();
         double max_edep = 0;
```
and this on top of PR base (main):
```diff
diff --git a/src/algorithms/calorimetry/CalorimeterHitDigi.cc b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
index 6fe18466..eff6b9c9 100644
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -110,6 +110,14 @@ std::unique_ptr<edm4hep::RawCalorimeterHitCollection> CalorimeterHitDigi::single
         // Note: juggler internal unit of energy is dd4hep::GeV
         const double eDep    = ahit.getEnergy();
 
+        double time = std::numeric_limits<double>::max();
+        for (const auto& c : ahit.getContributions()) {
+            if (c.getTime() <= time) {
+                time = c.getTime();
+            }
+        }
+        if (time > m_cfg.capTime) continue;
+
         // apply additional calorimeter noise to corrected energy deposit
         const double eResRel = (eDep > m_cfg.threshold)
                                ? m_normDist(generator) * std::sqrt(
@@ -118,18 +126,8 @@ std::unique_ptr<edm4hep::RawCalorimeterHitCollection> CalorimeterHitDigi::single
                                     std::pow(m_cfg.eRes[2] / (eDep), 2)
                                  )
                                : 0;
-
         const double ped    = m_cfg.pedMeanADC + m_normDist(generator) * m_cfg.pedSigmaADC;
         const long long adc = std::llround(ped + eDep * (m_cfg.corrMeanScale + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC);
-
-        double time = std::numeric_limits<double>::max();
-        for (const auto& c : ahit.getContributions()) {
-            if (c.getTime() <= time) {
-                time = c.getTime();
-            }
-        }
-        if (time > m_cfg.capTime) continue;
-
         const long long tdc = std::llround((time + m_normDist(generator) * tRes) * stepTDC);
 
         if (eDep> 1.e-3) m_log->trace("E sim {} \t adc: {} \t time: {}\t maxtime: {} \t tdc: {} \t cell ID {}", eDep, adc, time, m_cfg.capTime, tdc, ahit.getCellID());
```
Then the only discrepancies are seen by capybara in ScFi and LFHCAL (that are, by construction, broken by the first test patch in this comment). This way, I conclude that this change does not affect the calorimeters that relied on `single_hits_digi` that is this PR removes.

PR #865, opened 2023-08-21
Unify factories

nathanwbrei:
This introduces JOmniFactory, a lightweight layer connecting JANA to generic algorithms. It is assumed multiple input data (controlled by input tags) which might be changed by user parameters.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Let's not use any names that refer to 'ePIC'. It virtually guarantees that no one else will use it.

wdconinc:
Something causes this:
```
terminate called without an active exception
```
[ref](https://github.com/eic/EICrecon/actions/runs/6951040518/job/18912605741?pr=865#step:5:567)

but only when using the full geometry...

wdconinc:
Seems to be caused by the CalorimeterClusterRecoCoG. Reproduce it by `-Ppodio:output_include_collections=B0ECalTruthClusters`.

wdconinc:
In my tests, there are two issues with CalorimeterClusterRecoCoG (the only one that uses the algorithms interface):
- `m_cluster_output().get()` and `m_assoc_output().get()` when entering the `process` function are `nullptr`, and algorithms wraps the required output collections in a `gsl::not_null`. That is causing an exception.
- The exception 'somehow' is not handled correctly, and leads to the `terminate called without an active exception`.

After this diagnostic round, I'm moving on to trying to find solutions for the first issue. I don't think I'm the best qualified to handle the second issue.

src/extensions/jana/JOmniFactoryGeneratorT.h
```diff
@@ -0,0 +1,120 @@
...
+            factory->SetApplication(m_app);
+            factory->SetPluginName(this->GetPluginName());
+            factory->SetFactoryName(JTypeInfo::demangle<FactoryT>());
+            // factory->SetTag(wiring.m_tag);
+            // We do NOT want to do this because JMF will use the tag to suffix the collection names
+            // TODO: NWB: Change this in JANA
+            factory->config() = wiring.m_default_cfg;
+
+            // Set up all of the wiring prereqs so that Init() can do its thing
+            // Specically, it needs valid input/output tags, a valid logger, and
```
wdconinc, line 107:
```suggestion
            // Specifically, it needs valid input/output tags, a valid logger, and
```

wdconinc:
Capybara sees no meaningful diffs, e.g. https://github.com/eic/EICrecon/actions/runs/6984054765/job/19006840648#step:9:106 has only collection IDs due to #1138.

src/tests/omnifactory_test/JOmniFactoryTests.cc
```diff
@@ -0,0 +1,210 @@
...
+        m_init_call_count++;
+        logger()->info("Calling BasicTestAlg::Configure");
+    }
+
+    void ChangeRun(int64_t run_number) {
+        m_changerun_call_count++;
+        logger()->info("Calling BasicTestAlg::ChangeRun");
+    }
+
+    void Process(int64_t run_number, uint64_t event_number) {
```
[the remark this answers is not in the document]
wdconinc:
```suggestion
    // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
    void Process(int64_t run_number, uint64_t event_number) {
```

src/extensions/jana/JOmniFactory.h
```diff
@@ -0,0 +1,442 @@
...
+
+#pragma once
+
+/**
+ * Omnifactories are a lightweight layer connecting JANA to generic algorithms
+ * It is assumed multiple input data (controlled by input tags)
+ * which might be changed by user parameters.
+ */
+
+#include <datamodel_glue.h>
```
wdconinc, line 13:
```suggestion
#include "datamodel_glue.h"
```

src/extensions/jana/JOmniFactory.h
```diff
@@ -0,0 +1,442 @@
...
+#include <JANA/JMultifactory.h>
+#include <JANA/JEvent.h>
+#include <spdlog/spdlog.h>
+#include "extensions/spdlog/SpdlogExtensions.h"
+#include <services/log/Log_service.h>
```
wdconinc, line 18:
```suggestion
#include "services/log/Log_service.h"
```

src/extensions/jana/JOmniFactory.h
```diff
@@ -0,0 +1,442 @@
...
+        m_app->SetDefaultParameter(m_prefix + ":InputTags", default_input_collection_names, "Input collection names");
+        m_app->SetDefaultParameter(m_prefix + ":OutputTags", default_output_collection_names, "Output collection names");
+
+        // Set input collection names
+        if (m_inputs.size() != default_input_collection_names.size()) {
+            throw JException("JOmniFactory '%s': Wrong number of input collection names: %d expected, %d found.",
+                             m_prefix.c_str(), m_inputs.size(), default_input_collection_names.size());
+        }
+        size_t i = 0;
+        for (auto* input : m_inputs) {
```
wdconinc, line 365: C++20 now:
```suggestion
        for (size_t i = 0; auto* input : m_inputs) {
```

src/extensions/jana/JOmniFactory.h
```diff
@@ -0,0 +1,442 @@
...
+           input->collection_name = default_input_collection_names[i++];
+        }
+
+        // Set output collection names and create corresponding helper factories
+        if (m_outputs.size() != default_output_collection_names.size()) {
+            throw JException("JOmniFactory '%s': Wrong number of output collection names: %d expected, %d found.",
+                             m_prefix.c_str(), m_outputs.size(), default_output_collection_names.size());
+        }
+        i = 0;
+        for (auto* output : m_outputs) {
```
wdconinc, line 375:
```suggestion
        for (size_t i = 0; auto* output : m_outputs) {
```

src/extensions/jana/JOmniFactoryGeneratorT.h
```diff
@@ -0,0 +1,120 @@
...
+        std::vector<std::string> m_default_output_tags;
+        FactoryConfigType m_default_cfg; /// Must be properly copyable!
+    };
+
+    struct UntypedWiring {
+        std::string m_tag;
+        std::vector<std::string> m_default_input_tags;
+        std::vector<std::string> m_default_output_tags;
+        std::map<std::string, std::string> m_config_params;
+    };
```
wdconinc, line 29: How about
```suggestion
    struct StructWiring {
        std::string m_tag;
        std::vector<std::string> m_default_input_tags;
        std::vector<std::string> m_default_output_tags;
        FactoryConfigType m_default_cfg; /// Must be properly copyable!
    };
    
    struct TypedWiring {
        std::string m_tag;
        std::vector<std::string> m_default_input_tags;
        std::vector<std::string> m_default_output_tags;
        std::map<std::string, PropertyValueType> m_config_params;
    };

    struct UntypedWiring {
        std::string m_tag;
        std::vector<std::string> m_default_input_tags;
        std::vector<std::string> m_default_output_tags;
        std::map<std::string, std::string> m_config_params;
    };
```
with `using PropertyValueType = typename algorithms::PropertyValue;` from https://github.com/eic/algorithms/blob/main/core/include/algorithms/property.h#L33. That can add more type-safety for when desired.

Unfortunately no `FactoryT::PropertyValue`...

wdconinc: Related to this: https://github.com/eic/EICrecon/pull/1154, where std::map<std::string, PropertyValueType> m_config_params is used since it allows setting defaults without parsing.

src/extensions/jana/JOmniFactory.h
wdconinc: This should be applied throughout: `".h"` for in-tree, `<.h>` for system.

wdconinc: 4a6ffa7ca0398533c71192ff107f64906220d0b0

src/extensions/jana/JOmniFactory.h
wdconinc: dbc0fe8ac986aec22e5eeb20f667033947a39f26

src/extensions/jana/JOmniFactoryGeneratorT.h
nathanwbrei: Yeah I think that's great. Hopefully this will solve the problem I was facing where having an open universe of possible parameter types was preventing me from cleanly separating Omnifactories from Algorithms without resorting to Concepts.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
```diff
@@ -0,0 +1,52 @@
...
+        m_algo = std::make_unique<AlgoT>(GetPrefix());
+        m_algo->applyConfig(config());
+        m_algo->init(m_geoSvc().detector(), logger());
+    }
+
+    void ChangeRun(int64_t run_number) {
+    }
+
+    void Process(int64_t run_number, uint64_t event_number) {
+        m_cluster_output() = std::move(std::make_unique<PodioTypeMap<edm4eic::Cluster>::collection_t>());
```
nathanwbrei, line 44: I feel like creating the (initially empty) output collections should be responsibility of JOmnifactory itself, if the Algorithm isn't going to do it. That way we can be confident we will always have empty collections as opposed to missing collections.

src/extensions/jana/JOmniFactory.h
```diff
@@ -0,0 +1,453 @@
...
+        friend class JOmniFactory;
+
+        void Configure(JParameterManager& parman, const std::string& prefix) override {
+            parman.SetDefaultParameter(m_prefix + ":" + this->m_name, m_data, this->m_description);
+        }
+        void Configure(std::map<std::string, std::string> fields) override {
+            auto it = fields.find(this->m_name);
+            if (it != fields.end()) {
+                const auto& value_str = it->second;
+                if constexpr (10000 * JVersion::major
```
nathanwbrei, line 240: I don't like the fact that we have to do this at all, and I want to delete it as soon as possible, but it's a nice trick when we are in between versions like we are right now.

src/tests/omnifactory_test/JOmniFactoryTests.cc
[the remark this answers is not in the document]
nathanwbrei: I guess the "correct" fix is `struct EventKey { int64_t run_number, uint64_t event_number};`. 
Taking into consideration that other experiments have other levels in their event key such as luminosity blocks, I'd kind of be tempted to support arbitrary levels using something along the lines of: 
```std::variant<uint64_t[2], std::vector<uint64_t>> ```

However, this design makes it just as easy to mix up event number and run number as before, so maybe we should just go the strong types route. (I love strong types when I'm writing Haskell, but they feel wrong to me in C++ for some reason). 

Although if we are doing strong types, I think we should do this at the JANA level, if for no other reason than having run_number be int64_t but event_number be uint64_t drives me NUTS...

File: src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h (hunk 8 above)
Comment (wdconinc) [reply src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h:44]: Good point, I'll see if I can std::enable_if this in JOmniFactory based on some type in the algorithm.

File: src/extensions/jana/JOmniFactoryGeneratorT.h (hunk 7 above)
Comment (wdconinc) [reply src/extensions/jana/JOmniFactoryGeneratorT.h:29]: And what would be wrong with concepts anyway? ;-)
```cpp
    template < class T = FactoryConfigType >
        requires std::same_as<T, std::map<typename T::key_type, typename T::mapped_type, typename T::key_compare, typename T::allocator_type>>
    struct TypedWiring {
        std::string m_tag;
        std::vector<std::string> m_default_input_tags;
        std::vector<std::string> m_default_output_tags;
        std::map<std::string, typename FactoryConfigType::mapped_type> m_config_params;
    };
```

wdconinc: But realistically, without any way to test this TypedWiring struct here, I think this probably better postponed until https://github.com/eic/EICrecon/pull/1154, after merging here and rebasing there.

So, resolving here.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
wdconinc: This turns out to be harder than I thought, since the JOmniFactory doesn't know anything about the algorithm.

wdconinc: Addressed in [a42ba63](https://github.com/eic/EICrecon/pull/865/commits/a42ba6369499c754f3907ccaf89f326e1bef78e7). Not entirely happy with it, but good enough. Reasons:
- requires OutputBase to have additional public interface Reset(),
- Reset() is only used by PodioOutput,
- it requires the factories to expose AlgoT publicly, and we end up with `AlgoT::AlgoT::` as the type...

wdconinc:
No diffs, https://github.com/eic/EICrecon/actions/runs/7053374581/job/19200676466?pr=865#step:11:106, so this is good to merge, yay.

wdconinc (approved):

PR #869, opened 2023-08-24
Revert "feat: use podio collections in tracking digi/reco; rm unused interfaces (#840)"

wdconinc:
### Briefly, what does this PR introduce?
This reverts commit eb1387fcca29b2bae6e4c6bb02f1aab948c61bff and #840. It needs more work and more careful checking.

### What kind of change does this PR introduce?
- [x] Bug fix (issue: too many changes at once)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators @veprbl @kkauder 

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
Reverts to previous working version.

PR #870, opened 2023-08-24
fix: apply the specified time resolutions to tracking reco algorithm

wdconinc:
The time resolution specified on the command line or in the tracking.cc file was ignored for the tracker hit reconstruction (only). This fixes that issue so the timeError is set correctly when the hits are sent to ACTS for track fitting.

Changes in the output to be expected:
- timeError in B0, TOF endcap (all others are set to default values),
- potential for different track fit parameters downstream of these collections.

Ref https://github.com/eic/EICrecon/pull/840#issuecomment-1692340071, bullet 1.

TODO:
- [x] Merge #869 first

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @veprbl @kkauder 

No.

Yes, this will cause B0 and TOF time to be as set in tracking.cc, i.e.
```
src/detectors/B0TRK/B0TRK.cc:    hit_reco_cfg.time_resolution = 8;
src/detectors/ECTOF/ECTOF.cc:    hit_reco_cfg.time_resolution = 0.025;
```

### DIFF
--- a/src/algorithms/tracking/TrackerHitReconstruction.h
+++ b/src/algorithms/tracking/TrackerHitReconstruction.h
@@ -41,6 +41,9 @@ namespace eicrecon {
         /// Get a configuration to be changed
         eicrecon::TrackerHitReconstructionConfig& getConfig() {return m_cfg;}
 
+        /// Set a configuration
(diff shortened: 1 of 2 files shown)


ajentsch (dismissed):
I will need to update the B0 timing value - something I forgot when I updated the segmentation to match the AC-LGAD assumption.

veprbl (approved):

PR #879, opened 2023-08-29
Common transfer matrix algorithm

simonge:
Moves the common components of the Roman Pots and Off Momentum reconstruction into a single algorithm with configuration. Keeping the implementation as close as possible to the current method

- [ ] Bug fix (issue #__)
- [x] New feature (issue #721 )
- [ ] Documentation update
- [ ] Other: __

No
No


src/algorithms/fardetectors/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# Then correctly sets sources for ${_name}_plugin and ${_name}_library targets
+# Adds headers to the correct installation directory
+plugin_glob_all(${PLUGIN_NAME})
+
+# Find dependencies
+plugin_add_dd4hep(${PLUGIN_NAME})
+plugin_add_cern_root(${PLUGIN_NAME})
+plugin_add_event_model(${PLUGIN_NAME})
+
+# Add include directories
```
wdconinc, line 19: Remove commented blocks.

src/algorithms/fardetectors/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+cmake_minimum_required(VERSION 3.16)
```
wdconinc, line 1: Not needed in this module.

src/algorithms/fardetectors/MatrixTransferStatic.cc
```diff
@@ -0,0 +1,154 @@
...
+
+}
+
+std::unique_ptr<edm4eic::ReconstructedParticleCollection> eicrecon::MatrixTransferStatic::produce(const edm4hep::SimTrackerHitCollection& rawhits) {
+
+  auto outputParticles = std::make_unique<edm4eic::ReconstructedParticleCollection>();
+
+  //---- begin Reconstruction code ----
+
+  double goodHitX[2] = {0.0, 0.0};
```
wdconinc, line 54: Use one of the Vector3f data types in the data model.

src/algorithms/fardetectors/MatrixTransferStatic.cc
```diff
@@ -0,0 +1,154 @@
...
+    auto cellID = h.getCellID();
+    // The actual hit position in Global Coordinates
+    auto gpos = m_cellid_converter->position(cellID);
+    // local positions
+    auto volman = m_detector->volumeManager();
+    auto local = volman.lookupDetElement(cellID);
+
+    auto pos0 = local.nominal().worldToLocal(dd4hep::Position(gpos.x(), gpos.y(), gpos.z())); // hit position in local coordinates
+
+    if(!goodHit2 && gpos.z()/dd4hep::mm > m_cfg.hit2minZ && gpos.z()/dd4hep::mm < m_cfg.hit2maxZ){
```
wdconinc, line 72: This should be done with a plane in the cellID, not a z range.

src/algorithms/fardetectors/MatrixTransferStatic.cc
```diff
@@ -0,0 +1,154 @@
...
+    double YL[2] = {goodHitY[0] - m_cfg.local_y_offset, goodHitY[1] - m_cfg.local_y_offset};
+
+    double base = goodHitZ[1] - goodHitZ[0];
+
+    if (base == 0) {
+      m_log->info("Detector separation = 0! Cannot calculate slope!");
+    }
+    else{
+
+      double Xip[2] = {0.0, 0.0};
```
wdconinc, line 109: Similarly Vector2f here.

src/algorithms/fardetectors/MatrixTransferStatic.cc
```diff
@@ -0,0 +1,154 @@
...
+      double Xrp[2] = {XL[1], (1000 * (XL[1] - XL[0]) / (base)) - m_cfg.local_x_slope_offset}; //- _SX0RP_;
```
wdconinc, line 110: Magic constant 1000 is dd4hep::mm?

src/algorithms/fardetectors/MatrixTransferStatic.cc
```diff
@@ -0,0 +1,154 @@
...
+      // Polar Angle and deltaP at the IP
+
+      for (unsigned i0 = 0; i0 < 2; i0++) {
+	for (unsigned i1 = 0; i1 < 2; i1++) {
+	  Xip[i0] += aXinv[i0][i1] * Xrp[i1];
+	  Yip[i0] += aYinv[i0][i1] * Yrp[i1];
+	}
+      }
+
+      // convert polar angles to radians
```
wdconinc, line 124: Avoid hard coding units.

src/algorithms/fardetectors/MatrixTransferStatic.cc
```diff
@@ -0,0 +1,154 @@
...
+      reconTrack.setType(0);
+      reconTrack.setMomentum({prec});
+      reconTrack.setEnergy(std::hypot(edm4eic::magnitude(reconTrack.getMomentum()), m_cfg.partMass));
+      reconTrack.setReferencePoint({refPoint});
+      reconTrack.setCharge(m_cfg.partCharge);
+      reconTrack.setMass(m_cfg.partMass);
+      reconTrack.setGoodnessOfPID(1.);
+      reconTrack.setPDG(m_cfg.partPDG);
+      //reconTrack.covMatrix(); // @TODO: Errors
+      outputParticles->push_back(edm4eic::ReconstructedParticle(reconTrack));
```
wdconinc, line 148: Avoid copy constructor.
```suggestion
      outputParticles->push_back(reconTrack);
```
Or even use the collection's `create()` with forwarded arguments to create in place. Or use `emplace_back`.

src/algorithms/fardetectors/MatrixTransferStatic.h
```diff
@@ -0,0 +1,75 @@
...
+#include "MatrixTransferStaticConfig.h"
+
+namespace eicrecon {
+
+  class MatrixTransferStatic {
+  public:
+    MatrixTransferStatic() = default;
+
+    //----- Define constants here ------
+    std::string m_readout;
```
wdconinc, line 30: Remove unused members.

src/algorithms/fardetectors/MatrixTransferStatic.h
```diff
@@ -0,0 +1,75 @@
...
+    void setGeoConverter(std::shared_ptr<const dd4hep::rec::CellIDPositionConverter> id_conv) {m_cellid_converter=id_conv;}
+
+    // Set position convertor
+    void setDetector(dd4hep::Detector* det) {m_detector=det;}
+
+    // Get a configuration to be changed
+    eicrecon::MatrixTransferStaticConfig& getConfig() {return m_cfg;}
+
+    // Sets a configuration (config is properly copyible)
+    eicrecon::MatrixTransferStaticConfig& applyConfig(eicrecon::MatrixTransferStaticConfig cfg) { m_cfg = cfg; return m_cfg;}
```
wdconinc, line 63: Use `WithPODConfig` mixin.

src/algorithms/fardetectors/MatrixTransferStaticConfig.h
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2022, 2023, Christopher Dilks, Luigi Dello Stritto
```
wdconinc, line 2: Probably not the right names?

src/algorithms/fardetectors/MatrixTransferStaticConfig.h
```diff
@@ -0,0 +1,65 @@
...
+#pragma once
+
+#include <spdlog/spdlog.h>
+
+namespace eicrecon {
+  struct MatrixTransferStaticConfig {
+
+    float partMass  {0.938272};
+    float partCharge{1};
+    float partPDG   {2122};
```
wdconinc, line 13: PDG code is always integral, right? Long long should work, right?

src/algorithms/fardetectors/MatrixTransferStaticConfig.h
```diff
@@ -0,0 +1,65 @@
...
+    double hit2maxZ{0};
+
+    std::string readout{""};
+/*       double aX[2][2] = {{2.102403743, 29.11067626}, */
+/* 			 {0.186640381, 0.192604619}}; */
+/*       double aY[2][2] = {{0.0000159900, 3.94082098}, */
+/* 			 {0.0000079946, -0.1402995}}; */
+
+/*       // ------------------------------------------------ */
+
```
wdconinc, line 39: Remove commented blocks.

src/detectors/RPOTS/RPOTS.cc
```diff
@@ -8,15 +8,37 @@
 
 #include "extensions/jana/JChainFactoryGeneratorT.h"
```
wdconinc, line 9: Do we need a separate plugin for rpots and offmtrk?

src/detectors/RPOTS/RPOTS.cc
```diff
@@ -8,15 +8,37 @@
...
+    recon_cfg.nomMomentum          =  275.0;      //in GEV --> exactly half of the top energy momentum (for proton spectators from deuteron breakup)
+
+    recon_cfg.hit1minZ = 25099.0;
+    recon_cfg.hit1maxZ = 26022.0;
+    recon_cfg.hit2minZ = 27099.0;
+    recon_cfg.hit2maxZ = 28022.0;
+
+    recon_cfg.readout              = "ForwardRomanPotHits";
+
+    app->Add(new JChainFactoryGeneratorT<MatrixTransferStatic_factory>({"ForwardRomanPotHits"},"ForwardRomanPotRecParticles",recon_cfg));
```
wdconinc, line 41: Bonus points for writing this as a multifactory :-)

src/algorithms/CMakeLists.txt
```diff
@@ -5,3 +5,4 @@ add_subdirectory( tracking )
 add_subdirectory( pid )
 add_subdirectory( digi )
 add_subdirectory( reco )
+add_subdirectory( fardetectors )
```
wdconinc, line 8: We've been trying to keep algorithms organized by function rather than region. Would this fit in `reco`?

src/algorithms/fardetectors/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+plugin_add(${PLUGIN_NAME} WITH_STATIC_LIBRARY)
+
+# The macro grabs sources as *.cc *.cpp *.c and headers as *.h *.hh *.hpp
+# Then correctly sets sources for ${_name}_plugin and ${_name}_library targets
+# Adds headers to the correct installation directory
+plugin_glob_all(${PLUGIN_NAME})
+
+# Find dependencies
+plugin_add_dd4hep(${PLUGIN_NAME})
+plugin_add_cern_root(${PLUGIN_NAME})
```
wdconinc, line 14: This doesn't use ROOT classes, so probably doesn't need to link against ROOT.

src/algorithms/fardetectors/MatrixTransferStatic.cc
```diff
@@ -0,0 +1,154 @@
...
+
+#include "MatrixTransferStatic.h"
+
+void eicrecon::MatrixTransferStatic::init(std::shared_ptr<spdlog::logger> &logger) {
+
+  m_log = logger;
+
+  // do not get the layer/sector ID if no readout class provided
+  if (m_cfg.readout.empty()){m_log->error("READOUT IS EMPTY!"); return;}
+
```
wdconinc, line 14: Unused.
```suggestion
```

wdconinc (commented):
Two more comments.

src/algorithms/fardetectors/MatrixTransferStatic.cc
simonge: I agree that this should make proper use of the geometry but I would rather leave any implementation changes for another PR even if they don't effect the functionality. I believe @ajentsch is working on updated versions of this.
Specifically for this comment, the current readout structure for the Off Momentum and Roman Pots in the simulation could do with an overhaul to converge, the Off Momentum layer is always "1".

src/detectors/RPOTS/RPOTS.cc
simonge: I'm not sure, what would be the arguments for/against combining them?

src/algorithms/CMakeLists.txt
simonge: Maybe at the moment but it really needs to be sub divided into digi, tracking and reco in the future. I have been using fardetectors as a placeholder for the LowQ2 development too, currently containing an alternative tracking pathway to acts.

src/detectors/RPOTS/RPOTS.cc
wdconinc: Easier to maintain fewer files and plugins. But we shouldn't put completely unrelated code in one plugin either because then you lose the flexibility of loading one detector chain without needing to load the other. Since the rpots and offmtrk are closely related, having them in one plugin makes sense to me. If there's going to be a next layer that combines results from both, even more so.

src/algorithms/fardetectors/MatrixTransferStatic.cc
ajentsch: Yes.

src/algorithms/fardetectors/MatrixTransferStatic.cc
ajentsch: The matrix and offsets assume certain units are being used (because they are used to calculate the matrix in the first place).

src/detectors/RPOTS/RPOTS.cc
ajentsch: Don't forget we will be adding a whole new setup for reconstruction, and it will start with the RP first. It makes much more sense while we are actively developing to keep them separate. This algorithm is not a fancy one, and not hard to maintain, anyhow. Information from OMD and RP will not really be used together to reconstruct a particle, except in cases of cross-calibration, but that depends heavily on the vacuum system design.

wdconinc:
@simonge The pre-commit test at https://results.pre-commit.ci/run/github/512187504/1693496555.n76hxKPqRoKdK0SsrI10eg is failing due to some spelling checks.

wdconinc (approved):

PR #880, opened 2023-08-30
fix(dRICH): `CentralCKFTrajectories` -> `CentralCKFActsTrajectories`

c-dilks:
Fix dRICH track propagation by using `CentralCKFActsTrajectories` instead of `CentralCKFTrajectories`

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
yes, should restore functionality to dRICH

### DIFF
--- a/src/detectors/DRICH/DRICH.cc
+++ b/src/detectors/DRICH/DRICH.cc
@@ -109,7 +109,7 @@ extern "C" {
     // charged particle tracks
     app->Add(new JChainMultifactoryGeneratorT<RichTrack_factory>(
           "DRICHTracks",
-          {"CentralCKFTrajectories"},
+          {"CentralCKFActsTrajectories"},
           {"DRICHAerogelTracks", "DRICHGasTracks"},
           track_cfg,
           app
--- a/src/detectors/DRICH/README.md
+++ b/src/detectors/DRICH/README.md
@@ -30,7 +30,7 @@ flowchart TB
   subgraph Inputs
     direction LR
     SimHits(<strong>DRICHHits</strong><br/>MC dRICH photon hits<br/>edm4hep::SimTrackerHit):::col
-    Trajectories(<strong>CentralCKFTrajectories</strong><br/>ActsExamples::Trajectories):::col
+    Trajectories(<strong>CentralCKFActsTrajectories</strong><br/>ActsExamples::Trajectories):::col
     MCParts(<strong>MCParticles</strong><br/>MC True Particles<br/>edm4hep::MCParticles):::col
   end
 


wdconinc:
As long as there is nothing in the default output files that relies on the DRICH code, we are going to keep breaking the DRICH unintentionally because it is not tested for... So, you could (have) avoid(ed) this by adding `DRICHAerogelTracks` and `DRICHGasTracks` to the [JEventProcessorPODIO.cc](https://github.com/eic/EICrecon/blob/main/src/services/io/podio/JEventProcessorPODIO.cc) list, probably even as part of this PR.

c-dilks:
> As long as there is nothing in the default output files that relies on the DRICH code, we are going to keep breaking the DRICH unintentionally because it is not tested for... So, you could (have) avoid(ed) this by adding `DRICHAerogelTracks` and `DRICHGasTracks` to the [JEventProcessorPODIO.cc](https://github.com/eic/EICrecon/blob/main/src/services/io/podio/JEventProcessorPODIO.cc) list, probably even as part of this PR.

- tried before: https://github.com/eic/EICrecon/pull/655
- or https://github.com/eic/EICrecon/pull/734 which awaits https://github.com/eic/EICrecon/pull/829

wdconinc:
https://github.com/eic/EICrecon/pull/829 is on my list to review and merge this week...

c-dilks:
https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/merge_requests/293 is also ready (I just realized I never marked it "ready"), but is only useful if someone routinely checks the plots. This particular track propagation issue just showed up as empty plots.

c-dilks:
> > only useful if someone routinely checks the plots
> 
> We have several benchmarks that fail if there are no entries when there are entries expected.

We should add this feature, this MR is just a starting point. I keep trying to push @chchatte92 to improve the benchmarks, since he now does quite a lot more by hand :)

chchatte92:
> > > only useful if someone routinely checks the plots
> > 
> > 
> > We have several benchmarks that fail if there are no entries when there are entries expected.
> 
> We should add this feature, this MR is just a starting point. I keep trying to push @chchatte92 to improve the benchmarks, since he now does quite a lot more by hand :)

Absolutely. Let's say, I will add at least these three-four features by next week to the benchmarks
- Gaussian Fits for all the Residuals and theta distributions. 
- Apart from SPE Cherenkov angle as a function of phi, also SPE resolution.
- Poisson Fit to the NPE distribution 

Does it sound ok?

c-dilks:
Testing this fix with benchmarks using https://github.com/eic/drich-dev/pull/98 since it seems I never added them to the CI config.

And adding benchmarks to the CI config with https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/merge_requests/306

c-dilks:
https://github.com/eic/drich-dev/issues/103 blocks me from testing eta and momentum scans, but at least for single fixed momenta, dRICH PID appears to be recovered after this PR.

wdconinc (approved):
Yup.

PR #881, opened 2023-08-31
reco: add *ChargedJets

veprbl:
This adds charged jets that may be more usable for physics studies until hadronic subtraction is implemented.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @ruse-traveler @bspage912

### DIFF
--- a/src/algorithms/reco/ChargedParticleSelector.h
+++ b/src/algorithms/reco/ChargedParticleSelector.h
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023 Dmitry Kalinkin
+
+#pragma once
+
+#include <memory>
+
+#include <edm4hep/MCParticleCollection.h>
+#include <spdlog/logger.h>
+
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+  class ChargedParticleSelector : public WithPodConfig<NoConfig> {
+
+  private:
+
+    std::shared_ptr<spdlog::logger> m_log;
+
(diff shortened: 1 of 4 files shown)


src/algorithms/reco/ChargedParticleSelector.h
ruse-traveler, line 1: I think this is a reasonable addition (I imagine having a specific collection of charged MC particles will be useful in other contexts), but for my own understanding: what's the rationale for using a separate algorithm in this case rather than applying a condition in the factory?

For example, we pick out the final state particles to feed into the jet reconstruction in the factory itself. Should we (eventually) use an algorithm to produce a collection of final state particles and then run the factory over that final state collection?

ruse-traveler (approved):
Looks good!

src/algorithms/reco/ChargedParticleSelector.h
veprbl: I was asking this myself too. In principle, we could have added particle selection criteria as a configuration flag to the jet algorithm. That's a fine way to do it too. One argument for the current way is that for detector jets we rely on `ReconstructedChargedParticles`, so it is natural to implement the way I did, consistently as for the generator jets.

I am open to applying the condition in the factory based on the configuration.

ruse-traveler: Gotcha! I also like the approach here since it parallels the reconstructed jets, so I'd say let's stick with this! If we need a more specialized set of particles down the road, then we could explore adding selection criteria to the configuration flags...

PR #883, opened 2023-08-31
Update BHCal energy thresholds

ruse-traveler:

This PR updates the BHCal minimum energy thresholds applied during digitizaiton and clustering based on discussions from the calorimeter and background working groups (see [this spreadsheet](https://docs.google.com/spreadsheets/d/1s8oXj36SqIh7TJeHFH89gQ_ayU1_SVEpWQNkx6sETKs/edit?pli=1#gid=238482234)).

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: Updating default values

No.

Yes. Only tiles with at least 5 MeV deposited in them will be digitized and passed on to reconstruction.

### DIFF
--- a/src/detectors/BHCAL/BHCAL.cc
+++ b/src/detectors/BHCAL/BHCAL.cc
@@ -23,6 +23,7 @@ extern "C" {
           {
             .eRes = {},
             .tRes = 0.0 * dd4hep::ns,
+            .threshold = 5.0 * dd4hep::MeV,
             .capADC = 65536,
             .capTime = 100, // given in ns, 4 samples in HGCROC
             .dyRangeADC = 1.0 * dd4hep::GeV,
@@ -74,7 +75,7 @@ extern "C" {
             .localDistXY = {15*dd4hep::mm, 15*dd4hep::mm},
             .dimScaledLocalDistXY = {50.0*dd4hep::mm, 50.0*dd4hep::mm},
             .splitCluster = false,
-            .minClusterHitEdep = 3.0 * dd4hep::MeV,
+            .minClusterHitEdep = 5.0 * dd4hep::MeV,
             .minClusterCenterEdep = 30.0 * dd4hep::MeV,
             .transverseEnergyProfileMetric = "globalDistEtaPhi",
             .transverseEnergyProfileScale = 1.,


wdconinc:
Diffs limited to HcalBarrel collections, e.g. https://github.com/eic/EICrecon/actions/runs/6040073978/job/16391225109?pr=883#step:11:72. E.g. `HcalBarrelClusters.energy` (https://github.com/eic/EICrecon/actions/runs/6040073978/job/16391225109?pr=883#step:11:3371):
```
  [[0.0333, 0.0384, 0.253, 0.129, 0.0458], [...], ..., [14.3, 0.0638, ..., 0.332]]
  [[0.0338, 0.0412, 0.255, 0.123, 0.0458], ..., [14.3, 0.062, ..., 0.0578, 0.33]]
```

Though I am a bbit confused about how individual cluster energies can _decrease_ (0.129 -> 0.123 GeV, 0.0638 -> 0.062 GeV) when both the threshold and minimum energy deposition are _increased_.

ruse-traveler:
> Though I am a bbit confused about how individual cluster energies can _decrease_ (0.129 -> 0.123 GeV, 0.0638 -> 0.062 GeV) when both the threshold and minimum energy deposition are _increased_.

Ah! To me it looks like a couple ~3 MeV tiles are cut out that were included before (since the old thresholds were at 1 KeV for digitization and 3 MeV for clustering)...

wdconinc:
Wait, I wrote it down wrong. How can the energy of the cluster _increase_ when we cut out more hits by _increasing_ thresholds, while keeping the same number of clusters?

src/detectors/BHCAL/BHCAL.cc
```diff
@@ -23,6 +23,7 @@ extern "C" {
           {
             .eRes = {},
             .tRes = 0.0 * dd4hep::ns,
+            .threshold = 5.0 * dd4hep::MeV,
```
veprbl, line 26: There is a prior work https://github.com/eic/EICrecon/pull/674 which took another approach of setting threshold on ADC quantity at the hit reconstruction instead of digitization.

I don't quite understand why they went with `thresholdFactor` instead of `thresholdValue`. Those are interchangeable, but latter doesn't involve the expected pedestal sigma:
https://github.com/eic/EICrecon/blob/0218bf17021eea7586254fb1be571ad2bedbb699/src/algorithms/calorimetry/CalorimeterHitReco.cc#L25 I do agree with their approach more, for the reasons mentioned in https://chat.epic-eic.org/main/pl/szkezeezwjnb3fiaiihg8ze87o

This one should be close to being equivalent. I do have some questions to the implementation of this parameter. Look at it:
https://github.com/eic/EICrecon/blob/2c7550a21588f59bf889253f32a0dfeb43ccdeaa/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L144-L152
It does not physically removes the hit, just zeroes the physical energy contribution and keeps the contribution of noise in place. I don't understand what that is supposed to model.

wdconinc: > just zeroes the physical energy contribution and keeps the contribution

Just guessing, but zero acts like an overflow bin then?

veprbl: > > just zeroes the physical energy contribution and keeps the contribution
> 
> Just guessing, but zero acts like an ~overflow~underflow bin then?

Formally, the normal distribution has infinite tails, so a `ped` can be arbitrarily large. I guess, it can't surpass a 36-sigma cut in Sc-Fi PR :D

ruse-traveler:
> Wait, I wrote it down wrong. How can the energy of the cluster _increase_ when we cut out more hits by _increasing_ thresholds, while keeping the same number of clusters?

Oh! That is very strange... Do the increases all seem to be on the order of a few MeV or so?

> It does not physically removes the hit, just zeroes the physical energy contribution and keeps the contribution of noise in place. I don't understand what that is supposed to model.

Neither do I... 🤔

wdconinc:
After a walk, how about this explanation. Two nearby clusters, one high energy and one low energy, have a couple of low energy hits between them. Previously those hits were allocated in one way. Removing a hit changes the allocation. A small decrease in the high energy cluster can cause an increase in the low energy cluster through this reallocation. Maybe that's what's happening.

veprbl:
Don't see a big difference at a macroscopic level:
https://veprbl.github.io/capybara-reports/b4b1909887f299e6d78aef53f67d8cf2/

wdconinc (approved):
I'm going to approve since this is a fine way to change the thresholds, even considering @veprbl's comments which can be considered for future reference I guess. Still will need someone to agree with me and resolve the conversation.

ruse-traveler:
> After a walk, how about this explanation. Two nearby clusters, one high energy and one low energy, have a couple of low energy hits between them. Previously those hits were allocated in one way. Removing a hit changes the allocation. A small decrease in the high energy cluster can cause an increase in the low energy cluster through this reallocation. Maybe that's what's happening.

Makes sense to me! And looking at @veprbl's capybara plots, it looks like there's some fluctuations in the nhits which might corroborate that 👀

wdconinc:
Merge queues have a 60 minute timeout on jobs. Changed it to 360 minute. Not sure why it needs to be 60 minutes....

PR #885, opened 2023-08-31
fix: require EDM4eic 2.1 minimum

wdconinc:
We require to have a one-to-many relation to TrackParameters in Trajectory, not a one-to-one relation. That requires 2.1.0 at least.

- [x] Bug fix (issue https://github.com/eic/EICrecon/pull/860#discussion_r1311913936)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes, require EDM4eic 2.1.0 version upgrade.

No.


veprbl (approved):
Can confirm that this fails early, as intended.

veprbl:
I took liberty to remove a check that now is always true.

PR #890, opened 2023-09-01
Silicon thresholds

kkauder:
Moving silicon detector thresholds documented in https://docs.google.com/spreadsheets/d/1s8oXj36SqIh7TJeHFH89gQ_ayU1_SVEpWQNkx6sETKs/edit#gid=238482234
from npsim into digitization.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes.

### DIFF
--- a/src/detectors/BTRK/BTRK.cc
+++ b/src/detectors/BTRK/BTRK.cc
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "SiBarrelRawHits",
         {"SiBarrelHits"},
         {"SiBarrelRawHits"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+            // .timeResolution = 10,
+        },
         app
     ));
 
--- a/src/detectors/BTRK/BTRK.cc
+++ b/src/detectors/BTRK/BTRK.cc
@@ -23,7 +23,7 @@ void InitPlugin(JApplication *app) {
         {"SiBarrelRawHits"},
         {
             .threshold = 0.65 * dd4hep::keV,
-            // .timeResolution = 10,
+	    .timeResolution = 10,
         },
         app
     ));
--- a/src/detectors/BTRK/BTRK.cc
+++ b/src/detectors/BTRK/BTRK.cc
@@ -23,7 +23,6 @@ void InitPlugin(JApplication *app) {
         {"SiBarrelRawHits"},
         {
             .threshold = 0.65 * dd4hep::keV,
-	    .timeResolution = 10,
         },
         app
     ));
--- a/src/detectors/BVTX/BVTX.cc
+++ b/src/detectors/BVTX/BVTX.cc
@@ -23,7 +23,7 @@ void InitPlugin(JApplication *app) {
         {"BarrelVertexRawHits"},
         {
             .threshold = 0.65 * dd4hep::keV,
-	    .timeResolution = 10,
+            .timeResolution = 10,
         },
         app
     ));
--- a/src/detectors/BVTX/BVTX.cc
+++ b/src/detectors/BVTX/BVTX.cc
@@ -23,7 +23,6 @@ void InitPlugin(JApplication *app) {
         {"BarrelVertexRawHits"},
         {
             .threshold = 0.65 * dd4hep::keV,
-            .timeResolution = 10,
         },
         app
     ));
--- a/src/detectors/BVTX/BVTX.cc
+++ b/src/detectors/BVTX/BVTX.cc
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "BarrelVertexRawHits",
         {"VertexBarrelHits"},
         {"BarrelVertexRawHits"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+            // .timeResolution = 10,
+        },
         app
     ));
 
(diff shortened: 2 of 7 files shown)


src/detectors/BTRK/BTRK.cc
```diff
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "SiBarrelRawHits",
         {"SiBarrelHits"},
         {"SiBarrelRawHits"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+            // .timeResolution = 10,
```
wdconinc, line 26: Remove commented lines.
```suggestion
```

src/detectors/BVTX/BVTX.cc
```diff
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "BarrelVertexRawHits",
         {"VertexBarrelHits"},
         {"BarrelVertexRawHits"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+            // .timeResolution = 10,
```
wdconinc, line 26:
```suggestion
```

src/detectors/ECGEM/ECGEM.cc
```diff
@@ -25,7 +25,10 @@ void InitPlugin(JApplication *app) {
             "GEMTrackerEndcapHits3"
         },
         {"GEMRawHit"},
-        {}, // default config
+        {
+            .threshold = 0.25 * dd4hep::keV,
+            // .timeResolution = 10,
```
wdconinc, line 30:
```suggestion
```

src/detectors/ECTRK/ECTRK.cc
```diff
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "EndcapTrackerDigiHit",
         {"TrackerEndcapHits"},
         {"EndcapTrackerDigiHit"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+            // .timeResolution = 10,
```
wdconinc, line 26:
```suggestion
```

kkauder: I didn't see any field that said ten, but I'll put it in.

veprbl: I believe, the suggestion was to remove the commented lines (keeping timeResolution at 0).

wdconinc: Yes, though the default is 10, I think. (Not easy to verify right now.)

kkauder: It's 8 actually, for no discernible reason.
https://github.com/eic/EICrecon/blob/main/src/algorithms/digi/SiliconTrackerDigiConfig.h

src/detectors/BTRK/BTRK.cc
```diff
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "SiBarrelRawHits",
         {"SiBarrelHits"},
         {"SiBarrelRawHits"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+	    .timeResolution = 10,
```
wdconinc, line 26: So, this is right now a dangerous way to leave it: you add an explicit time resolution to digitization, but don't add one to reconstruction. Right now this might work because you set it to the default value. Someone might reasonably think that they can change the time resolution to something else. But that would put the time resolution used on digitization out of sync with the time resolution used in reconstruction. And that would give us incorrect reconstructed times.

wdconinc (changes requested):

src/detectors/BTRK/BTRK.cc
wdconinc: And that's not just happening here.

kkauder:
https://github.com/eic/EICrecon/commit/b0168d4ad1eb7c24ac4921bea2c9f32416092a0d

src/detectors/ECTRK/ECTRK.cc
```diff
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "EndcapTrackerDigiHit",
         {"TrackerEndcapHits"},
         {"EndcapTrackerDigiHit"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+	    .timeResolution = 10,
```
wdconinc, line 26: And what about this one?

kkauder:
Good call.

src/detectors/BVTX/BVTX.cc
```diff
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "BarrelVertexRawHits",
         {"VertexBarrelHits"},
         {"BarrelVertexRawHits"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+            .timeResolution = 10,
```
wdconinc, line 26:
```suggestion
```

src/detectors/ECTRK/ECTRK.cc
```diff
@@ -21,7 +21,10 @@ void InitPlugin(JApplication *app) {
         "EndcapTrackerDigiHit",
         {"TrackerEndcapHits"},
         {"EndcapTrackerDigiHit"},
-        {}, // default config
+        {
+            .threshold = 0.65 * dd4hep::keV,
+            .timeResolution = 10,
```
wdconinc, line 26:
```suggestion
```

wdconinc (approved):

PR #892, opened 2023-09-03
feat(ci): publish GitHub pages site with PR information (from fork)

wdconinc:
This modifies the approach we use for publishing the github-pages documentation to allow also publishing from pull requests while maintaining the main branch as reference top level url.

This works by doing some less linear things with artifacts than we usually do. Instead of building docs, then creating the github-pages artifact, and then deploying this, we now pull artifacts from other pipelines. In particular, we pull the docs from the last complete pipeline on main (with error on failure) and from the last successful pipeline on all open PRs (taking care to use the not yet published docs for the current branch, whether main or PR), then stage this all in a github-pages-staging artifact that is added to by multiple jobs, from which we then create the github-pages artifact, and then deploy. This allows us to put the PRs under e.g. `pr/891` etc.

The real reason for this is not that we want to have doxygen and rst docs for all these PRs (though it does make review of docs changes easier). It also allows us to write CI results to these areas, without needing to find another hosting solution. It could accommodate the unit test reports and capybara bokeh dashboards, for example.

Demonstration at:
- https://github.com/wdconinc/EICrecon/pull/5/ (this PR in my fork)
- https://wdconinc.github.io/EICrecon/#/ (docs from main)
- https://wdconinc.github.io/EICrecon/pr/5/#/ (docs from PR 5)

When we list the open PRs, we only have a token that has permission for querying this info in the current repository. That's why I had to demonstrate this in my fork and it only finds PRs against that fork. It is possible that this PR will not get the permissions to be able to identify all open PRs in the eic/EICrecon repository; we'll see.

Current size of docs per PR is about 20 MB. The maximum size of a GitHub Pages site is 1 GB, so we can accommodate 50 open PRs.

TODO:
- [x] relax github-pages environment protection to allow deployment from all branches, not just main

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -530,7 +530,7 @@ jobs:
         id: query
         with:
           route: GET /repos/{repo}/pulls?state=open
-          repo: "eic/EICrecon" # FIXME: ${{ github.repository }}
+          repo: ${{ github.repository }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Remap open PRs
@@ -549,7 +549,6 @@ jobs:
     steps:
       - uses: dawidd6/action-download-artifact@v2
         with:
-          repo: eic/EICrecon # FIXME ${{ github.repository }}
           pr: ${{ matrix.pr }}
           path: publishing_docs/pr/${{ matrix.pr }}/
           name: docs
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -579,7 +579,6 @@ jobs:
           apk add tar bash
         # FIXME bash not really required: see https://github.com/actions/upload-pages-artifact/pull/14
       - uses: actions/upload-pages-artifact@v1
-        if: github.ref == 'refs/heads/main'
         with:
           path: publishing_docs/
           retention-days: 7
@@ -587,7 +586,6 @@ jobs:
   deploy-docs:
     needs:
       - collect-docs
-    if: github.ref == 'refs/heads/main'
     permissions:
       pages: write
       id-token: write
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -521,6 +521,30 @@ jobs:
           path: publishing_docs/
           if-no-files-found: error
 
+  list-open-prs:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


wdconinc:
Currently, several (44...) warnings are produced in the CI pipeline:
Screenshot_20230903-101904

- [ ] the deploy-docs error will be resolved by the TODO item,
- [ ] the list-open-prs job depends on https://github.com/nickofthyme/object-remap, which needs
  - [ ] an update of node and,
  - [ ] shift away from set-output,
  - but the unexpected inputs warning can't be avoided with the environment variable trick at https://github.com/octokit/request-action/issues/26#issuecomment-996992989 due to the wildcard,
- [ ] clang-tidy always warns when no source files have been modified, we probably shouldn't emit a warning,
- [ ] get-docs-from-open-prs emits warnings when no artifact is found and when no files are uploaded, and we should probably ignore these warnings.

wdconinc:
We may be able to avoid the object-remap with only object filters, https://docs.github.com/en/actions/learn-github-actions/expressions#object-filters

.github/workflows/linux-eic-shell.yml
```diff
@@ -514,27 +511,110 @@ jobs:
...
+      matrix: ${{ fromJson(needs.list-open-prs.outputs.json) }}
+      fail-fast: false
+    steps:
+      - uses: dawidd6/action-download-artifact@v2
+        if: github.event.pull_request.number != matrix.pr
+        with:
+          pr: ${{ matrix.pr }}
+          path: publishing_docs/pr/${{ matrix.pr }}/
+          name: docs
+          if_no_artifact_found: warn
```
wdconinc, line 556:
```suggestion
          if_no_artifact_found: ignore
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -514,27 +511,110 @@ jobs:
...
         with:
-          name: jana.dot
-          path: publishing_docs/dot/
+          name: docs
+          path: publishing_docs/pr/${{ matrix.pr }}/
+      - uses: actions/upload-artifact@v3
+        with:
+          name: github-pages-staging
+          path: publishing_docs/
+          if-no-files-found: warn
```
wdconinc, line 566:
```suggestion
          if-no-files-found: ignore
```

wdconinc:
Closing in favor of #923.

Issue #875, opened 2023-08-28
`[FATAL] Detected timeout in worker! Stopping.` does not result in a non-zero exit code


veprbl:
This makes it impossible to trust that the succeeding jobs processed all events. I just wasted few days trying to compare incomplete output files.

wdconinc:
I've been calling for turning this off since day 1. It has not ever been useful, and it has only led to frustration! If something fails, fail already please.

I think we may be able to inject the appropriate flags in the eicrecon source file. Some judicious injections in `options` here: https://github.com/eic/EICrecon/blob/main/src/utilities/eicrecon/eicrecon.cc#L58.

nathanwbrei:
The purpose of the timeouts is to catch things like deadlock and livelock without relying on the cluster's job timeout mechanism. Namely, it could get expensive for us. That being said, timeout should definitely return an error code.

wdconinc:
@veprbl Is this still an issue or did #1003 resolve this? Easiest way to trigger a timeout is to run in valgrind :-/

wdconinc:
With the following diff to trigger a timeout:
```diff
diff --git a/src/algorithms/calorimetry/CalorimeterHitDigi.cc b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
index fb747972..9f9bb6ef 100644
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -10,6 +10,8 @@
 // Author: Chao Peng
 // Date: 06/02/2021
 
+#include <chrono>
+#include <thread>
 
 #include "CalorimeterHitDigi.h"
 
@@ -93,6 +95,9 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
 std::unique_ptr<edm4hep::RawCalorimeterHitCollection> CalorimeterHitDigi::process(const edm4hep::SimCalorimeterHitCollection &simhits)  {
     auto rawhits = std::make_unique<edm4hep::RawCalorimeterHitCollection>();
 
+    using namespace std::chrono_literals;
+    std::this_thread::sleep_for(20000ms);
+
     // find the hits that belong to the same group (for merging)
     std::unordered_map<uint64_t, std::vector<std::size_t>> merge_map;
     std::size_t ix = 0;
[...]

PR #894, opened 2023-09-03
fix: disable timeouts entirely (by default)

wdconinc:
This disables jana event and warmup timeouts entirely. It doesn't solve the exit code in #875, but it would just have just hung instead.

- [x] Bug fix (issue #875)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @veprbl not sure where you encountered the issue exactly, but would this have resulted in at least foregrounding the issue better?

No.

No.


veprbl:
I'm hesitant to go forward with this, as it does not resolve the underlying issue with the timeouts. (those may be re-enabled eventually?)

wdconinc:
I just can't find where we should handle the exception, or why we exit with 0 even when we end due to an exception...

wdconinc:
Not necessary anymore. Timeouts are returning exit code 2 now, and we don't need to highlight them anymore by disabling timeouts.

PR #914, opened 2023-09-05
Changed kinematic calculations to use edm4eic vector utils in ParticlesWithPID

simonge:
The std::hypot method used to define p_mag in ParticlesWithPID.cc by default returns a float. This is insufficient precision when looking at particles travelling very close to the beam axis.

This PR changes to using the edm4eic vector utils methods which return doubles.

Additionally the eta and phi tolerance values provided to match the MC and reconstructed particles do not make sense for the particles along the beamline so if both are below -5 they are assumed to match.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
no

### DIFF
--- a/src/algorithms/pid/ParticlesWithPID.cc
+++ b/src/algorithms/pid/ParticlesWithPID.cc
@@ -79,9 +79,9 @@ namespace eicrecon {
                     continue;
                 }
 
-                const auto p_mag = std::hypot(p.x, p.y, p.z);
-                const auto p_phi = std::atan2(p.y, p.x);
-                const auto p_eta = std::atanh(p.z / p_mag);
+                const auto p_mag = edm4eic::magnitude(p);
+                const auto p_phi = edm4eic::angleAzimuthal(p);
+                const auto p_eta = edm4eic::eta(p);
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
Thanks!

src/algorithms/pid/ParticlesWithPID.cc
```diff
@@ -92,6 +92,12 @@ namespace eicrecon {
                                    deta < m_cfg.etaTolerance &&
                                    dsphi < sinPhiOver2Tolerance;
 
+                // Matching kinematics with the static variables doesn't work at low angles and within beam divergence
+                // TODO - Maybe reconsider variables used or divide into regions
+                if ((p_eta < -5) && (edm4eic::eta(mom) < -5)) {
```
wdconinc, line 97: Should this equally treat far backward and far forward?

simonge: Probably, I have added this now.

wdconinc: @ajentsch @rahmans1 fyi; I think this only affects the B0 Acts tracks, right?

wdconinc (approved):

PR #915, opened 2023-09-05
Revert "fix: require EDM4eic 2.1 minimum"

veprbl:
Reverts eic/EICrecon#885

find_package seems to follow semver in the most strict sense: version 3.0.0 is assumed to be incompatible if we have asked for 2.1.0


wdconinc:
This is a setting (the default) of the CMake configuration that EDM4eic installs. If we want it to consider 3.0.0 compatible, we can make sure EDM4eic tells find_package that. I don't think that's what we want though, since we should treat 3.0.0 as incompatible...

wdconinc:
https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file 

We could use AnyNewerVersion instead of SameMajorVersion.

veprbl:
Great suggestion. Let me try.

veprbl:
Closing in favor of eic/EDM4eic#48

PR #919, opened 2023-09-05
Added LowQ2 plugin with hit digitization

simonge:
### Briefly, what does this PR introduce?
Adds a basic LowQ2 plugin which only digitizes the simulated hits using the standard silicon digitization factory

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
no

### Does this PR change default behavior?
Adds digitized lowq2 tagger hits to output

PR #921, opened 2023-09-05
Add simple tagger tracker hit clustering

simonge:
Adds hit clustering and hit point to the LowQ2 tagger reconstruction.
- [x] Requires #919 first.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
Adds clustered lowq2 tracker hit points to the output


src/algorithms/fardetectors/FarDetectorTrackerClusterConfig.h
```diff
@@ -0,0 +1,21 @@
...
+    std::string moduleField{"module"};
+    std::string layerField{"layer"};
+    std::string xField{"x"};
+    std::string yField{"y"};
+
+    int n_module{2};
+    int n_layer{4};
+
+    double time_limit{10};
+
```
wdconinc, line 19: It would be useful to add documentation in the comments as to what these values are representing, and what units they are in (`time_limit`).

src/detectors/LOWQ2/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+
```
wdconinc, line 1:
```suggestion
```

src/factories/fardetectors/FarDetectorTrackerCluster_factoryT.h
```diff
@@ -0,0 +1,77 @@
+// Created by Simon Gardner
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
```
wdconinc, line 3: This directory includes LGPL3 factories. They were written from scratch as multifactories, not based on the older factories elsewhere. If you used other factories in this directory as a model, the LGPL3 license applies and should be in the header.

src/algorithms/fardetectors/FarDetectorTrackerCluster.h
```diff
@@ -0,0 +1,41 @@
...
+
+  public:
+
+    /** One time initialization **/
+    void init(const dd4hep::Detector* det,
+              std::shared_ptr<const dd4hep::rec::CellIDPositionConverter>,
+              std::shared_ptr<spdlog::logger>);
+
+    /** Event by event processing **/
+    std::unique_ptr<edm4hep::TrackerHitCollection> produce(const edm4eic::RawTrackerHitCollection &inputhits);
```
wdconinc, line 26: Is it intentional that you skip the hit reconstruction step here? Normally we have the sequence `SimTrackerHit` -> `RawTrackerHit` -> `TrackerHit` -> `Cluster`. The hit reconstruction turns digitized hits back into something that has a position and energy deposition, not just an ADC channel under `charge`. This is relevant since the `charge` field includes a pedestal that will bias your clustering by over-weighting the lower energy outer hits in a cluster.

wdconinc (commented):
Main comment is that we likely want a hit reconstruction step before doing the clustering.

src/algorithms/fardetectors/FarDetectorTrackerCluster.h
simonge: There are several bits of the current early tracking chain and datamodel which I'm not very uncomfortable with, it feels like it's been built around getting the central silicon hits into the acts framework without much scope for generalisation. I'll try and summarize here but most points/questions don't belong in this PR.

1. A parameterization of charge sharing amongst pixels in silicon detectors (very technology dependent) needs to be added before digitization. Most of the current hits in the simulation will be singles but for e.g. AC-LGADs (I believe) this will be O(10+) more. [side note - is this being included in the DAQ rates?]. This factory is mostly a placeholder for when this can be introduced and different alternatives tested (including something through acts)
2.  The SiliconTrackerDigi algorithm doesn't allow for more than a single hit in a cell per event, this won't work with detectors with high occupancy over the few micro-second events, some integration time parameter could be added. Other configurable parameters for the digitization of each detector would be nice for studying but not necessary.
3.  The Clustering here is really an alternative to the TrackerHit step used in the rest of the trackers, in SiliconTrackerDigi there isn't currently any pedestal introduced and I would assume this would be subtracted and gain corrected by a calibration as part of a future digitization implementation. I don't see the benefit to converting back to positions and charge deposited for individual hits, clustering is done by cellid not position and the real TrackerHit energy and position can be a non-linear result of the clustering of neighbouring pixels.
4. The TrackerHitReconstruction algorithm doesn't currently feel like it's contributing anything, I feel like it it misusing the edm4eic::TrackerHit type as it contains the position and position error in two separate frames so if someone tried pass the collection somewhere else it isn't self consistent. Instead the edm4hep::TrackerHitPlane type looks like it keeps information on both local and global but is missing the nice timeError field.

I might be missing something else about how the data model is designed but it would be nice to keep some more relationships/associations along the way.

I'd be happy to be told I've interpreted something wrongly

TLDR; this is an alternative trackerHit step not skipping it. (Until there are improved digitization models)

simonge: I see the charge sharing is included in the document, I should have checked first.

wdconinc:
@simonge Are there additional changes you want to make to this in light of the tracking data model discussions we had 10 days ago?

wdconinc:
Would you be able to rebase this branch locally since the button here is not possible (but only if the conflicts are minor, which I suspect to be the case)?

simonge:
> @simonge Are there additional changes you want to make to this in light of the tracking data model discussions we had 10 days ago?

I don't think there was anything specific that can be changed here until any changes to the data model are made. I haven't compared efficiency of the algorithm to the previous juggler SimpleClustering one for calorimeters, but like that I would anticipate this being superseded once someone gets the chance to investigate further.

wdconinc:
> I don't think there was anything specific that can be changed here until any changes to the data model are made.

Who's working on those data model changes, i.e. drafting a PR of what would be required to change? I forget what we decided in our meeting about this. @sly2j ?

simonge:
> > I don't think there was anything specific that can be changed here until any changes to the data model are made.
> 
> Who's working on those data model changes, i.e. drafting a PR of what would be required to change? I forget what we decided in our meeting about this. @sly2j ?

I thought I'd drafted the data model PR but doesn't appear I have. The main change was to TrackerHit adding a vector of RawTrackerHit as associations and changing the position and position error to relate to the 2d positions on the plane, then adding the global 3d position too. I might have forgotten some other points we discussed.

I would additionally suggest changing the name of RawTrackerHit to RawTrackerSignal as it's not really a hit at this point but this comes more into the unexplored realm of detailed digitization.

I can't picture how the proposed Measurement2D is different or would be formed from multiple TrackerHits in the same tracker plane, I'll try and follow this up on the mattermost thread rather than here.

I'm hoping to have something to present at the meeting next Wednesday.

wdconinc:
While we're reviewing old PRs, how is this one doing? Should we also bring it up to date, taking advantage of the data model changes of the fall?

wdconinc:
Nevermind, this is actually rather up to date :-)

src/factories/fardetectors/FarDetectorTrackerCluster_factoryT.h
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023 Wouter Deconinck
```
wdconinc, line 2:
```suggestion
// Copyright (C) 2023 Simon Gardner
```

src/algorithms/fardetectors/FarDetectorTrackerClusterConfig.h
```diff
@@ -0,0 +1,20 @@
...
+    // Readout identifiers for dividing detector
+    std::string readout{""};
+    std::string moduleField{"module"};
+    std::string layerField{"layer"};
+    std::string xField{"x"};
+    std::string yField{"y"};
+
+    // Timing limit to add a hit to a cluster
+    double time_limit{10};
```
wdconinc, line 17: Can we use time units here?

simonge:
> While we're reviewing old PRs, how is this one doing? Should we also bring it up to date, taking advantage of the data model changes of the fall?

I had updated this one as a platform for testing out the templated splitting and merging factories.

src/algorithms/digi/CollectionCollector.h
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
```
veprbl, line 1: Use existing version from meta/ ?

src/algorithms/digi/SplitGeometry.h
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
```
veprbl, line 1: Use existing version from meta/ ?

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -30,6 +34,63 @@ extern "C" {
...
+
+        app->Add(new JOmniFactoryGeneratorT<FarDetectorTrackerCluster_factory>(
+            outputClusterTag,
+            {inputHitTag},
+            {outputClusterTag},
+            {
+              .readout = "TaggerTrackerHits",
+              .xField  = "x",
+              .yField  = "y",
+              .time_limit = 10 * dd4hep::ns,
```
veprbl, line 80: Same

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -30,6 +34,63 @@ extern "C" {
...
+         },
+         app
+      )
+    );
+
+    std::vector<std::string> layerClusters;
+
+    // Clustering of hits in each layer
+    for(int mod_id : moduleIDs){
+      for(int lay_id : layerIDs){
```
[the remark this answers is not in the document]
simonge: How expensive is splitting into the graph nodes? I feel this ideally would be split prior to digitization and remain split until tracks are made to avoid lots of unnecessary combinatoric comparisons. Although for some reason I'm apposed to splitting it up in the simulation readout.

I was considering dropping this clustering step in the analysis for the time being anyway. A geant4 hit is only associated to a single segmentation cell even if the track passes through several and there is no charge sharing implemented at the moment. As far as I'm aware this is also the only place using the ROOT RVec which is very convenient but might not be in line with the philosophy.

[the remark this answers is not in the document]
simonge: Alternatively to splitting the computational graph, the input and output collection list could both be variadic and looping over the collections within the algorithm. Would that be a neater solution?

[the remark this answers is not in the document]
veprbl: I don't know how bad are the combinatorics in this case. If this is done purely for optimization, this might be premature. Ideally, we go for the MVP first and add complexity as needed.

Regarding not needing clustering, let me know if we should skip this and review the linear fitting PR instead..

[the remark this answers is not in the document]
simonge: Might as well get it in if it's almost ready. It could provide a nice example to actually see the effects of implementing the charge sharing. I've swapped to looping over the collections inside the algorithm, although the other approach should still work too.

I realised I was just merging the clusters again at the bottom which was meant to originally test that the merger worked I've now removed that.

veprbl:
@simonge Do you think you could come up with a unit test for the algorithm? This should be similar to
https://github.com/eic/EICrecon/blob/main/src/tests/algorithms_test/calorimetry_CalorimeterIslandCluster.cc

simonge:
> @simonge Do you think you could come up with a unit test for the algorithm? This should be similar to https://github.com/eic/EICrecon/blob/main/src/tests/algorithms_test/calorimetry_CalorimeterIslandCluster.cc

I will try and do this more as an exercise for myself to understand the unit tests. I was anticipating this being replaced by an equivalent acts algorithm later if it contains the right kind of configuration for our detectors.

simonge:
> So you continue working on this PR or we merge this and then you refactor and contribute tests?

Tests and refactoring are now added.

veprbl (approved):
Can't really test - none of our DIS configurations seem to produce tagger hits (or we are missing them in geometry?).
Otherwise, LGTM.

simonge:
The Low-Q2 tagger acceptance needs events below Q2 <0.1 GeV. Do we not currently have a test using the min bias events?

veprbl:
> The Low-Q2 tagger acceptance needs events below Q2 <0.1 GeV. Do we not currently have a test using the min bias events?

No, but, could you please add some? It would be a great help with maintenance if we have those.

PR #922, opened 2023-09-06
fix: avoid including Navigator.hpp in tracking.cc

wdconinc:
Because JANA2 redefines `alignas` https://github.com/JeffersonLab/JANA2/blob/master/src/libraries/JANA/Utils/JResourcePool.h#L43-L46, we have been including an Acts header before any JANA headers in tracking.cc, just to get ACTS stuff loaded before JANA breaks it. Instead of loading Navigator (and all its Eigen dependencies) we can also just load the only relevant header: `boost/container/small_vector.hpp`. But really, this should be fixed in JANA, at https://github.com/JeffersonLab/JANA2/issues/238. `alignas` is a [formal part of the C++11 standard](https://en.cppreference.com/w/cpp/language/alignas) and should not be modified.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/global/tracking/tracking.cc
+++ b/src/global/tracking/tracking.cc
@@ -3,10 +3,9 @@
 //
 //
 
-#include <boost/container/small_vector.hpp>
-
 #include <JANA/JApplication.h>
 #include <JANA/JEvent.h>
+#undef alignas
 
 #include "extensions/jana/JChainFactoryGeneratorT.h"
 #include "extensions/jana/JChainMultifactoryGeneratorT.h"
--- a/src/global/tracking/tracking.cc
+++ b/src/global/tracking/tracking.cc
@@ -3,7 +3,7 @@
 //
 //
 
-#include <Acts/Propagator/Navigator.hpp>
+#include <boost/container/small_vector.hpp>
 
 #include <JANA/JApplication.h>
 #include <JANA/JEvent.h>
--- a/src/global/tracking/tracking.cc
+++ b/src/global/tracking/tracking.cc
@@ -5,7 +5,6 @@
 
 #include <JANA/JApplication.h>
 #include <JANA/JEvent.h>
-#undef alignas
 
 #include "extensions/jana/JChainFactoryGeneratorT.h"
 #include "extensions/jana/JChainMultifactoryGeneratorT.h"


src/global/tracking/tracking.cc
```diff
@@ -3,10 +3,9 @@
 //
 //
 
-#include <Acts/Propagator/Navigator.hpp>
-
 #include <JANA/JApplication.h>
 #include <JANA/JEvent.h>
+#undef alignas
```
veprbl, line 8: Let's maintain forward-compatibility:
```suggestion
// Workaround until https://github.com/JeffersonLab/JANA2/pull/239
#ifdef alignas
#undef alignas
#endif
```

wdconinc: "`#undef` has no effect if the name is not a macro." (https://gcc.gnu.org/onlinedocs/cpp/Undefining-and-Redefining-Macros.html)

wdconinc: This commit was mainly intended as a test to see if this approach even resolves the issue with running here. I think having to add `#undef` everywhere does not scale.

wdconinc: Test succeeded, by the way: https://github.com/eic/EICrecon/actions/runs/6102136482/job/16560729578?pr=922#step:5:862. 

So, at least adding `#undef` everywhere would be a solution. But then we need to add it **everywhere**.

wdconinc: I think the better approach is to continue our separation of JANA2 and Acts. What is ultimately the reason for the issue in this file (tracking.cc) is that `TrackerSourceLinker_factory` needs to include the header for the type of data it produces. That is still an Acts structure, not podio. So, we end up having to include a bunch of Acts headers, which is what ultimately leads to the problem. If we turn `TrackerSourceLinks` into podio types (see EDM4hep and its plugin in the newer Acts versions) then we avoid this problem because then JANA2 and Acts headers don't ever need to share a translation unit.

wdconinc: We can potentially forward declare our way out of this too, but I wasn't successful with that last night, before I really understood this issue.

wdconinc: Implemented in #933. We can now be sure `alignas` is not defined.

wdconinc:
```suggestion
```

veprbl (approved):

Issue #927, opened 2023-09-07
Compiler warning in src/tests/algorithms_test/pid_MergeParticleID.cc


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? eic-shell
- Any special versions of ROOT or Geant4? none

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Build EICrecon (maybe only with clang?).

### Expected Result: (what do you expect when you execute the steps above)
No warnings.

### Actual Result: (what do you get when you execute the steps above)
```
/home/wdconinc/git/EICrecon/.worktree/main/src/tests/algorithms_test/pid_MergeParticleID.cc:128:3: warning: non-void lambda does not return a value in all control paths [-Wreturn-type]
  };
  ^
/home/wdconinc/git/EICrecon/.worktree/main/src/tests/algorithms_test/pid_MergeParticleID.cc:140:47: note: in instantiation of function template specialization 'CATCH2_INTERNAL_TEST_0()::(anonymous class)::operator()<std::unique_ptr<edm4eic::CherenkovParticleIDCollection>, edm4eic::MutableTrackSegment>' requested here
    auto pid_0  = find_cherenkov_pid_for_track(result, tracks->at(0));
                                              ^
1 warning generated.
```

PR #929, opened 2023-09-07
fix: Compiler warning in src/tests/algorithms_test/pid_MergeParticleID.cc

c-dilks:
Resolve warning: non-void lambda does not return a value in all control paths.

- [x] Bug fix (issue #927)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no

no

### DIFF
--- a/src/tests/algorithms_test/pid_MergeParticleID.cc
+++ b/src/tests/algorithms_test/pid_MergeParticleID.cc
@@ -125,7 +125,7 @@ TEST_CASE("the PID MergeParticleID algorithm runs", "[MergeParticleID]") {
         return obj;
     }
     FAIL("ERROR: cannot find CherenkovParticleID given track");
-    if(coll->empty())
+    if(coll->size() == 0)
       throw std::runtime_error("empty collection used in pid_MergeParticleID::find_cherenkov_pid_for_track");
     return coll->at(0);
   };
--- a/src/tests/algorithms_test/pid_MergeParticleID.cc
+++ b/src/tests/algorithms_test/pid_MergeParticleID.cc
@@ -125,6 +125,9 @@ TEST_CASE("the PID MergeParticleID algorithm runs", "[MergeParticleID]") {
         return obj;
     }
     FAIL("ERROR: cannot find CherenkovParticleID given track");
+    if(coll->size() == 0)
+      throw std::runtime_error("empty collection used in pid_MergeParticleID::find_cherenkov_pid_for_track");
+    return coll->at(0);
   };
 
   // additive weights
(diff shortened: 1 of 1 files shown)


src/tests/algorithms_test/pid_MergeParticleID.cc
```diff
@@ -125,6 +125,9 @@ TEST_CASE("the PID MergeParticleID algorithm runs", "[MergeParticleID]") {
         return obj;
     }
     FAIL("ERROR: cannot find CherenkovParticleID given track");
+    if(coll->size() == 0)
```
wdconinc, line 128: How nitpicky should I be on code that should never run?
```suggestion
    if(coll->empty())
```

wdconinc (approved):

veprbl (dismissed):

src/tests/algorithms_test/pid_MergeParticleID.cc
```diff
@@ -125,6 +125,9 @@ TEST_CASE("the PID MergeParticleID algorithm runs", "[MergeParticleID]") {
         return obj;
     }
     FAIL("ERROR: cannot find CherenkovParticleID given track");
+    if(coll->empty())
```
veprbl, line 128: Surprisingly, despite PODIO declaring that they try to mimic STL collections, there is no `empty()`: https://github.com/AIDASoft/podio/blob/master/python/templates/Collection.h.jinja2

Let's revert
```suggestion
    if(coll->size() == 0)
```

veprbl (approved):

src/tests/algorithms_test/pid_MergeParticleID.cc
wdconinc: Ah shoot, yes, it's not even the first time I've suggested this in a code review and we've had to revert.

wdconinc: https://github.com/AIDASoft/podio/issues/476

veprbl:
Evicted from the merge queue: https://github.com/eic/EICrecon/actions/runs/6115159554/job/16598662329

wdconinc:
> Evicted from the merge queue

But clearly unrelated. This must be due to something else (and it must affects main too). Only other change is eic-shell got updated with the new EDM4eic data model, v3.0.1, and we now patch JANA2. My money is on the data model...

veprbl:
> > Evicted from the merge queue
> 
> But clearly unrelated. This must be due to something else (and it must affects main too). Only other change is eic-shell got updated with the new EDM4eic data model, v3.0.1, and we now patch JANA2. My money is on the data model...

Look at frame `#5` https://github.com/eic/EICrecon/actions/runs/6115159554/job/16598662329#step:5:715
could be a coincidence, of course

veprbl:
Though, indeed, this PR does not touch algorithm itself, and tests should not even get loaded in eicrecon-gun job.

PR #931, opened 2023-09-07
fix: remove non-POD-ness from OrthogonalTrackSeedingConfig

wdconinc:
The OrthogonalTrackSeedingConfig is now fully POD, which means that including the configuration does not incur the wrath/weight of all of Acts. The TrackSeeding algorithm now takes care of turning POD configs into the Acts config objects (much like how this is done in the other tracking algorithms, though the orthogonal seeding takes it to a new level of course).

Only necessary headers are now included in the POD config, and the headers in TrackSeeding.h and TrackSeeding.cc were reevaluated as well due to this shuffle. Two future FIXMEs have been added since the older Acts version we use has missing includes that have since been added. Check back when we get to Acts 23.0.0 and 27.2.0.

No output collection changes at all expected due to this PR.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @ShujieL @osbornjd 

No.

No.


osbornjd (dismissed):
Looks like a simple refactor to me, which looks fine as far as I can tell

wdconinc:
@osbornjd Thanks for the review. Had to request another one after fixing a conflict.

osbornjd (approved):
Merge conflicts don't change my review 😁

wdconinc:
> Merge conflicts don't change my review 😁

Tell github that :-)

PR #933, opened 2023-09-08
fix: trigger compilation error when `alignas` is a macro

wdconinc:
When `alignas` is overridden as a macro, we want to fail compilation. This allows us to ensure that this code is only used with patched JANA2 versions or future releases which don't define `alignas` as a macro.

- [x] Bug fix (issue https://github.com/JeffersonLab/JANA2/issues/238)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/utilities/eicrecon/eicrecon_cli.h
+++ b/src/utilities/eicrecon/eicrecon_cli.h
@@ -6,6 +6,11 @@
 
 #include <JANA/JApplication.h>
 
+#ifdef alignas
+// FIXME may be removed when minimum version in CMakeLists.txt includes the PR below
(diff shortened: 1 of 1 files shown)


src/utilities/eicrecon/eicrecon_cli.h
```diff
@@ -6,6 +6,11 @@
 
 #include <JANA/JApplication.h>
 
+#ifdef alignas
+// FIXME may be removed when minimum version in CMakeLists.txt includes the PR below
+#error JANA defines alignas macro; for patch see https://github.com/JeffersonLab/JANA2/pull/239
+#enfif
```
veprbl, line 12:
```suggestion
#endif
```

veprbl (approved):

PR #936, opened 2023-09-08
fix: replace GetCollectionBase with GetCollection<T>

wdconinc:
Back in the early JMultifactory days, we were not able to use `GetCollection<T>`, so we ended up with many `static_cast<const T::collection_t*>(GetCollectionBase())` workarounds. As of JANA 2.1.0 (https://github.com/JeffersonLab/JANA2/pull/215) we can now use the `GetCollection<T>` syntax.

- [x] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

  - @nathanwbrei Is there a particular reason for the difference between `event->GetCollection` defined on JEvent, but `SetCollection` as a JMultifactory method?

No.

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,7 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 
 # Check and print what JANA2 is used
-find_package(JANA 2.0.8 REQUIRED)
+find_package(JANA 2.1.0 REQUIRED)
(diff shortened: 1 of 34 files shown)


veprbl:
capybara on CI says this produces bunch of empty collections

wdconinc:
Ah, maybe this isn't ready for primetime then :-)

wdconinc:
Closing in favor of #1268

Issue #842, opened 2023-08-11
Add dependencies to RPATH by default?


veprbl:
Users may reasonably expect that the configuring dependencies provides not just compile-time header paths, but also the runtime library paths https://chat.epic-eic.org/main/pl/qsa4ya8qdpb35cztq6z6csqu1y

We could support that by providing and enabling some support for RPATH(RUNPATH preferred?).

wdconinc:
I'm working on this on the container side too. If we install EICrecon as all other packages, then we'll pick this up for free (since all spack packages are installed with rpath). It also has other benefits such as consistency (no dedicated build infrastructure in the container for just EICrecon) and better caching.

But we indeed need to support (and default to) rpath linking (and suitably adapted on macos too).

wdconinc:
In the container we now use rpath for all EICrecon libraries and plugins.

wdconinc:
Not quite closed. I think we want rpath linking for local installs too, right?

wdconinc:
Fixed by #938

PR #938, opened 2023-09-09
fix: enable rpath on build and install linking

wdconinc:
Enable rpath/runpath linking. The goal of this PR is to add the dependency libraries (JANA, podio, EDM4hep, EDM4eic) with their paths into rpath and runpath, and to have all internal library dependencies be explicit as well. Ultimately, no LD_LIBRARY_PATH should need to be set.

And the same on MacOS...

- [x] Bug fix (issue #842)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


CMakeLists.txt
```diff
@@ -82,6 +82,22 @@ if(NOT DEFINED PLUGIN_LIBRARY_OUTPUT_DIRECTORY)
...
+
+# Add the automatically determined parts of the RPATH
+# which point to directories outside the build tree to the install RPATH
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
+# This PROJECT_SOURCE_DIR at this point corresponds to repository root.
+# Saving it as EICRECON_SOURCE_DIR so can be used for jana_plugin.cmake
+# to correctly set plugin includes (and in other places as needed)
+set(EICRECON_SOURCE_DIR ${PROJECT_SOURCE_DIR})
+
```
veprbl, line 100: There is already EICRecon_SOURCE_DIR https://cmake.org/cmake/help/latest/command/project.html#command:project
and EICRECON_SOURCE_DIR from #441
```suggestion
```

CMakeLists.txt
```diff
@@ -82,6 +82,22 @@ if(NOT DEFINED PLUGIN_LIBRARY_OUTPUT_DIRECTORY)
 endif()
 message(STATUS "${CMAKE_PROJECT_NAME}: PLUGIN_LIBRARY_OUTPUT_DIRECTORY: ${PLUGIN_LIBRARY_OUTPUT_DIRECTORY}")
 
+# When building, don't use the install RPATH already
+# (but later on when installing)
+set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
```
veprbl, line 87: You explain what it does, but why not have rpath available in the build directory?

CMakeLists.txt
```diff
@@ -82,6 +82,22 @@ if(NOT DEFINED PLUGIN_LIBRARY_OUTPUT_DIRECTORY)
...
+set(CMAKE_INSTALL_RPATH
+    "${CMAKE_INSTALL_PREFIX}/${PLUGIN_LIBRARY_OUTPUT_DIRECTORY};${CMAKE_INSTALL_PREFIX}/${PLUGIN_OUTPUT_DIRECTORY}"
```
veprbl, line 89: We can drop JANA_PLUGIN_PATH after that?

veprbl (dismissed):

CMakeLists.txt
wdconinc: Ah, right. Copy/paste mistake.

CMakeLists.txt
veprbl: I see. Looks like they don't do dlopen without a full path, so rpath can't apply. That is not very convenient, IMO.

wdconinc: It might be useful to come up with a proposal as to how JANA2 should find/load plugins if/when `JANA_PLUGIN_PATH` is not set. One way could be to use the rpath/runpath of the current executable. Another may be requiring LD_PRELOAD or something.

In any case, I'm going to resolve this conversation.

CMakeLists.txt
veprbl: True, we don't want our CMAKE_INSTALL_PREFIX to be in CMAKE_BUILD_RPATH. I guess, running from build directory might be unusable also because it doesn't have a structure that JANA_PLUGIN_PATH can point to. Thanks for explaining. Marking this resolved.

veprbl (approved):
This does not break anything on macOS (although also not needed there - all libraries are resolved by full "install name" path).

wdconinc:
Sounds a bit like this experimental feature in spack: https://spack.readthedocs.io/en/latest/config_yaml.html#shared-linking-bind

That is something we've looked into because of the cvmfs-mounted singularity images. Why link against a remote filesystem's generic /use/local/lib (even as rpath) when you can avoid the directory lookup with a direct path?

I haven't looked into how it is implemented, but I think it's some elfutils rewriting that is needed.

PR #939, opened 2023-09-09
feat(ci): no cancel-in-progress on main

wdconinc:
This will prevent main pipelines from being interrupted by newer merges. This is becoming relevant in times of frequent merges. Because we do one or two extra things on main (like not just running clang-tidy on diffs only), the merge queue will complete faster than the previous main pipeline, so we risk just canceling every main pipeline...

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -12,7 +12,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref_name == 'main' }}
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref_name == 'main' }}
```
wdconinc, line 15:
```suggestion
  cancel-in-progress: ${{ github.ref_name != 'main' }}
```

veprbl (approved):
Good idea!

PR #940, opened 2023-09-09
feat: allow configuring WITH_SHARED_LIBRARY and WITHOUT_PLUGIN

wdconinc:
Allow to create shared instead of static algorithms libraries, and allow skipping the creation of an algorithms plugin.

Depends on:
- [x] https://github.com/eic/EICrecon/pull/942 (cherry-picked)
- [x] https://github.com/eic/EICrecon/pull/938 (cherry-picked)

- [x] Bug fix (issue #628 again)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
@veprbl Any idea why llvm-cov fails when we don't build the algorithms as plugins? The rest of the reconstruction doesn't use the algorithms as plugins but through the libraries.

wdconinc:
Btw, this is the solution to removing the duplicate objects.

wdconinc:
Comparison this branch (https://github.com/eic/EICrecon/actions/runs/6140234598/job/16658926482) vs main (https://github.com/eic/EICrecon/actions/runs/6138018645):
- build gcc release duration: 24 min 46 sec vs 32 min 51 sec
- build gcc release install artifact: 194 MB in 280 files vs 406 MB in 287 files (both after compression, 654 MB vs 1.63 GB before compression)

veprbl (approved):
Compiles on macOS. Ship it.

PR #942, opened 2023-09-10
fix(ci): don't use pre-filled JANA_PLUGIN_PATH with EICrecon plugins

wdconinc:
`JANA_PLUGIN_PATH` in eic-shell (and therefore CI environments) is prepopulated with `/usr/local/lib/EICrecon/plugins:/usr/local/plugins`. This means that removing (intentionally or otherwise) a plugin in a PR will still cause it to be found in `/usr/local/lib/EICrecon/plugins`. We need `/usr/local/plugins` though, because that contains the `janadot` plugin to create callgraphs.

There's an issue reported to JANA2 (https://github.com/JeffersonLab/JANA2/issues/136) that points out that installing plugins in a shared namespace at `/usr/local/plugins` may lead to issues and does not conform to the linux FHS. But at least EICrecon plugins are not installed there, otherwise we would have exactly this issue.

- [x] Bug fix (issues, potentially)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -310,7 +310,7 @@ jobs:
         setup: /opt/detector/setup.sh
         run: |
           export DETECTOR_CONFIG=${DETECTOR}_${{ matrix.detector_config }}
-          export JANA_PLUGIN_PATH=$PWD/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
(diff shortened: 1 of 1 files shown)


wdconinc:
This will avoid mistakes like https://github.com/eic/EICrecon/actions/runs/6137527478/job/16653814205?pr=940#step:5:126 sneaking through without failures.

veprbl (approved):
Makes sense.

PR #943, opened 2023-09-11
fix: install EICrecon libraries by default in `lib/EICrecon/libraries`, not `lib`

wdconinc:
Instead of installing EICrecon libraries into `lib` (essentially `/usr/local/lib`), we are polluting a global namespace without providing an interface definition. The libraries are only valid in the context of plugins that use them. In that sense they are not suitable for `lib` but are better treated as custom library objects that should not be assumed to be usable by other programs.

- [x] Bug fix (issue: don't install into lib/)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl:
This is not FHS-compliant. The only way to not pollute the library namespace is to give them better names (libeicrecon_log.so?).

wdconinc:
It is acceptable under the FHS: https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s06.html:
> Applications may use a single subdirectory under /usr/lib. If an application uses a subdirectory, all architecture-dependent data exclusively used by the application must be placed within that subdirectory. 
I think we fall well within "exclusively used" here...

Better naming would also a good option. I wouldn't mind having libalgorithms_*, libeicrecon_*, etc... It's mainly liblog and libpodio which are clashing.

PR #954, opened 2023-09-13
pr-backport.yml: add github token

veprbl:
Can't use github user to submit PRs with Github Actions

This is needed to avoid situations like #952


veprbl:
Demo: https://github.com/veprbl/EICrecon/pull/10

wdconinc (approved):
Thanks for figuring this out. Do we want the backport token locally per repository or is there a benefit to having it organization-wide?

PR #955, opened 2023-09-13
fix(ci): use npsim instead of ddsim

wdconinc:
### Briefly, what does this PR introduce?
We are trying to exercise the whole pipeline, but not throwing any optical photons. This means the dRICH does not currently see anything and we don't test the reconstruction and particle matching.

This is expected to result in 1) changes, 2) slower simulation running, 3) dragons.

### What kind of change does this PR introduce?
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #961, opened 2023-09-13
fix: use xyz_g[3] in richgeo/ReadoutGeo.cc

wdconinc:
This fixes the memory overrun in e.g. https://github.com/eic/EICrecon/actions/runs/6176955368/job/16769656988.

Fixes: c9ceddc0275c911065d9c3658a51d646264ca63e ('feat: SiPM pixel gap cuts')

### DIFF
--- a/src/services/geometry/richgeo/ReadoutGeo.cc
+++ b/src/services/geometry/richgeo/ReadoutGeo.cc
@@ -119,7 +119,7 @@ dd4hep::Position richgeo::ReadoutGeo::GetSensorLocalPosition(CellIDType cellID, 
   auto context = m_cellid_converter->findContext(cellID);
 
   // transformation vector buffers
-  double xyz_l[3], xyz_e[3], xyz_g[2];
+  double xyz_l[3], xyz_e[3], xyz_g[3];
   double pv_g[3], pv_l[3];
 
   // get sensor position w.r.t. its parent


wdconinc:
Cherry-picked from #955 since it fixes the error, but the CI changes aren't needed in v1.5.

c-dilks (approved):

c-dilks:
Good catch. I'm not comfortable with how `ReadoutGeo::GetSensorLocalPosition` is written... there must be a simpler way to do this.

The only test of the application of this method that we have is https://github.com/eic/drich-dev/blob/main/src/test_pixel_gap_cuts.cpp which runs in `drich-dev` CI workflows. It should be moved to `reconstruction_benchmarks`, but I never did this since we didn't have EICrecon-dependency there. Maybe it could be a unit test here, but one needs to look at the output image.

A simpler way to test this is to turn on/off the pixel gap cuts for the dRICH; you should have ~12% less photoelectrons with the cuts on, but this doesn't tell you if the gap mask geometry is correct.

wdconinc:
I'll leave that up to @chchatte92 to consider for future algorithmic and CI improvements. At least this fix should get us going again in production... Thanks for approving!

chchatte92:
Hello colleagues, I am out of Trieste for a couple of days and I don't have my Computer with me. I will try to when I am back on Monday.

chchatte92:
Hello colleagues, I am out of Trieste for a couple of days and I don't have my Computer with me. I will try to when I am back on Monday.

epic-capybara:
Successfully created backport PR for `v1.5`:
- #964

c-dilks:
> The only test of the application of this method that we have is https://github.com/eic/drich-dev/blob/main/src/test_pixel_gap_cuts.cpp which runs in `drich-dev` CI workflows. It should be moved to `reconstruction_benchmarks`, but I never did this since we didn't have EICrecon-dependency there. Maybe it could be a unit test here, but one needs to look at the output image.

test results still look good! Pixel gap masking is the only main thing we use this method for. (@chchatte92 this is for reference, showing what it's supposed to look like)

2023-09-14-134355_930x668_scrot

PR #964, opened 2023-09-14
[Backport v1.5] fix: use xyz_g[3] in richgeo/ReadoutGeo.cc

epic-capybara:
# Description
Backport of #961 to `v1.5`.

PR #966, opened 2023-09-14
linux-eic-shell.yml: enable on maintenance branches

veprbl:
Addresses an issue brought up in https://github.com/eic/EICrecon/pull/964#issuecomment-1720001570

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - 'v1.*' # maintenance branches
     tags:
       - '*'
   pull_request:
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - 'v1.*' # maintenance branches
```
wdconinc, line 7:
```suggestion
      - 'v*.*' # maintenance branches
```
?

veprbl: Apparently, it even supports some regex-like https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
```suggestion
      - 'v[0-9]+.[0-9]+' # maintenance branches
```

wdconinc: Yeah, I wasn't sure. I thought it just did limited globbing.

wdconinc (approved):

epic-capybara:
Git push to origin failed for v1.5 with exitcode 1

PR #975, opened 2023-09-15
feat(ci): run all CI pipelines in multi-threaded mode

wdconinc:
We get 2 cores on (default) GitHub actions nodes. This enables eicrecon multi-threading on both cores. It will hopefully surface any multi-threading issues soon enough (preferably in this PR...). It will also help us assess how much faster two-threaded running is compared to single-threaded (local tests indicate there is no benefit in wall-time, so...).

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Not sure that two threads are having any impact here, and it is also not reflected in the performance summary. If it takes the same time on two threads as on one thread, then I'd expect the efficiency to be half for two threads. @nathanwbrei ?

### Two threads 
https://github.com/eic/EICrecon/actions/runs/6992069680/job/19023290209#step:7:2747
Screenshot_20231125-184132

### One thread
https://github.com/eic/EICrecon/actions/runs/6991121189/job/19021445198#step:7:2831
Screenshot_20231125-184113

nathanwbrei:
Hey, at least it didn't crash horribly :D 

Often these cloud environments use cgroups to restrict your process to one physical cpu even if `lscpu` claims you have a bunch. JANA will launch however many threads you tell it, or how many threads it thinks it can use, but if the OS restricts you there won't be a speedup. Should be able to test this by benchmarking JTest: `jana -Pplugins=JTest -b`.

Or there's still a "barrier event" being imposed somewhere somehow. I'd check the event source for a `SetBarrierEvent()`, or a `FinishEvent()` override that does something tricky, or us setting the `jana:limit_events_in_flight` parameter.

wdconinc:
GitHub actions exposes two cores for use in CI pipelines, ref: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.

wdconinc:
Confirmed that the throughput should double from 1 to 2 threads, https://github.com/eic/EICrecon/actions/runs/6992832850/job/19024861801#step:3:167.

wdconinc:
No `FinishEvent`, no `SetBarrierEvent`, no `limit_events_in_flight`.

wdconinc:
There is the `mutex` that covers the entire JEventProcessorPODIO Process stage... I suspect we can narrow that scope.

nathanwbrei:
> There is the `mutex` that covers the entire JEventProcessorPODIO Process stage... I suspect we can narrow that scope.

That would do it! At first glance the only things we should be protecting are:
```c++
    if (m_is_first_event) {
        FindCollectionsToWrite(event);
    }
```
But sadly also this, at the bottom:
```
    m_writer->writeFrame(*frame, "events", m_collections_to_write);
    m_is_first_event = false;
```

There's a whole fun discussion waiting to be had about how we can remove that mutex entirely. Which is something I really want to do, simply for the sake of "good design". Of course our complicated `CollectionsToWrite` logic makes this a lot harder. I would love to refactor the whole method, but we'd have to figure out which of the weird choices we made are requirements, and which were hacks to get the thing working.

wdconinc:
#1069 goes a long way.

wdconinc:
Against #1069 I'm having some luck with
```diff
diff --git a/src/services/io/podio/JEventProcessorPODIO.cc b/src/services/io/podio/JEventProcessorPODIO.cc
index 0a275360..1001f2d0 100644
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -271,8 +271,9 @@ void JEventProcessorPODIO::FindCollectionsToWrite(const std::shared_ptr<const JE
 
 void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent> &event) {
 
-    std::lock_guard<std::mutex> lock(m_mutex);
+    std::unique_lock lock(m_mutex, std::defer_lock);
     if (m_is_first_event) {
+        lock.lock();
         FindCollectionsToWrite(event);
     }
 
@@ -295,8 +296,10 @@ void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent> &event) {
     // Note that collections MUST be present in frame. If a collection is null, the writer will segfault.
     const auto* frame = event->GetSingle<podio::Frame>();
 
+    lock.lock();
     m_writer->writeFrame(*frame, "events", m_collections_to_write);
     m_is_first_event = false;
+    lock.unlock();
 
     // Print the contents of some collections, just for debugging purposes
```
That doesn't remove mutex, but avoid locks during the algorithms, and only locks when accessing shared resources.

wdconinc:
Ready for review. I think this is now sufficiently stable for merging into main.
- running without errors is not enforced for MT test runs since we may have some issues that may only show up periodically, and we don't want to break main reference workflows
- MT running will not be using the same PRNG sequence, so statistical consistency only is expected in capybara diffs (which are taken with respect to single threaded reference)

simonge:
Looks good and I think I understand the capybara differences, although I'm not certain how to approach reviewing this. I will just ask some questions which might be more JANA/performance related than perhaps this PR.

At what level is this parallelizing the running? What is happening to the random generator, is it getting copied and using the same/different seed for each thread or getting shared?

Why does the PhysicsEventSource appear to report worse performance?

https://github.com/eic/EICrecon/actions/runs/15087557296/job/42412094018#step:7:1462
https://github.com/eic/EICrecon/actions/runs/15087557296/job/42412094007#step:7:1421

wdconinc:
> At what level is this parallelizing the running?

JANA2 creates a full set of factories for each thread, which analyze individual events. Each thread therefore analyzes full events independently of other threads. Reading (event source) and writing (event processor) is done serially in a single main thread and needs to be guarded. See e.g. https://jeffersonlab.github.io/JANA2/#/concepts?id=multithreading-and-factories for a graphic.

In the JANA2 code this looks like this: https://github.com/JeffersonLab/JANA2/blob/133cd22d3b3847b23dbd5f6a8569bcf47bf25b99/src/libraries/JANA/JApplication.cc#L198-L199
- algorithms are initialized with their provided parameters,
- then the workers are scaled to the number of threads.

The scaling itself is implemented at https://github.com/JeffersonLab/JANA2/blob/133cd22d3b3847b23dbd5f6a8569bcf47bf25b99/src/libraries/JANA/Engine/JExecutionEngine.cc#L102.

> What is happening to the random generator, is it getting copied and using the same/different seed for each thread or getting shared?

There are a few different approaches right now to the random generators. Mostly, we have algorithms that own their own random generator, which is going to be local to each thread. The only algorithm that runs by default with a fixed seed (PhotoMultiplierHitDigi) will run with the same sequence of random numbers for all threads.

There are some algorithms that have initial support for RandomSvc, which is intended to provide reproducible random numbers for events, even if it is queried from different threads independently. That's to some extent planned for the `algorithms` interface itself, but it is also possible we just implement it within EICrecon directly.

> Why does the PhysicsEventSource appear to report worse performance?

I don't want to over-interpret, but what you're looking at is the latency, so how long before the event is provided starting from the call to it (i.e. read time). Since event reading is serialized, in some cases another thread will already be reading an event, and the caller will not just get latency from read time, but also from wait time.

When we go to more and more threads, there will presumably be more latency on reading (while, in ideal scaling) there will be equal time between reads within a thread. When we get to the point where read latency is so bad that it dominates the throughput, then we will be spending more time waiting for other events to read than benefiting from multi-threading.

simonge:
> When we go to more and more threads, there will presumably be more latency on reading (while, in ideal scaling) there will be equal time between reads within a thread. When we get to the point where read latency is so bad that it dominates the throughput, then we will be spending more time waiting for other events to read than benefiting from multi-threading.

Off topic again - The reason I was looking at this is the event source reading is often a huge bottleneck in running local tests of small subsets of algorithms. Compared to equivalent analysis directly in ROOT it is significantly slower. It looks like the most recent version of podio v1.3 might be able to significantly speed up these tests by allowing reading of only the requested collections rather than the whole frame.

wdconinc:
> It looks like the most recent version of podio v1.3 might be able to significantly speed up these tests by allowing reading of only the requested collections rather than the whole frame.

That would need some kind of support in JANA2 that may not be trivial.

simonge:
> > It looks like the most recent version of podio v1.3 might be able to significantly speed up these tests by allowing reading of only the requested collections rather than the whole frame.
> 
> That would need some kind of support in JANA2 that may not be trivial.

Would you not just need to pass getEntry the list of collection names in the graph JANA expects to come from the source?

wdconinc:
> Would you not just need to pass getEntry the list of collection names in the graph JANA expects to come from the source?

Yeah, but JANA doesn't know which collection names are needed in advance. In essence, JANA only knows which output collections are going to be produced. The rest is derived through the execution graphs. Resolving in advance which input collections are needed likely means running a full 1st event and assuming that the rest is similar. I assume we read a full frame into memory for that reason. Maybe we only read a collection into memory whenever it is (first) needed, but we might be losing on block reads what we gain in reduced reads for all practical scenarios.

src/algorithms/pid/IrtCherenkovParticleID.cc
```diff
@@ -170,6 +170,7 @@ void IrtCherenkovParticleID::process(const IrtCherenkovParticleID::Input& input,
     auto irt_particle = std::make_unique<ChargedParticle>();
 
     // loop over radiators
+    std::lock_guard<std::mutex> lock(m_pid_radiators_mutex);
```
simonge, line 173: Could you add a comment summarizing what this is protecting

simonge (commented):
Everything looks reasonable in the code, diffs and output from the checks. Someone else with understanding between what the different sanitizers achieve might also need to comment for approval.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -530,23 +512,14 @@ void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent>& event) {
...
-
-  // TODO: NWB: We need to actively stabilize podio collections. Until then, keep this around in case
-  //            the writer starts segfaulting, so we can quickly see whether the problem is unstable collection IDs.
-  /*
-    m_log->info("Event {}: Writing {} collections", event->GetEventNumber(), m_collections_to_write.size());
-    for (const std::string& collname : m_collections_to_write) {
-        m_log->info("Writing collection '{}' with id {}", collname, frame->get(collname)->getID());
-    }
-    */
+  lock.lock();
```
veprbl, line 520: Can we turn this into guard_lock if we add a scope?

wdconinc: Yes, this use of a deferred lock is probably largely historical now.

wdconinc:
> Nice! The code and diffs look good to me. Like @simonge, I'm not quite sure what to expect from the different sanitizers, but I think this is good from my perspective after all the checks pass!

We've been working with the ROOT devs on the failing checks.

wdconinc:
> > [...], but I think this is good from my perspective after all the checks pass!
> 
> We've been working with the ROOT devs on the failing checks.

After the latest ROOT upgrade with patch that should resolve the data race, this seems to pass again. I'll rerun it a few times just in case; these are not deterministic bugs.

wdconinc:
The ROOT cling write lock issue has been worked around here, and we're rolling out a patch. I think this PR is now stable enough for production.

ruse-traveler (dismissed):
Still looks good to me!

src/utilities/eicrecon/eicrecon.cc
```diff
@@ -54,6 +55,8 @@ std::vector<std::string> EICRECON_DEFAULT_PLUGINS = {
 };
 
 int main(int narg, char** argv) {
+  ROOT::EnableThreadSafety();
```
veprbl, line 58: We might want to make this conditional on nthreads, but it doesn't need to be this PR.

wdconinc: I don't know if we can do this early enough in EICrecon. `nthreads` is a JApplication parameter, and immediately after it is parsed, the topology is created and execution engine splits off the threads. I don't see how we can call `ROOT::EnableThreadSafety()` conditionally on `nthreads`, but still before we split those threads. Short of doing our own argv parsing.

I think the cost should be minimal in single-threaded running, since by definition we should not have any waiting due to the mutexes.

wdconinc: https://github.com/eic/EICrecon/issues/1905

PR #978, opened 2023-09-16
fix: use RPATH in build tree, unset LD_LIBRARY_PATH for ctest

wdconinc:
This will (hopefully) fix the ctest failures in CI. We were not explicit about how we want rpath to be handled in the build tree (now we are), but more importantly on binaries the rpath settings in CMake set the RUNPATH elf record, which is lower precedence than the entries in LD_LIBRARY_PATH. That means that any executable (such as the test executable) will still defer to LD_LIBRARY_PATH, if set. So, we unset it for ctest.

This doesn't affect our other executable, eicrecon, because that isn't actually linked against anything (or nothing that is internal to our source tree). Instead it uses the dlopen to plugins explicitly in JANA_PLUGIN_PATH. When those are opened, the rpath to their dependencies is respected.

- [x] Bug fix (issue, see also #976)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -74,7 +74,8 @@ jobs:
       with:
         platform-release: "jug_xl:nightly"
         run: |
-          (cd build && ctest -V)
+          # Disable LD_LIBRARY_PATH to ensure R(UN)PATH use
+          (cd build && LD_LIBRARY_PATH="" ctest -V)
     - uses: actions/upload-artifact@v3
       with:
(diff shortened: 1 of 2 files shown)


wdconinc:
Seems to have worked https://github.com/eic/EICrecon/actions/runs/6208094301/job/16853929886?pr=978

wdconinc:
Doesn't quite explain why this worked before. There is likely still something wrong here but at least it gets of pipelines going again.

veprbl (approved):

wdconinc:
> Doesn't quite explain why this worked before.

We used to use static libraries, not shared libraries. So everything got resolved before it was installed through inclusion in the plugins themselves.

PR #984, opened 2023-09-17
fix: avoid undefined unsigned integer behavior in pid size checks

wdconinc:
This avoids undefined behavior flagged by ubsan in the statement `std::size_t n_tracks = -1;` and in the signed/unsigned integer comparisons in the pid system, which caused at least MergeTracks to always fail. It also adds output on the problematic sizes, e.g.
```
[DRICH:DRICHMergedTracks] [error] cannot merge input track collections with different sizes 3, 5
[DRICH:DRICHIrtCherenkovParticleID] [error] radiators have differing numbers of TrackSegments 3, 5, 0
```

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, should now attempt MergeTracks.

### DIFF
--- a/src/algorithms/pid/IrtCherenkovParticleID.cc
+++ b/src/algorithms/pid/IrtCherenkovParticleID.cc
@@ -134,7 +134,7 @@ std::map<std::string, std::unique_ptr<edm4eic::CherenkovParticleIDCollection>> e
 
   // loop over charged particles ********************************************
   m_log->trace("{:#<70}","### CHARGED PARTICLES ");
-  std::size_t num_charged_particles = in_charged_particle_size_distribution.begin()->second;
+  std::size_t num_charged_particles = in_charged_particle_size_distribution.begin()->first;
   for(long i_charged_particle=0; i_charged_particle<num_charged_particles; i_charged_particle++) {
     m_log->trace("{:-<70}", fmt::format("--- charged particle #{} ", i_charged_particle));
 
--- a/src/algorithms/pid/IrtCherenkovParticleID.cc
+++ b/src/algorithms/pid/IrtCherenkovParticleID.cc
@@ -3,6 +3,8 @@
 
 #include "IrtCherenkovParticleID.h"
 
+#include <fmt/ranges.h>
+
 // AlgorithmInit
 //---------------------------------------------------------------------------
 void eicrecon::IrtCherenkovParticleID::AlgorithmInit(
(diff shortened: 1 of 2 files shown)


c-dilks:
> ```[DRICH:DRICHIrtCherenkovParticleID] [error] radiators have differing numbers of TrackSegments 3, 5, 0```

Do we understand what causes these now, [e.g. in `dis` jobs](https://github.com/eic/EICrecon/actions/runs/6215248891/job/16868173379#step:5:695)?

I suspect these are charged particles which are outside the aerogel acceptance, but still inside the gas acceptance. After https://github.com/eic/EICrecon/pull/735 we simply don't have aerogel `TrackkPoint`s for such tracks, hence the error (prior behavior was to propagate and fail, but still produce a track with zero `TrackPoints` (wasting time))

c-dilks:
There was a bit of uncertainty whether we care about dRICH PID for such tracks or not... seems at some point we stopped caring and forgot to test this region.

chchatte92:
This region is critical. Namely, there the spherical abberation is large so we have to limit ourselves only low momentum PID from gas. The low p tracks will have larger multiple scattering for larger angles. Hence Chris and myself discussed and decided to give low priority to the region.

chchatte92:
Can we get an have an estimate of the eta of the tracks?

src/algorithms/pid/IrtCherenkovParticleID.cc
```diff
@@ -117,20 +119,22 @@ std::map<std::string, std::unique_ptr<edm4eic::CherenkovParticleIDCollection>> e
...
+    std::transform(in_charged_particles.begin(), in_charged_particles.end(),
+      std::back_inserter(in_charged_particle_sizes),
+      [](const auto& in_charged_particle) { return in_charged_particle.second->size(); });
+    m_log->error("radiators have differing numbers of TrackSegments {}", fmt::join(in_charged_particle_sizes, ", "));
+    return result;
   }
 
   // loop over charged particles ********************************************
   m_log->trace("{:#<70}","### CHARGED PARTICLES ");
+  std::size_t num_charged_particles = in_charged_particle_size_distribution.begin()->second;
```
c-dilks, line 137:
```suggestion
  std::size_t num_charged_particles = in_charged_particle_size_distribution.begin()->first;
```

c-dilks:
> Can we get an have an estimate of the eta of the tracks?

Use the first (or any) trackpoint's (x,y,z). I believe there is trace logging in `TrackPropagator` that could be used, or just add a quick test here in `IrtCherenkovParticleID` just to see (`if error, print eta`).

c-dilks:
This is a bad issue though: if there are _any_ tracks which only pass through gas and not aerogel, no dRICH PID will be done for the event!

c-dilks:
A quick fix to restore behavior before #735 would be to write out a `TrackSegment` with zero `TrackPoints` for such tracks, if and only if the track would pass through the backplane of the _gas_; see the filter surface implementation in #735, which depends on the radiator (aerogel has smaller radius than gas, try making both filters be the gas back plane disc).

wdconinc:
To the PR hijackers: I think this would make for an excellent discussion in a dedicated issue ;-)

c-dilks:
> To the PR hijackers: I think this would make for an excellent discussion in a dedicated issue ;-)

Of course! I haven't done any testing, just thinking aloud here. @chchatte92 can you open an issue and take care of it, if there indeed is one?

c-dilks (approved):

PR #986, opened 2023-09-17
feat(ci): correct LD_LIBRARY_PATH prefix and add job step to print `ldd *.so`

wdconinc:
This adds a job step in `build` and in `eicrecon-gun` to print `ldd` for all regular libraries and plugins.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/asan.supp
+++ b/.github/asan.supp
@@ -1,2 +1,2 @@
 interceptor_via_lib:libCore.so
-interceptor_via_fun:podio::ROOTFrameReader::openFiles
+interceptor_via_lib:libpodioRootIO.so
--- a/.github/asan.supp
+++ b/.github/asan.supp
@@ -1,1 +1,2 @@
 interceptor_via_lib:libCore.so
+interceptor_via_fun:podio::ROOTFrameReader::openFiles
--- a/.github/asan.supp
+++ b/.github/asan.supp
@@ -1,2 +1,1 @@
 interceptor_via_lib:libCore.so
-interceptor_via_lib:libpodioRootIO.so
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -69,6 +69,13 @@ jobs:
           # install this repo
           CC="${{ matrix.CC }}" CXX="${{ matrix.CXX }}" CXXFLAGS="${{ matrix.CXXFLAGS }}" cmake -B build -S . -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ matrix.CMAKE_BUILD_TYPE }} -DUSE_ASAN=ON -DUSE_TSAN=OFF -DUSE_UBSAN=OFF
           cmake --build build -- -j 2 install
+    - name: Check dynamic library loader paths
+      uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: /opt/detector/setup.sh
+        run: |
+          ldd lib/*.so lib/EICrecon/plugins/*.so
     - name: Run testsuite
       uses: eic/run-cvmfs-osg-eic-shell@main
       with:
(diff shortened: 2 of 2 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -69,6 +69,16 @@ jobs:
...
           cmake --build build -- -j 2 install
+    - name: Check dynamic library loader paths
+      uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: /opt/detector/setup.sh
+        run: |
+          for lib in lib/*.so lib/EICrecon/plugins/*.so ; do
+            readelf -d $lib
+            ldd $lib
```
wdconinc, line 80:
```suggestion
            ldd -r $lib
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -306,6 +316,13 @@ jobs:
       with:
         name: sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root
     - uses: cvmfs-contrib/github-action-cvmfs@v3
+    - name: Check dynamic library loader paths
+      uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: /opt/detector/setup.sh
+        run: |
+          ldd lib/*.so lib/EICrecon/plugins/*.so
```
wdconinc, line 325:
```suggestion
          ldd -r lib/*.so lib/EICrecon/plugins/*.so
```

wdconinc:
No more incorrect libraries loaded now.

.github/workflows/linux-eic-shell.yml
```diff
@@ -306,13 +317,22 @@ jobs:
       with:
         name: sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root
     - uses: cvmfs-contrib/github-action-cvmfs@v3
+    - name: Check dynamic library loader paths
+      uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: /opt/detector/setup.sh
+        run: |
+          export LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH
```
veprbl, line 326: `export LD_LIBRARY_PATH=${LD_LIBRARY_PATH//\/usr\/local\/lib}`

wdconinc: 1. We can't filter out `/usr/local/lib` entirely since it then doesn't find dd4hep and podio
2. We can't filter out with this pattern because it also mangles `/usr/local/lib/root` which is required for root
3. We can't construct LD_LIBRARY_PATH ourselves entirely because it needs the part in /opt/detector/epic-nightly/lib (which we can hardcode, but probably not ideal)

wdconinc: I was trying to construct it from nothing, and ended up with
```
export LD_LIBRARY_PATH=$PWD/lib:/opt/detector/epic-nightly/lib:$(spack location -i podio)/lib:$(spack location -i dd4hep)/lib:/usr/local/lib/root
```

wdconinc: That just looked more like a headache to maintain... even for a short time.

wdconinc: Actually, what I had tried before is:
```
export LD_LIBRARY_PATH=$PWD/lib:/opt/detector/epic-${DETECTOR_VERSION}/lib:$(spack location -i podio)/lib:$(spack location -i dd4hep)/lib:$(spack location -i root)/lib/root
```
which works, but it just doesn't look that nice (especially not to copy it multiple times in the code, and we can't even use it in the `env:` section on account of the `$( )` calls).

veprbl (dismissed):

.github/asan.supp
```diff
@@ -1 +1,2 @@
 interceptor_via_lib:libCore.so
+interceptor_via_lib:libpodioRootIO.so
```
wdconinc, line 2:
```suggestion
```

veprbl (approved):

PR #987, opened 2023-09-17
feat: enable ubsan by default in CI, with print_stacktrace=1 and suppressions

wdconinc:
This enables ubsan by default in CI, with some more undefined behaviors (float-divide-by-zero, unsigned-integer-overflow, local-bounds). It has two suppressions already, but more will no doubt be needed to limit the output to what we control. There is no way to have an exitcode != 0 when undefined behavior is found, so this depends a bit on people checking the output periodically.

TODO:
- [x] #989


wdconinc:
One more spurious ubsan stack trace at https://github.com/eic/EICrecon/actions/runs/6224924646/job/16895351952?pr=987#step:5:608

veprbl (approved):

PR #989, opened 2023-09-18
fix: avoid division by zero by not setting weighted position if sum of weights is zero

wdconinc:
In the center of gravity weighting, when all cluster hits have zero weight (logWeightBase is set too low, e.g. ZDC Ecal), then none of the hits are added in the weighted sum and the sum of weights is zero. Normalizing by the sum of weights leads to division by zero on a zero position at that point. Instead of division by zero we just leave the position unset (i.e. still at zero).

- [x] Bug fix (issue: division by zero)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -199,10 +199,10 @@ std::optional<edm4eic::Cluster> CalorimeterClusterRecoCoG::reconstruct(const edm
   }
   if (tw == 0.) {
     m_log->warn("zero total weights encountered, you may want to adjust your weighting parameter.");
-  } else {
-    cl.setPosition(v / tw);
-    cl.setPositionError({}); // @TODO: Covariance matrix
+    return {};
   }
+  cl.setPosition(v / tw);
+  cl.setPositionError({}); // @TODO: Covariance matrix
 
   // Optionally constrain the cluster to the hit eta values
   if (m_cfg.enableEtaBounds) {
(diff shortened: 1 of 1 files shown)


src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -194,9 +194,10 @@ edm4eic::Cluster CalorimeterClusterRecoCoG::reconstruct(const edm4eic::ProtoClus
   }
   if (tw == 0.) {
     m_log->warn("zero total weights encountered, you may want to adjust your weighting parameter.");
+  } else {
+    cl.setPosition(v / tw);
+    cl.setPositionError({}); // @TODO: Covariance matrix
```
veprbl, line 199: Should we fill the gaps with NaN values? Position {0,0,0} would not be correct, right?

wdconinc: I didn't want to introduce any changes here, which is why I kept it at zero which is what the code currently does (setPosition fails, excecution continues without setting position).

I think the correct thing here is to not have a cluster. I would prefer not to write NaN since then all code downstream has to prepare for accepting NaNs, and NaNs infect everything in accumulated operations like averages.

wdconinc: What actually happens here is that all hits are so low in energy that they don't rise above the specified logWeightBase value.  I think that would qualify as "there was no cluster after all".

wdconinc: We'd probably have to change the function to return `std::optional<Cluster>`.

wdconinc: See #1000

wdconinc:
@Chao1009 @veprbl Can we approve and merge this to clear some of the backlog?

Chao1009 (approved):

PR #967, opened 2023-09-14
fix: disable DRICH radiator and photon vertex cheat code

wdconinc:
### Briefly, what does this PR introduce?
This disables the cheatTrueRadiator and cheatPhotonVertex codes in the DRICH IrtCherenkovParticleID_factory configuration. This should reduce the number of warnings about non-optical photon hits on the DRICH sensors.

### What kind of change does this PR introduce?
- [x] Bug fix (issue #689)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators @chchatte92 @c-dilks 

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
Yes. No more cheating.

PR #991, opened 2023-09-18
fix(dRICH): use gas radiator backplane for track propagation filter

### LINKED ISSUE #990: Error: [DRICHIrtCherenkovParticleID] radiators have differing numbers of TrackSegments
- Which branch (often `main` for latest released): `main`
- Which version (or `HEAD` for the most recent on git): `HEAD`
- Any specific OS or system where the issue occurs? CI
- Any special versions of ROOT or Geant4? no

1. run standard reconstruction for DIS events (see CI job `eicrecon-dis`)

No errors from dRICH reconstruction

https://github.com/eic/EICrecon/actions/runs/6215248891/job/16868173477#step:5:700
```
Error: DRICHMergedTracks] [error] input track collections do not have the same size; cannot merge
Error: DRICHIrtCherenkovParticleID] [error] radiators have differing numbers of TrackSegments
```

### Context:

From #984 comments:

> I suspect these are charged particles which are outside the aerogel acceptance, but still inside the gas acceptance. After https://github.com/eic/EICrecon/pull/735 we simply don't have aerogel TrackkPoints for such tracks, hence the error (prior behavior was to propagate and fail, but still produce a track with zero TrackPoints (wasting time))

> This is a bad issue though: if there are any tracks which only pass through gas and not aerogel, no dRICH PID will be done for the event!

> A quick fix to restore behavior before https://github.com/eic/EICrecon/pull/735 would be to write out a TrackSegment with zero TrackPoints for such tracks, if and only if the track would pass through the backplane of the gas; see the filter surface implementation in https://github.com/eic/EICrecon/pull/735, which depends on the radiator (aerogel has smaller radius than gas, try making both filters be the gas back plane disc).

c-dilks:
Resolve #990 

We use the `Acts` surface at max $z$ as the "filter surface" to decide whether or not to proceed with track propagation to the sequence of surfaces in the dRICH radiators (see #735). For aerogel, that surface is slightly before the acrylic filter, and for gas that surface is just before the vessel exit window.

This PR resolves #990 by changing the filter surface to be the surface at max $z$ for gas, for _both_ aerogel and gas track propagation attempts. Prior to this PR, a charged particle with $\eta$ less than the max $\eta$ of aerogel would not produce any propagated track points and no `TrackSegment` object would be written out; however, there would still be a `TrackSegment` for the gas radiator, since the track still passes through gas, thus causing the errors reported in #990: this charged particle would have 0 `TrackSegment`s for aerogel, and 1 `TrackSegment` for gas and the merged (gas+aerogel).

With this PR, charged particle with such $\eta$ will still attempt to propagate to the aerogel planes, but all of them will fail* and the output `TrackSegment` will have zero `TrackPoint`s. Since there is still a `TrackSegment` in the output, the errors in #990 no longer appear and IRT proceeds happily**.  

*these failures waste CPU time, so this could be improved further, e.g., by checking _both_ the gas's and aerogel's max $z$ plane

**this PR is a 'quick' solution; perhaps a better, but more difficult solution is refactoring `IrtCherenkovParticleID` (and probably `MergeTracks` and `MergeParticleIDs`) so that it can handle the case of non-equal-sized input `TrackSegment` collections. 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
yes, stop ignoring such events as described in #990.

### DIFF
--- a/src/global/pid/RichTrack_factory.cc
+++ b/src/global/pid/RichTrack_factory.cc
@@ -70,6 +70,18 @@ void eicrecon::RichTrack_factory::Process(const std::shared_ptr<const JEvent> &e
     }
   }
 
+  // choose the filter surface: all charged particles that pass through the dRICH vessel will pass
+  // through the backplane of the gas radiator
+  std::shared_ptr<Acts::Surface> filter_surface;
+  for(auto& [output_tag, radiator_tracking_planes] : m_tracking_planes) {
+    if(richgeo::ParseRadiatorName(output_tag, m_log) == richgeo::kGas) {
+      filter_surface = radiator_tracking_planes.back();
+      break;
+    }
+  }
+  if(!filter_surface)
+    throw JException("cannot find filter surface for RICH track propagation");
+
   // run track propagator algorithm, for each radiator
   for(auto& [output_tag, radiator_tracking_planes] : m_tracking_planes) {
     try {
@@ -77,7 +89,7 @@ void eicrecon::RichTrack_factory::Process(const std::shared_ptr<const JEvent> &e
       auto result = m_propagation_algo.propagateToSurfaceList(
           trajectories,
           radiator_tracking_planes,
-          radiator_tracking_planes.back(), // `filterSurface`: assumes projectivity to radiator back-plane
+          filter_surface,
           track_point_cut,
           true
           );


c-dilks:
Can't test `reconstruction_benchmarks`, 502 error (and we still need to make them run in CI anyway).

wdconinc:
> 502 error

Server upgrade now complete. Try again.

chchatte92:
Hello colleagues, I did a quick test. I shot 500, 20 GeV single pion events. Uniform in eta 1.3 to 2.0. 
Here are the results:
a. W/o the gas radiator back plane we have  332 rings in aerogel and 341 for gas. (The current main branch EICRecon)
image

image

They are not exactly the same in number though. 
b. W/ this implementation now I have again 332 aerogel rings but 
image
but, 488 gas rings
image
The ring resolutions are comparable.

chchatte92:
I think I will narrow down the eta range and make another test. But, one thing is clear to me that a vast number of rings w/ lower number of photons passing only through the gas volume are rejected.

chchatte92 (commented):
I clearly see increase in the number of gas rings. However, I think that asking a point exactly at the filter maybe a bit too optimistic. Should I try something like that?
0.5*(filter_surface+radiator_tracking_planes.back())
This means at least a point should be there at the mid point of the vessel.

c-dilks:
NPE is independent of the changes in this PR, since that is produced by `PhotoMultiplierHitDigi`, which runs parallel to the `TrackPropagator` algorithm (this PR modifies the corresponding factory, `RichTracks_factory`). I do not expect NPE to change here... 

How does the reconstructed Cherenkov angle look for events with $\eta < \eta^\mbox{min}_\mbox{aerogel}$? Prior to this PR we should be getting no PID for such events (since that is the bug).

chchatte92:
> NPE is independent of the changes in this PR, since that is produced by `PhotoMultiplierHitDigi`, which runs parallel to the `TrackPropagator` algorithm (this PR modifies the corresponding factory, `RichTracks_factory`). I do not expect NPE to change here...

Hi Chris, Yes I know (that the Npe is partly independent of this PR). I wanted to demonstrate the number of entries of the Npe distribution. As the entries of such distribution is the accepted rings/tracks for the PID. W/ this PR we increase the accepted gas rings/tracks.

c-dilks:
> Should I try something like that?
> 0.5*(filter_surface+radiator_tracking_planes.back())

To do this it would be better to use one of the existing ACTS surfaces in the middle, but this would not solve our problem here, since that would miss the lowest $\eta$ trajectories.

On the other hand, mostly all charged particles which pass through either radiator of the dRICH must pass through the back z-plane. There are some charged particles at even lower $\eta$ which will only pass through the gas radiator and then the vessel cylinder (the outermost radial wall), but their Cherenkov photons will (mostly) not hit the spherical mirrors, so we just ignore those, assuming those are for hpDIRC PID.

c-dilks:
> Yes I know (that the Npe is partly independent of this PR)

What part of NPE depends on this PR?

c-dilks:
Ah, sorry you are looking at the NPE estimate from the Cherenkov PID... another NPE; so this does indeed depend on track propagation....

chchatte92:
> To do this it would be better to use one of the existing ACTS surfaces in the middle, but this would not solve our problem here, since that would miss the lowest

Can you please tell me how can I know which are the existing surfaces?

c-dilks:
> Can you please tell me how can I know which are the existing surfaces?

See `ActsGeo` in the `richgeo` service: https://github.com/eic/EICrecon/tree/main/src/services/geometry/richgeo

chchatte92:
> I would therefore expect a definite increase in the number of _entries_ in the NPE histograms, which you observe, since this PR re-enables PID in the gas-only region. I would not expect much change in the mean NPE, which you also observe.
 
 Yes. 

> I also don't understand the increase in NPE∼0 events; however, we know that #967 may have messed up IRT for gas (well, not the fault of disabling cheating but rather something else we never figured out yet).

My explanation is the following. The tracks I am shooting is a mixture of tracks in a broad eta bin. Some tracks will have  passed through the gas, might have produced some photons but essentially did not convert. These tracks will be there and will generate 0 mrad ring angle or no ring angle. Although valid tracks (a large source of background for below threshold antiprotons which will be above kaon threshold).  We see a rise in the 0 mrad ring angles as well. With exactly same entries.

src/global/pid/RichTrack_factory.cc
```diff
@@ -70,14 +70,26 @@ void eicrecon::RichTrack_factory::Process(const std::shared_ptr<const JEvent> &e
     }
   }
 
+  // choose the filter surface: all charged particles that pass through the dRICH vessel will pass
+  // through the backplane of the gas radiator
+  std::shared_ptr<Acts::Surface> filter_surface;
+  for(auto& [output_tag, radiator_tracking_planes] : m_tracking_planes) {
+    if(richgeo::ParseRadiatorName(output_tag, m_log) == richgeo::kGas) {
+      filter_surface = radiator_tracking_planes.back();
```
c-dilks, line 78: With our current geometry, this change would get _all_ charged particles which pass through either radiator, including the ones that exit out the vessel radial wall (which we may not care about).
```suggestion
      filter_surface = radiator_tracking_planes.front();
```

c-dilks: @chchatte92 I leave it up to you to decide which we want.

chchatte92: > @chchatte92 I leave it up to you to decide which we want.

Let's stick to `filter_surface=radiator_tracking_planes.back()` . We have intuitive feeling for this scenario.

wdconinc (commented):
Can you write more description in the PR? I.e. what is this doing to address #990.

c-dilks:
> Can you write more description in the PR? I.e. what is this doing to address #990.

Updated.

wdconinc:
Comment on "these failures waste CPU time". We were (before this PR) spending 40% of total CPU time on dRICH. This seems to have increased it (small statistics warning) to 50% (https://github.com/eic/EICrecon/actions/runs/6237737385/job/16932405028?pr=991#step:5:3138).

We have as informal rule used (and previously enforced against other subsystems) that a single subsystem may not use more than 80% of resources (memory, event size, running time). So, go make something else slower now or make the dRICH faster again ;-)

wdconinc (approved):
I am a bit torn about the change to using the shared_ptr here. It takes what was a guaranteed valid filter surface when passed to `propagateToSurfaceList` and turns it into a possible nullptr that relies on the if throw to not pass the nullptr.

Maybe at some point soon we can start using cppgsl and its not_null<T*> in the algorithms (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-nullptr).

c-dilks:
No surprise on the CPU time increase, at least that's more evidence this PR is working correctly :)

We do 35 (!) track propagations per charged particle (relevant snippet below). It used to be 15 (5 for aerogel, 10 for gas), but I think we turned it up because it didn't hurt (us, in our solo testing), and we were desparate to resolve issues in the gas PID (now resolved by something unrelated). Perhaps @chchatte92 could try turning it down and seeing if things still look okay (that is out of scope of this PR, make a new one). 

https://github.com/eic/EICrecon/blob/926b6722d9e8c19f53fe3c61a676e6d2be3f979c/src/detectors/DRICH/DRICH.cc#L66-L69

chchatte92:
> . Perhaps @chchatte92 could try turning it down and seeing if things still look okay (that is out of scope of this PR, make a new one).

Doing.

PR #992, opened 2023-09-18
ACTSGeo_service.cc: rename geom const materials-map -> material-map

veprbl:
The https://github.com/eic/epic/pull/494 is not merged at this time. Let's update that one to provide both constants, but going forward use proper meaning one.


wdconinc (approved):
This inconsistency came from ATHENA days. I think we originally had material-maps.json, then @sly2j said that's the wrong name and we ended up with materials-map.cbor, and we just kinda gave up on it then. Thanks for fixing it!

PR #994, opened 2023-09-19
fix: iwyu header fixes

wdconinc:
Automatic fixes obtained with:
```
iwyu_tool -p build -o iwyu -- -Xiwyu --cxx17ns -Xiwyu --mapping_file=$PWD/.github/iwyu.imp | fix_include --blank_lines --nosafe_headers --reorder --separate_project_includes="<tld>" --keep_iwyu_namespace_format
```

- [x] Bug fix (issue: header cleanup)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2022 Sylvester Joosten, Chao, Chao Peng, Whitney Armstrong, Dhevan Gangadharan
 
+#include <Evaluator/DD4hepUnits.h>
 /*
  *  Reconstruct the cluster with Center of Gravity method
  *  Logarithmic weighting is used for mimicking energy deposit in transverse direction
@@ -9,14 +10,29 @@
  */
 #include <boost/algorithm/string/join.hpp>
 #include <boost/range/adaptor/map.hpp>
-#include <fmt/format.h>
+#include <edm4eic/CalorimeterHitCollection.h>
+#include <edm4eic/vector_utils_legacy.h>
+#include <edm4hep/CaloHitContributionCollection.h>
+#include <edm4hep/MCParticleCollection.h>
+#include <edm4hep/Vector3f.h>
+#include <fmt/core.h>
+#include <podio/ObjectID.h>
+#include <podio/RelationRange.h>
+#include <Eigen/Core>
+#include <Eigen/Eigenvalues>
+#include <Eigen/Householder>
+#include <Eigen/Jacobi>
+#include <cctype>
+#include <complex>
+#include <exception>
+#include <limits>
 #include <map>
-#include <Eigen/Dense>
-
-#include <Evaluator/DD4hepUnits.h>
-#include <edm4hep/MCParticle.h>
+#include <new>
+#include <type_traits>
+#include <vector>
 
 #include "CalorimeterClusterRecoCoG.h"
+#include "algorithms/calorimetry/CalorimeterClusterRecoCoGConfig.h"
 
 namespace eicrecon {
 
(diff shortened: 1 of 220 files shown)


veprbl:
Regarding introduction of the forward declarations. The https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md#why-forward-declare suggests that in additional to parsing speed win, it may allow to avoid some ODR rule violations, which I don't think is as robust. As this is a developing codebase, it would be easier for users to having to pick and switch between forward declaration and full include depending on whether they use any methods/members from a class or not.

wdconinc:
The problem I see with fwd decl is mainly one of developer education: when do you need to include the full header, when do you include the fwd header (spdlog/fwd.h), and when do you just use a fwd decl yourself? I think that because this tool won't give suggested changes like clang-tidy does it will appear as magic to some, which will lead to frustration. Not a reason not to do it, but we need to think about how to train people and possibly how to make the iwyu suggestions clear. Right now I don't have a good solution for that (short of using capybara to create a PR against the PR).

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -1,22 +1,48 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2022 Sylvester Joosten, Chao, Chao Peng, Whitney Armstrong, Dhevan Gangadharan
 
+#include <DD4hep/Detector.h>
+#include <Eigen/src/Eigenvalues/EigenSolver.h>
+#include <Eigen/src/Householder/BlockHouseholder.h>
+#include <Eigen/src/Householder/Householder.h>
+#include <Eigen/src/Jacobi/Jacobi.h>
```
wdconinc, line 8: Not sure why this isn't caught by
```
{ include: ["@<Eigen/src/Householder/.*>", "private", "<Eigen/Householder>", "public"] },
```

veprbl:
Yes. This is basically my point, we'd confuse people unnecessarily. We can always enable forward declarations later.

src/algorithms/calorimetry/CalorimeterHitDigi.h
```diff
@@ -13,17 +13,17 @@
 
 #pragma once
 
+#include <stdint.h>
```
wdconinc, line 16: Boost dislikes `cstdint`

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
```diff
@@ -10,6 +10,32 @@
...
+#include <DD4hep/Detector.h>
+#include <DD4hep/IDDescriptor.h>
+#include <DD4hep/Objects.h>
+#include <DD4hep/Readout.h>
+#include <DD4hep/VolumeManager.h>
+#include <DDRec/CellIDPositionConverter.h>
+#include <DDSegmentation/BitFieldCoder.h>
+#include <Evaluator/DD4hepUnits.h>
+#include <Math/GenVector/DisplacementVector3D.h>
+#include <bits/utility.h>
```
wdconinc, line 24: Avoid low level headers

src/algorithms/tracking/IterativeVertexFinder.cc
```diff
@@ -5,32 +5,51 @@
...
 #include <Acts/Vertexing/HelicalTrackLinearizer.hpp>
+#include <Acts/Vertexing/HelicalTrackLinearizer.ipp>
 #include <Acts/Vertexing/ImpactPointEstimator.hpp>
+#include <Acts/Vertexing/ImpactPointEstimator.ipp>
 #include <Acts/Vertexing/IterativeVertexFinder.hpp>
-#include <Acts/Vertexing/LinearizedTrack.hpp>
+#include <Acts/Vertexing/IterativeVertexFinder.ipp>
 #include <Acts/Vertexing/Vertex.hpp>
-#include <Acts/Vertexing/VertexFinderConcept.hpp>
+#include <Acts/Vertexing/Vertex.ipp>
```
wdconinc, line 29: Avoid ipp headers.

wdconinc:
Closing in favor of new PR.

PR #997, opened 2023-09-19
fix: include edm4hep collection header

wdconinc:
This just applies a simple change to verify that ci-iwyu in #755 works as intended.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Huh, https://github.com/eic/EICrecon/actions/runs/6241994577/job/16945765072?pr=997. Unexpected headers.

wdconinc:
This now produces the correct suggestions: https://github.com/eic/EICrecon/actions/runs/6503105546/job/17663822601?pr=997, adds the patch to the artifacts.

There is an additional snag, which is that iwyu fixes are not necessarily immediately stable wrt rerunning iwyu: if you run iwyu again after fixes, it may suggest more fixes. We may need to loop until stable and hope it's not bistable or periodic...

wdconinc:
`ci-iwyu` in #755 now runs iwyu until it doesn't apply any more changes.

wdconinc:
This looks better now: https://github.com/eic/EICrecon/actions/runs/6510717157/job/17686511123?pr=997#step:14:11, with reduced Eigen headers.

wdconinc:
Not needed anymore now that #755 has been merged.

PR #999, opened 2023-09-20
Reducing number of track projections inside dRICH

chchatte92:
Currently we take 35 points inside the dRICH for the track propagation per charged particle. This increases the CPU time. This PR (a draft) includes 25 points instead of 35 points, 5 points for aerogel and 20 points for the gas.  This draft PR is in response to the suggestion: https://github.com/eic/EICrecon/pull/991#issuecomment-1726523966

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: __

This Draft PR is in response to the above mentioned comments from Chris Dilks. 

No. 
No.


chchatte92:
Hi Chris, I have checked that reducing the number of track points inside the gas from 30 to 20 does not change the ring resolutions. But there maybe a tiny difference (in my laptop with 500 tracks)  in process time.
with 30 points: 
image
 with 20 points:
image

Does it make sense to reduce these 20 points even lower?

chchatte92:
If I move to 10 track points we see almost no changes in the theta residual or number of photons etc but the computation time further reduces.

src/detectors/DRICH/DRICH.cc
```diff
@@ -66,7 +66,7 @@ extern "C" {
     // track propagation to each radiator
     RichTrackConfig track_cfg;
     track_cfg.numPlanes.insert({ "Aerogel", 5  });
-    track_cfg.numPlanes.insert({ "Gas",     30 });
+    track_cfg.numPlanes.insert({ "Gas",     20 });
```
c-dilks, line 69:
```suggestion
    track_cfg.numPlanes.insert({ "Gas",     10 });
```

wdconinc:
Is this something we can move from draft to approval to merged over the next days?

chchatte92:
> Is this something we can move from draft to approval to merged over the next days?

Hi Wouter, I am making some final tests and we can go for a final PR by tomorrow. Thanks.

chchatte92:
Dear @wdconinc and @c-dilks ,
  I have checked the performance of dRICH and they look good. We can stick to the 10 segments for gas (original IRT segmentation) and can go for a merge.

wdconinc (approved):

Issue #996, opened 2023-09-19
Rename JDD4hep_service to DD4hep_service (not a JANA2 component)


wdconinc:
**Is your feature request related to a problem? Please describe.**
User was looking for documentation about the JDD4hep_service in https://www.jlab.org/JANA/jana_doc_latest/index.html, where nothing relevant is available.

**Describe the solution you'd like**
Removing the `J` will make it clear this is NOT a JANA2 component.

PR #1001, opened 2023-09-20
fix: rename JDD4hep_service to DD4hep_service

wdconinc:
Rename JDD4hep_service to DD4hep_service to avoid the implication that JDD4hep_service is part of JANA2.

TODO:
- [ ] #1002 

- [x] Bug fix (issue #996)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/detectors/FOFFMTRK/OffMomentumReconstruction_factory.cc
+++ b/src/detectors/FOFFMTRK/OffMomentumReconstruction_factory.cc
@@ -24,7 +24,7 @@ namespace eicrecon {
         m_readout = m_input_tag;
 
         app->SetDefaultParameter(param_prefix+"readoutClass", m_readout);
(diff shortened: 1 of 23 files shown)


dhevang:
Looks like a simple search/replace.  Everybody with their own analysis/reco plugin would have to search and replace this as well.  Question though, JDD4hep_service inherits from JService, so are we sure that there's nothing Jana related to this class?  There are 2 virtual public functions.  However, they appear to unrelated to Jana as well.  Just making sure that this name change doesn't open a can of worms.

wdconinc:
It's a Jana service which is why it inherits JService, but it's not part of Jana.

As for the end user code, I'm pretty sure we break this all the time already and the explicit contract with those users is that any end user code should make its way into the EICrecon central repository.

dhevang (approved):

PR #1002, opened 2023-09-20
fix: set LD_LIBRARY_PATH for benchmark tests

wdconinc:
### Briefly, what does this PR introduce?
This fell bewteen the cracks. Leads to errors like https://github.com/eic/EICrecon/actions/runs/6249479829/job/16966875149#step:5:182.

### What kind of change does this PR introduce?
- [x] Bug fix (issue: https://github.com/eic/EICrecon/actions/runs/6249479829/job/16966875149#step:5:182)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1003, opened 2023-09-20
eicrecon_cli: return exit code when exception is detected

veprbl:
Resolves issue
https://github.com/eic/EICrecon/pull/1002#issuecomment-1727866709
seen in https://github.com/eic/EICrecon/actions/runs/6249479829/job/16966883258#step:5:182


wdconinc:
Fails as intended ;-)

wdconinc (approved):

Issue #949, opened 2023-09-12
Heisenbug (crash) on macOS


veprbl:
Crashes after several events
```
Thread model: pthreads
0x70000ecb2000: 
       `- JSignalHandler::handle_sigsegv(int, __siginfo*, void*) (0x1027e06ac)
        `- _sigtramp (0x7fff20776d7d)
         `- 4   ???                                 0x00007f89bc790e50 0x0 + 140229549297232 (0x7f89bc790e50)
          `- JFactoryPodioT<edm4hep::CaloHitContribution>::ClearData() (0x1072bccd9)
           `- JFactorySet::Release() (0x10277e042)
            `- JEventPool::get(unsigned long) (0x10278d727)
             `- JEventSourceArrow::execute(JArrowMetrics&, unsigned long) (0x10278cd5a)
              `- JWorker::loop() (0x102790a0c)
               `- void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (JWorker::*)(), JWorker*> >(void*) (0x102791c2e)
                `- _pthread_start (0x7fff207318fc)
                 `- thread_start (0x7fff2072d443)
```
The only workaround is to compile with -DUSE_ASAN=ON, then it doesn't crash at all.

_Originally posted by @veprbl in https://github.com/eic/EICrecon/issues/944#issuecomment-1714400411_

veprbl:
A similar bug on Linux (somehow, with edm4eic and eicrecon's factory):
```
Thread model: pthreads
140736862168640: 
       `- JSignalHandler::handle_sigsegv(int, siginfo_t*, void*) (0x7ffff7f6b721)
        `- /lib/libc.so.6 (0x7ffff563dbf0)
         `- edm4eic::ReconstructedParticle::~ReconstructedParticle() (0x7ffff4ef0d78)
          `- eicrecon::JFactoryPodioT<edm4eic::ReconstructedParticle>::ClearData() (0x7ffff379d3d0)
           `- JFactorySet::Release() (0x7ffff7eb5982)
            `- JEventSourceArrow::execute(JArrowMetrics&, unsigned long) (0x7ffff7edaf28)
             `- JWorker::loop() (0x7ffff7ee0614)
              `- /lib/libstdc++.so.6 (0x7ffff5adf3d4)
               `- /lib/libc.so.6 (0x7ffff5688e86)
[...]

Issue #837, opened 2023-08-08
There is `JFactoryPodioT` in JANA2 and EICrecon codebases


veprbl:
This may cause a confusion between the two classes when stacktraces are printed. Perhaps, we don't need both anymore.

veprbl:
```diff
--- eicrecon/src/services/io/podio/JFactoryPodioT.h
+++ jana2/src/libraries/JANA/Podio/JFactoryPodioT.h
@@ -1,17 +1,39 @@
 
-
 // Copyright 2023, Jefferson Science Associates, LLC.
 // Subject to the terms in the LICENSE file found in the top-level directory.
 
-#pragma once
+
+#ifndef JANA2_JFACTORYPODIOT_H
+#define JANA2_JFACTORYPODIOT_H
 
 #include <JANA/JFactoryT.h>
-#include <JANA/JEvent.h>
 #include <podio/Frame.h>
-#include "datamodel_glue.h"
 
+template <typename S> struct PodioTypeMap;
+
+/// The point of this additional base class is to allow us _untyped_ access to the underlying PODIO collection,
+/// at the cost of some weird multiple inheritance. The JEvent can trigger the untyped factory using Create(), then
+///
+class JFactoryPodio {
+protected:
+    const podio::CollectionBase* mCollection = nullptr;
+    bool mIsSubsetCollection = false;
+    podio::Frame* mFrame = nullptr;
+
+private:
+    // Meant to be called internally, from JMultifactory
+    friend class JMultifactory;
+    void SetFrame(podio::Frame* frame) { mFrame = frame; }
+
+    // Meant to be called internally, from JEvent:
+    friend class JEvent;
+    const podio::CollectionBase* GetCollection() { return mCollection; }
+
+public:
+    // Meant to be called from ctor, or externally, if we are creating a dummy factory such as a multifactory helper
+    void SetSubsetCollection(bool isSubsetCollection=true) { mIsSubsetCollection = isSubsetCollection; }
+};
 
-namespace eicrecon {
 
 template <typename T>
 class JFactoryPodioT : public JFactoryT<T>, public JFactoryPodio {
@@ -111,11 +133,8 @@
     this->mData.clear();
     this->mCollection = nullptr;  // Collection is owned by the Frame, so we ignore here
     this->mFrame = nullptr;  // Frame is owned by the JEvent, so we ignore here
-    if (this->mStatus != JFactory::Status::Uninitialized) {
-        this->mStatus = JFactory::Status::Unprocessed;
[...]

PR #1008, opened 2023-09-21
Fix crashes in JFactoryPodioT::ClearData()

veprbl:
We are observing inconsistent crashes in delete operator invocation in ClearData() on certain events. These faults occur on macOS with a high rate, specifically for CalorimeterHitContribution. We also have reports of some percent of reconstruction jobs (on Linux) crashing specifically on ReconstructedParticle.

There might be an issue with PODIO object clones, they should be pointing to the same object data. There doesn't seem to be any facility for reference counting.

A slight complication with addressing this is that the failure comes in the implementation FactoryPodioT from JANA2 libraries. That factory is implicitly instantiated as a result of EICrecon calling the `JEvent::InsertCollectionAlreadyInFrame` method in `JEventSourcePODIO`. This PR switches to EICrecon variant of the same code to have a chance of handling the issue locally.

It is not clear if mData is used or just populated for compatibility. This is an attempt at disabling the functionality in order to mitigate the issue.

Resolves: #949
cc @wdconinc @rahmans1


src/services/io/podio/JEventSourcePODIO.cc
```diff
@@ -9,6 +9,7 @@
 // it is limited to processing only a single event at a time.
```
veprbl, line 9: These changes would need to be duplicated to the JEventSourceLegacyPODIO.cc

nathanwbrei:
mData is populated for compatibility so that the old-style `JFactoryT<T>::Set()`, `JEvent::Get<T>()` work as expected. If you use `{Set,Get}Collection` everywhere instead you should be fine. 
However we should be trying to get rid of these hacked-up JANA facilities and upstream the long-term fixes.`mData` can't be the root cause of the heisenbug. PODIO shouldn't need reference counting because the underlying data is owned by the collection, not the lightweight object wrappers.

nathanwbrei:
Are these failures all happening with subset collections? Could it be that PODIO is failing when deleting the object wrappers that point to an already-deleted collection?

veprbl:
> mData is populated for compatibility so that the old-style `JFactoryT<T>::Set()`, `JEvent::Get<T>()` work as expected. If you use `{Set,Get}Collection` everywhere instead you should be fine.

That was my working assumption. Once CI runs again on this PR, we can see if this breaks anything (the `JEvent::Get` should be mostly phased out by now).

> However we should be trying to get rid of these hacked-up JANA facilities and upstream the long-term fixes.

Sure, here I'm going oppositely to #837.
Would be nice to be able to supply custom dependencies in our CI.

> `mData` can't be the root cause of the heisenbug.

It shouldn't be, yes.

> PODIO shouldn't need reference counting because the underlying data is owned by the collection, not the lightweight object wrappers.

This is good to know. I wonder if there can be some sort of lifetime issues if the input EDM4hep collections are directly saved into EDM4eic.

veprbl:
Even though we don't use mData and don't plan to. Maybe, let's keep the "feature" by attempting the "unlink" strategy.

veprbl:
JANA PR https://github.com/JeffersonLab/JANA2/pull/243

veprbl:
@wdconinc Review?

wdconinc:
Is this able to run with the LSAN enforcement we have in place in main?

wdconinc (approved):
Tested locally. No LSAN warnings emitted (I was just worried that PODIO wasn't going to do the deep clean that ClearData was doing before).

veprbl:
> Is this able to run with the LSAN enforcement we have in place in main?

This is designed to not leak memory, at least in EICrecon (JANA, in general, is another story https://github.com/JeffersonLab/JANA2/pull/243#issuecomment-1745477936).

Issue #971, opened 2023-09-15
Implement `Acts::Logging::OutputPrintPolicy` to funnel output to spdlog.


veprbl:
The current ACTS logging bypasses the spdlog. That results in an inconsistent formatting.

While implementing this we might as well attempt filtering known uninformative messages like #512 and the now used track propagation.

The #133 suggests there is a similar issue for DD4hep.

wdconinc:
I was looking into this, but I don't know if this is possible. We would need to provide an std::ostream& from spdlog to the Acts lines in e.g. CKFTracking.cc, such as
```cpp
ACTS_LOCAL_LOGGER(Acts::getDefaultLogger("CKFTracking Logger", logLevel, &std::cout));
```
I don't think spdlog provides this, only an ostream sink.

wdconinc:
Fixed by #1012

PR #1012, opened 2023-09-23
feat: use our spdlog::logger for Acts logging

wdconinc:
This introduces `SpdlogPrintPolicy` that is used to tell Acts to use the framework logger.

This already works, but probably makes some code obsolete, which I'll evaluate later. 

- [x] Bug fix (issue #971)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/extensions/spdlog/SpdlogToActs.h
+++ b/src/extensions/spdlog/SpdlogToActs.h
@@ -67,6 +67,13 @@ class SpdlogPrintPolicy final : public Acts::Logging::OutputPrintPolicy {
     explicit SpdlogPrintPolicy(std::shared_ptr<spdlog::logger> out, std::vector<std::regex> suppressions = {})
     : m_out(out), m_suppressions(suppressions) {}
 
+    /// @brief destructor
+    ~SpdlogPrintPolicy() {
+      if (m_suppressions_count > 0) {
+        m_out->info("last message repeated {} times", m_suppressions_count);
+      }
+    }
+
     /// @brief flush the debug message to the destination stream
     ///
     /// @param [in] lvl   debug level of debug message
@@ -74,7 +81,7 @@ class SpdlogPrintPolicy final : public Acts::Logging::OutputPrintPolicy {
     void flush(const Level& lvl, const std::string& input) final {
       for (const auto& suppression : m_suppressions) {
         if (std::regex_search(input, suppression)) {
-          return;
+          if (++m_suppressions_count > 1) return;
         }
       }
       m_out->log(ActsToSpdlogLevel(lvl), input);
@@ -111,13 +118,13 @@ class SpdlogPrintPolicy final : public Acts::Logging::OutputPrintPolicy {
       return std::make_unique<SpdlogPrintPolicy>(m_out);
     };
   #endif
-
   private:
     /// pointer to destination output stream
     std::shared_ptr<spdlog::logger> m_out;
 
     /// regexes for messages to be suppressed
     std::vector<std::regex> m_suppressions;
+    std::size_t m_suppressions_count{0};
 };
 
 inline std::unique_ptr<const Acts::Logger> getSpdlogLogger(
--- a/src/extensions/spdlog/SpdlogToActs.h
+++ b/src/extensions/spdlog/SpdlogToActs.h
@@ -18,7 +18,7 @@ namespace eicrecon {
 using namespace Acts::Logging;
 
 using SpdlogToActsLevel_t = boost::bimap<spdlog::level::level_enum, Acts::Logging::Level>;
-static SpdlogToActsLevel_t kSpdlogToActsLevel = boost::assign::list_of<SpdlogToActsLevel::relation>
+static SpdlogToActsLevel_t kSpdlogToActsLevel = boost::assign::list_of<SpdlogToActsLevel_t::relation>
   (spdlog::level::trace, Acts::Logging::VERBOSE)
   (spdlog::level::debug, Acts::Logging::DEBUG)
   (spdlog::level::info, Acts::Logging::INFO)
--- a/src/extensions/spdlog/SpdlogToActs.h
+++ b/src/extensions/spdlog/SpdlogToActs.h
@@ -24,8 +24,7 @@ static SpdlogToActsLevel_t kSpdlogToActsLevel = boost::assign::list_of<SpdlogToA
   (spdlog::level::info, Acts::Logging::INFO)
   (spdlog::level::warn, Acts::Logging::WARNING)
   (spdlog::level::err, Acts::Logging::ERROR)
-  (spdlog::level::critical, Acts::Logging::FATAL)
-  (spdlog::level::off, Acts::Logging::FATAL);
+  (spdlog::level::critical, Acts::Logging::FATAL);
 
 inline Acts::Logging::Level SpdlogToActsLevel(spdlog::level::level_enum input) {
   try {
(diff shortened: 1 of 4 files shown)


src/extensions/spdlog/SpdlogToActs.h
```diff
@@ -11,31 +11,110 @@
...
+    ///
+    /// @pre @p out is non-zero
+    explicit SpdlogPrintPolicy(std::shared_ptr<spdlog::logger> out) : m_out(out) {}
+
+    /// @brief flush the debug message to the destination stream
+    ///
+    /// @param [in] lvl   debug level of debug message
+    /// @param [in] input text of debug message
+    void flush(const Level& lvl, const std::string& input) final {
+      m_out->info(input);
```
wdconinc, line 65:
```suggestion
      m_out->log(ActsToSpdlogLevel(lvl), input);
```

src/extensions/spdlog/SpdlogToActs.h
```diff
@@ -11,31 +11,110 @@
...
-                return Acts::Logging::FATAL;
-            case spdlog::level::n_levels:
-                [[fallthrough]];
-            default:
-                break;
-        }
-
-        auto err_msg = fmt::format("SpdlogToActsLevel don't know this log level: '{}'", input);
-        throw JException(err_msg);
+
```
wdconinc, line 14: This should probably be a boost bimap with much simpler to and from functions.

Ref https://www.boost.org/doc/libs/1_82_0/libs/bimap/doc/html/boost_bimap/examples/simple_bimap.html

src/extensions/spdlog/SpdlogToActs.h
```diff
@@ -4,38 +4,118 @@
...
+using SpdlogToActsLevel_t = boost::bimap<spdlog::level::level_enum, Acts::Logging::Level>;
+static SpdlogToActsLevel_t kSpdlogToActsLevel = boost::assign::list_of<SpdlogToActsLevel_t::relation>
+  (spdlog::level::trace, Acts::Logging::VERBOSE)
+  (spdlog::level::debug, Acts::Logging::DEBUG)
+  (spdlog::level::info, Acts::Logging::INFO)
+  (spdlog::level::warn, Acts::Logging::WARNING)
+  (spdlog::level::err, Acts::Logging::ERROR)
+  (spdlog::level::critical, Acts::Logging::FATAL)
+  (spdlog::level::off, Acts::Logging::FATAL);
```
veprbl, line 28: Will this even work? https://www.boost.org/doc/libs/1_73_0/libs/bimap/doc/html/boost_bimap/the_tutorial/differences_with_standard_maps.html

veprbl: Use `std::option<Acts::Logging::Level>` or give up on boost?

src/extensions/spdlog/SpdlogToActs.h
```diff
@@ -4,38 +4,118 @@
...
+    const std::string& name,
+    const Acts::Logging::Level& lvl,
+    std::shared_ptr<spdlog::logger> log) {
+
+  auto output = std::make_unique<LevelOutputDecorator>(
+      std::make_unique<NamedOutputDecorator>(
+          std::make_unique<TimedOutputDecorator>(
+              std::make_unique<SpdlogPrintPolicy>(log)),
+          name));
+  auto print = std::make_unique<DefaultFilterPolicy>(lvl);
```
veprbl, line 116: Could keep unfiltered and let spdlog do it.

src/extensions/spdlog/SpdlogToActs.h
wdconinc: Ah, didn't even notice the double right value. We should be able to have a unique mapping.

wdconinc: Surely doesn't work right now. All acts logs are gone entirely.

wdconinc: Maybe we just don't need the bidirectional map anymore. We used to need it to translate the spdlog level to a corresponding acts level. Now we can probably just use a unidirectional map from acts to spdlog and be done with it.

src/extensions/spdlog/SpdlogToActs.h
wdconinc: I don't think acts provides a filter policy that doesn't do anything. We'd have to write one. Not hard of course.

veprbl: Perhaps, just passing `Acts::Logging::VERBOSE` would give us what we want.

src/extensions/spdlog/SpdlogToActs.h
```diff
@@ -4,38 +4,113 @@
...
+    };
+  #endif
+
+  private:
+    /// pointer to destination output stream
+    std::shared_ptr<spdlog::logger> m_out;
+};
+
+inline std::unique_ptr<const Acts::Logger> getSpdlogLogger(
+    const std::string& name,
```
veprbl, line 106: Now unused
```suggestion
```

wdconinc: ... yeah, needs changes elsewhere too.

wdconinc:
Now a new regex filter in the output policy (we can't put this in the filter policy, because that doesn't have access to the message). With this new regex filter, the "No tracks found" lines are all suppressed, but CKFTracking doesn't print anything else inside the Acts routines, so it doesn't mean it's working as intended... or just always suppressing. I guess I could test that at a verbose level. This also probably needs a summary line when the logger is destructed (e.g. "regex xyz suppressed 1000 lines"), but that requires putting the logger in a class member (should be done anyway instead of creating a new logger on every CKFTracking produce call).

Some Acts logging still sneaks through outside of spdlog:
```
[INFO] Status: 10 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 10 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 10 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 11 events processed  1.0 Hz (0.0 mHz avg)
[INFO] Status: 11 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 11 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 11 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 11 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 11 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 12 events processed  1.0 Hz (0.0 mHz avg)
13:31:26    HelTrkLinPro   ERROR     Propagation reached the step count limit of 1000 (did 1000 steps)
[INFO] Status: 12 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 12 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 12 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 12 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 12 events processed  0.0 Hz (0.0 mHz avg)
[INFO] Status: 12 events processed  0.0 Hz (0.0 mHz avg)
```

src/extensions/spdlog/SpdlogToActs.h
```diff
@@ -4,38 +4,124 @@
 
 #pragma once
```
veprbl, line 6: Code should become MPL https://eic-code-browser.sdcc.bnl.gov/lxr/source/include/Acts/Utilities/Logger.hpp

veprbl (approved):

src/extensions/spdlog/SpdlogToActs.h
wdconinc: I kept the acts filter as well as the spdlog filter.

src/extensions/spdlog/SpdlogToActs.h
wdconinc: Bidirectional map retained for now, but duplication in right side removed.

Issue #667, opened 2023-05-15
Allow collections written to an output file to be reread to continue analysis.


simonge:
Currently EICrecon output files, generally, cannot be reread by EICrecon to add to an analysis.

There appear to be at least 2 issues (maybe more but I can't get past the second one).

1. EICrecon requires the EventHeader collection to be present, this is currently not kept in the default list.
2. Trying to read a new collection which wasn't present in the original epic output file returns e.g.

```
- JFactoryPodioT<edm4eic::RawTrackerHit>* JEvent::InsertCollectionAlreadyInFrame<edm4eic::RawTrackerHit>(PodioTypeMap<edm4eic::RawTrackerHit>::collection_t const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) (0x7f9e563d4695)
    - void InsertingVisitor::operator()<edm4eic::RawTrackerHitCollection>(edm4eic::RawTrackerHitCollection const&) (0x7f9e563bde3a)
     - VisitPodioCollection<InsertingVisitor>::operator()(InsertingVisitor&, podio::CollectionBase const&) (0x7f9e563b183a)
      - JEventSourcePODIO::GetEvent(std::shared_ptr<JEvent>) (0x7f9e563a847f)
       - JEventSource::DoNext(std::shared_ptr<JEvent>) (0x7f9e602d3c43)
        - JEventSourceArrow::execute(JArrowMetrics&, unsigned long) (0x7f9e602dc265)
         - JWorker::loop() (0x7f9e602e14d7)
          - /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x7f9e5fc564a3)
           - /lib/x86_64-linux-gnu/libc.so.6 (0x7f9e5fa07fd4)
            - /lib/x86_64-linux-gnu/libc.so.6 (0x7f9e5fa885bc)
```

My primary intention for this is training and testing neural networks faster, but other uses where only a subset of the data is needed to test different approaches quickly without having to run through all of the collections will be faster too. I'm sure there are other useful reasons to implement this as well.

wdconinc:
Was this the error you were encountering here? https://github.com/eic/EICrecon/actions/runs/6285667201/job/17068353375#step:7:509
```
[FATAL] Detected exception in worker! Re-throwing on main thread.
----------------------------------------------------------
Exception: Bad event headers: Entry 0 contains 0 items, but 1 expected.
  Plugin:         podio
  Component:      JEventSourcePODIO
[...]

PR #1014, opened 2023-09-23
feat: add two-stage test job (digitization/reconstruction); write EventHeader by default

wdconinc:
This adds a test job that runs digitization and reconstruction in separate eicrecon steps, with only a raw file artifact as communication between them. This should allow for ensuring #667 is continuously tested.

It is likely that this test won't work until some additional work is completed either here or elsewhere, depending on the scope of that extra work.

The test is limited to the barrel ecal because we currently have to specify all required raw collections as command line arguments, and this seems representative enough. Tracking systems are more collections, and none of the other calorimeter systems have deeply involved experts, so that's how I ended up with barrel ecal.

Quality of life ideas:
- easier way to specify output selection from the default list, e.g. by regex (I would have picked `.*Raw.*` if that was possible)

- [x] Bug fix (issue #667)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -299,6 +299,61 @@ jobs:
         path: sim_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root
         if-no-files-found: error
 
+  eicrecon-two-stage-running:
+    runs-on: ubuntu-latest
+    needs:
+    - build
+    - npsim-gun
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


wdconinc:
@rahmans1 @T-Britton We had on our roadmap somewhere the separation of digitization and reconstruction in simulation production. This should allow that to happen but not quite yet in a scalable way (you have to list all collections, and then it's still possible you missed one or an algorithm changes to require a new dependency). I think we may be able to roll out something like this in October or November, at least on a testing basis?

veprbl (approved):

PR #1016, opened 2023-09-24
fix: don't hide all the warnings and errors in TrackPropagation.cc

wdconinc:
It is not ok to set the log level to FATAL just so all the warnings and errors are hidden (in https://github.com/eic/EICrecon/pull/862). There is a reason we have these warnings and errors, and if the output looks like below then it needs to be looked at. 

```
[tracking:CentralCKFTrajectories] [warning] No tracks found
[tracking:CentralCKFTrajectories] [info] last message repeated 11 times
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[INFO] Status: 6 events processed  0.2 Hz (0.0 mHz avg)
[tracking:CentralCKFTrajectories] [warning] No tracks found
[tracking:CentralCKFTrajectories] [info] last message repeated 23 times
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
[DRICH:DRICHTracks] [error] Propagation reached the step count limit of 1000 (did 1000 steps)
[DRICH:DRICHTracks] [error] Step failed with EigenStepperError:3: Step size adjustment exceeds maximum trials
```

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackPropagation.cc
+++ b/src/algorithms/tracking/TrackPropagation.cc
@@ -210,7 +210,7 @@ namespace eicrecon {
         Stepper stepper(magneticField);
         Propagator propagator(stepper);
 
-        Acts::Logging::Level logLevel = Acts::Logging::FATAL;
+        Acts::Logging::Level logLevel = eicrecon::SpdlogToActsLevel(m_geoSvc->getActsRelatedLogger()->level());
         ACTS_LOCAL_LOGGER(Acts::getDefaultLogger("ProjectTrack Logger", logLevel));
 
         Acts::PropagatorOptions<> options(m_geoContext, m_fieldContext, Acts::LoggerWrapper{logger()});
--- a/src/algorithms/tracking/TrackPropagation.cc
+++ b/src/algorithms/tracking/TrackPropagation.cc
@@ -32,6 +32,8 @@
 
 #include "ActsGeometryProvider.h"
 
+#include "extensions/spdlog/SpdlogToActs.h"
+
 #include <edm4eic/vector_utils.h>
 
 


veprbl:
Some "errors" will not be fixable, as the track may simply miss the target surface. The best we could do is to silence the message and try to record some statistics (% of failures).

wdconinc:
Yeah, some of the tracks won't reach the surface, sure, but timing out after 1000 steps for many tracks in every event should not be ok and needs to be looked at. Are we even still inside the experimental hall then?

wdconinc:
If these error messages weren't hidden, we would not have enabled DRICH reconstruction in the ongoing campaign.

wdconinc:
I've got some stats functionality based on https://github.com/eic/EICrecon/pull/1012/commits/1ce35174c91abc7835d34f9b6549d3b522f2ad6c

wdconinc:
This PR is also included, in a way, in #1012 which used the spdlog stream and levels for acts logging.

c-dilks:
I was wondering why those errors were suddenly suppressed...

DRICH errors are what I meant by "wasting time" in #991, which was a quick solution but not a great one.

c-dilks:
The DRICH needs some way of knowing whether a charged particle passed through it or not before attempting track propagation. Hopefully there is a better solution than track propagation to some specific 'filter' surface.

wdconinc:
> Hopefully there is a better solution than track propagation to some specific 'filter' surface.

Yeah, what we should be doing is maybe propagating all tracks to the edge of a detector cylinder where we care about them, and keep track of the surfaces that are crossed, not propagating all tracks multiple times to a finite subset of these surfaces... which is what we're doing now, right?

c-dilks:
Let's discuss in #1017

wdconinc:
> The best we could do is to silence the message and try to record some statistics (% of failures).

This is possible in #1012 but because the DRICH uses TrackPropagation with a local logger created for each propagation individually we can't present the kind of suppression stats as we do for CKFTracking in e.g. https://github.com/eic/EICrecon/actions/runs/6295052149/job/17094395487#step:5:740. It would just print the line ""&lt;pattern&gt;" suppressed 1 times" for each line it currently prints...

In any case, closing this since after #1012 the logs now have all output this PR would have added.

PR #1019, opened 2023-09-25
MatchClusters_factory: switch to use PODIO collections as input

veprbl:
One of the few factories with incomplete porting to GetCollection/SetCollection instead of Get/Set with STL vectors.


wdconinc (approved):
Looks good to me!

PR #1020, opened 2023-09-25
src/benchmarks: deprecate JEventProcessorSequential::PrefetchT in favor of PODIO collections

veprbl:
PrefetchT relies on compatibility interface that copies PODIO pointer objects into `std::vector<> JFactory::mData` member. We'd like for all of our code to work with PODIO collections directly.

### DIFF
--- a/src/benchmarks/reconstruction/femc_studies/femc_studiesProcessor.cc
+++ b/src/benchmarks/reconstruction/femc_studies/femc_studiesProcessor.cc
@@ -19,11 +19,10 @@
 #include <DDRec/CellIDPositionConverter.h>
 
 // Include appropriate class headers. e.g.
-#include <edm4hep/SimCalorimeterHit.h>
-#include <edm4hep/MCParticle.h>
-#include <edm4eic/CalorimeterHit.h>
-#include <edm4eic/Cluster.h>
-#include <edm4eic/ProtoCluster.h>
+#include <edm4hep/SimCalorimeterHitCollection.h>
+#include <edm4hep/MCParticleCollection.h>
+#include <edm4eic/CalorimeterHitCollection.h>
+#include <edm4eic/ClusterCollection.h>
 #include <edm4eic/vector_utils.h>
 
 #include <JANA/JApplication.h>
@@ -232,18 +231,18 @@ void femc_studiesProcessor::Process(const std::shared_ptr<const JEvent>& event) 
   // ===============================================================================================
   // process MC particles
   // ===============================================================================================
-  auto mcParticles = event -> Get<edm4hep::MCParticle>("MCParticles");
+  const auto &mcParticles = *static_cast<const edm4hep::MCParticleCollection*>(event->GetCollectionBase("MCParticles"));
   double mceta    = 0;
   double mcphi    = 0;
   double mcp      = 0;
   double mcenergy = 0;
   int iMC         = 0;
-  for (auto mcparticle : mcParticles) {
-    if (mcparticle->getGeneratorStatus() != 1)
+  for (const auto mcparticle : mcParticles) {
+    if (mcparticle.getGeneratorStatus() != 1)
       continue;
-    const auto& mom = mcparticle->getMomentum();
+    const auto& mom = mcparticle.getMomentum();
     // get particle energy
-    mcenergy = mcparticle->getEnergy();
+    mcenergy = mcparticle.getEnergy();
     //determine mceta from momentum
     mceta = -log(tan(atan2(sqrt(mom.x * mom.x + mom.y * mom.y), mom.z) / 2.));
     // determine mcphi from momentum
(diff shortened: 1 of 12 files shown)


src/benchmarks/reconstruction/lfhcal_studies/lfhcal_studiesProcessor.cc
```diff
@@ -313,15 +312,15 @@ void lfhcal_studiesProcessor::Process(const std::shared_ptr<const JEvent>& event
...
+  const auto &simHits = *static_cast<const edm4hep::SimCalorimeterHitCollection*>(event->GetCollectionBase(nameSimHits));
+  for (const auto caloHit : simHits) {
+    float x         = caloHit.getPosition().x / 10.;
+    float y         = caloHit.getPosition().y / 10.;
+    float z         = caloHit.getPosition().z / 10.;
+    uint64_t cellID = caloHit.getCellID();
+    float energy    = caloHit.getEnergy();
     double time = std::numeric_limits<double>::max();
-    for (const auto& c : caloHit->getContributions()) {
+    for (const auto c : caloHit.getContributions()) {
```
wdconinc, line 323: Maybe -Wpedantic but should we keep the reference?
```suggestion
    for (const auto& c : caloHit.getContributions()) {
```
(applies in other places too)

veprbl: Good catch! Although, I only see one place where this needed to be fixed.

src/benchmarks/reconstruction/femc_studies/femc_studiesProcessor.cc
```diff
@@ -482,15 +481,15 @@ void femc_studiesProcessor::Process(const std::shared_ptr<const JEvent>& event)
     m_log->info("-----> found {} clusters" , clusters_calo.size());
     hRecNClusters_E_eta->Fill(mcenergy, clusters_calo.size(), mceta);
     int iCl = 0;
-    for (auto& cluster : clusters_calo) {
+    for (const auto cluster : clusters_calo) {
```
wdconinc, line 484: Applies here.

src/benchmarks/reconstruction/femc_studies/femc_studiesProcessor.cc
```diff
@@ -482,15 +481,15 @@ void femc_studiesProcessor::Process(const std::shared_ptr<const JEvent>& event)
...
       if (iCl < maxNCluster && enableTreeCluster){
         t_fEMC_cluster_E[iCl]       = (float)cluster.cluster_E;
         t_fEMC_cluster_NCells[iCl]  = (int)cluster.cluster_NTowers;
         t_fEMC_cluster_Eta[iCl]     = (float)cluster.cluster_Eta;
         t_fEMC_cluster_Phi[iCl]     = (float)cluster.cluster_Phi;
       }
       hRecClusterEcalib_E_eta->Fill(mcenergy, cluster.cluster_E/mcenergy, mceta);
-      for (auto & cluster_tower : cluster.cluster_towers){
+      for (const auto cluster_tower : cluster.cluster_towers){
```
wdconinc, line 492: Here.

src/benchmarks/reconstruction/lfhcal_studies/lfhcal_studiesProcessor.cc
```diff
@@ -625,10 +624,10 @@ void lfhcal_studiesProcessor::Process(const std::shared_ptr<const JEvent>& event
   hRecFNClusters_E_eta->Fill(mcenergy, iClF, mceta);
   // fill hists for highest Island cluster
   iClF          = 0;
-  for (auto& cluster : lfhcalClustersF) {
+  for (const auto cluster : lfhcalClustersF) {
```
wdconinc, line 627: Here.

src/benchmarks/reconstruction/lfhcal_studies/lfhcal_studiesProcessor.cc
```diff
@@ -643,31 +642,31 @@ void lfhcal_studiesProcessor::Process(const std::shared_ptr<const JEvent>& event
 
   if (enableECalCluster){
     try {
-      auto fEMCClustersF = event->Get<edm4eic::Cluster>("EcalEndcapPClusters");
+      const auto &fEMCClustersF = *static_cast<const edm4eic::ClusterCollection*>(event->GetCollectionBase("EcalEndcapPClusters"));
       m_log->info("-----> found fEMCClustersF:" , fEMCClustersF.size());
-      for (auto& cluster : fEMCClustersF) {
+      for (const auto cluster : fEMCClustersF) {
```
wdconinc, line 647: Here.

wdconinc (commented):
Marked some more. As I said, it's pedantic and not performance critical of course.

wdconinc:
(Really, I just want to be able to say I have reviewed code from the Canadian seat of government, Parliament Hill.)

src/benchmarks/reconstruction/femc_studies/femc_studiesProcessor.cc
veprbl: But the PODIO collection iterators don't return a reference https://github.com/AIDASoft/podio/blob/babec745c12727cd5da590abbe1e3cf1ab8743ad/python/templates/macros/iterator.jinja2#L33-L36

wdconinc: Right. I keep thinking they are more std::vector than they actually are.

wdconinc (approved):

PR #1021, opened 2023-09-25
fix: better suppression of log messages in TrackPropagation

wdconinc:
There is a lot of output in the TrackPropagation algorithm, and it is not easy to suppress using #1012 because it comes from Acts calls in multiple functions. That is because #1012 only introduces local loggers, not class-wide loggers (which is intentional because class-wide suppression statistics are not as useful as event-scoped suppression statistics, which requires local logger that are destructed at the end of the event).

This moves the Acts::Logger in TrackPropagation to class scope so it can be created in a different function scope than where it is used. This requires then that we move (back?) to using propagateMany instead of just propagate (with loops in the factory). This brings factory and algorithm slightly more in line with current practice of passing collections instead of passing std::vector<T*>. It does require more input, which is currently initialized in the factory instead of the algorithm, where (arguably) it belong.

TODO:
- [ ] pass m_target_surface_ID and m_target_detector_ID as std::vector parallel to target_surface_list in propagateMany (this PR)
- [ ] migrate TrackPropagation_factory::SetPropagationSurfaces to the algorithm (different PR)

- [x] Bug fix (issue: too much output from TrackPropagation)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


src/algorithms/tracking/TrackPropagation.h
```diff
@@ -44,8 +44,8 @@ namespace eicrecon {
 
         /** Propagates a collection of trajectories to a given surface
          * @remark: being a simple wrapper of propagate(...) this method is more suitable for factories */
-        std::vector<std::unique_ptr<edm4eic::TrackPoint>> propagateMany(std::vector<const ActsExamples::Trajectories *> trajectories,
-                                                         const std::shared_ptr<const Acts::Surface> &targetSurf);
+        std::unique_ptr<edm4eic::TrackSegmentCollection> propagateMany(std::vector<const ActsExamples::Trajectories *> trajectories,
+                                                         const std::vector<std::shared_ptr<Acts::Surface>> &target_surface_list);
```
c-dilks, line 48: So `propagateMany` now propagates to a list of surfaces, as `propagateToSurfaceList` does? If so, consider combining these methods.

wdconinc: That's the idea, yeah. There are two users of this TrackPropagation algorithm right now: RICH projections to radiators and track projections to calorimeters. Aligning the methodology will only facilitate future maintenance (not to mention avoid duplicate evaluations). They differ in how the surfaces are passed, with the RICH using its ActsGeo interface in the RichGeo service and the TrackPropagation_factory doing hard-coded string lookups in the factory. In both cases it's non-ideal how much is done outside the algorithm, in my opinion. But a first step is to have both of these methodologies use the same algorithm interface/entry point, while keeping the current surface construction unchanged.

c-dilks: > non-ideal how much is done outside the algorithm

`RichTrack_factory` is also guilty here (which uses `richgeo`'s `ActsGeo`). For the RICHes our preference was to keep anything related to the geometry in `richgeo` to keep subsystem dependence out of the algorithms. At the moment only the dRICH uses `richgeo`, but the doors are open for the pfRICH (and possibly DIRC); the pfRICH already has an (untested) IRT geometry representation.

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -54,26 +54,43 @@ namespace eicrecon {
...
 
-            auto result = propagate(traj, targetSurf);
-            if(!result) continue;
+            auto this_propagated_track = propagated_tracks->create();
+            for(size_t isurf = 0; auto surf: target_surface_list) {
+
+                auto prop_point = propagate(traj, surf);
+                if(!prop_point) continue;
+#if EDM4EIC_VERSION_MAJOR >= 3
+//                prop_point->surface = m_target_surface_ID[isurf];
```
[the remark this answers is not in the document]
wdconinc: The reason for no diffs is because of the disabled lines in https://github.com/eic/EICrecon/pull/1021/files#diff-8d72010a235f53e207b8ccc1bccc1579c12a373bc56bf3dbc13798aa45827564L294-L297.

src/algorithms/tracking/TrackPropagation.cc
```diff
@@ -54,26 +54,43 @@ namespace eicrecon {
...
+//                prop_point->system = m_target_detector_ID[isurf];
+#endif
```
wdconinc, line 82:
```suggestion
```
Unneeded if we store this info inside the surfaces, after #1057.

PR #832, opened 2023-08-06
fix: avoid compiling all source files twice, in library and in plugin

wdconinc:
### Briefly, what does this PR introduce?
Before this PR, we compiled most if not all source files twice:
```console
$ find build -name "InclusiveKinematicsTruth_factory.*"
build/src/global/reco/CMakeFiles/reco_library.dir/InclusiveKinematicsTruth_factory.cc.o.d
build/src/global/reco/CMakeFiles/reco_library.dir/InclusiveKinematicsTruth_factory.cc.o
build/src/global/reco/CMakeFiles/reco_plugin.dir/InclusiveKinematicsTruth_factory.cc.o.d
build/src/global/reco/CMakeFiles/reco_plugin.dir/InclusiveKinematicsTruth_factory.cc.o
```
This PR creates the plugin from `${PLUGIN_NAME].cc`, linked against the library which is created from all other `*.cc` files (but excludes the plugin file). This PR also changes all libraries to shared.

Comparison in sizes:
- `build` from 598 MB to 412 MB
- `lib` from 297 MB to 119 MB

Comparison in compilation times:
- before:
```console
real    8m17.809s
user    52m34.625s
sys     2m47.702s
```
- after:
```console
real    3m57.501s
user    27m8.591s
sys     1m22.740s
```

TODO:
- [ ] Ensure strong rpath linking from plugin to library if both are in the same directory
- [ ] Fix `bin/eicmkplugin.py` (and put in a test)
- [ ] Fix llvm-cov job

### What kind of change does this PR introduce?
- [x] Bug fix (issue #518)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1023, opened 2023-09-25
fix: remove all odr violations, and enforce them in ASAN_OPTIONS

wdconinc:
This removes the remaining [one-definition rule (odr)](https://en.cppreference.com/w/cpp/language/definition) violations in the CKFTracking (tracking.so plugin and library), and the reco.so plugin and library. It also renames the misnamed janadot.cc to janatop.cc. This now allows us to turn on the detect_odr_violations in ASAN_OPTIONS.

The next steps could be to remove the flags PLUGIN_USE_CC_ONLY and PLUGIN_GLOB_ALL again, and default to PLUGIN_USE_CC_ONLY (which would make plugin and library based on entirely distinct source files). However, that's not quite working yet (for reasons I have to understand and resolve, are limited to services, and which don't introduce odr violations). In the mean time we should push this into main and prevent breakage from getting added inadvertently.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
> The next steps could be to remove the flags PLUGIN_USE_CC_ONLY and PLUGIN_GLOB_ALL again

Done. There are now three options:
- plugin only (default): all *.cc added into plugin (e.g. src/detectors/)
- plugin WITH_LIBRARY: name.cc into plugin, all the other *.cc into library (e.g. src/global/tracking)
- 'plugin' WITH_LIBRARY WITHOUT_PLUGIN: name.cc unused, all the other *.cc into library (e.g. src/algorithms)

wdconinc:
When you compile less, it also generally goes faster and you end up with smaller binaries. Who would have thought?

Screenshot_20230925-173728

Down from
Screenshot_20230806-161459
in the beginning of August https://github.com/eic/EICrecon/pull/832#issuecomment-1666969945

ruse-traveler (approved):
LGTM!

cmake/jana_plugin.cmake
```diff
@@ -118,8 +122,11 @@ macro(plugin_glob_all _name)
 
     # But... GLOB here makes this file just hot pluggable
     file(GLOB LIB_SRC_FILES CONFIGURE_DEPENDS *.cc *.cpp *.c)
-    #file(GLOB PLUGIN_SRC_FILES CONFIGURE_DEPENDS ${_name}.cc)
-    file(GLOB PLUGIN_SRC_FILES CONFIGURE_DEPENDS *.cc *.cpp *.c)
+    if(${_name}_WITH_LIBRARY)
+        file(GLOB PLUGIN_SRC_FILES CONFIGURE_DEPENDS ${_name}.cc)
+    else()
+        file(GLOB PLUGIN_SRC_FILES CONFIGURE_DEPENDS *.cc *.cpp *.c)
```
veprbl, line 128: `${_name}.cc` is a subset of `*.cc`

wdconinc: See line 152.

veprbl: I didn't realize that was in. Nevermind this then.

wdconinc: We can still fix issues even if it is in.

veprbl: Would be silly of me to leave messages if it wasn't fixable.

I think, the logic was and still is a bit complicated. If I understand correctly, the reason to move implementation of plugins into second-level libraries is for reuse of factory definitions, right? What if we've marked all factory symbols as invisible (they don't need to be exported as calls are routed via vtable). Then we could, at least, go back to static linking.

Also I'm not fond of eicrecon_* cmake functions. Stuff should be either a plugin or a library. We can have a single helper function for making a JANA plugin, but all the rest can be normal cmake stuff operating on single target at a time.

I'm sure I'm missing some important technical obstacles, but these are my current thoughts.

wdconinc: I don't like the plugin/factory separation either. But it's what we have and I've tried to rewrite it twice into one or the other and failed both times. It's getting closer to being where I want it to be, at least. I don't think there should be any directory that needs to have both a plugin and a library.

I don't like the CMake macros all that much either, but if we ask for people to copy all that stuff all the time whenever they create a plugin or a library, then I don't want to have to clean up unnecessary or incorrect target_link_libraries all over. I think it at least isolates everything and makes sure it's consistent.

PR #1025, opened 2023-09-26
Enable user configurability of jet reconstruction

ruse-traveler:

This PR updates the interfaces to jet reconstruction algorithm so as to allow users to change parameters of the algorithm from the command line, via a configuration file, or through initializer lists. For example:

`eicrecon -Preco:ReconstructedJets:minCstPt=2`

will raise the minimum pT of objects fed to the jet algorithm to 2 GeV/c for the reconstructed jet collection.

The utilized jet algorithm, recombination scheme, and area type can also be set in the same manner. If any of these are set to an unknown value (e.g. `-Preco:ReconstructedJets:jetAlgo=mystery_algo`), the EICrecon algorithm will issue a warning and run with a default value instead.

- [ ] Bug fix (issue #__)
- [X ] New feature (issue #901 )
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/algorithms/reco/JetReconstruction.cc
```diff
@@ -4,26 +4,44 @@
...
+      m_cfg.jetAlgo = m_defaultFastjetOpts.jetAlgo;
+    }
+    if (m_mapRecombScheme.find(m_cfg.recombScheme) == m_mapRecombScheme.end()) {
+      m_log->warn(" Unknown recombination scheme \"{}\" specified! Setting scheme to default and proceeding.", m_cfg.recombScheme);
+      m_cfg.recombScheme = m_defaultFastjetOpts.recombScheme;
+    }
+    if (m_mapAreaType.find(m_cfg.areaType) == m_mapAreaType.end()) {
+      m_log->warn(" Unknown area type \"{}\" specified! Setting type to default and proceeding.", m_cfg.areaType);
+      m_cfg.areaType = m_defaultFastjetOpts.areaType;
+    }
```
veprbl, line 39: If user's choice of option is not found, don't replace it with something else. Throw an exception (we commonly use JException), let it complain loudly.

bspage912:
I will try to look at this tomorrow, but I have a question on the expected behavior. If I execute the command you list above "eicrecon -Preco:ReconstructedJets:minCstPt=2", does this modify the existing branches created using  ReconstructedJets_factory - currently ReconstructedJets and ReconstructedChargedJets, or does it create a new branch(es) with this specific change? One could for example think of wanting to generate a separate branches, each with a different R, for comparisons. Does this PR allow that functionality?

veprbl:
@bspage912 The :ReconstructedJets: part determines which "factory tag" (collection name) you are setting options for. Different radii can be setup in the arguments to the factory generator in reco.cc.

src/algorithms/reco/JetReconstruction.cc
ruse-traveler: On 2nd thought: I like that approach. My 1st thought was to keep it "not so intrusive", but if a user is actively specifying something other than the default configuration, then they probably want loud complaints. I'll swap out the warns for errors and have it throw an exception!

ruse-traveler:
> I will try to look at this tomorrow, but I have a question on the expected behavior. If I execute the command you list above "eicrecon -Preco:ReconstructedJets:minCstPt=2", does this modify the existing branches created using ReconstructedJets_factory - currently ReconstructedJets and ReconstructedChargedJets, or does it create a new branch(es) with this specific change? One could for example think of wanting to generate a separate branches, each with a different R, for comparisons. Does this PR allow that functionality?

Exactly as @veprbl said! That command will change the min cst. pt for the existing reconstructed jets collection. And while I don't think we have any tools in place yet that would dynamically generate a new branch in events based on commands like that one, this PR does make it easier for a user to add their own factory!

Previously, since the parameters were hard coded into the algorithm, all factories using the `JetReconstruction` algorithm had to use those. With this PR, we can set the parameters (preferably via initializer lists in `reco.cc`, but also possibly from the command line) independently for each factory.

So one way users could add their own branch is by adding a new factory to `reco.cc`, e.g.

```
app->Add(new JChainMultifactoryGeneratorT<GeneratedJets_factory>(
                  "MyGeneratedJets",
                  {"MCParticles"},
                  {"MyGeneratedJets"},
                  {
                    .jetAlgo="ee_kt_algorithm"
                  },
                  app
));
```

And then recompile and rerun! (I suspect -- but haven't tried it out yet -- you could also do something similar with a plugin processor rather than modifying `reco.cc`...)

veprbl (approved):

bspage912 (approved):
Sorry I'm late. I think the functionality is what we need and if Dima is ok with the code and implementation - I'm fine.

PR #1031, opened 2023-09-28
feat(ci): test eicmkplugin in CI

wdconinc:
This adds a CI test for the eicmkplugin script (under the philosophy that if it's not tested, it is probably broken).

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -386,7 +386,7 @@ jobs:
           export JANA_PLUGIN_PATH=$PWD/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
           chmod a+x bin/*
           $PWD/bin/eicmkplugin.py MyCustomPlugin
-          cmake -S MyCustomPlugin -B MyCustomPlugin/build
+          cmake -S MyCustomPlugin -B MyCustomPlugin/build -DUSER_PLUGIN_OUTPUT_DIRECTORY=$PWD/lib/EICrecon/plugins
           cmake --build MyCustomPlugin/build --target install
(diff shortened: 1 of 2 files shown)


wdconinc:
Marking as draft since entirely untested...

wdconinc:
We could also stick this in the build job and then run it along with the other benchmark plugins in their test job.

wdconinc:
`target_compile_definitions({0}_plugin PUBLIC HAVE_PODIO)` should not be necessary in plugins linked against EICrecon, and not even in EICrecon linked against JANA2. This would be a good case for a set_target_property on the JANA::jana2_*_lib in their src/libraries/JANA/JANAConfig.cmake.in, so it is set by all downstream users, or even better setting this initially where the target is originally created at https://github.com/JeffersonLab/JANA2/blob/ea9ebf5aa2f0e00db587cf08ec106030188e54b7/src/libraries/JANA/CMakeLists.txt#L172 instead of using a project-wide set_compile_definitions.

wdconinc:
This has passed the latest checks (and failed the last pipeline due to the LSAN crash hopefully fixed in #1048): it is ready for review. We may not have many users using this boilerplate script for their plugins, but as long as we ship it we should probably keep it in our tests.

ruse-traveler (approved):
Looks good! I actually use this script pretty frequently, so I'm glad to see it get validated 😅

wdconinc:
> I actually use this script pretty frequently

Interesting... Can you write some comments on how you use it, for which scenarios, and whether there are specific testing scenarios we should include? Is this a methodology that is inspired by some other BNL experiments?

ruse-traveler:
> > I actually use this script pretty frequently
> 
> Interesting... Can you write some comments on how you use it, for which scenarios, and whether there are specific testing scenarios we should include? Is this a methodology that is inspired by some other BNL experiments?

Sure! It's not so much inspired by other BNL experiments, but I find it really convenient for when I need to make a short plugin to test something specific; e.g. recently I wanted to take a look at our reco particle-truth associations, so I used this to get the plugin skeleton and go from there.

wdconinc:
> short plugin to test something specific

Does it make sense to create a place where to put these? Something like an EICrecon-contrib repository?

ruse-traveler:
> > short plugin to test something specific
> 
> Does it make sense to create a place where to put these? Something like an EICrecon-contrib repository?

Oh! That could be very useful! Would this be a (more-or-less) unprotected repo for users to add plugins? We have something similar for sPHENIX: it's a repository where anyone part of the sPHENIX org can push their code that collects things like analysis code.

PR #1033, opened 2023-09-30
remove unused OffMomentumReconstruction_factory.*, RomanPotsReconstruction_factory.*

veprbl:
Trying to ensure that nothing in our codebase uses `JFactoryPodioT<T>::Set`.

This were replaced with `MatrixTransferStatic_factoryT` in #879


simonge (approved):
Good to me, I don't think there's a good reason to keep them around

Issue #560, opened 2023-03-21
IRT-environment related error message when running eicrecon


rahmans1:
### Environment: (where does this bug occur, have you tried other environments)
eic-shell nightly container

eic-info
 - Spack: 0.19.1 (5e0d2107348eed6cbe6deca43a30f5b06c5e40af)
 - Python: 3.10.8
 - Platform: linux-debian-zen2
 - Concretizer: clingo
 - dloznd2 abseil-cpp@20220623.0+ipo+shared build_system=cmake build_type=Release cxxstd=11
 - khlknat acts@21.1.0~alignment~analysis~autodiff~benchmarks~cuda+dd4hep~edm4hep+examples+fatras~fatras_geant4+geant4~hepmc3+identification~integration_tests+ipo+json~legacy~odd~onnx~profilecpu~profilemem~pythia8+python~svg~sycl+tbb+tgeo~unit_tests build_system=cmake build_type=Release log_failure_threshold=MAX
 - at6y22c acts-dd4hep@1.0.1+ipo build_system=cmake build_type=Release
 - tjaj7zo actsvg@0.4.26+examples+ipo+meta build_system=cmake build_type=Release
 - odnziwj afterburner@0.1.2+ipo+root+zlib build_system=cmake build_type=Release
 - 2m4azot aida@3.2.1 build_system=generic
 - fso55ya assimp@5.2.4+ipo+shared build_system=cmake build_type=Release
 - lolqzlz berkeley-db@18.1.40+cxx~docs+stl build_system=autotools patches=26090f4,b231fcc
 - xzngwjj boost@1.80.0+atomic+chrono~clanglibcpp~container~context~contract~coroutine+date_time~debug+exception~fiber+filesystem+graph~graph_parallel~icu+iostreams~json+locale+log+math~mpi+multithreaded~nowide~numpy~pic+program_options+python+random+regex+serialization+shared+signals~singlethreaded~stacktrace+system~taggedlayout+test+thread+timer~type_erasure~versionedlayout+wave build_system=generic cxxstd=98 patches=a440f96,b8569d7 visibility=hidden
 - r3scpwi bzip2@1.0.8~debug~pic+shared build_system=generic
 - gzfdj3r cairo@1.16.0+X+fc+ft+gobject+pdf~png~svg build_system=autotools patches=7c4da77
 - y7irjob clhep@2.4.6.0~cms+ipo build_system=cmake build_type=Release cxxstd=17
[...]

PR #1036, opened 2023-10-02
fix: rm independent geometry reading in RICHGeo

wdconinc:
This removes the unused (deprecated) IrtGeo interfaces that load their own compact file. This is in particular relevant because RichGeo cleans up the geometry regardless with a `destroyInstance` that is applied on both the DD4hep_service and RichGeo_service geometries. That could potentially cause a race condition when both DD4hep_service and RichGeo_service try to destroy the same geometry instance (it probably is happening already but is caught by the try catch around both instances). This change also allows us to move towards more const correctness in the treatment of DD4hep geometries (i.e. not passing the geometry as a non-const pointer).

- [x] Bug fix (issue: `destroyInstance` in RichGeo_service also applies on DD4hep_service geometry)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @c-dilks @chchatte92 

No.

No.

### DIFF
--- a/src/services/geometry/richgeo/IrtGeo.cc
+++ b/src/services/geometry/richgeo/IrtGeo.cc
@@ -10,18 +10,6 @@ richgeo::IrtGeo::IrtGeo(std::string detName_, dd4hep::Detector *det_, std::share
   Bind();
 }
 
-// alternate constructor: use compact file for DD4hep geometry (backward compatibility)
-richgeo::IrtGeo::IrtGeo(std::string detName_, std::string compactFile_, std::shared_ptr<spdlog::logger> log_) :
-  m_detName(detName_), m_log(log_)
-{
-  // build DD4hep detector from compact file
-  m_det = &dd4hep::Detector::getInstance();
-  m_det->fromXML(compactFile_);
-
-  // set IRT and DD4hep geometry handles
-  Bind();
-}
-
 // Bind() -----------------------------------------
 // set IRT and DD4hep geometry handles
 void richgeo::IrtGeo::Bind() {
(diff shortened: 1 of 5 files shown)


wdconinc:
Maybe we can evaluate removing the try-catch in:
```
DD4hep_service::~DD4hep_service(){
    try {
        if(m_dd4hepGeo) m_dd4hepGeo->destroyInstance();
        m_dd4hepGeo = nullptr;
    } catch (...) {}
}
```

c-dilks:
As long as `capybara` finds no diffs, this is fine with me.

Time to also remove [`create_irt_auxfile`](https://github.com/eic/drich-dev/blob/5568a41e97b23c329f4ae02decd82795c5dc2674/src/create_irt_auxfile.cpp), which is the only reason these alternate constructors existed, having supported our transition away from Juggler.

If IRT auxfile creation is still needed by anyone, it is straightforward to write an EICrecon plugin to do this, following [`create_irt_auxfile`](https://github.com/eic/drich-dev/blob/5568a41e97b23c329f4ae02decd82795c5dc2674/src/create_irt_auxfile.cpp).

c-dilks (approved):

c-dilks:
> it is straightforward 

Haha, not quite. The `spack` build of IRT does not have ROOT dictionary generation:

- https://github.com/eic/irt/pull/32
- https://github.com/eic/EICrecon/issues/560

So anyone wanting to make IRT auxfiles will need to re-build `irt` with dictionary generation enabled, then rebuild EICrecon... or fix the `spack` build problem for real.

(cc. @chchatte92)

wdconinc:
Is there any other reason for enabling `root_io` on `irt` in the container?

c-dilks:
> Is there any other reason for enabling `root_io` on `irt` in the container?

Negative, that's why I turned it off. I suspect there is some hard-coded path issue in the dictionary generation. I was never able to reproduce the issue with local (non-`spack`) builds.

Issue #1040, opened 2023-10-03
Release notes are not structured enough


wdconinc:
**Is your feature request related to a problem? Please describe.**
Users don't find the release notes very useful. There is too much noise and not easy to find points relevant to them.

**Describe the solution you'd like**
Maybe a way to improve this is to add headers or sections by tags we already assign. This seems possible in config at: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

**Describe alternatives you've considered**
Hand-curating (bleh)...

PR #1041, opened 2023-10-03
feat(ci): release notes definition

wdconinc:
Sets up release notes to have categories. Note sure how to test...

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1040)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,32 @@
+changelog:
+  exclude:
+    authors:
+      - 'app/pre-commit-ci'
+  categories:
+    - title: 🕸️ Tracking
+      labels:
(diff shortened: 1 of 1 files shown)


veprbl (approved):
Seems to work when playing in https://github.com/eic/EICrecon/releases/new.
We might have to reorder the list later, if that could allow us to define a pecking order (bug beats far-forward beats tracking).

wdconinc:
> We might have to reorder the list later

Yeah, I see now that it doesn't want to list things twice, so we end up with some PRs under Tracking that are more general. Well, I'll consider it an improvement over the status quo.

PR #1035, opened 2023-10-02
fix: allow for reordered DD4hep_service geometry arguments

wdconinc:
### Briefly, what does this PR introduce?
This fixes the DD4hep_service to allow querying the geometry in any order, and avoids race conditions in the `detector()` and `cellIDPositionConverter()` calls.

TODO:
- [x] revert 167ae084f932502799ace697ccf0c9d30e1a5308

### What kind of change does this PR introduce?
- [x] Bug fix (issue #1032)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

Issue #1032, opened 2023-09-29
Order of geometry arguments from DD4hep_service in algorithm init matters


simonge:
A bit of a confusing bug as it relies on a number of conditions when multiple collections are output but when only writing one the argument list for initializing an algorithm e.g. FarDetectorTrackerCluster/MatrixTransferStatic needs to take:
`geoSvc->cellIDPositionConverter()` 
before 
`geoSvc->detector()`
(from the DD4hep_service)

The other ordering results in the error at the bottom here despite no changes in the code other than the order of the init arguments.

If a collection with the correct ordering is also included in the output **and** comes first alphabetically everything runs as expected.

Searching the code looks like these two algorithms are the only place this is currently used. The solution might be to remove the cellIDPositionConverter from the DD4hep_service and take it from the detector inside the algorithm as it is done everywhere else i.e.

`m_converter = std::make_shared<const dd4hep::rec::CellIDPositionConverter>(const_cast<dd4hep::Detector&>(*detector));
`
### Environment: 
On the main branch but testing using add_simple_tracker_hit_clustering branch in the eic-shell

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Run eicrecon on the add_simple_tracker_hit_clustering branch with only `-Ppodio:output_include_collections=TaggerTrackerClusterPositions `
2. Runs ok if calling `-Ppodio:output_include_collections=TaggerTrackerClusterPositions,ForwardOffMRecParticles`
3. But not if renamed 'ForwardOffMRecParticles->ZForwardOffMRecParticles'
-----
4. Same on main branch but changing the order of MatrixTransferStatic init to
```
void eicrecon::MatrixTransferStatic::init(const dd4hep::Detector* det,
					  std::shared_ptr<const dd4hep::rec::CellIDPositionConverter> id_conv,
                                          std::shared_ptr<spdlog::logger> &logger) {
```
6. Then running with `-Ppodio:output_include_collections=ForwardOffMRecParticles`

### Expected Result: (what do you expect when you execute the steps above)
Output with the required collection.

### Actual Result: (what do you get when you execute the steps above)
```
Thread model: pthreads
140287080666816: 
       `- JSignalHandler::handle_sigsegv(int, siginfo_t*, void*) (0x7f973d29bff6)
        `- /lib/x86_64-linux-gnu/libc.so.6 (0x7f973cd7efd0)
[...]

PR #1045, opened 2023-10-04
fix: use unique_ptr for geometry in DD4hep_service, hand out not_null const pointers

wdconinc:
While looking into #1032, #1035, #1036 I started realizing two things:
- cell ID position converters are created by algorithms from geometry instead of doing this once in the service,
- there is no const protection in the detector pointers passed to algorithms (i.e. an algorithm could change geometry),
- there is no protection against null pointers being passed from geometry.

The first two points are related since the interface for creating a cell ID position converter requires a non-const detector (see also https://github.com/AIDASoft/DD4hep/pull/1148).

This PR centralizes the creation of cell ID position converters, clarifies ownership through unique_ptrs, and makes sure no null pointers are passed to the algorithms. In places where the constructor gets the const not_null pointer, we can also impose this in the class on the relevant member variable.

- [x] Bug fix (issue: avoid creation of cell ID position converters in algorithms)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators:
  - @simonge reordered MatrixTransferStatic init arguments
  - @c-dilks @chchatte92 this also affects RichGeo_service etc.

Yes, now requires GSL (which has always been in eic-shell and is a requirement).

No.


veprbl:
Do we really want to smear non-null check across our interfaces? Is it not enough to just ensure that the geometry service doesn't return null.

wdconinc:
> Do we really want to smear non-null check across our interfaces? Is it not enough to just ensure that the geometry service doesn't return null.

It's heavy on notation, I admit, but the requirement for a not_null pointer is in this case imposed by the algorithms, not by the service. The DD4hep_service can satisfy the demand by promising to return a not_null pointer. Putting the requirement in the algorithms makes clear to anyone working in those algorithms that `detector` can be relied upon to be not_null (yeah, this benefit it kinda negated by the whole init-function and clearer in the RichGeo_service).

wdconinc:
@veprbl I removed the gsl::not_null on all algorithms init calls in https://github.com/eic/EICrecon/pull/1045/commits/b9b97f4ccbba7bb7a20c9c3ae27a44701e06232b

This doesn't really add much utility on the init calls if we can't set it on the class members. I've kept it in DD4hepBfield since there we construct with the detector geometry provided as constructor argument (similar to RichGeo_services).

c-dilks:
The dRICH benchmarks look okay with this PR.

src/algorithms/calorimetry/CalorimeterHitDigi.h
```diff
@@ -13,6 +13,7 @@
 
 #pragma once
 
+#include <gsl/gsl>
```
veprbl, line 16: Unused? (and ditto for other occurrences)

src/services/geometry/dd4hep/DD4hep_service.cc
```diff
@@ -30,21 +29,22 @@ DD4hep_service::~DD4hep_service(){
...
 /// Return pointer to the cellIDPositionConverter object.
 /// Call Initialize if needed.
 //----------------------------------------------------------------
-std::shared_ptr<const dd4hep::rec::CellIDPositionConverter>
-DD4hep_service::cellIDPositionConverter() {
+gsl::not_null<const dd4hep::rec::CellIDPositionConverter*>
+DD4hep_service::converter() {
     std::call_once(init_flag, &DD4hep_service::Initialize, this);
-    return (m_cellid_converter);
+    return (m_cellid_converter.get());
```
veprbl, line 47:
```suggestion
    return m_cellid_converter.get();
```

veprbl:
> @veprbl I removed the gsl::not_null on all algorithms init calls in [b9b97f4](https://github.com/eic/EICrecon/commit/b9b97f4ccbba7bb7a20c9c3ae27a44701e06232b)
> 
> This doesn't really add much utility on the init calls if we can't set it on the class members. I've kept it in DD4hepBfield since there we construct with the detector geometry provided as constructor argument (similar to RichGeo_services).

Yes. That was my point, that it's probably the return values where gsl::not_null makes sense (I don't claim any deep philosophical understanding of GSL, though).

src/algorithms/calorimetry/CalorimeterHitDigi.h
wdconinc: Yeah, didn't include that in my sed. Let me add (subtract) them.

src/services/geometry/dd4hep/DD4hep_service.cc
```diff
@@ -30,21 +29,22 @@ DD4hep_service::~DD4hep_service(){
 /// Return pointer to the dd4hep::Detector object.
 /// Call Initialize if needed.
 //----------------------------------------------------------------
-dd4hep::Detector* DD4hep_service::detector() {
+gsl::not_null<const dd4hep::Detector*>
+DD4hep_service::detector() {
     std::call_once(init_flag, &DD4hep_service::Initialize, this);
-    return (m_dd4hepGeo);
+    return (m_dd4hepGeo.get());
```
wdconinc, line 35:
```suggestion
    return m_dd4hepGeo.get();
```

veprbl (approved):
Diff LGTM, capybara happy

PR #1046, opened 2023-10-04
Rearrange the position calls in MatrixTransferStatic algorithm

simonge:
Hopefully increases the efficiency of the MatrixTransferStatic algorithm by converting 2D->3D coordinates rather than 2D->3D->2D as CellDPositionConverter does. Makes use of the readout passed as config so it doesn't need to look this up every call from the CellID.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: Hopefully streamlines the algorithm a bit. Very minor improvement.

NA
NA


wdconinc:
@simonge Do you want to revive this (fix conflicts) or is this something that can be closed due to other activity?

simonge:
Might as well try and bring it up to date. Hopefully in a few minutes...

wdconinc:
I assume you don't expect any differences due to this change? https://eicrecon.epic-eic.org/pr/1046/capybara/rec_dis_18x275_minQ2=1000_craterlake/index.html#ForwardRomanPotRecParticles: green before, red after.

simonge:
> I assume you don't expect any differences due to this change? https://eicrecon.epic-eic.org/pr/1046/capybara/rec_dis_18x275_minQ2=1000_craterlake/index.html#ForwardRomanPotRecParticles: green before, red after.

I presume there isn't such a report from when this was last up to date?

wdconinc:
Unfortunately not (we don't keep the graphics output, but even the logs are expired, e.g. https://github.com/eic/EICrecon/actions/runs/6409910580/job/17402104851?pr=1046).

simonge:
The problem here is I don't completely understand the translation between segmentation position, detector element position and global position. As it is, this tried to go straight from segmentation to global, which appears to be wrong. By the time it works as expected the code is less clear and any efficiency increase is negligible. 

Functions could be added to dd4hep to more easily access the intermediate translation steps without doing them twice but it's not worth dwelling on as this method is hopefully being superseded.

PR #1050, opened 2023-10-05
fix: use subset collection when collecting hits in TrackSeeding

wdconinc:
This fixes a regression introduced in #983 that caused realistic track seeding to fail. Essentially, you cannot just add an object from one podio collection into another podio collection "by reference", i.e. without cloning. You can do that with pointers in std::vector, which was done here before #983. We can achieve the same behavior in podio, but we have to be more intentional about it and define a subset collection.

Of course, podio complains loudly when you try to add an object from one collection into another collection. It emits an exception: `Object already in a collection. Cannot add it to a second collection`. But JANA2 eats that exception and doesn't even print it :angry:

TODO:
- [x] check diffs

- [x] Bug fix (issue: no track seeding results after #983)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @ShujieL 

No.

Yes, CentralTrackSeedingResults and dependents are restored.

### DIFF
--- a/src/global/tracking/TrackSeeding_factory.cc
+++ b/src/global/tracking/TrackSeeding_factory.cc
@@ -63,6 +63,7 @@ void eicrecon::TrackSeeding_factory::Process(const std::shared_ptr<const JEvent>
     // Collect all hits
     // FIXME Collection is better done with a TrackerHitCollector factory
     edm4eic::TrackerHitCollection total_hits;
+    total_hits.setSubsetCollection();
 
     for(auto input_tag: GetInputTags()) {
         auto hits = static_cast<const edm4eic::TrackerHitCollection*>(event->GetCollectionBase(input_tag));


wdconinc:
@nathanwbrei Why does an exception emitted in a JChainFactoryT::Process() get silently absorbed? Can we at least print that somewhere, even if we discard the factory and continue running? Is this a policy in EICrecon JChainFactoryT/JFactoryPodioT or in JANA2 (from what I can tell, it's in JANA2 somewhere)?

ShujieL (approved):

nathanwbrei:
> @nathanwbrei Why does an exception emitted in a JChainFactoryT::Process() get silently absorbed? Can we at least print that somewhere, even if we discard the factory and continue running? Is this a policy in EICrecon JChainFactoryT/JFactoryPodioT or in JANA2 (from what I can tell, it's in JANA2 somewhere)?

Replied here: https://github.com/eic/EICrecon/issues/1052

Issue #1052, opened 2023-10-05
Exception emitted in a JChainFactoryT::Process() gets silently absorbed


veprbl:
@nathanwbrei Why does an exception emitted in a JChainFactoryT::Process() get silently absorbed? Can we at least print that somewhere, even if we discard the factory and continue running? Is this a policy in EICrecon JChainFactoryT/JFactoryPodioT or in JANA2 (from what I can tell, it's in JANA2 somewhere)?

_Originally posted by @wdconinc in https://github.com/eic/EICrecon/issues/1050#issuecomment-1747880942_

veprbl:
Related to #625?

nathanwbrei:
I think this is indeed related to #625, more specifically the extremely weird way in which the JEventProcessorPodio used to handle "missing" collections. Definitely not something we want. Feel free to remove and see what happens.

PR #1053, opened 2023-10-05
fix: print exceptions when they happen, and summarize failed collections at the end

wdconinc:
Instead of silently eating the exception, this modifies the JEventProcessorPODIO to print the failed collection and its exception when it first happens, and then prints the failed collections again at the end, all using the error log stream. That results in log streams such as:
```
[acts_init] [info] DD4Hep geometry converted!
[acts_init] [info] Checking surfaces...
[acts_init] [info] Loading magnetic field...
[acts_init] [info] ActsGeometryProvider initialization complete
[JEventProcessorPODIO] [error] Failed to get CentralTrackSeedingResults due to exception: Object already in a collection. Cannot add it to a second collection
[INFO] Status: 7 events processed  7.0 Hz (0.1 Hz avg)
[INFO] Status: 21 events processed  14.0 Hz (0.4 Hz avg)
[INFO] Status: 33 events processed  12.0 Hz (0.7 Hz avg)
[INFO] Status: 45 events processed  12.0 Hz (0.9 Hz avg)
[INFO] Status: 61 events processed  16.0 Hz (1.2 Hz avg)
[INFO] Status: 70 events processed  9.0 Hz (1.3 Hz avg)
[INFO] Status: 80 events processed  10.0 Hz (1.5 Hz avg)
[INFO] Status: 89 events processed  9.0 Hz (1.6 Hz avg)
[INFO] Status: 98 events processed  9.0 Hz (1.7 Hz avg)
[INFO] Status: 100 events processed  8.8 Hz (1.8 Hz avg)
[INFO] All workers have stopped.
[INFO] Merging threads ...
[JEventProcessorPODIO] [error] Failed to get the following collections at least once:
[JEventProcessorPODIO] [error] Failed collection 'CentralTrackSeedingResults'
[INFO] JArrow: Finalized JEventProcessor 'JEventProcessorPODIO'
```

We may be able to go further and not allow any exceptions at all now. I don't think we need that functionality anymore since we now expect algorithms to catch their exceptions and we don't have any unused factories.

- [x] Bug fix (issue #1052)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


nathanwbrei:
Some thoughts:

1. Thanks for moving the log messages to the spdlog. I'm tempted to add the option to redirect all JANA-internal logging to spdlog. (This might be a good task for our new hire, Ayan...) It would be nice to have as many messages as possible formatted the same for the sake of searching and filtering them. 

2. I personally wouldn't mind going even further and terminating execution upon encountering an exception. (I honestly thought you or Dmitry had already done this!) It's not all that possible to catch and recover from an exception when stuff is happening asynchronously and on different threads. Maybe we could add an option to terminate on exception simply for the CI.

wdconinc:
> Maybe we could add an option to terminate on exception simply for the CI.

I was thinking in this direction as well. Essentially I was going to suggest a boolean parameter `allow_failing_collections` inside JEventProcessorPODIO that configures whether to fail on exceptions (well, rethrows the exception, and JANA2 should then fail).

wdconinc:
> I'm tempted to add the option to redirect all JANA-internal logging to spdlog.

I think it would be interesting to develop an interface where someone can provide a JANA2 compatible logger that is then used for all JANA2 logging (and if not provided it falls back to the current default). We already do something like that with Acts::Logger. For JANA2 my question would be when we pass this to JANA2: we can't exactly wait until we have a Log_service.

wdconinc:
Move Nathan's comment 2 to separate issue, https://github.com/eic/EICrecon/issues/1056, to be implemented in different PR.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -281,41 +281,47 @@ void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent> &event) {
     //            that are determined by hash, we have to ensure they are reproducible
     //            even if the collections are filled in unpredictable order (or not at
     //            all). See also below, at "TODO: NWB:".
```
veprbl, line 283: From CI, it doesn't look like any exceptions are triggered by missing collection id's. So we can delete this, perhaps?

src/services/io/podio/JEventProcessorPODIO.h
```diff
@@ -32,4 +32,6 @@ class JEventProcessorPODIO : public JEventProcessor {
     std::vector<std::string> m_collections_to_write;  // derived from above config. parameters
     std::vector<std::string> m_collections_to_print;
 
+  private:
+    std::set<std::string> m_failed_collections;
```
veprbl, line 36: I've submitted a few PRs for HcalEndcapP. With those merged, we could fail immediately instead of just making warnings.

src/services/io/podio/JEventProcessorPODIO.cc
wdconinc: I am certainly open to making it a requirement that factories have to catch exceptions due to missing collections, instead of catching this in the event processor.

veprbl: I meant something like #1069

Issue #796, opened 2023-07-25
LFHCAL special treatment in CalorimeterHitReco


wdconinc:
Generic algorithms only work if there are no lines like
```cpp
        if (m_cfg.readout == "LFHCALHits" && m_cfg.sampFracLayer[0] != 0.){
```
in the algorithm source code. This is fragile because it depends on the detector name (which may change) and it distributes the configuration across different sources files instead of keeping it manifest in a place (configuration file, conditions database) where it can be defined.

Proposal: remove the `LFHCALHits` comparison and generalize layer treatment for any detector that provides a configuration where it is set.

PR #1054, opened 2023-10-05
fix: remove adhoc LFHCAL treatment in CalorimeterHitReco

wdconinc:
This removes adhoc LFHCAL treatment in CalorimeterHitReco and implements a configurable approach that uses previously implemented layer ID resolution. Instead of `if (m_cfg.readout == "LFHCALHits")`, this now:
- uses the cellID decoder `id_dec` defined on init intsead of a local version,
- doesn't depend on fragile testing `sampFracLayer[0] != 0.` but instead `sampFracLayer.empty()`,
- single place to do the energy reconstruction based on the varying sampFrac,
- single place to do the layer field id decoding.

TODO:
- [x] check diffs in CI: LFHCALRecHits.layer is now filled too

- [x] Bug fix (issue #796)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @FriederikeBock 

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitReco.cc
@@ -118,7 +118,6 @@ std::unique_ptr<edm4eic::CalorimeterHitCollection> CalorimeterHitReco::process(c
     // number is encountered disable this algorithm. A useful message
     // indicating what is going on is printed below where the
     // error is detector.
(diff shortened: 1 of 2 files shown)


veprbl:
Causes small numerical differences in the values, but according to the report https://veprbl.github.io/capybara-reports/3975c00465424eaba3e03815ab56705b/, the actual distributions are unchanged

veprbl (approved):

Issue #930, opened 2023-09-07
Track{Projector,Propagation}: investigate why we are unable to access ACTS surface id


veprbl:
https://github.com/eic/EICrecon/blob/e5f0a9df377a9f1f43a1addb42d6bb3712f281ac/src/algorithms/tracking/TrackProjector.cc#L151

https://github.com/eic/EICrecon/blob/e5f0a9df377a9f1f43a1addb42d6bb3712f281ac/src/algorithms/tracking/TrackPropagation.cc#L294

Read access fails under ASAN, write access causes segfault. Needs investigation.

See also #862

wdconinc:
I can't reproduce this locally (clang, clang++, Debug, ASAN + UBSAN). Is this still an issue for you? Maybe the ODR violation or memory leak fixes addressed this?

I'm going to create a branch from this PR to see if maybe CI has trouble with it.

PR #1055, opened 2023-10-05
fix: reenable storing geometryId into TrackPoint surface field

wdconinc:
This reenables the geometryId call to store the surface field of the TrackPoint. It was previously having issues with ASAN.

- [x] Bug fix (issue #930)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackProjector.cc
+++ b/src/algorithms/tracking/TrackProjector.cc
@@ -148,7 +148,7 @@ namespace eicrecon {
                 const float pathLength = static_cast<float>(trackstate.pathLength());
                 const float pathLengthError = 0;
 
-                uint64_t surface = 0; // trackstate.referenceSurface().geometryId().value(); FIXME - ASAN is not happy with this
+                uint64_t surface = trackstate.referenceSurface().geometryId().value();
                 uint32_t system = 0;
 
                 // Store track point
--- a/src/algorithms/tracking/TrackPropagation.cc
+++ b/src/algorithms/tracking/TrackPropagation.cc
@@ -292,7 +292,7 @@ namespace eicrecon {
         m_log->trace("    loc err = {:.4f}", static_cast<float>(covariance(Acts::eBoundLoc0, Acts::eBoundLoc1)));
 
 #if EDM4EIC_VERSION_MAJOR >= 3
-        uint64_t surface = 0; // targetSurf->geometryId().value(); // FIXME - ASAN is not happy with this
+        uint64_t surface = targetSurf->geometryId().value();
         uint32_t system = 0; // default value...will be set in TrackPropagation factory
 #endif
 


veprbl:
This is odd. I thought I saw it previously fail on CI, before commenting this out. And I definitely did see this locally. Let me test.

src/algorithms/tracking/TrackProjector.cc
```diff
@@ -148,7 +148,7 @@ namespace eicrecon {
                 const float pathLength = static_cast<float>(trackstate.pathLength());
                 const float pathLengthError = 0;
 
-                uint64_t surface = 0; // trackstate.referenceSurface().geometryId().value(); FIXME - ASAN is not happy with this
+                uint64_t surface = trackstate.referenceSurface().geometryId().value();
                 uint32_t system = 0;
```
wdconinc, line 152: Per https://github.com/acts-project/acts/blob/020a076bebfea8f9c01a4fb0813f5bcdb495590c/Core/include/Acts/Geometry/GeometryIdentifier.hpp#L61-L62, we should use the `extra` bit fields in `geometryId` to store this subsystem ID... Which means we didn't really need the `system` field...

```suggestion
                uint32_t system = trackstate.referenceSurface().geometryId().extra();
```

This would allow us to stuff the info from https://github.com/eic/EICrecon/blob/main/src/global/tracking/TrackPropagation_factory.cc#L165-L166 into the geometry ID, making that a bit cleaner.

wdconinc: #1057

wdconinc: ... storing the 8-bit DD4hep detector subsystem ID in the 8-bit `extra` field of `geometryId` is not as easy as I thought it would be. We can pass a GeometryIdentifierHook to Acts::convertDD4hepDetector which can write the `extra` field, but it doesn't seem to have access at the point when it is called to the DD4hep geometry objects. So, we only seem to be able to use the other `geometryId` fields to create the `extra` field, or we have to bundle an Acts::Surface-to-detector-subsystem-ID map in the hook.

And if we're going to have to create a map in ActsGeometryProvider, we might as well set the map after geometry creation and use it to set the `TrackPoint::system` field here based on the `volume` field in `geometryId` IFF `extra` is not set.

veprbl: Sure, I was talking about this when `system` field was added.

wdconinc: Sorry, must have missed that. Figured out the GeometryIdentifierHook in #1058.

veprbl: The `GeometryIdentifierHook` I didn't know about at the time. I thought there was no facility to actually support this.

wdconinc: It certainly isn't well advertised. But it seems to have been added exactly for this purpose.

veprbl (approved):
This is no longer showing up as a segfault on my system. No ASAN issue in CI. Also I remember that I saw nothing in ACTS code that would have prevented this from working. So, let's reenable.

src/algorithms/tracking/TrackProjector.cc
veprbl: Yeah, nice find.

PR #1057, opened 2023-10-06
feat: add system and layer ID to Acts::Surface GeometryIdentifier

wdconinc:
The Acts::Surface geometry identifier is intended to store information we are currently passing through two other lists in the TrackPropagation factory. This fills the geometry identifier field with:
- the 'surface' as layer ID,
- the 'detector' as extra ID (which is inteded for subsystem identifiers).

These changes will allow us to get rid of the extra vectors that are passed to the algorithm, which in turn allows us to use the same algorithm for real and constructed surfaces.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/global/tracking/TrackPropagation_factory.cc
+++ b/src/global/tracking/TrackPropagation_factory.cc
@@ -78,12 +78,15 @@ void eicrecon::TrackPropagation_factory::SetPropagationSurfaces() {
     const double BEMC_halfz = (std::max(m_geoSvc->detector()->constant<double>("EcalBarrelBackward_zmax"),
                             m_geoSvc->detector()->constant<double>("EcalBarrelForward_zmax")) / dd4hep::mm) * extend * Acts::UnitConstants::mm;
     auto BEMC_Trf         = transform * Acts::Translation3(Acts::Vector3(0, 0, 0));
-    auto m_BEMC_prop_surface1  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R, BEMC_halfz);
-    auto m_BEMC_prop_surface2  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R + ECAL_avgClusterDepth, BEMC_halfz);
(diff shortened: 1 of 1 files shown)


src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -78,12 +78,15 @@ void eicrecon::TrackPropagation_factory::SetPropagationSurfaces() {
...
-    auto m_BEMC_prop_surface1  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R, BEMC_halfz);
-    auto m_BEMC_prop_surface2  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R + ECAL_avgClusterDepth, BEMC_halfz);
-    m_target_surface_list.push_back(m_BEMC_prop_surface1);
+    auto BEMC_prop_surface1  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R, BEMC_halfz);
+    auto BEMC_prop_surface2  = Acts::Surface::makeShared<Acts::CylinderSurface>(BEMC_Trf, BEMC_R + ECAL_avgClusterDepth, BEMC_halfz);
+    auto BEMC_system_id = m_geoSvc->detector()->constant<uint32_t>("ECalBarrel_ID");
+    BEMC_prop_surface1->assignGeometryId(Acts::GeometryIdentifier().setExtra(BEMC_system_id).setLayer(1));
+    BEMC_prop_surface2->assignGeometryId(Acts::GeometryIdentifier().setExtra(BEMC_system_id).setLayer(2));
+    m_target_surface_list.push_back(BEMC_prop_surface1);
     m_target_detector_ID.push_back(m_geoSvc->detector()->constant<uint32_t>("ECalBarrel_ID"));
```
wdconinc, line 87: Yes, I know this can use `BEMC_system_id` but these lines will just disappear next anyway.

veprbl (approved):
This is in line with my original suggestion about filling information to ACTS Ids.

PR #1062, opened 2023-10-09
Remove factories for HcalEndcapP

veprbl:
That detector was removed from the geometry configurations long time ago.

cc #1053


src/detectors/FHCAL/FHCAL.cc
```diff
@@ -19,196 +17,6 @@ extern "C" {
...
-              .sampFrac = 0.033,
-              .logWeightBase = 6.2,
-              .depthCorrection = 0.0,
-              .enableEtaBounds = false,
-            },
-            app   // TODO: Remove me once fixed
-          )
-        );
-
-        app->Add(new JChainMultifactoryGeneratorT<CalorimeterHitDigi_factoryT>(
```
wdconinc, line 117: Don't we need to keep the insert still?

veprbl: Good catch. We should keep it, actually.

wdconinc (approved):
Looks good to me now. For consistency in software, at some future we may rename the LFHCAL collections back to HcalEndcapP. But for now we should get rid of unused factories.

Issue #696, opened 2023-06-03
Need implementation of SiPM photon counter smearing


veprbl:
**Is your feature request related to a problem? Please describe.**
Current implementation of energy-dependent smearing parametrization
https://github.com/eic/EICrecon/blob/8050426d3ec2c0ee94ad5f60705a1042f5551a35/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L145-L148
requires tuning, and procedures for that are not obvious. Smearing models grounded in our understanding of the hardware can be implemented instead.

**Describe the solution you'd like**
We would like to get an explicit signal smearing behavior expected for MPPC's, e.g. using formalizm described in reference
https://arxiv.org/pdf/1510.01102.pdf
This should account for low photon statistics and saturation effects.

**Describe alternatives you've considered**
Do sophisticated studies outside of the reconstruction framework, and tune tune resolution parameters to reproduce their results.

**Additional context**

ruse-traveler:
Comment for posterity: I've added a flag indicating that this is pertinent to identified TDR wants from the calorimetry side. Namely, this might be relevant for developing HGROC-specific digitization models, which multiple subsystems expressed a desire for...

PR #1064, opened 2023-10-09
CalorimeterHitDigi: add photon counter based readout

veprbl:
This implements modes to add additional signal smearing associated with the scintillation process in calorimeters. The SiPM mode allows to study effect of saturation for a given material, light collection efficiency and type of sensor.

The part with binomial distribution sampling is a SiPM formalism from Jin Huang's work done for sPhenix:
https://github.com/sPHENIX-Collaboration/coresoftware/commit/f746f130c11adfdb70b1978a5d3ebf1b258fe1e8

Resolves: #696


src/detectors/EEMC/EEMC.cc
```diff
@@ -23,6 +23,13 @@ extern "C" {
           {
             .eRes = {0.0 * sqrt(dd4hep::GeV), 0.02, 0.0 * dd4hep::GeV},
             .tRes = 0.0 * dd4hep::ns,
+            .readoutType = CalorimeterHitDigiConfig::kSipmReadout,
+            .lightYield = 300. / dd4hep::MeV,
+            // See simulation study by A. Hoghmrtsyan https://indico.bnl.gov/event/20415/
+            // This includes quantum efficiency of the SiPM
+            .photonDetectionEfficiency = 17. / 300.,
+            // S14160-6015PS, 4 sensors per cell
+            .numEffectiveSipmPixels = 159565 * 4,
```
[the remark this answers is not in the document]
veprbl:
```suggestion
            .numEffectiveSipmPixels = 159565ULL * 4,
```

src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -205,21 +213,50 @@ void CalorimeterHitDigi::process(const CalorimeterHitDigi::Input& input,
...
+      std::poisson_distribution<> n_photons_detected_dist(n_photons_mean);
+      const long long int n_photons_detected = n_photons_detected_dist(m_generator);
+      const long long int n_max_photons =
+          m_cfg.dyRangeADC * m_cfg.lightYield * m_cfg.photonDetectionEfficiency;
+      trace("n_photons_detected {}", n_photons_detected);
+      adc = std::max(std::llround(ped + n_photons_detected * corrMeanScale_value * (1.0 + eResRel) /
+                                            n_max_photons * m_cfg.capADC),
+                     0LL);
+    } else if (readoutType == kSipmReadout) {
+      const long long int n_photons = edep * m_cfg.lightYield;
```
akioogawa, line 241: I guess it depends on what resolution smearing be eResRel represents with this change... But shouldn't SiPM saturation include resolution smearing by (1.0 + eResRel)? If so
  n_photons_mean = edep * (1.0 + eResRel) * m_cfg.lightYield * m_cfg.photonDetectionEfficiency;
then
  adc = std::max(std::llround(ped + n_photons_detected * corrMeanScale_value / n_max_photons * m_cfg.capADC), 0LL);

The same for PoissonPhotonReadout too.

veprbl: So my understanding is that eResRel is used to fudge calorimeter resolutions to expected ones. Then it makes sense to apply it after all non-linearities so that those parameters can be extracted by quadrature substraction of observed resolution from a target resolution. Does that make sense? I'm happy to reorder it the other way, if you prefer. No other detector uses eRelRel (*except B0 which does photon smearing with it, so it won't enable both*).

src/detectors/EEMC/EEMC.cc
```diff
@@ -115,40 +108,25 @@ void InitPlugin(JApplication* app) {
...
-      {.energyWeight = "log", .logWeightBase = 4.6}, app));
-
   app->Add(new JOmniFactoryGeneratorT<CalorimeterClusterRecoCoG_factory>(
-#if EDM4EIC_VERSION_MAJOR >= 8
-      "EcalEndcapNClustersWithoutPIDAndShapes",
-#else
-      "EcalEndcapNClustersWithoutShapes",
-#endif
+      "EcalEndcapNClusters",
       {
```
wdconinc, line 119: We really have to figure out why (and to what extent) these reverts have happened.

wdconinc (approved):

PR #1067, opened 2023-10-10
Implement jet PODIO relations

ruse-traveler:

This PR rewrites the jet reconstruction algorithms and factories in order to accommodate PODIO relations between the jets and their constituents. [Previously,](https://wiki.bnl.gov/EPIC/index.php?title=JetsHF) the constituents' kinematics were stored in the same output collection as the jets and could be identified via their `type` member. Now, only jets are stored in the output collection and their constituents are identified via standard PODIO relations.

Note that in the rewrite, the jet algorithm now processes a generic EDM4EIC or EDM4HEP collection rather than a list of 4-momenta.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #902 )
- [ ] Documentation update
- [ ] Other: __

- [X ] Documentation has been added / updated (the Jet/HF wiki page will be updated on merging)

Yes. Any code making use of the generated and reconstructed jet collections will need to be updated to

1. stop checking the jet `type` for if the object is a jet or constituent;
2. refer to constituents via PODIO relations.

Yes. Only jets will be stored in the output collections.

### DIFF
--- a/src/algorithms/reco/JetReconstruction.cc
+++ b/src/algorithms/reco/JetReconstruction.cc
@@ -46,33 +46,38 @@ namespace eicrecon {
       m_log->error(" Unknown area type \"{}\" specified!", m_cfg.areaType);
       throw JException(out.what());
     }
-  }
+  }  // end 'init(std::shared_ptr<spdlog::logger>)'
 
 
 
-  std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::process(
-    const std::vector<const edm4hep::LorentzVectorE*> momenta) {
+  template <typename T> std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::process(const T& input_collection) {
 
     // Store the jets
     std::unique_ptr<edm4eic::ReconstructedParticleCollection> jet_collection { std::make_unique<edm4eic::ReconstructedParticleCollection>() };
 
+    // extract input momenta and collect into pseudojets
+    std::vector<PseudoJet> particles;
+    for (unsigned iInput = 0; const auto& input : *input_collection) {
+
+      // get 4-vector
+      const auto& momentum = input.getMomentum();
+      const auto& energy = input.getEnergy();
+      const auto* lorentz = new edm4hep::LorentzVectorE(momentum.x, momentum.y, momentum.z, energy);
+
+      // Only cluster particles within the given pt Range
+      if ((lorentz->pt() > m_cfg.minCstPt) && (lorentz->pt() < m_cfg.maxCstPt)) {
+        particles.emplace_back(lorentz->px(), lorentz->py(), lorentz->pz(), lorentz->e());
+        particles.back().set_user_index(iInput);
+      }
+      ++iInput;
+    }
+
     // Skip empty
-    if (momenta.empty()) {
+    if (particles.empty()) {
       m_log->trace("  Empty particle list.");
       return jet_collection;
     }
-
-    m_log->trace("  Number of particles: {}", momenta.size());
-
-    // Particles for jet reconstrution
-    std::vector<PseudoJet> particles;
-    for (const auto &mom : momenta) {
-
-      // Only cluster particles within the given pt Range
-      if ((mom->pt() > m_cfg.minCstPt) && (mom->pt() < m_cfg.maxCstPt)) {
-        particles.emplace_back(mom->px(), mom->py(), mom->pz(), mom->e());
-      }
-    }
+    m_log->trace("  Number of particles: {}", particles.size());
 
     // Choose jet and area definitions
     JetDefinition jet_def(m_mapJetAlgo[m_cfg.jetAlgo], m_cfg.rJet, m_mapRecombScheme[m_cfg.recombScheme]);
@@ -90,36 +95,21 @@ namespace eicrecon {
 
       m_log->trace("  jet {}: pt = {}, y = {}, phi = {}", i, jets[i].pt(), jets[i].rap(), jets[i].phi());
 
-      // Type = 0 for jets, Type = 1 for constituents
-      // Use PDG values to match jets and constituents
+      // create jet to store in output collection
       edm4eic::MutableReconstructedParticle jet_output = jet_collection->create();
-      jet_output.setType(0);
-      jet_output.setPDG(i);
       jet_output.setMomentum(edm4hep::Vector3f(jets[i].px(), jets[i].py(), jets[i].pz()));
       jet_output.setEnergy(jets[i].e());
       jet_output.setMass(jets[i].m());
 
-      // loop over constituents
+      // link constituents to jet kinematic info
       std::vector<PseudoJet> csts = jets[i].constituents();
       for (unsigned j = 0; j < csts.size(); j++) {
-
-        const double cst_pt = csts[j].pt();
-        m_log->trace("    constituent {}'s pt: {}", j, cst_pt);
-
-        // Type = 0 for jets, Type = 1 for constituents
-        // Use PDG values to match jets and constituents
-        edm4eic::MutableReconstructedParticle cst_output = jet_collection->create();
-        cst_output.setType(1);
-        cst_output.setPDG(i);
-        cst_output.setMomentum(edm4hep::Vector3f(csts[j].px(), csts[j].py(), csts[j].pz()));
-        cst_output.setEnergy(csts[j].e());
-        cst_output.setMass(csts[j].m());
-        //jet_output.addToParticles(cst_output);  // FIXME: global issue with podio reference
+        add_to_jet_kinematics(input_collection[csts[j].user_index()], jet_output);
       } // for constituent j
     } // for jet i
 
     // return the jets
     return jet_collection;
-  }
+  }  // end 'process(const T&)'
 
 }  // end namespace eicrecon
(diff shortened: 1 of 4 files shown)


src/algorithms/reco/JetReconstruction.h
```diff
@@ -27,13 +30,47 @@ namespace eicrecon {
...
+      // helper function to copy edm4hep::MCParticle onto edm4eic::ReconstructedParticle
+      // FIXME there is likely a far more elegant way of doing this...
+      edm4eic::ReconstructedParticle copy_mcparticle_onto_recoparticle(const edm4hep::MCParticle& mc) {
+
+        // calculate total energy
+        const float mom_squared = (std::pow(mc.getMomentum().x, 2) + std::pow(mc.getMomentum().y, 2) + std::pow(mc.getMomentum().z, 2));
+        const float energy = std::sqrt(std::pow(mc.getMass(), 2) + mom_squared);
+
+        // create reco particle
+        edm4eic::MutableReconstructedParticle mutable_reco;
```
veprbl, line 63: Does this even work? This particle is not added to any PODIO collection that is in the frame, that should result in an unusable object reference when saved to a file. Not quite sure yet what to advise here.

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -57,24 +57,18 @@ namespace eicrecon {
...
-            // select only final state charged particles
             const bool is_final_state = (particle.getGeneratorStatus() == 1);
-            if (!is_final_state) continue;
-
-            const auto& momentum = particle.getMomentum();
-            const auto& energy = particle.getEnergy();
-            momenta.push_back(new edm4hep::LorentzVectorE(momentum.x, momentum.y, momentum.z, energy));
-        }  // end particle loop
+            if (is_final_state) {
+                for_reconstruction -> push_back(particle);
```
veprbl, line 66:
```suggestion
                for_reconstruction->push_back(particle);
```

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -90,36 +95,21 @@ namespace eicrecon {
...
-        cst_output.setMass(csts[j].m());
-        //jet_output.addToParticles(cst_output);  // FIXME: global issue with podio reference
+        add_to_jet_kinematics(input_collection[csts[j].user_index()], jet_output);
       } // for constituent j
     } // for jet i
 
     // return the jets
     return jet_collection;
-  }
+  }  // end 'process(const T&)'
```
veprbl, line 113: The style creep of jet reco code is not left unnoticed ;)

veprbl (commented):
Very nice. I have one major concern about how RecoParticles derived from MCParticles are to be stored.

src/algorithms/reco/JetReconstruction.cc
ruse-traveler: 😉

src/algorithms/reco/JetReconstruction.h
ruse-traveler: TBH I'm not sure if this actually works yet (it probably doesn't) 😅 I'll be checking that soon.

This probably isn't the right way to go about it, but I couldn't think of another way to add an MCParticle to a ReconstructedParticle...

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -46,34 +46,39 @@ namespace eicrecon {
...
       throw JException(out.what());
     }
-  }
+  }  // end 'init(std::shared_ptr<spdlog::logger>)'
 
 
 
-  std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::process(
-    const std::vector<const edm4hep::LorentzVectorE*> momenta) {
+  template <typename T> std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::process(const T& input_collection) {
```
veprbl, line 53: There is no way for a compiler to figure out which instantiations it needs to emit in this translation unit, so you'll need to add these two lines below:
```c++
template std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::process(const edm4hep::MCParticleCollection& input_collection);
template std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::process(const edm4eic::ReconstructedParticleCollection& input_collection);
```

src/global/reco/GeneratedJets_factory.cc
```diff
@@ -61,24 +61,19 @@ namespace eicrecon {
...
-
-            // select only final state charged particles
             const bool is_final_state = (particle.getGeneratorStatus() == 1);
-            if (!is_final_state) continue;
-
-            const auto& momentum = particle.getMomentum();
-            const auto& energy = particle.getEnergy();
-            momenta.push_back(new edm4hep::LorentzVectorE(momentum.x, momentum.y, momentum.z, energy));
-        }  // end particle loop
+            if (is_final_state) {
```
wdconinc, line 70: Thanks for simplifying the factory for ReconstructedJets. But can we stick this here in the GeneratedJets algorithm too so the GeneratedJets factory is minimalistic again? I was just looking into converting these to JOmniFactory for better external wiring, and that's the main reason we would not be able to do this: the factory is too specific and we want it to be generic.

ruse-traveler: That makes sense! Truth be told, I think it would be very nice to use 1 generic factory for both the `GeneratedJets` and `ReconstructedJets`. Off the top of my head, I can think of 2 ways of doing this:

1. the 1st would be to move the `is_final_state` selection to the algorithm and set it up so that it only checks for this if the input is an `edm4hep::MCParticle` collection;

2. and the 2nd would be to make a "final state" selector (like the `ChargedParticleSelector` factory) that picks out only final state particles and saves them to a collection which would then be the input for the `GeneratedJets`....

In the case of the former, it might make sense to make the desired generator status a user-configurable parameter.

wdconinc: Instead of attaching to the MCParticles collection (of type MCParticleCollection) in GeneratedJets_factory, why not attach to the GeneratedParticles collection (of type ReconstructedParticleCollection) from the MC2SmearedParticle factory? Then both GeneratedJets and ReconstructedJets can both work on ReconstructedParticle collections, instead of having different types.

ruse-traveler: Ah! Using the `GeneratedParticles` collection would also render the issue with picking out only final state particles moot, right? That would let us move to all PODIO relations and vastly simplify the jet factory in one go...

That being said, I do agree with @veprbl that it would be nice to exercise the templated code. But it might be more prudent for us to focus on getting the Jet EDM in and come back to it afterwards...

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -1,13 +1,14 @@
...
+// Copyright (C) 2024 Derek Anderson, Zhongling Ji, Dmitry Kalinkin
 
 // class definition
 #include "JetReconstruction.h"
 
 // for error handling
 #include <JANA/JException.h>
-#include <Math/GenVector/LorentzVector.h>
 #include <edm4hep/Vector3f.h>
+#include <edm4hep/MCParticleCollection.h>
```
wdconinc, line 10:
```suggestion
#include <edm4hep/MCParticleCollection.h>// IWYU pragma: keep
```
Ref: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYUIsDifficult.md#handling-template-arguments

wdconinc:
@ruse-traveler Could you push this as a branch under the main repo, not a fork?

veprbl:
https://veprbl.github.io/capybara-reports/724db6f129d14c0296cfb929010d8112/

veprbl:
No difference in artifacts from my latest commit.

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -92,36 +99,23 @@ namespace eicrecon {
...
-        // Type = 0 for jets, Type = 1 for constituents
-        // Use PDG values to match jets and constituents
-        edm4eic::MutableReconstructedParticle cst_output = jet_collection->create();
-        cst_output.setType(1);
-        cst_output.setPDG(i);
-        cst_output.setMomentum(edm4hep::Vector3f(csts[j].px(), csts[j].py(), csts[j].pz()));
-        cst_output.setEnergy(csts[j].e());
-        cst_output.setMass(csts[j].m());
-        //jet_output.addToParticles(cst_output);  // FIXME: global issue with podio reference
+        jet_output.addToParticles(input_collection->at(csts[j].user_index()));
```
bspage912, line 111: Maybe an issue for another pull request, but do we want to store all constituents in the particles section? This is ok for for GeneratedJets of course, but right now ReconstructedJets constituents are tracks. And in the future, there will be a mix. Not even sure how to specify this from the factory.

bspage912 (approved):
Seems ok (at least I don't have the expertise to spot problems with the "omnification". I had the one question about how we link to the constituents, but this is likely another topic. Do the factories no longer need .cc files?

wdconinc:
> Do the factories no longer need .cc files?

No, if it all fits in a header then that's ok too. Only plugins need a cc file.

src/algorithms/reco/JetReconstruction.cc
veprbl: I think this is understood. We need to get back to discussion of EDM for the next iteration.

veprbl: There is no mix currently. We use MCParticles converted to ReconstructedParticles as there is no onetoone ralation for the prior.

PR #1068, opened 2023-10-10
Pdu surface normal

chchatte92:
Added new methods inside richgeo/IrtGeoDRICH to extract sensor normalZ

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No


src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -2,7 +2,7 @@
 // Subject to the terms in the LICENSE file found in the top-level directory.
 
 #include "IrtGeoDRICH.h"
-
+#include <IRT/CherenkovDetectorCollection.h>
```
c-dilks, line 5: already included in base class
```suggestion
```

c-dilks:
Rebase this:
```bash
git fetch origin
git rebase --onto origin/main e8415e9
```

chchatte92:
> Rebase this:
> 
> ```shell
> git fetch origin
> git rebase --onto origin/main e8415e9
> ```
> 
> then force push

Done!

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -186,7 +185,44 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
+    auto sensorSize      = m_det->constant<double>("DRICH_sensor_size") / dd4hep::mm;
+    for(auto const& [de_name, detSensor] : m_detRich.children()) {
+      if(de_name.find("sensor_de_"+secName)!=std::string::npos) {
+        // get sensor info
+        const auto sensorID      = detSensor.id();
+        const auto detSensorPars = detSensor.extension<dd4hep::rec::VariantParameters>(true);
+        if(detSensorPars==nullptr)
+          throw std::runtime_error(fmt::format("sensor '{}' does not have VariantParameters", de_name));
+        // - sensor surface position
+        auto posSensor = GetVectorFromVariantParameters<dd4hep::Position>(detSensorPars, "pos") / dd4hep::mm;
```
c-dilks, line 207: `posSensor` is not used in this method

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -186,7 +185,44 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
+        // - sensor orientation
+        auto normXdir = GetVectorFromVariantParameters<dd4hep::Direction>(detSensorPars, "normX");
+        auto normYdir = GetVectorFromVariantParameters<dd4hep::Direction>(detSensorPars, "normY");
+        auto normZdir = normXdir.Cross(normYdir); // sensor surface normal
+        // - surface offset, used to convert sensor volume centroid to sensor surface centroid
+        auto surfaceOffset = normZdir.Unit() * (0.5*sensorThickness);
```
c-dilks, line 213: `surfaceOffset` is not used.

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -186,7 +185,44 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
+  auto nSectors= m_det->constant<int>("DRICH_num_sectors");
+  TVector3 normX(1, 0,  0); // normal vectors
+  TVector3 normY(0, -1, 0);
+  TVector3 sensorNorm;
+  auto cellMask = uint64_t(std::stoull(m_det->constant<std::string>("DRICH_cell_mask")));
+  for (int isec = 0; isec < nSectors; isec++) {
+    std::string secName = "sec" + std::to_string(isec);
+    auto sensorThickness = m_det->constant<double>("DRICH_sensor_thickness") / dd4hep::mm;
+    auto sensorSize      = m_det->constant<double>("DRICH_sensor_size") / dd4hep::mm;
+    for(auto const& [de_name, detSensor] : m_detRich.children()) {
```
c-dilks, line 199: Looping over sensors is not really practical, and violates DRY. Please make use of `m_sensor_info`:
https://github.com/eic/EICrecon/blob/46f7d9a7fba79a1c2661897c8785a48cea36c3de/src/services/geometry/richgeo/IrtGeo.h#L62

You can add a new member to `richgeo::Sensor` for the surface norm:
https://github.com/eic/EICrecon/blob/46f7d9a7fba79a1c2661897c8785a48cea36c3de/src/services/geometry/richgeo/RichGeo.h#L20-L27

It gets filled here, where you already have `normZ`:
https://github.com/eic/EICrecon/blob/46f7d9a7fba79a1c2661897c8785a48cea36c3de/src/services/geometry/richgeo/IrtGeoDRICH.cc#L137-L146

chchatte92: So at `GetSensorSurface(CellIDType)`, I need to pass this `id & cellMask` somehow to get the sensor info. Right?

c-dilks: Yes, just use `m_sensor_info` for the lookup.

chchatte92: > Yes, just use `m_sensor_info` for the lookup.

is there a method called lookup? Or any example?

chchatte92: Ah it was stupid question again. m_sensor_info is a map.
so 
auto my_normal  =  m_senso_info->constant<dd4hep::Direction>("surface_normZdir"), if I insert the normZdir as sensor_normZdir inside Sensor class. Right?

c-dilks: The map is 1-to-1 sensor ID to a `richgeo::Sensor` object. Observe the usage of `m_sensor_info` in other places.

`cellID` masked by the cell mask is the sensor ID, if I recall correctly; double check how it is defined in `epic`.

chchatte92: > https://github.com/eic/EICrecon/blob/46f7d9a7fba79a1c2661897c8785a48cea36c3de/src/services/geometry/richgeo/RichGeo.h#L20-L27

@c-dilks what is this `int` here? I understood that the map is between sensor id and sensor info. Is it right?

chchatte92: > `cellID` masked by the cell mask is the sensor ID, if I recall correctly; double check how it is defined in `epic`.

It is like that, I was using it. `sensorID = cellID & cellmask`

chchatte92: So if I am not wrong this `DD4hep_to_IRT` is called independently. I mean I still can call some of the sensor info without creating objects which are defined inside `DD4hep_to_IRT`.

c-dilks: > @c-dilks what is this `int` here? I understood that the map is between sensor id and sensor info. Is it right?

Yes, the `int` is the sensor ID, serving as the map key. The map value is the sensor info, a `richgeo::Sensor` object.

c-dilks: > So if I am not wrong this `DD4hep_to_IRT` is called independently. I mean I still can call some of the sensor info without creating objects which are defined inside `DD4hep_to_IRT`.

It's called in the constructor: https://github.com/eic/EICrecon/blob/a73bc36547a8f5d81266234b5a17d60d8ace9139/src/services/geometry/richgeo/IrtGeoDRICH.h#L13-L14

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -186,7 +184,25 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
+
+  auto sensor_obj = sensor_info_it->second;
+  auto normZdir = sensor_obj.surface_normZdir;
+
+  sensorNorm.SetX(static_cast<double>(normZdir.x()));
+  sensorNorm.SetY(static_cast<double>(normZdir.y()));
+  sensorNorm.SetZ(static_cast<double>(normZdir.z()));
+  //std::cout<<sensorNorm.X()<<std::endl;
+  return sensorNorm;
+}
```
c-dilks, line 205: Remove unused variables and re-definitions, and protect against the case where the sensor cannot be found. This also switches to using `m_sensor_info.surface_offset.Unit()`, as suggested in other comment.
```suggestion
TVector3 richgeo::IrtGeoDRICH::GetSensorSurface(CellIDType id){
  TVector3 sensorNorm;
  auto cellMask = uint64_t(std::stoull(m_det->constant<std::string>("DRICH_cell_mask")));
  auto sID = id & cellMask;
  auto sensor_info_it = m_sensor_info.find(sID);
  if(sensor_info_it != m_sensor_info.end()) {
    auto sensor_obj = sensor_info_it->second;
    auto normZdir = sensor_obj.surface_offset.Unit();
    sensorNorm.SetX(static_cast<double>(normZdir.x()));
    sensorNorm.SetY(static_cast<double>(normZdir.y()));
    sensorNorm.SetZ(static_cast<double>(normZdir.z()));
  }
  else
    m_log->error("Cannot find sensor {} in IrtGeoDRICH::GetSensorSurface", id);

  return sensorNorm;
}
```

src/services/geometry/richgeo/IrtGeoDRICH.h
```diff
@@ -13,7 +13,7 @@ namespace richgeo {
       IrtGeoDRICH(gsl::not_null<const dd4hep::Detector*> det_, gsl::not_null<const dd4hep::rec::CellIDPositionConverter*> conv_, std::shared_ptr<spdlog::logger> log_) :
         IrtGeo("DRICH",det_,conv_,log_) { DD4hep_to_IRT(); }
       ~IrtGeoDRICH();
-
+    TVector3 GetSensorSurface(CellIDType);
```
c-dilks, line 16: Should we go with a clearer name?
```suggestion
    TVector3 GetSensorSurfaceNorm(CellIDType);
```

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -143,8 +141,8 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
         sensor_info.size             = sensorSize;
         sensor_info.surface_centroid = posSensor;
         sensor_info.surface_offset   = surfaceOffset;
+        sensor_info.surface_normZdir = normZdir;
```
c-dilks, line 144: I just realised `surfaceOffset` and `normZdir` are parallel... perhaps we just use `surfaceOffset.Unit()` downstream so we don't have to increase the size of `m_sensor_info` with duplicated information.

src/services/geometry/richgeo/RichGeo.h
```diff
@@ -24,6 +24,7 @@ namespace richgeo {
       double            size;
       dd4hep::Position  surface_centroid;
       dd4hep::Direction surface_offset; // surface centroid = volume centroid + `surface_offset`
+      dd4hep::Direction surface_normZdir; //normal Z direction of surface
```
c-dilks, line 27: Remove if you agree this is not needed (see other comment)
```suggestion
```

c-dilks (changes requested):

chchatte92 (commented):
Changed

chchatte92:
@c-dilks I changed all reused variables and used the sensor_offset.Unit as the direction vector.

src/services/geometry/richgeo/RichGeo.h
chchatte92: Removed and surface_offset is now used.

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -186,7 +183,24 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
   SetReadoutIDToPositionLambda();
 }
-
+TVector3 richgeo::IrtGeoDRICH::GetSensorSurfaceNorm(CellIDType id){
+  TVector3 sensorNorm;
+  auto cellMask = uint64_t(std::stoull(m_det->constant<std::string>("DRICH_cell_mask")));
+  auto sensor_info = this->m_sensor_info;
+  auto sID = id & cellMask;
+  auto sensor_info_it = sensor_info.find(sID);
+  if(sensor_info_it!=sensor_info.end()){
```
c-dilks, line 192:
```suggestion
  auto sID = id & cellMask;
  auto sensor_info_it = m_sensor_info.find(sID);
  if(sensor_info_it!=m_sensor_info.end()){
```

c-dilks:
Other than one more minor comment, this looks good to me!

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -186,7 +183,24 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
+    auto sensor_obj = sensor_info_it->second;
+    auto normZdir = sensor_obj.surface_offset.Unit();
+    sensorNorm.SetX(static_cast<double>(normZdir.x()));
+    sensorNorm.SetY(static_cast<double>(normZdir.y()));
+    sensorNorm.SetZ(static_cast<double>(normZdir.z()));
+  }
+  else
+    m_log->error("Cannot find sensor {} in IrtGeoDRICH::GetSensorSurface", id);
```
wdconinc, line 200: Throw exception instead of just print?

src/services/geometry/richgeo/IrtGeoDRICH.h
```diff
@@ -13,7 +13,7 @@ namespace richgeo {
       IrtGeoDRICH(gsl::not_null<const dd4hep::Detector*> det_, gsl::not_null<const dd4hep::rec::CellIDPositionConverter*> conv_, std::shared_ptr<spdlog::logger> log_) :
         IrtGeo("DRICH",det_,conv_,log_) { DD4hep_to_IRT(); }
       ~IrtGeoDRICH();
-
+    TVector3 GetSensorSurfaceNorm(CellIDType);
```
wdconinc, line 16: Can this be private only?

wdconinc:
You'll need to apply the iwyu patch that causes the failed check. We started requiring that, for now on a pr-by-pr basis while we implement it treewide.

src/services/geometry/richgeo/IrtGeoDRICH.h
c-dilks: No. Alternatively one can move this method's implementation to wherever it is needed; currently the only user is https://github.com/eic/drich-dev/pull/112.

src/services/geometry/richgeo/IrtGeoDRICH.cc
chchatte92: Hi @wdconinc, I can throw an exception. I was imagining the following:
`throw std::runtime_error("sensor not found in IrtGeoDRIC::GetSensorSurfaceNormal");`
But can we also print the `sensorID` that was not found in the exception? I can imagine to keep both lines inside the else condition.

chchatte92:
> You'll need to apply the iwyu patch that causes the failed check. We started requiring that, for now on a pr-by-pr basis while we implement it treewide.

Hi @wdconinc , I have the main branch merged to this branch, git pulled and made sure it compiles in my local branch; but still here the iwyu check fails. What should I do?

src/services/geometry/richgeo/IrtGeoDRICH.h
chchatte92: Shall I resolve this conversation?

chchatte92:
@wdconinc  and @c-dilks, seems like there are no more conflicts and failures. Shall we proceed to merge?

src/services/geometry/richgeo/IrtGeoDRICH.cc
wdconinc: You will have to put the logging line before the throw.

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -209,7 +203,26 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
+    auto normZdir = sensor_obj.surface_offset.Unit();
+    sensorNorm.SetX(static_cast<double>(normZdir.x()));
+    sensorNorm.SetY(static_cast<double>(normZdir.y()));
+    sensorNorm.SetZ(static_cast<double>(normZdir.z()));
+  }
+  else{
+    throw std::runtime_error("sensor not found in IrtGeoDRIC::GetSensorSurfaceNormal");
+    m_log->error("Cannot find sensor {} in IrtGeoDRICH::GetSensorSurface", id);
+  }
+    //std::cout<<sensorNorm.X()<<std::endl;
```
wdconinc, line 223:
```suggestion
```

src/services/geometry/richgeo/IrtGeoDRICH.cc
```diff
@@ -209,7 +203,25 @@ void richgeo::IrtGeoDRICH::DD4hep_to_IRT() {
...
+  if(sensor_info_it!=sensor_info.end()){
+    auto sensor_obj = sensor_info_it->second;
+    auto normZdir = sensor_obj.surface_offset.Unit();
+    sensorNorm.SetX(static_cast<double>(normZdir.x()));
+    sensorNorm.SetY(static_cast<double>(normZdir.y()));
+    sensorNorm.SetZ(static_cast<double>(normZdir.z()));
+  }
+  else{
+    throw std::runtime_error("sensor not found in IrtGeoDRIC::GetSensorSurfaceNormal");
+    m_log->error("Cannot find sensor {} in IrtGeoDRICH::GetSensorSurface", id);
```
wdconinc, line 221:
```suggestion
    m_log->error("Cannot find sensor {} in IrtGeoDRICH::GetSensorSurface", id);
    throw std::runtime_error("sensor not found in IrtGeoDRIC::GetSensorSurfaceNormal");
```

wdconinc (approved):

wdconinc:
@c-dilks Your changes-requested review is the only thing left blocking. Can you review, and approve if ok?

c-dilks (approved):

PR #1069, opened 2023-10-11
JEventProcessorPODIO: treat all unhandled exceptions as fatal

### LINKED ISSUE #625: An exception in factory `Init()` should cause termination with an error exit code

### LINKED ISSUE #1052: Exception emitted in a JChainFactoryT::Process() gets silently absorbed

veprbl:
We are now at a point where quality of the code in factories has been much improved. The exceptions can finally become failures to ensure that we catch the regressions early.

Resolves: #1052
Resolves: #625
Closes: #1053


wdconinc:
Is this something we are closer to being able to merge?

wdconinc:
What's holding this up now?

wdconinc:
@veprbl What's holding up this PR? It's a nice bit of simplification.

nathanwbrei:
Looks like the problem is that in the reconstruction step, the `EcalBarrelScFiClusters` algorithm needs `EcalBarrelScFiHits`, but they haven't been put in the file in the digitization step.

nathanwbrei:
I figure we should change CalorimeterClusterRecoCoG so that the `mchits` input is `std::optional`. Right now the algorithm expects the collection to exist, and simply be empty.

veprbl:
> Looks like the problem is that in the reconstruction step, the `EcalBarrelScFiClusters` algorithm needs `EcalBarrelScFiHits`, but they haven't been put in the file in the digitization step.

I'm confused about how this even works now.

nathanwbrei:
Have we been validating anywhere that this test produces meaningful output?

nathanwbrei:
More specifically, what happens downstream of:
```yaml
    - name: Upload reconstruction output
      uses: actions/upload-artifact@v3
      with:
        name: rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
        path: rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
        if-no-files-found: error
```

wdconinc:
> More specifically, what happens downstream of:

The only thing we do in CI is run a diff with the same file in the main branch. We don't validate performance or require certain performance parameters.

veprbl:
> More specifically, what happens downstream of:
> 
> ```yaml
>     - name: Upload reconstruction output
>       uses: actions/upload-artifact@v3
>       with:
>         name: rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
>         path: rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
>         if-no-files-found: error
> ```

Interestingly enough, what happens next is this artifact gets overwritten by a result of eicrecon-gun job:

```
# gets first step
capybara capy rev HEAD~ --artifact-name raw_e_1GeV_20GeV_craterlake.edm4eic.root
# should get the second step
capybara capy rev HEAD~ --artifact-name rec_e_1GeV_20GeV_craterlake.edm4eic.root
```

veprbl:
And, yes, @nathanwbrei, you were right, the second step "works" by simply producing an empty output collections:
image

veprbl:
I also discovered some older discussion in otherwise invalid PR https://github.com/eic/EICrecon/pull/1071

nathanwbrei:
@veprbl points out that the current failure is probably the same as #1226

.github/workflows/linux-eic-shell.yml
```diff
@@ -399,7 +399,7 @@ jobs:
...
           export JANA_PLUGIN_PATH=$PWD/install/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
-          $PWD/install/bin/eicrecon -Ppodio:output_include_collections=EventHeader,MCParticles,EcalBarrelScFiRawHits,EcalBarrelImagingRawHits -Ppodio:output_file=raw_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root -Pplugins=dump_flags,janadot -Pdump_flags:json=${{ matrix.particle }}_${{ matrix.detector_config }}_flags.json -Pjana:warmup_timeout=0 -Pjana:timeout=0
+          $PWD/install/bin/eicrecon -Ppodio:output_include_collections=EventHeader,MCParticles,EcalBarrelScFiRawHits,EcalBarrelScFiHits,EcalBarrelImagingRawHits,EcalBarrelImagingHits,EcalBarrelScFiHitsContributions,EcalBarrelImagingHitsContributions -Ppodio:output_file=raw_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root -Pplugins=dump_flags,janadot -Pdump_flags:json=${{ matrix.particle }}_${{ matrix.detector_config }}_flags.json -Pjana:warmup_timeout=0 -Pjana:timeout=0
```
veprbl, line 402: It's nice that this works, but this would make it into a non-test for two-stage processing. The `EcalBarrelScFiRawHits` can be reproduced from `EcalBarrelScFiHits`.

nathanwbrei: Agreed, but we can fix the test when we merge the JOmniFactory optional inputs

veprbl: I agree.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -2,14 +2,14 @@
...
 #include <JANA/JApplication.h>
-#include <JANA/JLogger.h>
 #include <JANA/Services/JParameterManager.h>
 #include <JANA/Utils/JTypeInfo.h>
 #include <fmt/core.h>
 #include <podio/CollectionBase.h>
 #include <podio/Frame.h>
 #include <spdlog/common.h>
 #include <exception>
+#include <iosfwd>
```
veprbl, line 12: @wdconinc yet another IWYU loop:
```diff
diff --git a/src/services/io/podio/JEventProcessorPODIO.cc b/src/services/io/podio/JEventProcessorPODIO.cc
index 0a27536..b4ee1c7 100644
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -2,6 +2,7 @@
 #include "JEventProcessorPODIO.h"
 
 #include <JANA/JApplication.h>
+#include <JANA/JLogger.h>
 #include <JANA/Services/JParameterManager.h>
 #include <JANA/Utils/JTypeInfo.h>
 #include <fmt/core.h>
@@ -9,7 +10,6 @@
 #include <podio/Frame.h>
 #include <spdlog/common.h>
 #include <exception>
-#include <iosfwd>
 
 #include "services/log/Log_service.h"
```

wdconinc: Is this a loop? It seems the clang-tidy-iwyu job completed in less than the timeout. I'm not going to claim to understand exactly why it suggests what it suggests all the time, but sometimes it seems merging main commits in a branch causes it to change its mind.

veprbl: Sure, I can assume that the new suggestion is up-to-date.

veprbl:
Let's merge this one. Easier to re-implement than rebase.

PR #1071, opened 2023-10-11
linux-eic-shell.yml: produce reco hits, not just copy raw hits in eicrecon-two-stage-running

veprbl:
Copying over the raw EDM4hep hits is a bit too convenient. The intention was to test the analysis continuation.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -333,7 +333,7 @@ jobs:
           export LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH
           export JANA_PLUGIN_PATH=$PWD/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
           chmod a+x bin/*
-          $PWD/bin/eicrecon -Ppodio:output_include_collections=EventHeader,MCParticles,EcalBarrelScFiRawHits,EcalBarrelImagingRawHits -Ppodio:output_file=raw_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root -Pplugins=dump_flags,janadot -Pdump_flags:json=${{ matrix.particle }}_${{ matrix.detector_config }}_flags.json -Pjana:warmup_timeout=0 -Pjana:timeout=0
+          $PWD/bin/eicrecon -Ppodio:output_include_collections=EventHeader,MCParticles,EcalBarrelScFiHits,EcalBarrelImagingHits -Ppodio:output_file=raw_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root -Pplugins=dump_flags,janadot -Pdump_flags:json=${{ matrix.particle }}_${{ matrix.detector_config }}_flags.json -Pjana:warmup_timeout=0 -Pjana:timeout=0
     - name: Upload digitization output
       uses: actions/upload-artifact@v3
       with:


veprbl:
As expected, this doesn't actually work:
```
Thread model: pthreads
140506247313088: 
       `- JSignalHandler::produce_overall_report[abi:cxx11]() (0x7fca6b9b6e6c)
        `- JSignalHandler::handle_sigsegv(int, siginfo_t*, void*) (0x7fca6b9b7ff6)
         `- /lib/x86_64-linux-gnu/libc.so.6 (0x7fca6905afd0)
          `- eicrecon::CalorimeterHitDigi::process(edm4hep::SimCalorimeterHitCollection const&) (0x7fca51372564)
           `- eicrecon::CalorimeterHitDigi_factoryT::Process(std::shared_ptr<JEvent const> const&) (0x7fca526c293e)
            `- JMultifactoryHelperPodio<edm4hep::RawCalorimeterHit>::Process(std::shared_ptr<JEvent const> const&) (0x7fca525c7d10)
             `- JFactory::Create(std::shared_ptr<JEvent const> const&) (0x7fca6b904cf8)
              `- JFactoryPodioT<edm4hep::RawCalorimeterHit>::Create(std::shared_ptr<JEvent const> const&) (0x7fca527327a5)
               `- JEvent::GetCollectionBase(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) const (0x7fca5b725ee2)
                `- eicrecon::CalorimeterHitReco_factoryT::Process(std::shared_ptr<JEvent const> const&) (0x7fca526c5a7a)
                 `- JMultifactoryHelperPodio<edm4eic::CalorimeterHit>::Process(std::shared_ptr<JEvent const> const&) (0x7fca525c81e0)
                  `- JFactory::Create(std::shared_ptr<JEvent const> const&) (0x7fca6b904cf8)
                   `- JFactoryPodioT<edm4eic::CalorimeterHit>::Create(std::shared_ptr<JEvent const> const&) (0x7fca5272b6d5)
                    `- JEvent::GetCollectionBase(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) const (0x7fca5b725ee2)
                     `- eicrecon::CalorimeterIslandCluster_factoryT::Process(std::shared_ptr<JEvent const> const&) (0x7fca526caac6)
                      `- JMultifactoryHelperPodio<edm4eic::ProtoCluster>::Process(std::shared_ptr<JEvent const> const&) (0x7fca525c86b0)
                       `- JFactory::Create(std::shared_ptr<JEvent const> const&) (0x7fca6b904cf8)
                        `- JFactoryPodioT<edm4eic::ProtoCluster>::Create(std::shared_ptr<JEvent const> const&) (0x7fca527245bf)
                         `- JEvent::GetCollectionBase(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) const (0x7fca5b725ee2)
                          `- eicrecon::CalorimeterClusterRecoCoG_factoryT::Process(std::shared_ptr<JEvent const> const&) (0x7fca526cf5cd)
                           `- JMultifactoryHelperPodio<edm4eic::Cluster>::Process(std::shared_ptr<JEvent const> const&) (0x7fca525c8b80)
                            `- JFactory::Create(std::shared_ptr<JEvent const> const&) (0x7fca6b904cf8)
                             `- JFactoryPodioT<edm4eic::Cluster>::Create(std::shared_ptr<JEvent const> const&) (0x7fca527449f1)
                              `- JEvent::GetCollectionBase(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) const (0x7fca5b725ee2)
                               `- eicrecon::EnergyPositionClusterMerger_factoryT::Process(std::shared_ptr<JEvent const> const&) (0x7fca4f7b3228)
                                `- JMultifactoryHelperPodio<edm4eic::MCRecoClusterParticleAssociation>::Process(std::shared_ptr<JEvent const> const&) (0x7fca525c9050)
                                 `- JFactory::Create(std::shared_ptr<JEvent const> const&) (0x7fca6b904cf8)
                                  `- JFactoryPodioT<edm4eic::MCRecoClusterParticleAssociation>::Create(std::shared_ptr<JEvent const> const&) (0x7fca5273a068)
                                   `- JEvent::GetCollectionBase(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) const (0x7fca5b725ee2)
                                    `- JEventProcessorPODIO::Process(std::shared_ptr<JEvent const> const&) (0x7fca419885ac)
                                     `- JEventProcessor::DoMap(std::shared_ptr<JEvent const> const&) (0x7fca6b93e72d)
                                      `- JEventProcessorArrow::execute(JArrowMetrics&, unsigned long) (0x7fca6b926625)
                                       `- JWorker::loop() (0x7fca6b931437)
                                        `- /lib/x86_64-linux-gnu/libstdc++.so.6 (0x7fca69ad44a3)
                                         `- /lib/x86_64-linux-gnu/libc.so.6 (0x7fca690a8044)
                                          `- /lib/x86_64-linux-gnu/libc.so.6 (0x7fca691285fc)
```

simonge:
> Copying over the raw EDM4hep hits is a bit too convenient. The intention was to test the analysis continuation.

Is it not the other way around. EcalBarrelScFiHits and EcalBarrelImagingHits are the EDM4hep "raw" hits while EcalBarrelScFiRawHits and EcalBarrelImagingRawHits are the post digitization step hits.

veprbl:
@simonge You must be right. I've misinterpreted a message about missing to mean that we were missing a reconstruction step
https://github.com/eic/EICrecon/actions/runs/6488263828/job/17620869065?pr=1069#step:7:518
But it's actually an explicit input for Associations
https://github.com/eic/EICrecon/blob/46f7d9a7fba79a1c2661897c8785a48cea36c3de/src/detectors/BEMC/BEMC.cc#L79-L84
I guess, that should mean that we can't do such test with included Associations, which is something we might have to face later.

kkauder:
@simonge Correct. It would be nice to rename the simple Hits into SimuHits or something like that to avoid confusion. RawHits are the objects that should be treated the same whether coming from digitized simulations or from real data.

simonge:
It's in the digitization factory where the associations aren't explicit and probably should be.
https://github.com/eic/EICrecon/blob/46f7d9a7fba79a1c2661897c8785a48cea36c3de/src/detectors/BEMC/BEMC.cc#L27-L30

Using the contributions:
https://github.com/eic/EICrecon/blob/46f7d9a7fba79a1c2661897c8785a48cea36c3de/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L122

simonge:
There are two issues this has brought to light although not directly critical for this test.
1.  Continued analysis of a collection requiring it's associations needs the associations to also be saved - can/should this be forced? Or might this result in everything being linked to each other anyway and saved whether asked for or not.
2. The association collection doesn't need to exist for the code to try and access it. For this example the error could be caught if the association collection is an input to the relevant calorimetry factories.

wdconinc:
In the days of juggler, this was addressed with flexibility by allowing truth associations to be optional. If they weren't in the input, an algorithm wouldn't produce output truth associations.

veprbl:
> In the days of juggler, this was addressed with flexibility by allowing truth associations to be optional. If they weren't in the input, an algorithm wouldn't produce output truth associations.

Can you elaborate? Was the optionality implemented in python? Was it implemented conditionally on presence of certain input collections?

wdconinc:
Defining an input or output collection in gaudi is by string of the collection name, just like other properties could be strings. When a string was passed for associations, we just got the data handle for that collection; if not, then no collections were used. See e.g. https://eicweb.phy.anl.gov/EIC/juggler/-/blob/v8.0.2/JugReco/src/components/ClusterRecoCoG.cpp?ref_type=tags#L116

Algorithms has a similar interface which allows for optional collections, see e.g. https://github.com/eic/algorithms/blob/main/calorimetry/src/ClusterRecoCoG.cpp#L56 (mind you, the input and output types are pointers and the optional ones can be nullptr).

I think a similar interface should be easily supportable in EICrecon if we allow the factory to catch the right exception, e.g. JException:: CollectionNotFound, and set the pointer to nullptr (and throw for other exceptions as we do now).

wdconinc:
It's not just something that would be nice to have, but it's a requirement that we are able to run without MC associations, as soon as we get test beam that we want to analyze with EICrecon, in fact.

veprbl:
> It's not just something that would be nice to have, but it's a requirement that we are able to run without MC associations, as soon as we get test beam that we want to analyze with EICrecon, in fact.

We are already able to run without MC associations - the factories that throw an exception are automatically excluded from writing (the logic is in the JEventProcessorPODIO).

> Defining an input or output collection in gaudi is by string of the collection name, just like other properties could be strings. When a string was passed for associations, we just got the data handle for that collection; if not, then no collections were used. See e.g. https://eicweb.phy.anl.gov/EIC/juggler/-/blob/v8.0.2/JugReco/src/components/ClusterRecoCoG.cpp?ref_type=tags#L116

So, I can see that a missing collection doesn't disable the collection, but leads to algorithm not populating the DataHandle in `::execute()`. I don't know if in JANA production of a collection is a requirement for successful execution of a factory with the same tag, but, I suspect, those two things may be not tied together (it might be that it returns nullptr from `GetCollectionBase`, like you suggest below). Then the same could be implemented in EICrecon. Then the question is how to handle missing data in the downstream factories? We would need to instrument them to also skip work if associations are not available.

> I think a similar interface should be easily supportable in EICrecon if we allow the factory to catch the right exception, e.g. JException:: CollectionNotFound, and set the pointer to nullptr (and throw for other exceptions as we do now).

Yeah, if we find that using exceptions for control flow is the JANA way, we might restrict the types of those that are used for that purpose.

wdconinc:
Probably a topic for discussion with @nathanwbrei at one of the next Wednesday meetings.

veprbl:
This was invalid, so let's close.

PR #1079, opened 2023-10-23
fix: IWYU pragmas and iwyu.imp fixes

wdconinc:
This adds some fixes to the IWYU treatment which is now enabled.
- IWYU always_keep on SpdlogFormatters.h ([iwyu fails to detect usage](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYUIsDifficult.md#who-is-responsible-for-dependent-template-types)),
- IWYU keep SpdlogFormatters.h while always_keep not available yet,
- IWYU keep known missing std headers in Acts,
- IWYU import fixes to avoid duplicate symbols due to inclusion of ipp AND hpp files,
- IRT include directories fix to avoid short-form include suggestions.

TODO:
- [x] verify that within this branch, after repeating `iwyu_tool | fix_include` until stable, we are left with a compilable source tree, see #1080 

- [x] Bug fix (issue: full-project IWYU results in non-compilable code)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/iwyu.imp
+++ b/.github/iwyu.imp
@@ -223,5 +223,18 @@
   #{ include: ["<stdlib.h>", "private", "<cstdlib>", "public"] },
   #{ include: ["<unistd.h>", "private", "<cunistd>", "public"] },
 
-  #{ include: ["@<(.*)\.ipp>", "private", "<\1.hpp>", "public"] }
+  #{ include: ["@<Acts/(.*)\.ipp>", "private", "<Acts/\1.hpp>", "public"] }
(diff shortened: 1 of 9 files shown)


wdconinc:
The clang-tidy-iwyu check fails here because the additional of the pragmas causes those files to be treated as a changed file, and the entire iwyu treatment is expected to be applied. Since the check is not required, this can still be applied. The full fixes are in #1080.

veprbl (approved):

PR #1085, opened 2023-10-27
feat: depend on algorithms; demo interface in CaloClusterRecoCoG

wdconinc:
This adds support for [algorithms](https://github.com/eic/algorithms) in EICrecon. It depends on getting the updated algorithms@main into eic-shell, to get the addition of `find_package(algorithms)` support. This does not add any 'real' algorithms support but demonstrates the accessibility of the interface in CalorimeterClusterRecoCoG.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,9 @@ add_compile_definitions(HAVE_PODIO)
 # TODO: NWB: Do this correctly in JANA via target_compile_definitions for v2.1.1
 # TODO: NWB: Maybe we want these to be prefixed, e.g. JANA2_HAVE_PODIO
 
+# Algorithms
+find_package(algorithms 1.0.0 REQUIRED Core)
+
 # PODIO, EDM4HEP, EDM4EIC event models
 find_package(Eigen3 REQUIRED)
 find_package(podio REQUIRED)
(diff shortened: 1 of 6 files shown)


wdconinc:
@nathanwbrei FYI, this should be the algorithms interface we want to move towards. I think an adiabatic change in the algorithms is possible, keeping the algorithms in EICrecon for now, and migrating to using algorithms as an interface definition for now. Two unimplemented aspects: 1) defining the properties in the algorithms themselves, and then visiting them doesn't have a clean way to connect with how we set defaults by structs (but could become a string map), 2) not sure yet how we provide the services.

wdconinc:
Zero diffs, e.g. https://github.com/eic/EICrecon/actions/runs/6844069779/job/18607746223?pr=1085#step:11:76.

nathanwbrei (approved):
I think this is another solid step forward towards framework-agnostic algorithms. I don't think anything here will conflict with JOmnifactory in its current form, and I see a path to extend JOmnifactory into a JAlgorithmFactory (hopefully with a better name though!) that can work with any Algorithm.

PR #1088, opened 2023-10-29
calorimetry_CalorimeterIslandCluster: add test case for adjacencyMatrix code path

veprbl:
Distance-based adjacency test is not the only code path. Using adjacencyMatrix is preferred, but was not tested until now. This implements a mock DD4hep detector that just provides the necessary readout encoder/decoder.


wdconinc (approved):
Looks good. I had something more expensive in mind for a mock detector and I'm really glad this can be done with just a few lines!

Issue #895, opened 2023-09-04
Calorimeter reconstruction thresholds don't match the electronics thresholds


veprbl:
| Detector  | Threshold [ADC]       | Gain [keV / ADC]      | Scale factor  | Threshold [keV]       | Spreadsheet [keV]     |
|-|-|-|-|-|-|
| HcalBarrel    | 11.000 [ADC]  | 1.5e+01 [keV / ADC]   | 1.0e+00       | 167.8 [keV]   | 5000 [keV]    |
| EcalEndcapPInsert     | 5.500 [ADC]   | 5.5e+00 [keV / ADC]   | 3.0e-02       | 30.2 [keV]    | ??? [keV]     |
| HcalEndcapPInsert     | -100.000 [ADC]        | 6.1e+00 [keV / ADC]   | 1.0e+00       | -610.4 [keV]  | 250 [keV]     |
| EcalBarrelImaging     | 42.000 [ADC]  | 3.7e-01 [keV / ADC]   | 1.0e+00       | 15.4 [keV]    | 15 [keV]      |
| HcalEndcapP   | 3.800 [ADC]   | 1.5e+01 [keV / ADC]   | 1.0e+00       | 58.0 [keV]    | undef [keV]   |
| EcalEndcapN   | 7.000 [ADC]   | 1.2e+03 [keV / ADC]   | 1.0e+00       | 8544.9 [keV]  | 5000 [keV]    |
| HcalEndcapN   | 2.200 [ADC]   | 3.5e+00 [keV / ADC]   | 1.0e+00       | 7.7 [keV]     | 170 [keV]     |
| EcalBarrelScFi        | 10.815 [ADC]  | 4.6e+01 [keV / ADC]   | 1.0e+00       | 495.1 [keV]   | 500 [keV]     |
| ZDCEcal       | 12.800 [ADC]  | 1.2e+01 [keV / ADC]   | 1.0e+00       | 158.1 [keV]   | ??? [keV]     |
| B0ECal        | 7.000 [ADC]   | 1.2e+03 [keV / ADC]   | 1.0e+00       | 8544.9 [keV]  | 1000 [keV]    |
| LFHCAL        | 3.800 [ADC]   | 1.5e+01 [keV / ADC]   | 1.0e+00       | 58.0 [keV]    | 5000 [keV]    |
[...]

PR #1095, opened 2023-10-31
Bhcal threshold

kkauder:
Updated BHCAL thresholds and digitization parameters from the [digitization table from 10/2](https://brookhavenlab.sharepoint.com/:x:/s/EICPublicSharingDocs/EeICxMJ9vGtPtyq5g4qzcSwBKCqkHDj06KDdSvYQ7pxqrQ?rtime=lD_5kMnT20g).

### Notes:
- This uses a cut on accumulated energy. Instead, you could set the energy cut to 0 and use `.thresholdValue=328` instead (or `.thresholdFactor = 164`). The difference between the two is that in the latter case, an ADC fluctuation can bring the result below/above threshold.
- I believe the 5 MeV threshold corresponds to a 150 MeV real deposition, accounting for sampling fraction. Just double-checking that that's correct.

- [x] Bug fix (issue #895 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: Maintenance

No.

Yes.

### DIFF
--- a/src/detectors/BHCAL/BHCAL.cc
+++ b/src/detectors/BHCAL/BHCAL.cc
@@ -34,7 +34,7 @@ extern "C" {
           {
             .eRes = {},
             .tRes = 0.0 * dd4hep::ns,
-            .threshold = 5.0 * dd4hep::MeV,
+            .threshold = 0.0, // Use ADC cut instead
             .capADC        = HcalBarrel_capADC,
             .capTime = 100, // given in ns, 4 samples in HGCROC
             .dyRangeADC    = HcalBarrel_dyRangeADC,
@@ -54,8 +54,8 @@ extern "C" {
             .pedMeanADC    = HcalBarrel_pedMeanADC,
             .pedSigmaADC   = HcalBarrel_pedSigmaADC, // not used; relying on energy cut
             .resolutionTDC = HcalBarrel_resolutionTDC,
-            .thresholdFactor = 0.0, // not used; relying on energy cut
-            .thresholdValue = 0.0, // not used; relying on energy cut
+            .thresholdFactor = 0.0, // not used; relying on flat ADC cut
+            .thresholdValue = 33, // pedSigmaADC + thresholdValue = half-MIP (333 ADC)
             .sampFrac = 0.033, // average, from sPHENIX simulations
             .readout = "HcalBarrelHits",
             .layerField = "tower",
--- a/src/detectors/BHCAL/BHCAL.cc
+++ b/src/detectors/BHCAL/BHCAL.cc
@@ -14,6 +14,7 @@
 #include "factories/calorimetry/CalorimeterHitReco_factoryT.h"
 #include "factories/calorimetry/CalorimeterIslandCluster_factoryT.h"
 #include "factories/calorimetry/CalorimeterTruthClustering_factoryT.h"
+#include "factories/calorimetry/CalorimeterTruthClustering_factoryT.h"
 
 extern "C" {
     void InitPlugin(JApplication *app) {
@@ -22,18 +23,25 @@ extern "C" {
 
         InitJANAPlugin(app);
 
+        // Make sure digi and reco use the same value
+        decltype(CalorimeterHitDigiConfig::capADC)        HcalBarrel_capADC = 65536; //65536,  16bit ADC
+        decltype(CalorimeterHitDigiConfig::dyRangeADC)    HcalBarrel_dyRangeADC = 1.0 * dd4hep::GeV;
+        decltype(CalorimeterHitDigiConfig::pedMeanADC)    HcalBarrel_pedMeanADC = 300;
+        decltype(CalorimeterHitDigiConfig::pedSigmaADC)   HcalBarrel_pedSigmaADC = 2;
+        decltype(CalorimeterHitDigiConfig::resolutionTDC) HcalBarrel_resolutionTDC = 1 * dd4hep::picosecond;
+
         app->Add(new JChainMultifactoryGeneratorT<CalorimeterHitDigi_factoryT>(
           "HcalBarrelRawHits", {"HcalBarrelHits"}, {"HcalBarrelRawHits"},
           {
             .eRes = {},
             .tRes = 0.0 * dd4hep::ns,
             .threshold = 5.0 * dd4hep::MeV,
-            .capADC = 65536,
+            .capADC        = HcalBarrel_capADC,
             .capTime = 100, // given in ns, 4 samples in HGCROC
-            .dyRangeADC = 1.0 * dd4hep::GeV,
-            .pedMeanADC = 10,
-            .pedSigmaADC = 2.0,
-            .resolutionTDC = 1.0 * dd4hep::picosecond,
+            .dyRangeADC    = HcalBarrel_dyRangeADC,
+            .pedMeanADC    = HcalBarrel_pedMeanADC,
+            .pedSigmaADC   = HcalBarrel_pedSigmaADC,
+            .resolutionTDC = HcalBarrel_resolutionTDC,
             .corrMeanScale = 1.0,
             .readout = "HcalBarrelHits",
           },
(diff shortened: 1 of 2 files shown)


johnlajoie:
I've sked Derek to take a look since he is the one who has worked on the BHCal digitization.  Since we read out individual tiles, the sampling has no meaning in this context (it's not a tower or some such). The 5 MeV is directly on the amount of energy deposited in the tile, which corresponds to about a half-MIP or ~20 SiPM pixels.

kkauder:
> I've asked Derek to take a look since he is the one who has worked on the BHCal digitization. Since we read out individual tiles, the sampling has no meaning in this context (it's not a tower or some such). The 5 MeV is directly on the amount of energy deposited in the tile, which corresponds to about a half-MIP or ~20 SiPM pixels.

What is the sampling fraction used for then?

If I follow the digitization here
https://github.com/eic/EICrecon/blob/main/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L171-L172
I get for a 5 MeV deposition
`adc = round(300 + 5 MeV * (1.0 / 1000 MeV) * 65536) = 628.`
Plugging that into 
https://github.com/eic/EICrecon/blob/main/src/algorithms/calorimetry/CalorimeterHitReco.cc#L171-L172
I get 
`energy = (628 - 300 ) / 65536 * 1000 MeV / (sampfrac=0.033) = 152 MeV`
In other words, the digi class cuts on and converts the energy as deposited, the reco class corrects that by the sampling fraction.

src/detectors/BHCAL/BHCAL.cc
```diff
@@ -42,13 +49,13 @@ extern "C" {
...
-            .dyRangeADC = 1.0 * dd4hep::GeV,
-            .pedMeanADC = 10,
-            .pedSigmaADC = 2.0,
-            .resolutionTDC = 1.0 * dd4hep::picosecond,
-            .thresholdFactor = 5.0,
-            .thresholdValue = 1.0,
+            .capADC        = HcalBarrel_capADC,
+            .dyRangeADC    = HcalBarrel_dyRangeADC,
+            .pedMeanADC    = HcalBarrel_pedMeanADC,
+            .pedSigmaADC   = HcalBarrel_pedSigmaADC, // not used; relying on energy cut
```
ruse-traveler, line 55: Apologies, but I wasn't able to make the background meeting yesterday. Was there a decision to use the energy threshold vs. the ADC threshold?

Since we'll cut on ADC in actual data, I think it would be better to move to relying on only the ADC threshold.

ruse-traveler (commented):
This all looks quite reasonable to me!

Like I mentioned in my comments above, I think it would be preferable to move to relying on only the ADC threshold, and the math you suggest above looks good.

Truth be told, though, I'm having a hard time telling if the changes I see in the artifacts are meaningful. So let me run a quick sanity check to make sure that using the above numbers without the energy threshold produce the same reconstructed hit spectrum.

src/detectors/BHCAL/BHCAL.cc
kkauder: There hasn't been a decision, just to go through the calo ones individually and where necessary/easy default to the energy cut. If you prefer the ADC cut, just give me the okay (and add the number to the spreadsheet), and I will make the change.

ruse-traveler: Ah! Gotcha! In that case, let's plan on going to ADC only. Let me double check that the 2 are indeed equivalent, and then we can make the changes accordingly...

ruse-traveler: Hi Kolja, I finished up running some checks. I've attached a plot comparing the reconstructed hit spectrum for different combinations of energy cuts/adc thresholds (made w/ 100 10 GeV single pi-).

I should have caught this earlier, but we do want to cut on a half-MIP (5 MeV) deposition without any adjustment for sampling fractions (like John noted). So I think I may be misunderstanding what these parameters are...

We've set the pedMeanADC to be 300 ADC, just under 5 MeV. Following the logic in the digi factory, the condition for zero-suppression is:

`amplitude < (threshold + pedMeanADC)`

which means that if we want to cut anything below 300 (333) ADC then we would need to set the threshold parameters to something like

```
thresholdFactor = 0 (1)
thresholdValue = 0 (31)
```

In the plot, the yellow line corresponds to to a zero-suppression of 300 (factor/value = 0) while the green line corresponds to a zero-suppression of 333 (factor = 1, value = 31).

So I was wondering: do `thresholdFactor` and `thresholdValue` correspond to physical quantities? As I understand it, `thresholdFactor` corresponds to an n-sigma cut on our pedestal distribution, but I'm confused about what `thresholdValue` corresponds to...

checkingCuts_withVal31fac1AndVal0Fac0_PR1095 bhcalRecHitsEne d3m11y2023

kkauder:
I put together a hopefully better explanation for yesterday's Simulation meeting, https://indico.bnl.gov/event/21036/
Basically, the important line is 
```
thresholdADC = m_cfg.thresholdFactor * m_cfg.pedSigmaADC + m_cfg.thresholdValue;
```
You can cut EITHER by specifying the number of SigmaADC's OR a fixed threshold value, one of the two should always be 0. So, you can set `thresholdFactor=3;  pedSigmaADC = 31; thresholdValue = 0;` or `thresholdFactor=0;  pedSigmaADC = 0; thresholdValue = 33;` to get to 333. 
It's all a bit redundant, but it's grown over time.

ruse-traveler:
> I put together a hopefully better explanation for yesterday's Simulation meeting, https://indico.bnl.gov/event/21036/ Basically, the important line is

Perfect! Thanks! That helps quite a lot!

> You can cut EITHER by specifying the number of SigmaADC's OR a fixed threshold value, one of the two should always be 0. So, you can set `thresholdFactor=3; pedSigmaADC = 31; thresholdValue = 0;` or `thresholdFactor=0; pedSigmaADC = [whatever is right, only used in digi]; thresholdValue = 33;` to get to 333. It's all a bit redundant, but it's grown over time.
 
Gotcha! In that case, let's go with the latter case (`pedSigmaADC = 300, thresholdFactor = 0, thresholdValue = 33`). Conceptually, that matches most closely with what we've been doing! I'll put the suggestions in the code...

src/detectors/BHCAL/BHCAL.cc
```diff
@@ -42,13 +49,13 @@ extern "C" {
...
+            .resolutionTDC = HcalBarrel_resolutionTDC,
+            .thresholdFactor = 0.0, // not used; relying on energy cut
+            .thresholdValue = 0.0, // not used; relying on energy cut
```
ruse-traveler, line 58: Suggestion:
```
.thresholdValue = 33.0, // pedSigmaADC + thresholdValue = half-MIP (333 ADC)

**Review by ruse-traveler — CHANGES_REQUESTED** [review]

File: src/detectors/BHCAL/BHCAL.cc [hunk 3]
```diff
@@ -22,18 +22,25 @@ extern "C" {
...
+        decltype(CalorimeterHitDigiConfig::pedMeanADC)    HcalBarrel_pedMeanADC = 300;
+        decltype(CalorimeterHitDigiConfig::pedSigmaADC)   HcalBarrel_pedSigmaADC = 2;
+        decltype(CalorimeterHitDigiConfig::resolutionTDC) HcalBarrel_resolutionTDC = 1 * dd4hep::picosecond;
+
         app->Add(new JChainMultifactoryGeneratorT<CalorimeterHitDigi_factoryT>(
           "HcalBarrelRawHits", {"HcalBarrelHits"}, {"HcalBarrelRawHits"},
           {
             .eRes = {},
             .tRes = 0.0 * dd4hep::ns,
             .threshold = 5.0 * dd4hep::MeV,
```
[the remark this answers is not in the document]
kkauder: These values are still needed for digitization. Everything else, I believe I implemented, but please double-check.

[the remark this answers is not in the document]
ruse-traveler: Whoops! I accidentally commented on the wrong lines 😅 

Looks great!

ruse-traveler (approved):

PR #1097, opened 2023-11-02
linux-eic-shell.yml: fix cancel-in-progress

veprbl:
It looks like workflows on main continue to be cancelled: https://github.com/eic/EICrecon/actions/runs/6711622872/job/18239385242

This is yet another attempt at fixing this in the series of ... #968 #998.


wdconinc (commented):
Interesting approach, as far as I understand it.

wdconinc (approved):

PR #1099, opened 2023-11-02
Make sure only one of thresholdFactor & thresholdValue are non-zero

kkauder:
They're (kind of) mutually exclusive.

Do NOT merge this PR until all detectors are in compliance. Or comment out the `throw` to see what is non-compliant. Currently:
LUMISPECCAL:EcalLumiSpecRecHits
https://github.com/eic/EICrecon/issues/1155

Done:
B0ECAL:B0ECalRecHits
EEMC:EcalEndcapNRecHits
FEMC:EcalEndcapPRecHits
FEMC:EcalEndcapPInsertRecHits
BHCAL:HcalBarrelRecHits
EHCAL:HcalEndcapNRecHits
FHCAL:LFHCALRecHits

- [x] Other: Improvement

Yes. Code will break while `.cc` files in src/detector are non-compliant.

Yes. Code will break while `.cc` files in src/detector are non-compliant.


kkauder:
Because of the throw, the checks cannot work until all detectors are in compliance

wdconinc:
@kkauder What is the status of this PR? How many fixes are you still waiting for? Could you update the description at the top when some of them have been fixed?

wdconinc:
@kkauder Now that #1156 is merged, can we wrap up this PR?

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -44,6 +44,14 @@ void CalorimeterHitReco::init(const dd4hep::Detector* detector, const dd4hep::re
     m_log = logger;
 
     // threshold for firing
+    // Should set either m_cfg.thresholdFactor or m_cfg.thresholdValue, not both
+    if ( m_cfg.thresholdFactor > 0 &&  m_cfg.thresholdValue > 0){
+        m_log->error("thresholdFactor = {}, thresholdValue = {}. Only one of these should be non-zero.",
+                    m_cfg.thresholdFactor, m_cfg.thresholdValue);
+        // throw std::exception(fmt::format("CalorimeterHitReco: thresholdFactor = {}, thresholdValue = {}. Only one of these should be non-zero.",
+        //                                   m_cfg.thresholdFactor, m_cfg.thresholdValue));
+        throw;
```
wdconinc, line 53: Can you throw a std::exception with what arg, https://en.cppreference.com/w/cpp/error/exception? e.g.
```suggestion
        throw std::runtime_error("thresholdFactor and thresholdValue should not both be set");
```

wdconinc: I guess I see now that there are some comments to that regard. That commented code should be removed if it isn't going to be activated.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -44,6 +44,14 @@ void CalorimeterHitReco::init(const dd4hep::Detector* detector, const dd4hep::re
     m_log = logger;
 
     // threshold for firing
+    // Should set either m_cfg.thresholdFactor or m_cfg.thresholdValue, not both
+    if ( m_cfg.thresholdFactor > 0 &&  m_cfg.thresholdValue > 0){
```
wdconinc, line 48: Technically, these could be specified as negative (no unsigned float after IEEE 754).
```suggestion
    if ( m_cfg.thresholdFactor != 0 &&  m_cfg.thresholdValue != 0){
```

wdconinc: Another approach here commonly taken is:
```suggestion
    if ( m_cfg.thresholdFactor * m_cfg.thresholdValue != 0){
```
but that may be slightly worse performance-wise (like we care in an init method... but this is deeply trained into me, so I have to mention it).

src/algorithms/calorimetry/CalorimeterHitReco.cc
kkauder: Fair, but I think replacing it with a comment `// throw with an argument doesn't trigger abort` or something like that is appropriate (so that I don't get this question again in the future :) )

src/algorithms/calorimetry/CalorimeterHitReco.cc
kkauder: I'm actually hypersensitive to float `==` or `!=` comparisons, although probably not warranted for 0. I usually resort to something like `fabs(a-b)>1e-9`. Anyway, whichever you prefer is fine.

wdconinc: Yes, excellent instincts, but indeed not warranted for comparisons with default initialized values that can be represented exactly and cheaply in floating point.

wdconinc: If you want to use 1e-9, then at least use the epsilon from numerical_limits.

kkauder: Just an example, but yes, I've gone the extra step before to use numerical_limits. Not worth it here, 1e-6 should be fine (it's late, I confused 6 and 9 :) ) .

wdconinc:
@kkauder Can we wrap up this PR? Two outstanding conversations need resolving.

kkauder:
Should be good to go.

wdconinc (approved):
The issue with "throw with an argument doesn't trigger abort" should get fixed, though.

PR #1104, opened 2023-11-09
Updated ADC parameters and threshold for HCal (and ECal) inserts

kkauder:
Updated backward Insert ADC parameters from the [digitization table from 10/2](https://brookhavenlab.sharepoint.com/:x:/s/EICPublicSharingDocs/EeICxMJ9vGtPtyq5g4qzcSwBKCqkHDj06KDdSvYQ7pxqrQ?rtime=lD_5kMnT20g).

### Notes:
- Threshold cut is applied on the ADC level.
0.25 MeV particle:
adc = 50 + 0.25 / 200 * 32768 = 91

- [x] Bug fix (issue #895 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: Maintenance

No.

Yes.

### DIFF
--- a/src/detectors/FHCAL/FHCAL.cc
+++ b/src/detectors/FHCAL/FHCAL.cc
@@ -6,7 +6,7 @@
 #include <TString.h>
 #include <string>
 
-#include "algorithms/interfaces/WithPodConfig.h"
+#include "algorithms/calorimetry/CalorimeterHitDigiConfig.h"
 #include "extensions/jana/JChainMultifactoryGeneratorT.h"
 #include "factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h"
 #include "factories/calorimetry/CalorimeterHitDigi_factoryT.h"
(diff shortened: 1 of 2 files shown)


kkauder:
@rymilton Ping :) Sorry for the impatience, we'd like to roll this into a bugfix release as soon as possible.

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -22,30 +22,38 @@ extern "C" {
...
-            .pedSigmaADC = 10.,
-            .resolutionTDC = 10 * dd4hep::picosecond,
+            .capADC = HcalEndcapPInsert_capADC,
+            .dyRangeADC = HcalEndcapPInsert_dyRangeADC,
+            .pedMeanADC = HcalEndcapPInsert_pedMeanADC,
+            .pedSigmaADC = HcalEndcapPInsert_pedSigmaADC,
+            .resolutionTDC = HcalEndcapPInsert_resolutionTDC,
             .thresholdFactor = 0.,
-            .thresholdValue = -100.,
+            .thresholdValue = 41.0, // 0.25 MeV --> adc = 50 + 0.25 / 200 * 32768 = 91
```
rymilton, line 55: Could you please explain this line to me? Why do we use 0.25 MeV and 50 in adc =50 + ... ? And the threshold value is just the 0.25 / 200 * 32768 portion, or should it be 91? Sorry for the naive questions -- I'm not that familiar with these numbers.

rymilton (commented):
Overall looks good but I'd like clarification on the threshold value before approving.

src/detectors/FHCAL/FHCAL.cc
kkauder: The 50 is wrong (I looked at the wrong entry in the spreadhseet), but yes, for the threshold it just gets subtracted anyway, I just spelled it out here, because that's the adc calculation. In summary, the 41 remains correct. Let me plug in the other pedestalMean, just to be consistent.

kkauder: @rymilton Done. Thanks for catching the typo (only wrong in the comments, but I simplified that comment to match what you pointed out)

rymilton: Thanks for the explanation! That looks good to me then.

rymilton (approved):

PR #1106, opened 2023-11-11
fix: mv id() to getObjectID().index because id() returns ObjectID, not uint_t, in podio v0.17.1

wdconinc:
This adds forward compatibility with podio v0.17.1. See https://github.com/AIDASoft/podio/pull/493

- [x] Bug fix (issue: compilation falures with podio 0.17.1)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No. The modified algorithms are not relying on the most significant bits in the previous ObjectID uint_t that encode the collectionID.

### DIFF
--- a/src/algorithms/pid/MergeParticleID.cc
+++ b/src/algorithms/pid/MergeParticleID.cc
@@ -6,6 +6,7 @@
 #include <edm4eic/CherenkovParticleIDHypothesis.h>
 #include <edm4eic/TrackSegmentCollection.h>
 #include <fmt/core.h>
+#include <podio/ObjectID.h>
 #include <podio/RelationRange.h>
 #include <spdlog/common.h>
 #include <stddef.h>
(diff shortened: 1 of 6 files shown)


src/algorithms/pid/MergeParticleID.cc
```diff
@@ -94,7 +94,7 @@ std::unique_ptr<edm4eic::CherenkovParticleIDCollection> eicrecon::MergeParticleI
         m_log->error("PID object found with no charged particle");
         continue;
       }
-      auto id_particle = charged_particle_track_segment.id();
+      auto id_particle = charged_particle_track_segment.getObjectId().index;
```
wdconinc, line 97:
```suggestion
      auto id_particle = charged_particle_track_segment.getObjectID().index;
```

veprbl (approved):
For correctness sake, we should be leaning towards full object id comparisons where possible. This looks correct, still.

wdconinc:
> For correctness sake, we should be leaning towards full object id comparisons where possible.

Fortunately that is already the case in several other places. I didn't (need to) replace those cases because the ObjectID comparison works fine. It's only the places where the ObjectID is explicitly assumed to be a uint_t (for storing or debug output) that are modified here.

epic-capybara:
Successfully created backport PR for `v1.7`:
- #1109

PR #1108, opened 2023-11-14
services/io/podio: rely on ::collection_type feature from recent PODIO

veprbl:
This takes advantage of https://github.com/AIDASoft/podio/pull/465 to start deprecating one of the two responsibilities of the `make_datamodel_glue.py`. A similar change to JANA2 will be done independently. I suggest to keep the `#if` pragmas for now, in case we want to bisect possible issues with the latest PODIO.


src/services/io/podio/make_datamodel_glue.py
```diff
@@ -112,10 +114,16 @@ def AddCollections(datamodelName, collectionfiles):
...
     f.write('\n')
     f.write('#include <stdexcept>\n')
+    f.write('#include <podio/podioVersion.h>\n')
     f.write('#include <podio/CollectionBase.h>\n')
     f.write('\n')
 
-    f.write('\ntemplate <typename T> struct PodioTypeMap {};')
+    f.write('\ntemplate <typename T> struct PodioTypeMap {')
+    f.write('\n#if podio_VERSION >= PODIO_VERSION(0, 17, 0)')
+    f.write('\n    using collection_t = typename T::collection_type;')
```
wdconinc, line 123: Is this still needed? Do we still rely on `PodioTypeMap<T>::collection_t` anywhere after all other changes? Is that reliance in Jana2?

veprbl: Yes, in JANA2, for now.

wdconinc (approved):

PR #1110, opened 2023-11-14
feat: install capybara only when not available

wdconinc:
Instead of installing epic-capybara for every pipeline, this only installs when `capybara` is not already in the path.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -609,13 +609,6 @@ jobs:
       with:
         repository: 'eic/epic-capybara'
         path: epic-capybara
-    - name: Build and install epic-capybara
-      if: steps.download_previous_artifact.outputs.found_artifact == 'true'
-      run: |
-        pip install hatch
-        cd epic-capybara
-        hatch build
-        pip install dist/*.whl
     - name: Compare to previous artifacts (local capybara)
       uses: eic/run-cvmfs-osg-eic-shell@main
       if: steps.download_previous_artifact.outputs.found_artifact == 'true'
@@ -623,9 +616,6 @@ jobs:
         platform-release: "${{ env.platform-release }}"
         setup: "/opt/detector/epic-${{ env.detector-version }}/setup.sh"
         run: |
-          echo "::group::pip install"
-          python3 -m pip install -r ${{ github.workspace }}/.github/requirements.txt
-          echo "::endgroup::"
           echo "::group::diff ref -> new"
           python3 ${{ github.workspace }}/epic-capybara/bara.py ref/rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
           echo "::endgroup::"


.github/workflows/linux-eic-shell.yml
```diff
@@ -564,10 +572,7 @@ jobs:
         platform-release: "${{ env.platform-release }}"
         setup: /opt/detector/setup.sh
         run: |
-          echo "::group::pip install"
-          python3 -m pip install -r ${{ github.workspace }}/.github/requirements.txt
-          echo "::endgroup::"
-          python3 ${{ github.workspace }}/epic-capybara/bara.py ref/rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
+          epic-capybara bara ref/rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root rec_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root
```
veprbl, line 575: Unfortunately, bara can not yet replace bara.py, the new CLI argument set is much nicer, but the stdout output is not the same. I would prefer if we run both bara.py script and tool in parallel for a while.

.github/workflows/linux-eic-shell.yml
```diff
@@ -544,14 +544,22 @@ jobs:
...
     - name: Checkout epic-capybara
       uses: actions/checkout@v3
-      if: steps.download_previous_artifact.outputs.found_artifact == 'true'
+      if: steps.download_previous_artifact.outputs.found_artifact == 'true' && steps.check_capybara_available.outputs.found == 'false'
       with:
         repository: 'eic/epic-capybara'
         path: epic-capybara
     - name: Build and install epic-capybara
-      if: steps.download_previous_artifact.outputs.found_artifact == 'true'
+      if: steps.download_previous_artifact.outputs.found_artifact == 'true' && steps.check_capybara_available.outputs.found == 'false'
```
veprbl, line 562: Would it make sense to remove this entirely? In that case, will eic-container deploy nightly capybara automatically for us?

wdconinc: If we want to run the 'old' bara.py, then we need this. We don't install the 'old' versions in the container.

veprbl: The bara.py comes from the checkout above, there is no plan to install it with wheels.

wdconinc: `capybara` is now installed inside the container (by git hash, currently [3a6e6a9c75630fd5e7375a8bd7f9b5c20eb227cd](https://github.com/eic/epic-capybara/commit/3a6e6a9c75630fd5e7375a8bd7f9b5c20eb227cd)), so I'm going to remove this check (make it a requirement instead) and avoid the hatching of the a checked out version.

.github/workflows/linux-eic-shell.yml
wdconinc: Since `capybara` is centrally installed, all runtime dependencies are available. I'll remove the `pip install` of requirements and maintain the previous direct call of `bara.py`.

veprbl (approved):
This is long overdue.

PR #1112, opened 2023-11-15
fix(ci): upload build/ as tar.gz for speedup

wdconinc:
The way GitHub actions upload-artifact uploads each file individually, and stores them in a compressed archive on the server. This means that there are 1500+ files uploaded, each counting as an API call. It's also slow due to the way the protocols work. Compressing on the node, then uploading one file could be net faster.

- [x] Bug fix (issue: we spend more time uploading than compiling)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -98,7 +98,8 @@ jobs:
         run: |
           export LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH
           ctest --test-dir build -V
-    - uses: actions/upload-artifact@v3
+    - name: Upload install directory
+      uses: actions/upload-artifact@v3
       with:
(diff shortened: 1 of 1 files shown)


wdconinc:
Not really faster... so the only benefit is the API calls then...
- Elsewhere: 5m51s https://github.com/eic/EICrecon/actions/runs/6881921712/job/18719313049#step:11:93
- Here: 1m41s + 3m35s https://github.com/eic/EICrecon/actions/runs/6884230885/job/18726309891?pr=1112#step:12:119

Not sure if worth it; @veprbl what say you?

veprbl (approved):
I'd say .tar.gz is a preferable format

wdconinc:
Will apply the same treatment to install/ as well.

wdconinc:
Actually, let's apply the same treatment to install in a new pr. It will need a dedicated CMAKE_INSTALL_PREFIX.

PR #1123, opened 2023-11-18
fix(ci): don't upload unused build directory artifacts

wdconinc:
We spend several minutes just waiting for the build artifacts to get uploaded, and then we don't use them... This only uploads the build directory artifact that is used by the clang-tidy-iwyu job (which is, incidentally, associated with the fasted compilation job, the smallest build directory, and fastest upload). This will speed up when the build matrix completes, and when the next set of jobs can start running.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -216,7 +216,7 @@ jobs:
     - name: Download build artifact
       uses: actions/download-artifact@v3
       with:
-        name: build-clang-eic-shell-Debug
+        name: build-${{ env.clang-tidy-iwyu-CC }}-eic-shell-${{ env.clang-tidy-iwyu-CMAKE_BUILD_TYPE }}
         path: .
     - name: Uncompress build artifact
       run: tar -xaf build.tar.zst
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -116,9 +118,14 @@ jobs:
           !.git/
           !.ccache/
         if-no-files-found: error
+    # Only compress and upload build directory if we are going to use it later
     - name: Compress build directory
+      if: ${{ matrix.CC }} == ${{ env.clang-tidy-iwyu-CC }}
+       && ${{ matrix.CMAKE_BUILD_TYPE }} == ${{ env.clang-tidy-iwyu-CMAKE_BUILD_TYPE }}
```
wdconinc, line 124: Huh, it seems GitHub actions accepts multi-lines without throwing a syntax error, but then it just ignores the second line... Both clang entries in the matrix got their build directory uploaded...

wdconinc: And it uploaded the gcc build directories too. Something's wrong.

wdconinc: Fixed in ae79bc8.

wdconinc:
Ready for review. Does what it's intended to do: not upload useless things; speed up pipelines.

wdconinc:
Looks good, no diffs, https://github.com/eic/EICrecon/actions/runs/7055147653/job/19205768393?pr=1123#step:11:110.

PR #1124, opened 2023-11-18
fix(ci): don't compress build artifact (upload artifact does it anyway)

wdconinc:
In #1112 we gzipped the tar, but the upload-artifact step does the compression anyway (and we can't turn it off), so we should gain some speed by just letting upload-artifact do the compression and letting the tar be faster.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -116,13 +116,13 @@ jobs:
           !.git/
           !.ccache/
         if-no-files-found: error
-    - name: Compress build directory
-      run: tar -czf build.tar.gz build/
+    - name: Archive build directory
+      run: tar -cf build.tar build/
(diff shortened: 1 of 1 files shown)


wdconinc:
### Before:

- Total time: 4m 31s
- Normalized to install upload before: 4m 31s

### After:

- Total time: 4m 14s
- Normalized to install upload before: 5m 5s

So, turns out it's better to do the tar.gz compression locally, and then again on upload. Somewhat strange... 

Or maybe not. The compression is probably only done after the data is on the server and isn't included in the time of the step. So what determines the time is maybe just the size of files to be uploaded.

wdconinc:
Closing in favor of #1152 .

PR #1128, opened 2023-11-20
fix: collect min versions in CMakeLists.txt

wdconinc:
This defines all minimum dependency versions in one place, and uses them in all find_package calls (or, in the case of Acts, immediately after the find_package call).

- [x] Bug fix (issue: duplication of version limits, absence of version limits)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
We do need to start testing both oldest and newest versions in CI, so these versions mean something...

wdconinc:
@veprbl Another old PE revived for review and merging so we get it off the backlog.

veprbl:
This doesn't build. I also don't like the change too much, I think we are better without it. Feel free to fix, if you think it's worthwhile.

cmake/jana_plugin.cmake
```diff
@@ -345,7 +345,7 @@ endmacro()
 macro(plugin_add_event_model _name)
 
   if(NOT podio_FOUND)
-    find_package(podio ${podio_VERSION_MIN})
+    find_package(podio ${podio_VERSION_MIN} QUIET)
     if(NOT podio_FOUND)
       find_package(podio 1.0 REQUIRED)
```
veprbl, line 350: Version from variable?

wdconinc: Could too. I just copied the approach we have in edm4eic for consistency without our software stack.

wdconinc: Though I see the argument that the consistency could also be with what we do for Acts.

veprbl: Oh, nevermind. I get it. "1.0" is not driven by a minimal version, even though the code would need to be updated once `podio_VERSION_MIN` goes up to 1.

.github/workflows/linux-eic-shell.yml
```diff
@@ -133,8 +133,8 @@ jobs:
         platform-release: "${{ env.platform }}:${{ matrix.release }}"
         run: |
           # install this repo
-          CXX="${{ matrix.CXX }}" CXXFLAGS="${{ matrix.CXXFLAGS }}" cmake -B build -S . -DCMAKE_INSTALL_PREFIX=install -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ matrix.CMAKE_BUILD_TYPE }} -DUSE_ASAN=ON -DUSE_TSAN=OFF -DUSE_UBSAN=ON
-          cmake --build build -j $(getconf _NPROCESSORS_ONLN) --target install
+          CXX="${{ matrix.CXX }}" CXXFLAGS="${{ matrix.CXXFLAGS }}" cmake --warn-uninitialized -B build -S . -DCMAKE_INSTALL_PREFIX=install -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ matrix.CMAKE_BUILD_TYPE }} -DUSE_ASAN=ON -DUSE_TSAN=OFF -DUSE_UBSAN=ON
+          cmake  --build build -j $(getconf _NPROCESSORS_ONLN) --target install
```
veprbl, line 137:
```suggestion
          cmake --build build -j $(getconf _NPROCESSORS_ONLN) --target install
```

veprbl (approved):

veprbl (approved):
LGTM

Issue #255, opened 2022-10-20
Integrate `irt` for RICH PID


c-dilks:
**Is your feature request related to a problem? Please describe.**
Integrate the [standalone IRT library](https://github.com/eic/irt); this is to just get us started, where the first major step is adding `irt` as a dependency and showing it works.

**Describe the solution you'd like**
- [x] https://github.com/eic/irt/issues/3
- [x] https://github.com/eic/irt/issues/12
- [x] any [other issues](https://github.com/eic/irt/issues) that would need resolving before introducing `irt` as a dependency here
- [x] initial test: be able to generate the same geometry bindings between IRT and DD4hep that we use in Juggler

**Describe alternatives you've considered**

**Additional context**
This begins porting work from https://eicweb.phy.anl.gov/EIC/juggler/-/merge_requests/377

PR #1130, opened 2023-11-20
fix: link tests with --no-as-needed due to JANA #255

wdconinc:
This fixes an issue in the linking of algorithms_test which popped up today (and likely was always there...). Until https://github.com/JeffersonLab/JANA2/pull/255 is included, we install a libJANA.so with undefined references to podio because it doesn't include it in the NEEDED section. By passing `--no-as-needed` we don't strip the podio calls from the algorithms_test executables where they are unused.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/tests/algorithms_test/CMakeLists.txt
+++ b/src/tests/algorithms_test/CMakeLists.txt
@@ -11,6 +11,8 @@ add_executable(${TEST_NAME}
 
 # Explicit linking to podio::podio is needed due to https://github.com/JeffersonLab/JANA2/issues/151
 target_link_libraries(${TEST_NAME} PRIVATE Catch2::Catch2WithMain algorithms_calorimetry_library algorithms_pid_library podio::podio podio::podioRootIO)
+# As-needed fails due to absent podio symbols in libJANA.so (https://github.com/JeffersonLab/JANA2/pull/255)
+target_link_options(${TEST_NAME} PRIVATE "-Wl,--no-as-needed")
 
 # Install executable
 install(TARGETS ${TEST_NAME} DESTINATION bin)
--- a/src/tests/algorithms_test/CMakeLists.txt
+++ b/src/tests/algorithms_test/CMakeLists.txt
@@ -12,7 +12,9 @@ add_executable(${TEST_NAME}
 # Explicit linking to podio::podio is needed due to https://github.com/JeffersonLab/JANA2/issues/151
 target_link_libraries(${TEST_NAME} PRIVATE Catch2::Catch2WithMain algorithms_calorimetry_library algorithms_pid_library podio::podio podio::podioRootIO)
 # As-needed fails due to absent podio symbols in libJANA.so (https://github.com/JeffersonLab/JANA2/pull/255)
-target_link_options(${TEST_NAME} PRIVATE "-Wl,--no-as-needed")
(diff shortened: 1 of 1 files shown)


wdconinc:
@bleaktwig This should fix the issue you were seeing. Can you verify, and approve if it does? Then we can merge this before too many others run into the same issue.

src/tests/algorithms_test/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ add_executable(${TEST_NAME}
 
 # Explicit linking to podio::podio is needed due to https://github.com/JeffersonLab/JANA2/issues/151
 target_link_libraries(${TEST_NAME} PRIVATE Catch2::Catch2WithMain algorithms_calorimetry_library algorithms_pid_library podio::podio podio::podioRootIO)
+# As-needed fails due to absent podio symbols in libJANA.so (https://github.com/JeffersonLab/JANA2/pull/255)
+target_link_options(${TEST_NAME} PRIVATE "-Wl,--no-as-needed")
```
veprbl, line 15:
```suggestion
if (CMAKE_CXX_COMPILER_LINKER_ID MATCHES "GNU(gold)?")
  target_link_options(${TEST_NAME} PRIVATE "-Wl,--no-as-needed")
endif()
```
(untested)

veprbl (dismissed):
We should be patching JANA2, but this is probably a fine stopgap.

src/tests/algorithms_test/CMakeLists.txt
wdconinc: Added in CMake v3.29... one month ago. https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8885/diffs. Maybe not yet?

veprbl (approved):

bleaktwig:
> @bleaktwig This should fix the issue you were seeing. Can you verify, and approve if it does? Then we can merge this before too many others run into the same issue.

Sorry for being late to the party. This fix is not solving the issue between libJANA and podio on my machine, even after upgrading eic-shell, re-cloning the repository, and building from scratch. It can still be "solved" by leaving `src/tests` out of CMakeLists.txt tho.

The exact error thrown by cmake is:
```
/bin/ld: /usr/local/lib/libJANA.so: undefined reference to `podio::ROOTFrameWriter::ROOTFrameWriter(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
/bin/ld: /usr/local/lib/libJANA.so: undefined reference to `podio::ROOTFrameWriter::finish()'
/bin/ld: /usr/local/lib/libJANA.so: undefined reference to `podio::ROOTFrameWriter::writeFrame(podio::Frame const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
collect2: error: ld returned 1 exit status
```

bleaktwig:
It seems to me that the problem is in [this conditional statement](https://github.com/eic/EICrecon/commit/183aa5202ae40726fe7d1069822b0b8d01f48cbe#diff-30caccb48bc2a5ac56fb0ee97eeb14b91ac37b74cf09fe7cfb6df48ac41bc79eR15). Taking the `target_link_options` out of it fixes the problem. I'm not very familiar with cmake so as to provide a better solution but let me know if you need any other check.

PR #1132, opened 2023-11-21
algorithms_test/CMakeLists.txt: alternative option detection mechanism

veprbl:
- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@
 
 
 
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.18)
 
 project(EICRecon)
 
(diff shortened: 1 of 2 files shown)


veprbl:
I may have tried to use a feature [1] from an unreleased version of cmake.
https://cmake.org/cmake/help/git-master/variable/CMAKE_LANG_COMPILER_LINKER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_LINKER_ID

wdconinc (approved):

wdconinc:
> Capybara summary for PR 1132

I realized that capybara reports lump the gcc and clang test run results together, which is not ideal. Maybe we only need to run clang, not gcc.

PR #1133, opened 2023-11-21
Revert "fix: update CMAKE_CXX_STANDARD default to 20, now required by…

veprbl:
Acts doesn't have this requirement, we don't need it yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,14 +35,10 @@ include(CheckCXXCompilerFlag)
 include(cmake/git_version.cmake)
 set_git_version(CMAKE_PROJECT_VERSION)
 
-# Set default standard to C++20
-set(CMAKE_CXX_STANDARD_MIN 20)
-set(CMAKE_CXX_STANDARD ${CMAKE_CXX_STANDARD_MIN} CACHE STRING "C++ standard to be used")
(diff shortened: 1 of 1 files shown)


wdconinc:
See #1134.

wdconinc:
Propose to close this PR? Requiring C++20 is useful for various reasons in EICrecon: concepts, ranged-for transforms and filters, std::format,...

veprbl:
To me, it looks like most of the features are available in older compilers with C++17. My understanding is that requiring C++20 just declares a requirement for a full compliance, which is not necessary to use the features, but shuts down development with less than latest toolchains.

wdconinc:
I think declaring C++20 doesn't require a compiler with full compliance (case in point: I spent the weekend trying to upgrade clang to v16 because the C++20 it claims to support is incomplete enough to matter for gaudi and juggler). I see it as outlining a set of features that are acceptable to rely on.

If the older compilers have those features in C++17 mode (and when not requiring strict standard adherence), then they also have a flag for C++20 (or at least C++2a) support, which in that case should be used, I think.

veprbl:
My main concern with the switch was not as much about the need to use C++20, but rather the arbitrary timing with absence of a deprecation timeline. I guess, the answer to that is "it works in eic-shell", which is an okay standard, I suppose.

PR #1135, opened 2023-11-21
fix: set C++std to 20, allow higher; agree with Acts and ROOT

wdconinc:
Maybe this is a better approach too: set by default to 17, allow user to set higher, check that C++std agrees with what ROOT and Acts are assuming.

Question: should we automatically upgrade default 17 to 20 if we find that Acts or ROOT requires it?

- [x] Bug fix (issue: how to support C++17 and 20 at the same time)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,8 +35,8 @@ include(CheckCXXCompilerFlag)
 include(cmake/git_version.cmake)
 set_git_version(CMAKE_PROJECT_VERSION)
 
-# Set default standard to C++17
-set(CMAKE_CXX_STANDARD_MIN 17)
+# Set default standard to C++20
+set(CMAKE_CXX_STANDARD_MIN 20)
 set(CMAKE_CXX_STANDARD ${CMAKE_CXX_STANDARD_MIN} CACHE STRING "C++ standard to be used")
 if(CMAKE_CXX_STANDARD LESS CMAKE_CXX_STANDARD_MIN)
   message(FATAL_ERROR "Unsupported C++ standard: ${CMAKE_CXX_STANDARD} (at least ${CMAKE_CXX_STANDARD_MIN} required)")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,7 +152,7 @@ endif()
 # ActsCore
 get_target_property(ACTS_COMPILE_FEATURES ActsCore INTERFACE_COMPILE_FEATURES)
 if (NOT "cxx_std_${CMAKE_CXX_STANDARD}" IN_LIST ACTS_COMPILE_FEATURES)
-  message(FATAL_ERROR "Acts has not been built with the requested C++ standard.")
+  message(WARNING "Acts has not been built with the requested C++ standard ${CMAKE_CXX_STANDARD}.")
 endif()
 
 # ROOT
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -149,9 +149,19 @@ if(${Acts_VERSION} VERSION_LESS ${Acts_VERSION_MIN}
         AND NOT "${Acts_VERSION}" STREQUAL "9.9.9")
     message(FATAL_ERROR "Acts version ${Acts_VERSION_MIN} or higher required, but ${Acts_VERSION} found")
 endif()
+# ActsCore
+get_target_property(ACTS_COMPILE_FEATURES ActsCore INTERFACE_COMPILE_FEATURES)
+if (NOT "cxx_std_${CMAKE_CXX_STANDARD}" IN_LIST ACTS_COMPILE_FEATURES)
+  message(FATAL_ERROR "Acts has not been built with the requested C++ standard.")
```
veprbl, line 155: What is the worst that can happen if we make this a non-fatal error? There are some practical cases when compiling with differing standards may be a fine escape hatch. I see folks are adding such checks to their projects, but don't get what is the actual goal. And I'm not aware of any ABI incompatibility between the recent standards.

wdconinc: There are cases where C++20 support will cause the implementation of different interfaces to be activated, compared to C++17 support. When trying to link against it in the C++17 standard, you might end up with undefined references.

But realistically, I think the worst that can happen is probably just that warnings go ignored for too long...

Maybe a warning is enough?

veprbl: I would keep it a warning. I think, the API inconsistencies, that you mention, if present, will show up loudly during the compilation.

veprbl, line 155:
```suggestion
  message(WARNING "Acts has not been built with the requested C++ standard ${CMAKE_CXX_STANDARD}.")
```

CMakeLists.txt
```diff
@@ -149,9 +149,19 @@ if(${Acts_VERSION} VERSION_LESS ${Acts_VERSION_MIN}
...
+endif()
 
-# CERN ROOT
+# ROOT
 find_package(ROOT REQUIRED)
+# Check that ROOT is compiled with a modern enough c++ standard
+get_target_property(ROOT_COMPILE_FEATURES ROOT::Core INTERFACE_COMPILE_FEATURES)
+if (NOT "cxx_std_${CMAKE_CXX_STANDARD}" IN_LIST ROOT_COMPILE_FEATURES)
+  message(FATAL_ERROR "ROOT has not been built with the requested C++ standard.")
```
veprbl, line 163:
```suggestion
  message(WARNING "ROOT has not been built with the requested C++ standard ${CMAKE_CXX_STANDARD}.")
```

CMakeLists.txt
```diff
@@ -35,8 +35,8 @@ include(CheckCXXCompilerFlag)
 include(cmake/git_version.cmake)
 set_git_version(CMAKE_PROJECT_VERSION)
 
-# Set default standard to C++20
-set(CMAKE_CXX_STANDARD_MIN 20)
+# Set default standard to C++17
+set(CMAKE_CXX_STANDARD_MIN 17)
```
wdconinc, line 39: I'm wondering if we should keep the minimum at C++20. Already before the Acts and ROOT upgrades to C++20, we were using C++20 features in EICrecon.

wdconinc: Removed this change and keeping at C++20.

veprbl (approved):
The changes make sense.

Issue #1137, opened 2023-11-21
Capybara summary in PR clobbers clang and gcc sets


wdconinc:
> Capybara summary for PR 1132

I realized that capybara reports lump the gcc and clang test run results together, which is not ideal. Maybe we only need to run clang, not gcc.

_Originally posted by @wdconinc in https://github.com/eic/EICrecon/issues/1132#issuecomment-1821069684_

PR #1139, opened 2023-11-22
feat: only run clang eicrecon-{dis,gun}, no more gcc

wdconinc:
Instead of clobbering the summaries, we can just run clang only. There is not a lot of value in running both gcc and clang. Would reduce our CI pipelines too.

- [x] Bug fix (issue #1137)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
In my opinion, this is fine. We don't actually expect meaningful runtime differences between gcc and clang.

wdconinc:
I went back and forth on this, but it'd be a lot more invasive to add a `${{ matrix.CC }}` to every artifact name, and then somehow get capybara to work correctly too...

Issue #1126, opened 2023-11-18
Install ASAN suppressions with CMake 


wdconinc:
**Is your feature request related to a problem? Please describe.**
Since we have support for ASAN in CMake, we should also install the suppression files that are needed to run with ASAN. Currently we rely on having the repository and treat ?san.supp as CI-only resources.

**Describe the solution you'd like**
We could install the ?san.supp files into <prefix>/share/EICrecon/ when USE_ASAN is on.

PR #1140, opened 2023-11-22
fix: install {asan,lsan,tsan,ubsan}.supp with CMake; other CMake cleanup

wdconinc:
This installs the {asan,lsan,tsan,ubsan}.supp files into `share/EICrecon` when the installation has support for the respective sanitizers. At this point I'm leaving the *.supp files in .github/ since we don't have an equivalent other directory, and with the current practice of in-tree installations we don't really want to use share/ itself.

It also applies a few minor CMake cleanups:
- use `message(VERBOSE)` instead of an EICRECON flag; one can use `cmake --log-level VERBOSE` to print this,
- include GNUInstallDirs for consistent directory support (not yet propagated to plugins)
- use install DIRECTORY with FILE_PATTERN for cmake/*.cmake files instead of a GLOB

- [x] Bug fix (issue #1126)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/tsan.supp
+++ b/.github/tsan.supp
@@ -1,0 +1,1 @@
+# https://github.com/google/sanitizers/wiki/ThreadSanitizerSuppressions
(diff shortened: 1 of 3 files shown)


veprbl (approved):

veprbl:
LGTM

Issue #1145, opened 2023-11-23
TrackerMeasurementFromHits should use podio collections


wdconinc:
### Is your feature request related to a problem? Please describe.
The TrackerMeasurementFromHits algorithm currently uses podio pointers in std::vector, which bypasses the memory safety features of podio collections.

### Describe the solution you'd like
The TrackerMeasurementFromHits algorithm should use only podio collections.

PR #1146, opened 2023-11-23
fix: pass TrackerHitCollection to TrackerMeasurement, not std::vector<T*>

wdconinc:
Instead of passing a std::vector<T*> to the TrackerMeasurementFromHits algorithm, we should pass a TrackerHitCollection. This PR modifies the algorithm and factory. The factory accepts an array of input collections in a similar way as the TrackerHitCollector does, but that functionality is not currently used (since we use the TrackerHitCollector). Nevertheless, it means we also need to use a subset collection here.

- [x] Bug fix (issue #1145)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackerMeasurementFromHits.cc
+++ b/src/algorithms/tracking/TrackerMeasurementFromHits.cc
@@ -43,33 +43,30 @@ namespace eicrecon {
 }
 
 
-    std::unique_ptr<edm4eic::Measurement2DCollection> TrackerMeasurementFromHits::produce(std::vector<const edm4eic::TrackerHit*> trk_hits) {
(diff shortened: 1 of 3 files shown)


wdconinc:
No diffs in capybara beyond EventHeader.timestamp, e.g. https://github.com/eic/EICrecon/actions/runs/6974170948/job/19000567618?pr=1146#step:9:105.

wdconinc:
No diffs per capybara, e.g. https://github.com/eic/EICrecon/actions/runs/6983925082/job/19006466068#step:9:106. Only timestamp and collection ID due to #1138.

veprbl (approved):
Looks correct

PR #1148, opened 2023-11-24
feat: use algorithms:: interface CalorimeterHitDigi (incl RandomSvc)

wdconinc:
In the ongoing attempts at making the algorithms self-descriptive and conformant to the algorithms:: specification, this converts another one: CalorimeterHitDigi. This wasn't as straightforward as CalorimeterClusterRecoCoG because of the RandomSvc. No attempt is made to connect a JANA2 random service (in part because I don't think JANA2 has one; @nathanwbrei is that still true?). Due to the change of random generator, this is not expected to agree bit-for-bit with what is in main.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Tests for the changes have been added: algorithms_test/calorimetry_CalorimeterHitDigi.cc needed some modifications to deal with the unique pointer memory model.

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -111,13 +111,17 @@ void CalorimeterHitDigi::init(const dd4hep::Detector* detector, std::shared_ptr<
 }
 
 
-std::unique_ptr<edm4hep::RawCalorimeterHitCollection> CalorimeterHitDigi::process(const edm4hep::SimCalorimeterHitCollection &simhits)  {
-    auto rawhits = std::make_unique<edm4hep::RawCalorimeterHitCollection>();
+void CalorimeterHitDigi::process(
+      const CalorimeterHitDigi::Input& input,
+      const CalorimeterHitDigi::Output& output) const {
+
+    const auto [simhits] = input;
+    auto [rawhits] = output;
 
     // find the hits that belong to the same group (for merging)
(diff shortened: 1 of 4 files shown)


wdconinc:
Capybara finds diffs everywhere, but all with p ~ 1. Exceptions are the RawHits amplitudes, e.g. https://github.com/eic/EICrecon/actions/runs/6984604558/job/19007867565?pr=1148#step:9:499. I suspect that the KS test fails on discrete distributions, in particular when we may have an off-by-one binning error. Needs investigation.

src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -20,6 +20,7 @@
 #include <Evaluator/DD4hepUnits.h>
 #include <edm4hep/CaloHitContributionCollection.h>
 #include <fmt/core.h>
+#include <gsl/pointers>
```
veprbl, line 23: Odd that this is needed. Looking at the code, I assume it comes from
https://github.com/eic/algorithms/blob/77e78012a79e294fae39091bdee39ac8092f1472/core/include/algorithms/type_traits.h#L10

veprbl (approved):
LGTM

src/algorithms/calorimetry/CalorimeterHitDigi.cc
wdconinc: In iwyu we trust.

PR #1153, opened 2023-11-25
feat(ci): enable janatop for eicrecon runs

wdconinc:
This enables janatop in EICrecon runs in CI. I considered enabling it by default but there's already enough output in EICrecon. It is useful in CI, even though it requires active inspection to see if there was a change. When adding new algorithms, that should at least be easier now.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Nevermind, this is already active, https://github.com/eic/EICrecon/pull/963.

PR #1154, opened 2023-11-27
feat: use algorithms properties in CalorimeterClusterRecoCoG

wdconinc:
This PR converts CalorimeterClusterRecoCoG to using properties defined as algorithms::Property.

TODO:
- [ ] https://github.com/eic/EICrecon/pull/865#discussion_r1411292184

Cherry-picked d498abe05e2a249dfb553cee7f4256a2628e3d12 from https://github.com/eic/EICrecon/pull/1161.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/extensions/jana/JChainMultifactoryT.h
+++ b/src/extensions/jana/JChainMultifactoryT.h
@@ -25,7 +25,9 @@
 #include <JANA/JMultifactory.h>
 #include "algorithms/interfaces/WithPodConfig.h"
 
-using eicrecon::ConfigMap = std::map<std::string, algorithms::PropertyValue>;
+namespace eicrecon {
+    using ConfigMap = std::map<std::string, algorithms::PropertyValue>;
+}
 
 template <typename ConfigT = eicrecon::NoConfig>
 class JChainMultifactoryT : public JMultifactory {
--- a/src/extensions/jana/JChainMultifactoryT.h
+++ b/src/extensions/jana/JChainMultifactoryT.h
@@ -26,7 +26,7 @@
 #include "algorithms/interfaces/WithPodConfig.h"
 
 namespace eicrecon {
-    using ConfigMap = std::map<std::string, algorithms::PropertyValue>;
+    using ConfigMap = std::map<std::string_view, algorithms::PropertyValue>;
 }
 
 template <typename ConfigT = eicrecon::NoConfig>
--- a/src/extensions/jana/JChainMultifactoryT.h
+++ b/src/extensions/jana/JChainMultifactoryT.h
@@ -19,11 +19,13 @@
 #include <string>
 #include <vector>
 
+#include <algorithms/property.h>
 #include "datamodel_glue.h"
 #include <JANA/JEvent.h>
 #include <JANA/JMultifactory.h>
 #include "algorithms/interfaces/WithPodConfig.h"
 
+using eicrecon::ConfigMap = std::map<std::string, algorithms::PropertyValue>;
 
 template <typename ConfigT = eicrecon::NoConfig>
 class JChainMultifactoryT : public JMultifactory {
--- a/src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
+++ b/src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
@@ -59,15 +59,24 @@ class CalorimeterClusterRecoCoG_factoryT :
         // SpdlogMixin logger initialization, sets m_log
         InitLogger(app, GetPrefix(), "info");
 
-        app->RegisterParameter(param_prefix + ":samplingFraction", m_algo.getProperty<double>("samplingFraction"));
-        app->RegisterParameter(param_prefix + ":logWeightBase", m_algo.getProperty<double>("logWeightBase"));
-        app->RegisterParameter(param_prefix + ":energyWeight", m_algo.getProperty<std::string>("energyWeight"));
-        app->RegisterParameter(param_prefix + ":enableEtaBounds", m_algo.getProperty<bool>("enableEtaBounds"));
-
-        m_algo.setProperty("samplingFraction", app->GetParameterValue<double>(param_prefix + ":samplingFraction"));
-        m_algo.setProperty("logWeightBase", app->GetParameterValue<double>(param_prefix + ":logWeightBase"));
-        m_algo.setProperty("energyWeight", app->GetParameterValue<std::string>(param_prefix + ":energyWeight"));
-        m_algo.setProperty("enableEtaBounds", app->GetParameterValue<bool>(param_prefix + ":enableEtaBounds"));
+        // Initialize properties
+        for (const auto& [key, prop] : m_algo.getProperties()) {
+          std::visit(
+            [app, &m_algo = m_algo, &m_log = m_log,
+             param_prefix, key = key](auto&& val) {
+              using T = std::decay_t<decltype(val)>;
+              if constexpr (std::is_fundamental_v<T>
+                         || std::is_same_v<T, std::string>) {
+                std::string param = param_prefix + ":" + std::string(key);
+                app->RegisterParameter(param, m_algo.getProperty<T>(key));
+                m_algo.setProperty(key, app->GetParameterValue<T>(param));
+              } else {
+                m_log->warn("No support for parsing {} of type {}", key, typeid(val).name());
+              }
+            },
+            prop.get()
+          );
+        }
 
         m_algo.init(detector, logger());
     }
--- a/src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
+++ b/src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
@@ -29,6 +29,11 @@ class CalorimeterClusterRecoCoG_factoryT :
       DeclarePodioOutput<edm4eic::Cluster>(GetOutputTags()[0]);
       DeclarePodioOutput<edm4eic::MCRecoClusterParticleAssociation>(GetOutputTags()[1]);
 
+      // Initialize properties
+      for (const auto& [key, value] : cfg) {
+        m_algo.setProperty(key, value);
+      }
+
     }
 
     //------------------------------------------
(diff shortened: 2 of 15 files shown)


src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -22,13 +22,18 @@ class CalorimeterClusterRecoCoG_factoryT :
...
+    : JChainMultifactoryT<ConfigMap>(tag, input_tags, output_tags, cfg),
       m_algo(tag) {
 
       DeclarePodioOutput<edm4eic::Cluster>(GetOutputTags()[0]);
       DeclarePodioOutput<edm4eic::MCRecoClusterParticleAssociation>(GetOutputTags()[1]);
 
+      // Initialize properties
+      for (const auto& [key, value] : cfg) {
+        m_algo.setProperty(key, value);
+      }
```
wdconinc, line 35: This should be the other way around: loop over defined properties, look up in map, pop them from map. If map is not empty, warn or error about unused properties provided.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -48,15 +53,16 @@ class CalorimeterClusterRecoCoG_factoryT :
...
+        app->RegisterParameter(param_prefix + ":enableEtaBounds", m_algo.getProperty<bool>("enableEtaBounds"));
 
-        app->SetDefaultParameter(param_prefix + ":samplingFraction", cfg.sampFrac);
-        app->SetDefaultParameter(param_prefix + ":logWeightBase", cfg.logWeightBase);
-        app->SetDefaultParameter(param_prefix + ":energyWeight", cfg.energyWeight);
-        app->SetDefaultParameter(param_prefix + ":enableEtaBounds", cfg.enableEtaBounds);
+        m_algo.setProperty("samplingFraction", app->GetParameterValue<double>(param_prefix + ":samplingFraction"));
+        m_algo.setProperty("logWeightBase", app->GetParameterValue<double>(param_prefix + ":logWeightBase"));
+        m_algo.setProperty("energyWeight", app->GetParameterValue<std::string>(param_prefix + ":energyWeight"));
+        m_algo.setProperty("enableEtaBounds", app->GetParameterValue<bool>(param_prefix + ":enableEtaBounds"));
```
wdconinc, line 64: This is obviously ugly. It should also be a loop over defined properties. What prevents it from working is that SetDefaultParameter takes a `T& val` that `m_algo.getProperty<T>` can't satisfy. Or at least I didn't figure out yet how to get it to satisfy.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
wdconinc: The reason is that now we have to handle incorrect keys that can fail setProperty and throw. If we start from the defined keys, we don't have to worry about that; incorrect keys are just not found and left over.

wdconinc: https://github.com/eic/EICrecon/pull/1154/commits/e2cce491bf6a2db5539566963ac51c7440d49d19

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
wdconinc: https://github.com/eic/EICrecon/pull/1154/commits/b14d3dcfa9c24754922b5223e168501a1af057c7

wdconinc: Problem here was that JANA2 doesn't support std::vector<bool> (but instead std::array<T>) and the visit with a variant that includes std::vector<bool> trips up. We if constexpr our way around this now, but warn if we hit this situation.

src/extensions/jana/JChainMultifactoryT.h
```diff
@@ -19,11 +19,15 @@
...
 #include <vector>
 
+#include <algorithms/property.h>
 #include "datamodel_glue.h"
 #include <JANA/JMultifactory.h>
 #include "algorithms/interfaces/WithPodConfig.h"
 #include "extensions/jana/JChainFactoryT.h"  // Just to pull in struct NoConfig
 
+namespace eicrecon {
+    using ConfigMap = std::map<std::string, algorithms::PropertyValue>;
```
wdconinc, line 29: Shoulda used a string_view as these are all read only.
```suggestion
    using ConfigMap = std::map<std::string_view, algorithms::PropertyValue>;
```

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -22,13 +22,24 @@ class CalorimeterClusterRecoCoG_factoryT :
...
+        ConfigMap cfg)
+    : JChainMultifactoryT<ConfigMap>(tag, input_tags, output_tags, cfg),
       m_algo(tag) {
 
       DeclarePodioOutput<edm4eic::Cluster>(GetOutputTags()[0]);
       DeclarePodioOutput<edm4eic::MCRecoClusterParticleAssociation>(GetOutputTags()[1]);
 
+      // Initialize properties
+      for (const auto& [key, prop] : m_algo.getProperties()) {
+        if (cfg.count(std::string(key)) > 0) {
```
wdconinc, line 34: I feel like getting an iterator once would be better than triple lookup we do now.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -48,15 +59,25 @@ class CalorimeterClusterRecoCoG_factoryT :
...
-        auto cfg = GetDefaultConfig();
-
-        app->SetDefaultParameter(param_prefix + ":samplingFraction", cfg.sampFrac);
-        app->SetDefaultParameter(param_prefix + ":logWeightBase", cfg.logWeightBase);
-        app->SetDefaultParameter(param_prefix + ":energyWeight", cfg.energyWeight);
-        app->SetDefaultParameter(param_prefix + ":enableEtaBounds", cfg.enableEtaBounds);
+        // Initialize properties
+        for (const auto& [key, prop] : m_algo.getProperties()) {
+          std::visit(
+            [app, &m_algo = m_algo, &m_log = m_log,
```
wdconinc, line 65: It feels like we should be able to avoid the capture of m_algo here if we pass prop instead of prop.get().

wdconinc:
Cool, no diffs at https://github.com/eic/EICrecon/actions/runs/7015508837/job/19085214064?pr=1154#step:11:104

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
wdconinc: c4c6d86dd15fb4cff91a8d0faa617e1d9e490cf2

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
wdconinc: This isn't possible since visit needs the second argument to be the variant, not something that contains a variant...

wdconinc:
This is ready for review and merging.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -48,15 +60,25 @@ class CalorimeterClusterRecoCoG_factoryT :
...
+             param_prefix, key = key](auto&& val) {
```
veprbl, line 67: I might be wrong, but this could be just
```suggestion
            [app, &m_algo, &m_log, param_prefix, key](auto& val) {
```

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
```diff
@@ -48,15 +60,25 @@ class CalorimeterClusterRecoCoG_factoryT :
...
+              using T = std::decay_t<decltype(val)>;
+              if constexpr (std::is_fundamental_v<T>
+                         || std::is_same_v<T, std::string>) {
+                std::string param = param_prefix + ":" + std::string(key);
+                app->RegisterParameter(param, m_algo.getProperty<T>(key));
```
veprbl, line 72:
```suggestion
                app->RegisterParameter(param, val);
```

veprbl (dismissed):
This worked out nice. Having a map of properties helped a bunch.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h
wdconinc: "Class members cannot be captured explicitly by a capture without initializer (as mentioned above, only variables are permitted in the capture list)" (somewhere in the long page https://en.cppreference.com/w/cpp/language/lambda)

We could capture `this`, but I wanted to avoid that (more scope than needed). It would look like this:
```suggestion
            [app, this, param_prefix, key = key](auto&& val) {
```

wdconinc: `key` is not a variable either; it's a structured binding.

wdconinc: I looked up the structured bindings in captures rules, and it turns out that's allowed in C++20 now. About 2/3 down the page at https://en.cppreference.com/w/cpp/language/structured_binding.

veprbl:
The full error message is 4 gigabytes, but can be summarized as:
```
In file included from /home/runner/work/EICrecon/EICrecon/src/detectors/EEMC/EEMC.cc:13:
/home/runner/work/EICrecon/EICrecon/src/factories/calorimetry/CalorimeterClusterRecoCoG_factoryT.h:66:67: error: 'key' in capture list does not name a variable
            [app, &m_algo = m_algo, &m_log = m_log, param_prefix, key](auto&& val) {
                                                                  ^
```

veprbl:
There are new LSAN errors in CI.

wdconinc:
> There are new LSAN errors in CI.

New podio probably means the old suppressions don't work anymore...

wdconinc:
> There are new LSAN errors in CI.

Actually, this is more curious (and tl;dr no memory leaks). What happened is that within a pipeline one job checked the geometry version in eic-shell and then decided it didn't need to regenerate the simulation file (cache hit). Then the reconstruction job started in the next job, in its own eic-shell, and it had a newer geometry version due to the cvmfs sync, so it continued to run with the older simulation geometry input file but against newer geometry in memory.

The geometry difference was https://github.com/eic/epic/pull/603/files -> i.e. r,phi segmentation to x,y segmentation. That probably caused many warnings about missing cellIDs (including MaxCellIDerrors failures), and memory errors are probably associated with that. Probably ok to let this slide...

Last checks indicate no memory issues anymore: https://github.com/eic/EICrecon/actions/runs/7038493136/job/19156610394?pr=1154#step:8:2091

wdconinc:
And no diffs, e.g. https://github.com/eic/EICrecon/actions/runs/7050567792/job/19192283374?pr=1154#step:11:106. After https://github.com/eic/EICrecon/pull/1154/commits/9ee8a85d1f2b6760ac063dae2423f60efd2e15d5 we should get a clean clang-tidy-iwyu check too. And the eicmkplugin fix is in the merge queue.

So, good for review and merge, as far as I'm concerned.

wdconinc:
Deferred to #1165

Issue #1163, opened 2023-11-30
Regression: EICreconConfig.cmake not installed anymore


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? eic-shell

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. eicmkplugin.py CI is failing to find the EICreconConfig.cmake, https://github.com/eic/EICrecon/actions/runs/7040539144/job/19162384007

### Expected Result: (what do you expect when you execute the steps above)
1. eicmkplugin.py-generated CMakeLists should find EICreconConfig.cmake.
2. CI should check against installed prefix, not against system install of EICreconConfig.cmake

PR #1164, opened 2023-11-30
fix(ci): test eicmkplugin with EICrecon_ROOT=install/

wdconinc:
In CI, test the eicmkplugin with the local build, not system install. 

(Alternative is to modify eicmkplugin to write the install prefix into the CMakeList as a HINT, but that requires adding too much to eicmkplugin. It is also not a requirement so it would still fall back to system installs.)

- [x] Bug fix (issue #1163)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl:
I'm wondering how `/usr/local/lib/cmake` is being considered. Is it in the `CMAKE_PREFIX_PATH` environment variable? Or should we explore one of the options like https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH.html#variable:CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH ?

wdconinc:
This works now and is ready for review.

veprbl (approved):
Diff LGTM, the path in a warning now points to the correct EICrecon install path.

PR #1165, opened 2023-12-01
feat: use algorithms properties in CalorimeterClusterRecoCoG (2)

wdconinc:
This PR converts CalorimeterClusterRecoCoG to using properties defined as algorithms::Property. This is the second attempt at #1154, now based on JOmniFactory and ParameterRef.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -18,11 +18,18 @@
 #include "extensions/spdlog/SpdlogExtensions.h"
 #include "services/log/Log_service.h"
 
+#include <ranges>
 #include <string>
+#include <variant>
 #include <vector>
 
 using namespace eicrecon;
 struct EmptyConfig {};
+
+template<typename AlgoT, typename ConfigT>
+concept UsesAlgorithmsProperties = requires(AlgoT algo, ConfigT config) {
+    { algo.getProperties().at("key").get() } -> std::same_as<typename ConfigT::mapped_type>;
+};
 
 template <typename AlgoT, typename ConfigT=EmptyConfig>
 class JOmniFactory : public JMultifactory {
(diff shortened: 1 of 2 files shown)


src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
```diff
@@ -24,19 +27,41 @@ class CalorimeterClusterRecoCoG_factory : public JOmniFactory<CalorimeterCluster
...
     // Resource<DD4hep_service, const dd4hep::Detector*> m_detector {this, [](std::shared_ptr<DD4hep_service> s, int64_t run_nr){ return s->detector(); }}
 
 public:
     void Configure() {
-        m_algo = std::make_unique<AlgoT>(GetPrefix());
-        m_algo->applyConfig(config());
-        m_algo->init(m_geoSvc().detector(), logger());
+      m_algo = std::make_unique<AlgoT>(GetPrefix());
+
+      for (const auto& [key, prop] : m_algo->getProperties()) {
```
wdconinc, line 39: We need m_algo to exist before we can register the upstream parameters. We need to have ownership of the parameters in this factory since JOmniFactory does not clear the pointers in its m_parameters. We can probably do something like RegisterAlgorithmParameters(m_algo, m_parameters), but we'd still need to declare the vector<unique_ptr> as part of this algorithm. A mixin class may be another option.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
```diff
@@ -24,19 +27,41 @@ class CalorimeterClusterRecoCoG_factory : public JOmniFactory<CalorimeterCluster
...
+        std::visit(
+          [this, key = key](auto&& val) {
+            using T = std::decay_t<decltype(val)>;
+            if constexpr (std::is_fundamental_v<T>
+                       || std::is_same_v<T, std::string>) {
```
wdconinc, line 44: Vector is now also ok, I think.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
wdconinc: Moved this into JOmniFactory in [0211205](https://github.com/eic/EICrecon/pull/1165/commits/0211205b1a5caff56968fe543e503ca6686b850b)

wdconinc:
It is curious that this introduces the kind of diffs (e.g. https://github.com/eic/EICrecon/actions/runs/7072810354/job/19252231634#step:15:82) that are more typically assocated with double/float conversions. So either before or after there may be a conversion that shouldn't happen.

src/extensions/jana/JOmniFactory.h
```diff
@@ -259,6 +267,39 @@ class JOmniFactory : public JMultifactory {
...
     }
 
+    template <typename AlgorithmsT>
+        requires UsesAlgorithmsProperties<AlgorithmsT, ConfigT>
+    void RegisterAllParameters(std::unique_ptr<AlgorithmsT>& algo, std::vector<std::unique_ptr<ParameterBase>>& parameters) {
+        for (const auto& [key, prop] : algo->getProperties()) {
+            std::visit(
+                [this, key = key, &parameters](auto&& val) {
+                    using T = std::decay_t<decltype(val)>;
+                    if constexpr (std::is_fundamental_v<T> || std::is_same_v<T, std::string>) {
```
wdconinc, line 277: This is incomplete because JParameterManager can parse the algorithms::Property of type `T = std::vector<U>` where `U` is a fundamental type or std::string. However, we can't use
```suggestion
                    if constexpr (std::is_fundamental_v<T> || std::is_same_v<T, std::string>
                        || std::is_same_v<std::vector<typename T::value_type, T> && std::is_fundamental_v<typename T::value_type>) {
```
because it tries to apply this to the fundamental types too, and fails due to `T::value_type` not existing for e.g. `T = bool`. I'm not sure yet how to solve this.

wdconinc: @nathanwbrei @veprbl Maybe you have some insights on how to structure a type test here.

wdconinc:
> So either before or after there may be a conversion that shouldn't happen.

Actually, it's worse. Somewhere I am passing a reference to a copy of the property value in the algorithm, so the algorithm still operates with the original default... Back to the drawing board.

PR #1168, opened 2023-12-04
feat: convert factories from JChainMultifactory to JOmniFactory

wdconinc:
This converts several factories from JChainMultifactory to JOmniFactory. Instead of listing which ones are converted, it is easier to list which ones aren't and why, to keep as a check list for work to do:
- [ ] factories that produce subset collections cannot be specified as such using the PodioOutput syntax in JOmniFactory,
  - [x] TrackerMeasurementFromHits (I cheated and removed the subset collection since anyway unused) 
  - [ ] TrackerHitCollector
  - [ ] ChargedParticleSelector
- [ ] factories that don't produce podio collections require the use of the Output syntax in JOmniFactory,
  - [ ] CKFTracking writes ActsExamples::Trajectories and ActsExamples::ConstTrackContainer
  - [ ] RichTrack reads ActsExamples::Trajectories and ActsExamples::ConstTrackContainer
  - [ ] TrackPropagation reads ActsExamples::Trajectories and ActsExamples::ConstTrackContainer
  - [ ] TrackProjector reads ActsExamples::Trajectories and ActsExamples::ConstTrackContainer
  - [ ] IterativeVertexFinder reads ActsExamples::Trajectories and ActsExamples::ConstTrackContainer
- [ ] factories that try to be too clever and do stuff that should be in the algorithm,
  - [ ] GeneratedJets passes a vector of Lorentz vectors to the algorithm
  - [ ] ReconstructedJets passes a vector of Lorentz vectors to the algorithm
  - [ ] PhotoMultiplierHitDigi works differently depending on the name of the plugin
  - [ ] frankly pretty much all the RICH stuff
- [ ] 'variadic' factories that accept arbitrary vectors of input tags and collate them before passing to the algorithm,
  - [ ] MatchClusters
  - [ ] MergeTrack
  - [ ] ReconstructedElectrons
  - [ ] TrackerHitCollector

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


nathanwbrei:
That's a thorough accounting! What's the blockage regarding "factories that don't produce podio collections require the use of the Output syntax in JOmniFactory"? Does the `Output` syntax have a bug?

wdconinc:
> Does the `Output` syntax have a bug?

Nothing major, but enough friction to have me focus on the podio collections first. In particular, it is the `this->tag_name` access in `Output` that's not resolving. It seems to be due to a residual `tag_name` instead of `collection_name`. I'll file a separate PR for that.

wdconinc:
Local tests have this agreeing without diffs with main right now :-) Maybe we should just go ahead and merge it...

src/global/reco/InclusiveKinematicsDA_factory.h
```diff
@@ -13,36 +13,38 @@
...
-    protected:
-        InclusiveKinematicsDA m_inclusive_kinematics_algo;
-
-    };
+class InclusiveKinematicsDA_factory :
+        public JOmniFactory<InclusiveKinematicsDA_factory> {
+
+private:
+    using AlgoT = eicrecon::InclusiveKinematicsDA;
+    std::unique_ptr<AlgoT> m_algo;
```
wdconinc, line 26: @nathanwbrei The reason for this unique_ptr pattern is to allow instantiation of the algorithm with name set to `GetPrefix()` as required by the eic/algorithms interface. That interface allows setting name only once, on in instantiation.

veprbl (approved):
* capybara report shows no differences
* diff LGTM
* `capy pr 1168 --artifact-name pi_craterlake_flags.json | tail -n2 | xargs diff -u` shows only minor changes (ouuput/input tags now show correct defaults)

PR #1171, opened 2023-12-04
fix: Output uses collection_name; Input uses GetCollection in JOmniFactory.h

wdconinc:
Small fix to be able to use `Output` in JOmniFactory.

- [x] Bug fix (issue: https://github.com/eic/EICrecon/pull/1168#issuecomment-1839508786)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -121,11 +121,11 @@ public:
         friend class JOmniFactory;
 
         void CreateHelperFactory(JOmniFactory& fac) override {
-            fac.DeclareOutput<T>(this->tag_name);
+            fac.DeclareOutput<T>(this->collection_name);
         }
 
         void SetCollection(JOmniFactory& fac) override {
-            fac.SetData<T>(this->tag_name, this->data);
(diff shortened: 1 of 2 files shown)


nathanwbrei (dismissed):

src/extensions/jana/JOmniFactory.h
```diff
@@ -55,7 +55,7 @@ class JOmniFactory : public JMultifactory {
     private:
         friend class JOmniFactory;
 
-        void GetData(const JEvent& event) {
+        void GetCollection(const JEvent& event) {
```
nathanwbrei, line 58: Ah, the joys of heavily templated code...

wdconinc: True. It used to be that if your code compiled when you included a header, you could be pretty sure there were no unresolved symbols in there...

wdconinc:
Added Output test to unit test. Started with `std::string` but the typename is not going to be stable over time and depends on spacing in the template type... so it's not an `Output<edm4hep::SimCalorimeterHit>`, which is more predictable.

nathanwbrei (approved):

PR #1175, opened 2023-12-05
feat: use unique_ptr in MatchClusters; require all collections in reco.cc

wdconinc:
MatchClusters is trying to be clever in the factory, while at the same time assuming hardcoded collection names. This changes the MatchClusters factory to require MCParticles, ReconstructedChargedParticles, and ReconstructedChargedParticleAssociations as first collections, then a sequence of XClusters, XClusterAssociations (instead of building the Associations listing itself). This makes the reco.cc more complete and less dependent on MatchCluster_factory magic.

This also modifies MatchClusters the algorithm to return tuple<unique_ptr<Collection> x 2>.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/global/reco/MatchClusters_factory.cc
+++ b/src/global/reco/MatchClusters_factory.cc
@@ -30,10 +30,10 @@ namespace eicrecon {
     }
 
     void MatchClusters_factory::Process(const std::shared_ptr<const JEvent> &event) {
-        size_t i = 0;
-        auto mc_particles = static_cast<const edm4hep::MCParticleCollection*>(event->GetCollectionBase(GetInputTags()[i++]));
-        auto charged_particles = static_cast<const edm4eic::ReconstructedParticleCollection*>(event->GetCollectionBase(GetInputTags()[i++]));
-        auto charged_particle_assocs = static_cast<const edm4eic::MCRecoParticleAssociationCollection*>(event->GetCollectionBase(GetInputTags()[i++]));
+        size_t tag_ix = 0;
+        auto mc_particles = static_cast<const edm4hep::MCParticleCollection*>(event->GetCollectionBase(GetInputTags()[tag_ix++]));
+        auto charged_particles = static_cast<const edm4eic::ReconstructedParticleCollection*>(event->GetCollectionBase(GetInputTags()[tag_ix++]));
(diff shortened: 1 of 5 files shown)


wdconinc:
No diffs at https://github.com/eic/EICrecon/actions/runs/7110106939/job/19356172753?pr=1175#step:11:108.

src/global/reco/MatchClusters_factory.cc
```diff
@@ -23,36 +23,27 @@ namespace eicrecon {
...
+        auto mc_particles = static_cast<const edm4hep::MCParticleCollection*>(event->GetCollectionBase(GetInputTags()[tag_ix++]));
+        auto charged_particles = static_cast<const edm4eic::ReconstructedParticleCollection*>(event->GetCollectionBase(GetInputTags()[tag_ix++]));
+        auto charged_particle_assocs = static_cast<const edm4eic::MCRecoParticleAssociationCollection*>(event->GetCollectionBase(GetInputTags()[tag_ix++]));
 
         std::vector<const edm4eic::ClusterCollection*> cluster_collections;
         std::vector<const edm4eic::MCRecoClusterParticleAssociationCollection*> cluster_assoc_collections;
 
-        for(auto &input_tag: GetInputTags()) {
+        // FIXME Use proper stride and adjacent views in C++23
+        auto stride = [](int n) {
```
ruse-traveler, line 42: I'm actually not familiar with ranges! So for my own understanding, what's going on with this block?

ruse-traveler (commented):
Very nice! Aside from the use of strides on line 42 (which is due to me be unfamiliar with the library), I think these changes will make this factory a lot more transparent!

src/global/reco/MatchClusters_factory.cc
wdconinc: Since we only have C++20 support we can't use strided views yet (i.e. every n'th element) which is a C++23 feature. We do have filtered views in C++20, so this creates a filter predicate that is equivalent to a strided view: stride(n) will be true for each n'th call.

ruse-traveler: Ah! I see! So then the block below uses that pick up every other collection starting at tag_ix (i.e. the clusters). And then likewise for the associations. Thanks!!

ruse-traveler (approved):

PR #1177, opened 2023-12-06
feat: migrate PhotoMultiplierDigi_factory to JOmniFactory

wdconinc:
This migrates PhotoMultiplierDigi_factory to JOmniFactory. Couple of points of note:
- `src/global/digi` now has only headers, so the target `digi_library` is gone; probably could move `PhotoMultiplierDigi_factory` into `src/factories/digi` for clarity,
- JANA doesn't support parsing of `std::vector<std::pair<double, double>>` for quantum efficiency, so that's disabled in the ParameterRef list (@nathanwbrei OT, but having to specify a variable name for these `ParameterRef` seems useless; can we stick them in a container?),
- the passing of a lambda which captures `this` and calls `m_RichGeoSvc().GetReadoutGeo(GetPluginName())` which in turn has a `call_once` is janky (but not a change from previous practice...):
  - `m_RichGeoSvc()` is implicitly modifying the service with `call_once` (and apparently allowed to; @nathanwbrei can JANA enforce immutability of services?)
  - the RichGeo service as used by the DRICH plugin and the RichGeo service as used by the DIRC plugin are going to be different, maybe, and whoever calls `GetReadoutGeo()` first gets the `call_once` flag; that seems dangerous. Putting both PhotoMultiplierDigi_factories in the same plugin doesn't seem like it would be possible. The RichGeo likely needs to be able to support more than one ReadoutGeo and initialize them on demand. It probably needs to fill a map by detector name (not plugin name), with a map of `call_once` flags or so. @c-dilks any thoughts on where you were going with that?

I think for now this maintains the status-quo, but we need to find a better solution to this in a different PR that reworks this...

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
No diffs.

wdconinc:
> "keep detector-dependence out of the algorithms."

Definitely appreciate this design philosophy! (and the difficulties that come with it...)

veprbl (approved):
Diff LGTM

PR #1185, opened 2023-12-14
Updated track seeding globaltoLocal() conversion

bschmookler:
As discussed [here](https://indico.bnl.gov/event/21496/#1-comparison-study-between-act), in ACTS v.30 we find our call to the globalToLocal() fails. This causes the seed ACTS loc_a to always get set positive:

image1

Above plot is for single negative muons generated with p = [0.5,20] GeV/c and eta = [-4,4] from (x,y,z) = (0,0,0).

This causes some bias after the seeds are sent through the CKF:

image2

Above plot shows muons generated at three different z locations, all with (x,y) = (0,0).

First fix attempt set the direction based on the seed reconstruction, but did not change the [tolerance](https://acts.readthedocs.io/en/v9.0.0/api/variable_namespace_acts_1a2387e3079dc446e063259a3734adb2bd.html#_CPPv4N4Acts20s_onSurfaceToleranceE):

```
Acts::Vector3 direction(sin(theta)*cos(phi), sin(theta)*sin(phi), cos(theta));
auto local = perigee->globalToLocal(m_geoSvc->getActsGeometryContext(),
                                                            global, 
					                    direction
                                                          );
```

The seed loc_a asymmetry was improved but not fully removed:

image3

The bias after the CKF is largely (but not entirely) removed:

image4

Second fix attempt was to adjust the tolerance in addition to the above changes:

```
Acts::Vector3 direction(sin(theta)*cos(phi), sin(theta)*sin(phi), cos(theta));
auto local = perigee->globalToLocal(m_geoSvc->getActsGeometryContext(),
                                                           global, 
					                   direction,
					                   1E-3 //Set tolerance to 1.0 um -- 10 times default value
					                   );
```
However, setting a tolerance of 1 um gave the exact same result as the default 0.1 um tolerance. Not sure if this parameter is actually being used. Maybe should go back to default?

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Yes


src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -207,10 +207,16 @@ std::unique_ptr<edm4eic::TrackParametersCollection> eicrecon::TrackSeeding::make
...
-
+      //FIXME: localpos(0) can be either positive or negative
+      //Overwritten if call to globalToLocal is successful
       Acts::Vector2 localpos(sqrt(square(xypos.first) + square(xypos.second)), z0);
+
+      Acts::Vector3 direction(sin(theta)*cos(phi), sin(theta)*sin(phi), cos(theta));
+      auto local = perigee->globalToLocal(m_geoSvc->getActsGeometryContext(),
+                                          global,
+                                          direction,
+                                          1E-3 //Set tolerance to 1.0 um -- 10 times default value
```
wdconinc, line 218: Please use units.

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -207,10 +207,16 @@ std::unique_ptr<edm4eic::TrackParametersCollection> eicrecon::TrackSeeding::make
...
+                                         );
       if(local.ok())
```
wdconinc, line 220: If we can't do a successful `globalToLocal`, then we probably shouldn't add the seed with the default `localpos` value.

bschmookler:
> Try setting the tolerance larger than 1 micron. That is expecting our seeds to have a 1 micron (or less) resolution of the pointing resolution to line centered at (x,y) =(0,0) line along the z axis. I don't think we'll ever have that precision especially at the seed level... Previously the tolerance was effectively infinity since any track was accepted, so setting it to something like 1 cm will probably recover the previous behavior. You could test how small/large this value can actually be before you start losing track seeds and it could perhaps be a metric by which we can reject non-primary track seeds from the first reconstruction pass.

Thanks. I'll check this. The resolution plots are in mm, so that makes sense the tolerance needs to be set larger :).

src/algorithms/tracking/TrackSeeding.cc
bschmookler: I now put this as a parameter in the _OrthogonalTrackSeedingConfig.h_ file.

src/algorithms/tracking/TrackSeeding.cc
bschmookler: I now set some initial dummy values, which allow checking whether a seed can successfully call `globalToLocal`.
\

bschmookler:
Now that I set  `Acts::Vector2 localpos(-9999., -9999.);` before calling `globalToLocal`, I can easily see if the call to that method works.

Previously, when we had the momentum direction always set to (1,1,1), I see the method always fail:

image

If I set the correct direction,   `Acts::Vector3 direction(sin(theta)*cos(phi), sin(theta)*sin(phi), cos(theta));`, I checked what happens with the default tolerance. I see that for 100 events, we have 302 seeds, and all 302 seeds succeed when calling the `globalToLocal `method. This is still true when I change the tolerance to `10um, 100um, 1mm, and 1cm`.

bschmookler:
Now that the `globalToLocal `succeeds, I'm seeing if changing the tolerance to 1mm or 1 cm changes the distribution of the seed Loc-a resolution.

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -207,14 +207,19 @@ std::unique_ptr<edm4eic::TrackParametersCollection> eicrecon::TrackSeeding::make
       auto perigee = Acts::Surface::makeShared<Acts::PerigeeSurface>(Acts::Vector3(0,0,0));
       Acts::Vector3 global(xypos.first, xypos.second, z0);
 
-      auto local = perigee->globalToLocal(m_geoSvc->getActsGeometryContext(),
-                                          global, Acts::Vector3(1,1,1));
+      //Overwritten if call to globalToLocal is successful
+      Acts::Vector2 localpos(-9999., -9999.);
```
wdconinc, line 211: No, this is potentially even worse. It is even more wrong than the sqrt(x,y) you had before.

The point is that you don't need this: if the conversion is bad, then don't generate a seed at all.

wdconinc (changes requested):

src/algorithms/tracking/TrackSeeding.cc
osbornjd: It's not clear to me that if the conversion is bad we just shouldn't generate a seed from it. This could potentially remove highly displaced tracks from the seeding, i.e. those that are not compatible with a line surface at (0,0,0), which could be very secondary decays from Kshort decays or conversions or...

wdconinc: By writing 9999, it is now a bad unphysical seed. Maybe you could do better for all the reasons Joe mentioned (and even the sqrt(x,y) was better), but writing 9999 is not right and in that case you are better off not writing any seed at all.

bschmookler: I did a few checks. 

1) I generated 10k muon events with p = [0.5,20] GeV/c and eta = [-4,4] and phi = [-Pi,Pi] at each of the following settings and checked how often the `globalToLocal` succeeds. I did this with a 1 mm tolerance and the updated `direction` vector.

| Position (x,y,z) mm      | Number of seeds found | Number of seeds where `globalToLocal `fails
| ------------------------ | -------------------------- |------------------------------------------------
| (0,0,0)                          |       25774                       | 0 |
| (0,0,10)                        |       25482                       | 0 |
| (0,0,-10)                      |        25583                       | 0 |
| (1,0,0)                          |       25521                       | 0 |
| (10,0,0)                       |        5567                         | 0 |

So, for all these events, I never see a failure in `globalToLocal`.

2) For the case where the muon is generated from (0,0,0), I checked the Acts Loc-a distribution with a 0.1 um tolerance and a 1 mm tolerance. The results are identical:

0.1 micron tolerance:

image_1

1 mm tolerance:

image_2

I would have expected to see an effect at 1 mm based on the range of the distribution.

3) As to the question of what default value to set before calling globalToLocal, maybe it doesn't matter based on the results above. But since Acts `Loc-a` will either be `+hypot(x,y)` or `-hypot(x,y)`, then we can either set the sign randomly to avoid any bias

`m_rng.uniform_double<double>(-1,1) < 0 ? -1 : 1`

Or we can set the sign based on the phi value. For example, for the case when I generate the particle from `(1,0,0) mm` and then plot the seed `phi` vs. seed Acts `Loc-a`, I see the following:

image_3

So, if `phi` is less than 0, the Acts Loc-a will be positive. I think this should hold in general, but I'm not completely sure.

osbornjd: In this simulation, are all the tracks thrown from exactly (0,0,0) (or whatever vertex value you used), i.e. there is no vertex spread at all?

bschmookler: Yes, that is correct. If I throw from `(x,y,z) = (0,0,0) mm`, for example, all 10k muons are thrown from exactly `(0,0,0)`.

osbornjd: I assume also the muons were all the same charge? Such that if you flipped the charge the plot of loca vs phi would basically be inverted

bschmookler: I only used negative muons in the above simulations.

I'm not sure if for positive muons the `loc-a vs. phi` plot would invert, since the globalToLocal determines `Loc-a` only based on the PCA position and the momentum direction, not the charge. So, for positive muons thrown from the same point `(x,y,z) = (1,0,0) mm` and the same `phi`, both the PCA position and direction would be the same (ignoring any bending of the particle between the creation point and the PCA position).

[https://acts.readthedocs.io/en/latest/core/geometry/surfaces.html#line-surface](https://acts.readthedocs.io/en/latest/core/geometry/surfaces.html#line-surface)

bschmookler: If I instead throw 10k positive muons from `(x,y,z) = (1,0,0) mm`, I see the plot below. So the shape is the same for both charge values.

image

There are some seeds with phi reconstructed very poorly due to the charge being incorrectly determined, as was shown in one of the Track Reconstruction meetings. And this 'distortion' is at different phi values for positive and negative muons.

osbornjd: What is the global x-y distribution of the seeds? Is this "modulation" distribution a result of the way we project the seeds back to the (0,0) x-y origin and then transform to the local coordinate on the perigee? Or is this modulation a result of the actual PCA calculation in the global to local transform?

bschmookler: The modulation comes from the projection of the particle to the point of closest approach w.r.t the line surface, as shown in the diagram below:

pic0

The sign of `Loc-a` is then given by whether the particle is going clockwise or counterclockwise around the line surface. This can be seen in the code here:

https://github.com/acts-project/acts/blob/main/Core/src/Surfaces/LineSurface.cpp#L80-L123

So, if the particle is generated at `(x,y,z) = (+1,0,0) mm`, then it will have the modulation shown above, since positive `phi` corresponds to counterclockwise rotation.

If the particle is generated at `(x,y,z) = (-1,0,0) mm`, then positive `phi` means clockwise rotation and a positive 'Loc-a':

pic1

bschmookler: I think that the Acts code linked above explains why we always see `globalToLocal` succeed once we use the correct `direction` vector, and why the results are the same irrespective of the `tolerance` we set. 

When we fit the seed triplet, we determine the circle parameters -- radius and center. Then we use that to determine the coordinates of the PCA. We set the momentum direction as tangential to the circle at the PCA. So, by construction, the `seed` direction vector fulfills the `globalToLocal` tangential requirement.

osbornjd: Okay, thanks for all this digging. I understand your point now and I think that essentially solves the issue of what is going on. So to wrap this PR up and get it merged, I think there are two things that need to be done
1. The tolerance can be removed completely, since as you showed by construction we have no dependence on it since the seed momentum vector is by construction tangential to the circle and thus perpendicular to the line surface at the global PCA provided.
2. Change the setting of  the local position by default to some -9999 values to just an uninitialized 2vector. If the global to local transform fails, then we just discard the seed since by construction it should never fail as we already determine the PCA where Acts expects it to be (with the proper momentum vector). So we could do something like 
```
if(!local.ok())
{
  continue;
}

localpos = local.value();
....
````

Does that seem like a reasonable path forward?

bschmookler: Yes, sounds good. I'll implement this.

bschmookler: I made the updates. I confirmed that the new results are equivalent to those shown in the tables/plots in this thread.

wdconinc: @osbornjd Could you review and approve this, at your convenience?

wdconinc (approved):

osbornjd (approved):
Looks good to me, thanks for the detailed study @bschmookler

PR #1186, opened 2023-12-14
Implement Particle Flow Alpha

ruse-traveler:

This PR implements factories for particle flow (PF) and a rough, initial algorithm (labeled `PF Alpha`). While there are significant improvements to be made in the algorithm as written, this PR serves as a starting point for general PF development at ePIC.

The algorithm divides the central ePIC detector into three regions of pseudorapidity (negative, central, and positive) and then runs a particle flow scheme set by a user-configurable index to process the tracks, track projections, and ECal/HCal pair for each region. This design is intended to accommodate the different physics needs of each region, and to allow for flexibility in testing/benchmarking.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #900 )
- [ ] Documentation update
- [ ] Other: __

No.

Yes. By default, EICrecon will now run PF and save its output.


src/global/reco/ParticleFlow_factory.h
```diff
@@ -0,0 +1,92 @@
...
+// for geometry services
+#include <services/geometry/dd4hep/DD4hep_service.h>
+// necessary algorithms and configurations
+#include "algorithms/reco/ParticleFlow.h"
+#include "algorithms/reco/ParticleFlowConfig.h"
+
+namespace eicrecon {
+
+  class ParticleFlow_factory :
+          public JChainMultifactoryT<ParticleFlowConfig>,
```
wdconinc, line 23: Please use JOmniFactory for any new algorithms. Good example is src/factories/calorimetry/CalorimeterClusterRecoCoG.

ruse-traveler:
I'm opening this as a draft since I think this will need quite a bit of discussion.

One thing I'd be curious to hear people's thoughts on is if there's a better way of organizing the algorithm. As written, all of the helper functions, particle flow schemes (i.e. just `PFAlpha` right now), and plumbing are in a single class. I think it might be better for both maintainability and readability to move `PFAlpha` into a separate class which is then included as a member in the `ParticleFlow`, e.g.
```
class ParticleFlow<Etc>
  /* usual things */
  private:
    PFAlpha alpha;
}
```
Future particle flow schemes would then have their header and be included in a similar way. Then the helper functions could moved into a separate header accessible to all of the particle flow schemes.

However, I'm not sure how you would handle the configuration files and the output collection in a setup like that...

src/global/reco/ParticleFlow_factory.h
```diff
@@ -0,0 +1,92 @@
...
+      std::map<size_t, std::string> m_mapIndexToECalID = {
+        {m_const.iNegative, "ECalEndcapN_ID"},
+        {m_const.iCentral,  "ECalBarrel_ID"},
+        {m_const.iPositive, "ECalEndcapP_ID"}
+      };
+      std::map<size_t, std::string> m_mapIndexToHCalID = {
+        {m_const.iNegative, "HCalEndcapN_ID"},
+        {m_const.iCentral,  "HCalBarrel_ID"},
+        {m_const.iPositive, "HCalEndcapP_ID"}
+      };
```
wdconinc, line 88: This functionality should go into the algorithm. There should be nothing in the factory that differs between algorithms.

src/global/reco/reco.cc
```diff
@@ -185,5 +185,22 @@ void InitPlugin(JApplication *app) {
             app
     ));
 
+    app->Add(new JChainMultifactoryGeneratorT<ParticleFlow_factory>(
```
wdconinc, line 188: JOmniFactoryGeneratorT.

src/global/reco/ParticleFlow_factory.cc
```diff
@@ -0,0 +1,126 @@
...
+    app->SetDefaultParameter(param_prefix + ":hcalSumRadius", cfg.hcalSumRadius);
+    app->SetDefaultParameter(param_prefix + ":ecalFracSub",   cfg.ecalFracSub);
+    app->SetDefaultParameter(param_prefix + ":hcalFracSub",   cfg.hcalFracSub);
+
+    // initialize particle flow algorithm
+    m_pf_algo.applyConfig(cfg);
+    m_pf_algo.init(logger());
+
+    // initialize geometry service
+    m_geoSvc = app -> template GetService<DD4hep_service>();
```
wdconinc, line 44: This boilerplate will almost all go away in JOmniFactory.

src/global/reco/ParticleFlow_factory.h
```diff
@@ -0,0 +1,92 @@
...
+      // geometry service
+      std::shared_ptr<DD4hep_service> m_geoSvc;
+
+      // class-wide constants
+      const struct constants {
+        size_t nCaloPairs;
+        size_t iNegative;
+        size_t iCentral;
+        size_t iPositive;
+      } m_const = {3, 0, 1, 2};
```
wdconinc, line 68: Not sure I understand the point of this struct. Aren't you using this as a standard enum?

src/global/reco/ParticleFlow_factory.h
```diff
@@ -0,0 +1,92 @@
...
+
+      // map of calo collection tags, id names onto indices
+      std::map<std::string, size_t> m_mapCaloInputToIndex = {
+        {"EcalEndcapNClusters",    m_const.iNegative},
+        {"HcalEndcapNClusters",    m_const.iNegative},
+        {"EcalBarrelScFiClusters", m_const.iCentral},
+        {"HcalBarrelClusters",     m_const.iCentral},
+        {"EcalEndcapPClusters",    m_const.iPositive},
+        {"LFHCALClusters",         m_const.iPositive}
```
wdconinc, line 77: Hardcoded collection names should be avoided. You should not assume any collection names, and work only on the input you are given. At some point in the future, there may be multiple ways to collect clusters, and there may be multiple ParticleFlow paths that work on different cluster collections. This prevents that for working.

src/global/reco/ParticleFlow_factory.cc
```diff
@@ -0,0 +1,126 @@
...
+    // for calo inputs
+    std::vector<const edm4eic::ClusterCollection*> vecECalInput(m_const.nCaloPairs);
+    std::vector<const edm4eic::ClusterCollection*> vecHCalInput(m_const.nCaloPairs);
+
+    // grab and organize input collections
+    TrkInput trkInput;
+    for (const std::string& input_tag : GetInputTags()) {
+
+      // check if input is a track collection
+      if (input_tag.find("ReconstructedChargedParticles") != std::string::npos) {
```
wdconinc, line 69: No hardcoded collection names.

src/global/reco/ParticleFlow_factory.cc
```diff
@@ -0,0 +1,126 @@
...
+        trkInput.first = static_cast<const edm4eic::ReconstructedParticleCollection*>(event->GetCollectionBase(input_tag));
+        m_log->debug("Found input track collection {}", input_tag);
+        continue;
+      }
+
+      // check if input is a track projection collection
+      if (input_tag.find("TrackProjection") != std::string::npos) {
```
wdconinc, line 76: No hardcoded collection names.

src/global/reco/ParticleFlow_factory.cc
```diff
@@ -0,0 +1,126 @@
...
+
+  void ParticleFlow_factory::Process(const std::shared_ptr<const JEvent> &event) {
+
+    // for calo inputs
+    std::vector<const edm4eic::ClusterCollection*> vecECalInput(m_const.nCaloPairs);
+    std::vector<const edm4eic::ClusterCollection*> vecHCalInput(m_const.nCaloPairs);
+
+    // grab and organize input collections
+    TrkInput trkInput;
+    for (const std::string& input_tag : GetInputTags()) {
```
wdconinc, line 66: The looping over input tags gives you flexibility but that's not what we need here. I think you should enforce the interface: first input = ReconstructedParticleCollection, second input = TrackSegmentCollection, third input = ClusterCollection, etc.

src/global/reco/ParticleFlow_factory.cc
```diff
@@ -0,0 +1,126 @@
...
+      if (input_tag.find("TrackProjection") != std::string::npos) {
+        trkInput.second = static_cast<const edm4eic::TrackSegmentCollection*>(event->GetCollectionBase(input_tag));
+        m_log->debug("Found input track projection collection {}", input_tag);
+        continue;
+      }
+
+      // check if input is a calo collection
+      bool isECalCollect = (input_tag.find("Ecal") != std::string::npos);
+      bool isHCalCollect = ((input_tag.find("Hcal") != std::string::npos) || (input_tag.find("HCAL") != std::string::npos));
+      bool endsInClust   = (input_tag.find("Clusters") != std::string::npos);
```
wdconinc, line 85: Use the types of the collection, not the names (you already realize this since you have had to duplicate the Hcal and HCAL check because people can't and shouldn't be trusted to stick to a single name).

src/global/reco/ParticleFlow_factory.cc
```diff
@@ -0,0 +1,126 @@
...
+    std::vector<CaloInput> vecCaloInput(m_const.nCaloPairs);
+    for (size_t iCaloPair = 0; iCaloPair < m_const.nCaloPairs; iCaloPair++) {
+      vecCaloInput[iCaloPair] = std::make_pair(vecECalInput[iCaloPair], vecHCalInput[iCaloPair]);
+    }
+
+    // prepare vector of pairs of ECal-HCal IDs
+    VecCaloIDs vecCaloIDs(m_const.nCaloPairs);
+    for (size_t iCaloPair = 0; iCaloPair < m_const.nCaloPairs; iCaloPair++) {
+      const uint32_t idECal = m_geoSvc -> detector() -> constant<uint32_t>(m_mapIndexToECalID[iCaloPair]);
+      const uint32_t idHCal = m_geoSvc -> detector() -> constant<uint32_t>(m_mapIndexToHCalID[iCaloPair]);
```
wdconinc, line 114: If there is a lookup needed from collection to detector ID, it should be done in the algorithm, not the factory. If there is a lookup needed, I don't think you need to go through strings and constants, but you should be able to use the links to hits and their cell ID.

src/global/reco/ParticleFlow_factory.cc
```diff
@@ -0,0 +1,126 @@
...
+      vecCaloIDs[iCaloPair] = std::make_pair(idECal, idHCal);
+    }
+
+    // run algorithm
+    auto pf_objects = m_pf_algo.process(trkInput, vecCaloInput, vecCaloIDs);
```
wdconinc, line 119: We are moving to make this all a lot simpler so that what goes into this process is only (and can only be) the collections you specified in the reco.cc generation. Nothing more. If you put all this logic into the factory, someone else is going to have to strip it out again in a few weeks/months.

src/global/reco/ParticleFlow_factory.h
```diff
@@ -0,0 +1,92 @@
...
+          public SpdlogMixin {
+
+    public:
+
+      // aliases for brevity
+      using TrkInput     = std::pair<const edm4eic::ReconstructedParticleCollection*, const edm4eic::TrackSegmentCollection*>;
+      using CaloInput    = std::pair<const edm4eic::ClusterCollection*, const edm4eic::ClusterCollection*>;
+      using CaloIDs      = std::pair<uint32_t, uint32_t>;
+      using VecCaloInput = std::vector<CaloInput>;
+      using VecCaloIDs   = std::vector<CaloIDs>;
```
wdconinc, line 33: These types will have to be kept in sync with the ParticleFlow.h header. You include that header here (you have to) so reuse the types from that header instead of redefining them.

But... kinda irrelevant comment since this all should go into the algorithm anyway :-)

src/algorithms/reco/ParticleFlow.h
```diff
@@ -0,0 +1,168 @@
...
+       */
+      struct MergedCluster {
+        bool done = false;
+        int32_t pdg = 0;
+        float mass = 0.;
+        float charge = 0.;
+        float energy = 0.;
+        edm4hep::Vector3f momentum = {0., 0., 0.};
+        edm4hep::Vector3f weighted_position = {0., 0., 0.};
+        std::vector<edm4eic::Cluster> clusters;
```
wdconinc, line 70: I'm worried about the copying that's going to be happening here. Do we really need a separate struct for this that we then need to copy into and out of? If we use podio types we already have, then we don't need to worry about that. Is there something missing in the podio cluster types?

src/algorithms/reco/ParticleFlow.h
```diff
@@ -0,0 +1,168 @@
...
+      ProjectMap m_projMap;
+
+      // class-wide constants
+      const struct Constants {
+        size_t   nCaloPairs;
+        uint64_t innerSurface;
+        int32_t  idPi0;
+        float    massPiCharged;
+        float    massPi0;
+      } m_const = {3, 1, 111, 0.140, 0.135};
```
wdconinc, line 155: Make the configuration values in ParticleFlowConfig.h and avoid a struct here.

src/algorithms/reco/ParticleFlow.h
```diff
@@ -0,0 +1,168 @@
...
+      using TrkInput     = std::pair<const edm4eic::ReconstructedParticleCollection*, const edm4eic::TrackSegmentCollection*>;
+      using CaloInput    = std::pair<const edm4eic::ClusterCollection*, const edm4eic::ClusterCollection*>;
+      using CaloIDs      = std::pair<uint32_t, uint32_t>;
+      using VecCaloInput = std::vector<CaloInput>;
+      using VecCaloIDs   = std::vector<CaloIDs>;
+
+      // aliases for internal types
+      using ClustMap      = std::map<edm4eic::Cluster, bool>;
+      using TrackMap      = std::map<edm4eic::ReconstructedParticle, bool>;
+      using ProjectMap    = std::map<edm4eic::TrackSegment, bool>;
```
wdconinc, line 84: A map from thing to bool is not useful. A map is already that. If something is in the map, it is true; if it isn't, then it is false. Benefit of not defining a bool as second: you don't need to go an loop over everything first to set it to false. So, where `track_map[track] = false` use `track_map.erase(it)`.

src/algorithms/reco/ParticleFlow.cc
```diff
@@ -0,0 +1,881 @@
...
+  // --------------------------------------------------------------------------
+  //! Get Energy From Nearest Track Projection
+  // --------------------------------------------------------------------------
+  /*! Helper function to get the energy of track projection nearest to a
+   *  point (e.g. the position of calorimeter cluster).
+   */
+  float ParticleFlow::get_energy_of_nearest_projection(const ProjectionBundle& bundle, const edm4hep::Vector3f& position, const float mass) {
+
+    // identify nearest projection
+    float    dNearest = 9999.;
```
wdconinc, line 784: Arbitrary number, use std::numeric_limits.

src/algorithms/reco/ParticleFlow.cc
```diff
@@ -0,0 +1,881 @@
...
+   */
+  ParticleFlow::MergedCluster ParticleFlow::make_merged_cluster(
+    const bool done,
+    const int32_t pdg,
+    const float mass,
+    const float chrg,
+    const float ene,
+    const edm4hep::Vector3f mom,
+    const edm4hep::Vector3f pos,
+    const std::vector<edm4eic::Cluster> clusters
```
wdconinc, line 815: Copy is taken on input, avoid with const reference.

src/algorithms/reco/ParticleFlow.cc
```diff
@@ -0,0 +1,881 @@
...
+  //! Make Merged Cluster
+  // --------------------------------------------------------------------------
+  /*! Helper function to create a merged cluster.
+   */
+  ParticleFlow::MergedCluster ParticleFlow::make_merged_cluster(
+    const bool done,
+    const int32_t pdg,
+    const float mass,
+    const float chrg,
+    const float ene,
```
wdconinc, line 812: Redblity s nhancd whn w dnt rmv lters.

src/algorithms/reco/ParticleFlow.h
```diff
@@ -0,0 +1,168 @@
...
+      using PointAndFound = std::pair<edm4eic::TrackPoint, bool>;
```
wdconinc, line 85: A pair of something and bool for these purposes is `std::optional<>`.

ruse-traveler:
Closed ahead of refactoring WIP for modularity.

PR #1187, opened 2023-12-21
fix: rm Acts detail header for stable interface

wdconinc:
In https://github.com/acts-project/acts/pull/2617 the detail header will be removed.

- [x] Bug fix (issue: Acts v31 compatibility)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


osbornjd (approved):
Looks fine to me. For a future PR we may consider just going ahead and implementing our own custom detector element (that has the exact same functionality as the current detec, as this is how the alignment transforms are applied. I imagine it will inherit from [DD4hepDetectorElement](https://github.com/acts-project/acts/blob/main/Plugins/DD4hep/include/Acts/Plugins/DD4hep/DD4hepDetectorElement.hpp) in our case

PR #1188, opened 2023-12-21
fix: forward compatibility with Acts-v31

wdconinc:
~~Thanks to C++20 we can now do templated lambdas which don't require syntactic validity of all code paths. That makes forward/backward compatibility across Acts versions easier with constexpr lambdas.~~

Ifdefs on defined Acts major versions allow us to bypass code blocks that are not compilable or parsable under different versions of Acts.

Ref: https://github.com/acts-project/acts/pull/2649

- [ ] Bug fix (issue #__)
- [x] New feature (issue: Acts v31 compatibility)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/IterativeVertexFinder.cc
+++ b/src/algorithms/tracking/IterativeVertexFinder.cc
@@ -87,21 +87,19 @@ std::unique_ptr<edm4eic::VertexCollection> eicrecon::IterativeVertexFinder::prod
   VertexSeeder::Config seederCfg(ipEst);
   VertexSeeder seeder(seederCfg);
   // Set up the actual vertex finder
-  VertexFinder finder = [&]<unsigned int ActsVersion = Acts::Version>(){
-    if constexpr (ActsVersion >= 310000u) {
-      VertexFinder::Config finderCfg(std::move(vertexFitter), std::move(linearizer),
-                                     std::move(seeder), std::move(ipEst));
-      finderCfg.maxVertices                 = m_cfg.m_maxVertices;
-      finderCfg.reassignTracksAfterFirstFit = m_cfg.m_reassignTracksAfterFirstFit;
-      return VertexFinder(std::move(finderCfg));
-    } else {
-      VertexFinder::Config finderCfg(vertexFitter, std::move(linearizer),
-                                     std::move(seeder), ipEst);
-      finderCfg.maxVertices                 = m_cfg.m_maxVertices;
-      finderCfg.reassignTracksAfterFirstFit = m_cfg.m_reassignTracksAfterFirstFit;
-      return VertexFinder(finderCfg);
-    }
-  }();
+  #if Acts_VERSION_MAJOR >= 31
+  VertexFinder::Config finderCfg(std::move(vertexFitter), std::move(linearizer),
+                                 std::move(seeder), std::move(ipEst));
+  finderCfg.maxVertices                 = m_cfg.m_maxVertices;
+  finderCfg.reassignTracksAfterFirstFit = m_cfg.m_reassignTracksAfterFirstFit;
+  VertexFinder finder(std::move(finderCfg));
+  #else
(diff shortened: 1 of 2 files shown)


wdconinc:
Ha, gcc and clang disagree on the validity here. That's annoying. Probably https://stackoverflow.com/questions/55909018/an-if-constexpr-branch-does-not-get-discarded-inside-lambda-that-is-inside-a-t and some careful parsing of the standard language...

wdconinc:
This is one of the remaining two red boxes for clang's c++20 support, as shown at https://en.cppreference.com/w/cpp/compiler_support#cpp20 (it is [P0588R1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0588r1.html)).

osbornjd (commented):
What exactly does this get us? One less `std::move` for when we update to v31, after which we'll just delete the `#else`?

wdconinc:
The current code does not compile under v31 because some of these config objects are now required to be unique ptrs and need explicit move semantics (and under v30 the move semantics doesn't work because it's a call by reference that we can't move to).

In a world where we make more frequent Acts upgrades we can add the backwards compatibility support gradually and support at least compilation under more than just one version. It allows us to roll out new versions in a more user-friendly way: we can build a container with Acts v31 that has the same main version that is in the container with Acts v30.

wdconinc:
We can keep the v30 support until it is not maintainable anymore. At this point I imagine we might have a hard break at v32 again due to the changes in the dd4hep geometry conversion that are being merged now. We probably won't want to do ifdefs for large parts of the code base that would be required to support both sides of that transition.

veprbl:
I don't get why we need ifdef. I think xvalue implicitly casts to rvalue.

wdconinc:
Of course I tried that too as the very first thing...

wdconinc:
But I'd be much happier with a clean solution where the same code works for both, so please have at it!

src/algorithms/tracking/IterativeVertexFinder.cc
```diff
@@ -86,11 +86,15 @@ std::unique_ptr<edm4eic::VertexCollection> eicrecon::IterativeVertexFinder::prod
...
   VertexSeeder seeder(seederCfg);
   // Set up the actual vertex finder
-  VertexFinder::Config finderCfg(vertexFitter, std::move(linearizer),
-                                 std::move(seeder), ipEst);
+  VertexFinder::Config finderCfg(std::move(vertexFitter), std::move(linearizer),
+                                 std::move(seeder), std::move(ipEst));
   finderCfg.maxVertices                 = m_cfg.m_maxVertices;
   finderCfg.reassignTracksAfterFirstFit = m_cfg.m_reassignTracksAfterFirstFit;
+  #if Acts_VERSION_MAJOR >= 31
+  VertexFinder finder(std::move(finderCfg));
```
veprbl, line 94: I also don't understand why we need an `std::move` here:
https://github.com/acts-project/acts/blob/v30.3.2/Core/include/Acts/Vertexing/IterativeVertexFinder.hpp#L145
takes struct `Config &cfg`
https://github.com/acts-project/acts/blob/v31.2.0/Core/include/Acts/Vertexing/IterativeVertexFinder.hpp#L145
takes struct `Config cfg`

wdconinc: As the Acts PR in the summary says, this is because the config contains loggers in these config structs which are now (v31) unique_ptrs. If we pass by value we need a copy, which aren't allowed. That's why we love them.

veprbl: Thanks! I thought that it was requirement of the interface, but absence of a copy constructor is also a valid reason.

veprbl (approved):
This compiles with acts 31.2.0

wdconinc:
Thanks for reducing the amount of ifdefed code!

PR #1191, opened 2023-12-28
feat: set default CMAKE_BUILD_TYPE in CMakeLists.txt

wdconinc:
Instead of leaving CMAKE_BUILD_TYPE unset, this sets the default to `Release`, unless `.git` exists and it sets `Debug`. This avoids ending up with `noconfig` on target exports (or unoptimized compiles).

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, developers with `.git` will likely switch from unspecified defaults to actual `Debug` configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,9 +37,6 @@ set_git_version(CMAKE_PROJECT_VERSION)
 
 # Set a default build type if none was specified
 set(default_build_type "Release")
-if(EXISTS "${CMAKE_SOURCE_DIR}/.git")
-  set(default_build_type "Debug")
-endif()
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
   set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,20 @@ include(CheckCXXCompilerFlag)
 include(cmake/git_version.cmake)
 set_git_version(CMAKE_PROJECT_VERSION)
 
+# Set a default build type if none was specified
+set(default_build_type "Release")
+if(EXISTS "${CMAKE_SOURCE_DIR}/.git")
+  set(default_build_type "Debug")
+endif()
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
+  set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE
+      STRING "Choose the type of build." FORCE)
+  # Set the possible values of build type for cmake-gui
+  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+    "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+endif()
+
 # Set default standard to C++20
 set(CMAKE_CXX_STANDARD_MIN 20)
 set(CMAKE_CXX_STANDARD ${CMAKE_CXX_STANDARD_MIN} CACHE STRING "C++ standard to be used")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -35,6 +35,20 @@ include(CheckCXXCompilerFlag)
 include(cmake/git_version.cmake)
 set_git_version(CMAKE_PROJECT_VERSION)
 
+# Set a default build type if none was specified
+set(default_build_type "Release")
+if(EXISTS "${CMAKE_SOURCE_DIR}/.git")
+  set(default_build_type "Debug")
```
veprbl, line 41: This violates principle of least surprise. I don't think anyone needs Debug. One can add -g to CMAKE_CXX_FLAGS_RELEASE.

wdconinc: There's two issues here:
- do we want the default build type to be defined or up to local preferences?
- do we want the default to be Debug when `.git` exists?

It sounds like we agree on the first one pointing to `Release` by default. We can disagree on the second one.

But requiring users to add `-g` to `CMAKE_CXX_FLAGS_RELEASE` to get Debug mode is not canonical. In that case, I'd rather expect users to use `CMAKE_BUILD_TYPE=Debug`, but not default to `Debug` when `.git.` exists. Is that what you're suggesting?

veprbl: > There's two issues here:
> 
>     * do we want the default build type to be defined or up to local preferences?

I don't see the point in each codebase defining their own default, that's not very reliable. Users with local preferences can always define CMAKE_BUILD_TYPE. The rest can copy-paste `-DCMAKE_BUILD_TYPE=Release` from README.md.

> 
>     * do we want the default to be Debug when `.git` exists?

Everyone has .git. Package managers may not have it, but those programs would define it.

> 
> 
> It sounds like we agree on the first one pointing to `Release` by default. We can disagree on the second one.
> 
> But requiring users to add `-g` to `CMAKE_CXX_FLAGS_RELEASE` to get Debug mode is not canonical. In that case, I'd rather expect users to use `CMAKE_BUILD_TYPE=Debug`, but not default to `Debug` when `.git.` exists. Is that what you're suggesting?

I meant when "-g" is sometimes appended to CMAKE_CXX_FLAGS_RELEASE in CMakeFiles.txt.

wdconinc: > Users with local preferences can always define CMAKE_BUILD_TYPE. The rest can copy-paste -DCMAKE_BUILD_TYPE=Release from README.md.

Then I don't understand what your objection is to setting the default to Release since that seems to be what you think should happen by default.

Recap about what I see as the problem here: we don't define any default build type, and users who don't explicitly set one are ending up with something that is neither of Release, Debug, MinSizeRel, or RelWithDebInfo. They essentially end up with what the compiler picks if you don't specify anything, which can differ from compiler to compiler (obligatory reference to the intel compiler enabling optimizations including fastmath by default and resulting in incorrect and non-IEEE compliant results). These compiler defaults are neither optimal for either development nor for production use: the default with c++ in eic-shell ends up being unoptimized code without debug symbols (`-O0` without `-g`). This is neither ideal for development since it generates mostly useless backtraces, and it isn't useful for production use since even basic safe optimizations are not applied.

veprbl: > > Users with local preferences can always define CMAKE_BUILD_TYPE. The rest can copy-paste -DCMAKE_BUILD_TYPE=Release from README.md.
> 
> Then I don't understand what your objection is to setting the default to Release since that seems to be what you think should happen by default.

The objection was to using different defaults based on obscure conditions. Picking either Release or Debug is okay.

I was suggesting some meaningful improvements like explicitly providing option in the README, and/or using environment variable to provide consistent experience for all cmake work in the container. This is all in response to your points, as I understood them.

wdconinc: Ok, then we agree on the solution here. I'll remove the `.git` branch. I'd like to avoid environment variables but I'll update the README.md too.

CMakeLists.txt
```diff
@@ -35,6 +35,20 @@ include(CheckCXXCompilerFlag)
...
+endif()
```
wdconinc, line 42:
```suggestion
```

veprbl (approved):

PR #1195, opened 2023-12-31
feat: register DD4hep_service as algorithms::GeoSvc provider

wdconinc:
To be able to use the algorithms interface, including the service requirements defined upstream, we need to connect the EICrecon DD4hep_service to algorithms::GeoSvc as a provider. This does exactly that.

The new AlgorithmsInit_service depends on the DD4hep_service, Log_service, and we use the algorithms RandomSvc since we don't have one in EICrecon yet. 

TODO:
- [x] ~~depends on https://github.com/eic/algorithms/pull/14~~ (or not really, but would avoid a const_cast)
- [x] depends on https://github.com/eic/algorithms/pull/16


wdconinc:
No diffs, e.g. https://github.com/eic/EICrecon/actions/runs/7522587001/job/20474895740?pr=1195#step:15:82

src/services/algorithms/services/ServiceSvc_service.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
```
wdconinc, line 1:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2023, 2024 Wouter Deconinck, Sylvester Joosten

#pragma once
```

wdconinc:
This is ready for review and merging. I could propagate this to all algorithms in this PR but I would prefer to merge this so it's getting some exercise on one algorithm and then propagate it further in a different PR. 

This relates to #1197 which uses the default LogSvc in algorithms and will need this PR to have LogSvc be provided by Log_service.

src/factories/calorimetry/CalorimeterHitDigi_factory.h
```diff
@@ -32,13 +32,13 @@ class CalorimeterHitDigi_factory : public JOmniFactory<CalorimeterHitDigi_factor
     ParameterRef<std::vector<std::string>> m_fields {this, "signalSumFields", config().fields};
     ParameterRef<std::string> m_readout {this, "readoutClass", config().readout};
 
-    Service<DD4hep_service> m_geoSvc {this};
+    Service<ServiceSvc_service> m_serviceSvc {this};
```
wdconinc, line 35: Just want to call you attention to the new service line that would need to be added to factories to use this (and some changes in the algorithm itself).

src/services/algorithms/services/ServiceSvc_service.h
```diff
@@ -0,0 +1,70 @@
...
+
+        // Get services
+        m_log_service = srv_locator->get<Log_service>();
+        m_dd4hep_service = srv_locator->get<DD4hep_service>();
+
+        // Logger for ServiceSvc
+        m_log = m_log_service->logger("ServiceSvc");
+
+        // Register DD4hep_service as algorithms::GeoSvc
+        [[maybe_unused]] auto& geoSvc = algorithms::GeoSvc::instance();
```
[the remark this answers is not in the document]
wdconinc: This is still needed. Without it, there is no geoSvc instantiated. @sly2j Is this intended behavior? How does juggler get around this?

wdconinc:
No remaining diffs, https://eicrecon.epic-eic.org/pr/1195/capybara/rec_dis_18x275_minQ2=1000_craterlake/index.html

src/tests/algorithms_test/calorimetry_CalorimeterHitDigi.cc
```diff
@@ -44,7 +44,7 @@ TEST_CASE( "the clustering algorithm runs", "[CalorimeterHitDigi]" ) {
     cfg.pedMeanADC = 123;
     cfg.resolutionTDC = 1.0 * dd4hep::ns;
     algo.applyConfig(cfg);
-    algo.init(detector.get(), logger);
+    algo.init(logger);
```
veprbl, line 47: Here, the detector instance is trivial, but how would we go about providing it in other cases?

wdconinc: It's all being handled through the algorihtm, which is what specifies which services it needs. So, e.g. CalorimeterIslandCluster would get a class member `const algorithms::GeoSvc& m_geoSvc = algorithms::GeoSvc::instance()` which would then expose `m_geoSvc.detector()`. We can even cache that as `m_detector` and access everything the way we do now.

The new bit here is the ServiceSvc_service which connects DD4hep_service to algorithms::ServiceSvc so it knows that algorithms::GeoSvc is to be provided by DD4hep_service.

wdconinc: In a sense it is allowing the algorithm to express which services it needs, and allowing EICrecon to specify which services it can provide. This is better than having the interface of algorithms (i.e. `init()`) depend on having factories pass resources from the services.

veprbl: But this is not EICrecon, this is a unit test.

wdconinc: Ah, ok, I see what you are asking. Let me figure that one out and modify the test to use the (trivial) detector. We will need to provide a (trivial) ServiceSvc ourselves, I think.

veprbl: That's what concerns me a bit. If we need to rely on how symbols are resolved (to full ServiceSvc or mock one), that might be a bit fragile.

wdconinc: Fortunately the algorithms service infrastructure is much more lightweight than JANA's and doesn't depend on symbol resolution in the same way (or at all, depending on how specifically you are defining symbol resolution).

It's all based on a singleton pattern so in a single test process it is entirely independent of other services and it only relies on C++ language features, not dlopen. See 20aa4ca.

src/tests/algorithms_test/calorimetry_CalorimeterHitDigi.cc
```diff
@@ -29,6 +29,18 @@ TEST_CASE( "the clustering algorithm runs", "[CalorimeterHitDigi]" ) {
...
+  auto& serviceSvc = algorithms::ServiceSvc::instance();
+  [[maybe_unused]] auto& geoSvc = algorithms::GeoSvc::instance();
+  serviceSvc.setInit<algorithms::GeoSvc>([&detector](auto&& g) {
+    g.init(detector.get());
+  });
+  [[maybe_unused]] auto& randomSvc = algorithms::RandomSvc::instance();
+  serviceSvc.setInit<algorithms::RandomSvc>([](auto&& r) {
+    r.setProperty("seed", static_cast<size_t>(1));
+    r.init();
+  });
```
wdconinc, line 41:
```suggestion
  [[maybe_unused]] auto& randomSvc = algorithms::RandomSvc::instance();
  auto seed = Catch::getSeed();
  serviceSvc.setInit<algorithms::RandomSvc>([seed](auto&& r) {
    r.setProperty("seed", static_cast<size_t>(seed));
    r.init();
  });
```

wdconinc: Also needs `#include <catch2/catch_get_random_seed.hpp>` above the current context.

wdconinc: After 3.1.1, https://github.com/catchorg/Catch2/commit/dc001fa935d71b4b77f263fce405c9dbdfcbfe28

wdconinc: Requiring 3.1.1, released Oct 19, 2022. Upgrading nightly to include.

wdconinc: Unfortunately not finding any defines with the version in catch2 to make this work with pre-3.1.1 too...

src/services/algorithms/services/services.cc
```diff
@@ -0,0 +1,17 @@
+// Copyright 2022, David Lawrence
```
veprbl, line 1:
```suggestion
// Copyright 2024, Wouter Deconinck
```

src/services/algorithms/services/services.cc
```diff
@@ -0,0 +1,17 @@
...
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
+//
```
veprbl, line 4:
```suggestion
```

src/services/algorithms/services/services.cc
```diff
@@ -0,0 +1,17 @@
...
+#include <JANA/JApplication.h>
+#include <memory>
+
+#include "ServiceSvc_service.h"
+
+
+extern "C" {
+void InitPlugin(JApplication *app) {
+    InitJANAPlugin(app);
+    app->ProvideService(std::make_shared<ServiceSvc_service>(app) );
```
veprbl, line 15:
```suggestion
    app->ProvideService(std::make_shared<ServiceSvc_service>(app));
```

src/tests/algorithms_test/calorimetry_CalorimeterHitDigi.cc
```diff
@@ -22,13 +28,26 @@ using eicrecon::CalorimeterHitDigi;
...
-
   std::shared_ptr<spdlog::logger> logger = spdlog::default_logger()->clone("CalorimeterHitDigi");
   logger->set_level(spdlog::level::trace);
 
   auto detector = dd4hep::Detector::make_unique("");
 
+  auto& serviceSvc = algorithms::ServiceSvc::instance();
+  [[maybe_unused]] auto& geoSvc = algorithms::GeoSvc::instance();
+  serviceSvc.setInit<algorithms::GeoSvc>([&detector](auto&& g) {
+    g.init(detector.get());
```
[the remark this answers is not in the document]
wdconinc: `detector` has and retains ownership of the `unique_ptr`, and the `algorithms::GeoSvc` inside `serviceSvc` only gets a const raw pointer view into it through the `g.init` call inside the lambda. `algorithms::Geosvc` does not own the geometry if it initialized with a const raw pointer, and stores the geometry in a different owning unique_ptr if it has to initialize the goemetry itself. When `GeoSvc` is destroyed no cleanup of `detector` happens. `detector` is destroyed only once, when it goes out of scope when `TEST_CASE` is destroyed.

Ref: https://github.com/eic/algorithms/blob/main/core/include/algorithms/geo.h#L36 definition where const raw pointer is stored, and https://github.com/eic/algorithms/blob/main/core/src/geo.cpp#L11 where the actual handling of non-null const raw pointer is done.

wdconinc:
(Rebased earlier and lost some commits, but re-added them again.)

src/services/algorithms/services/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 3.16)
```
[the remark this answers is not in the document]
wdconinc: How about `src/services/algorithms/init/` with `algorithms_init.cc` and `AlgorithmsInit_service`? It would require the factory to have `Service<AlgorithmsInit_service> m_algorithmsInitSvc{this};`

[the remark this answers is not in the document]
wdconinc: Hmm, won't work due to `get_filename_component(PLUGIN_NAME ${CMAKE_CURRENT_LIST_DIR} NAME)` but I've always wanted to change that into the relative path under `src` with `/` changed to `_`.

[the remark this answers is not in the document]
wdconinc: How about `src/services/algorithms_init/` with `algorithms_init.cc` and `AlgorithmsInit_service`? That doesn't put some generic `init.so` plugin in a common dir.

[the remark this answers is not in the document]
veprbl: Sounds good to me!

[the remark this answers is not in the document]
wdconinc: Renamed in https://github.com/eic/EICrecon/pull/1195/commits/b91d559c6d57dea32b71b6808a7ca55e9bb0bb58 and https://github.com/eic/EICrecon/pull/1195/commits/07adf7ce7ed27d1ecc0bbf79b9c55c35c7caba30

veprbl (approved):
Thanks

Issue #1166, opened 2023-12-04
TrackParamTruthInitConfig assumes Acts unit system


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD

### Steps to reproduce: (give a step by step account of how to trigger the bug)
In https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackParamTruthInitConfig.h the vertex and momentum variables are encoded in the Acts unit system.

### Expected Result: (what do you expect when you execute the steps above)
Our convention in all other configs is to specify variables in the DD4hep unit system.

Issue #925, opened 2023-09-06
Adopt EDM4eic unit system constants


veprbl:
Since https://github.com/eic/EDM4eic/pull/35 is now available, we should look to adopt those constants instead of incorrect uses of `dd4hep::` constants (those work only coincidentally, for energy units). The https://github.com/eic/EICrecon/pull/671 gives more context on this. The #390 depends on this.

wdconinc:
FYI https://github.com/AIDASoft/podio/commit/60dcd12f062658cbd62442d83514b2eb9f140277

PR #1198, opened 2024-01-02
fix: convert TrackParamTruthInit to use DD4hep units

wdconinc:
Units are set (via command line or config files) assuming a certain unit system. We have chosen DD4hep for this in all other cases (in part because it comes with a parser). Having one algorithm use Acts units is potentially confusing since distances may be passed in the wrong units.

This PR converts TrackParamTruthInit to store the config variables in DD4hep units, and modifies the algorithm itself to interpret the config in DD4hep units. No changes are expected to the output.

No values are specified in the factory or in the plugin, so this is the only change that should be needed.

- [x] Bug fix (issue #1166)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -4,7 +4,7 @@
 
 #include "TrackParamTruthInit.h"
 
-#include <Evaluate/DD4hepUnits.hpp>
+#include <Evaluator/DD4hepUnits.hpp>
 #include <TParticlePDG.h>
 #include <edm4hep/Vector3d.h>
 #include <edm4hep/Vector3f.h>
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -91,7 +91,7 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
         // Insert into edm4eic::TrackParameters, which uses numerical values in its specified units
         auto track_parameter = track_parameters->create();
         track_parameter.setType(-1); // type --> seed(-1)
-        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y) / dd4hep::mm), static_cast<float>(v.z / dd4hep::mm)}); // 2d location on surface [mm]
+        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y)), static_cast<float>(v.z)}); // 2d location on surface [mm]
         track_parameter.setLocError({1.0, 1.0}); // sqrt(variance) of location [mm]
         track_parameter.setTheta(theta); // theta [rad]
         track_parameter.setPhi(phi); // phi [rad]
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -4,7 +4,7 @@
 
 #include "TrackParamTruthInit.h"
 
-#include <Acts/Definitions/Units.hpp>
+#include <Evaluate/DD4hepUnits.hpp>
 #include <TParticlePDG.h>
 #include <edm4hep/Vector3d.h>
 #include <edm4hep/Vector3f.h>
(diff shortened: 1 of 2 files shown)


src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -4,7 +4,7 @@
 
 #include "TrackParamTruthInit.h"
 
-#include <Acts/Definitions/Units.hpp>
+#include <Evaluate/DD4hepUnits.hpp>
```
wdconinc, line 7:
```suggestion
#include <Evaluator/DD4hepUnits.hpp>
```

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -4,7 +4,7 @@
 
 #include "TrackParamTruthInit.h"
 
-#include <Acts/Definitions/Units.hpp>
+#include <Evaluator/DD4hepUnits.hpp>
```
wdconinc, line 7:
```suggestion
#include <Evaluator/DD4hepUnits.h>
```

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -91,16 +86,16 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
...
+        const auto pinit = pmag * (1.0 + m_cfg.m_momentumSmear * m_normDist(generator));
 
         // Insert into edm4eic::TrackParameters, which uses numerical values in its specified units
         auto track_parameter = track_parameters->create();
         track_parameter.setType(-1); // type --> seed(-1)
-        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y)), static_cast<float>(v.z)}); // 2d location on surface [mm]
+        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y) / dd4hep::mm), static_cast<float>(v.z / dd4hep::mm)}); // 2d location on surface [mm]
         track_parameter.setLocError({1.0, 1.0}); // sqrt(variance) of location [mm]
-        track_parameter.setTheta(theta); //theta [rad]
+        track_parameter.setTheta(theta); // heta [rad]
```
wdconinc, line 96:
```suggestion
        track_parameter.setTheta(theta); // theta [rad]
```

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -91,16 +86,16 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
...
 
         // modify initial momentum to avoid bleeding truth to results when fit fails
-        const auto pinit = pmag*(1.0 + m_cfg.m_momentumSmear * m_normDist(generator));
+        const auto pinit = pmag * (1.0 + m_cfg.m_momentumSmear * m_normDist(generator));
 
         // Insert into edm4eic::TrackParameters, which uses numerical values in its specified units
         auto track_parameter = track_parameters->create();
         track_parameter.setType(-1); // type --> seed(-1)
-        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y)), static_cast<float>(v.z)}); // 2d location on surface [mm]
+        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y) / dd4hep::mm), static_cast<float>(v.z / dd4hep::mm)}); // 2d location on surface [mm]
```
wdconinc, line 94:
```suggestion
        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y)), static_cast<float>(v.z)}); // 2d location on surface [mm]
```

wdconinc:
No diffs.

veprbl:
> No diffs.

Is that with running capybara locally?
I see diffs on the web pages right now.

wdconinc:
Might be that webpages are out of sync. Last CI run eg https://github.com/eic/EICrecon/actions/runs/7503697056/job/20430403477 has no diffs.

veprbl (approved):
Diff LGTM. We should be transitioning to `edm4eic::unit`, but now is not the time.

wdconinc:
Definitely we should move to `edm4eic::units`, see https://github.com/eic/EICrecon/issues/925.

PR #1203, opened 2024-01-02
fix: eicmkplugin: target_link_library EICrecon::EICrecon interface lib

wdconinc:
This adds an effective target EICrecon::EICrecon that links all plugins and libraries for very broad-based plugin support, and uses this interface in the eicmkplugin test.

Also:
- algorithms::algocore should use namespace
- add a few more missing find_dependencies

- [x] Bug fix (issue: https://github.com/eic/EICrecon/actions/runs/7388613847/job/20100681742)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


cmake/jana_plugin.cmake
```diff
@@ -90,6 +100,8 @@ macro(plugin_add _name)
             EXPORT EICreconTargets
             DESTINATION ${PLUGIN_LIBRARY_OUTPUT_DIRECTORY}
         )
+        # Link library into interface library
+        target_link_libraries(${PROJECT_NAME} INTERFACE ${_name}_library)
```
veprbl, line 104: It would simpler to instead move
```
        target_include_directories(${_name}_plugin
          PUBLIC
            $<BUILD_INTERFACE:${EICRECON_SOURCE_DIR}/src>
            $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}>
        )
```
to apply to the `${PROJECT_NAME}`? The current way with indirection looks fancy, but none of the plugins/libraries present different interfaces, I reckon.

wdconinc: While we can add this to the PROJECT_NAME target, we would still need to add an install interface link library to the granular targets that are exported or they would be invalid and downstream would complain if it includes a build interface location. We can move away from exporting the plugins and libraries as targets but that's more complicated with the way it's set up now.

Issue #1176, opened 2023-12-06
Factories should use JOmniFactory


wdconinc:
Several factories need conversion from JChainMultifactory to JOmniFactory:
- [x] #1312
  - [x] TrackerMeasurementFromHits #1168 
  - [x] TrackerHitCollector #1282
  - [x] ChargedParticleSelector #1217 
- [x] factories that don't produce podio collections require the use of the Output syntax in JOmniFactory,
  - [x] CKFTracking writes ActsExamples::Trajectories and ActsExamples::ConstTrackContainer #1168 
  - [x] RichTrack reads ActsExamples::Trajectories and ActsExamples::ConstTrackContainer #1395
  - [x] TrackPropagation reads ActsExamples::Trajectories #1317
  - [x] TrackProjector reads ActsExamples::Trajectories #1168 
  - [x] IterativeVertexFinder reads ActsExamples::Trajectories #1168 
- [x] factories that try to be too clever and do stuff that should be in the algorithm,
  - [x] GeneratedJets passes a vector of Lorentz vectors to the algorithm: being worked on in #1067
  - [x] ReconstructedJets passes a vector of Lorentz vectors to the algorithm: being worked on in #1067
  - [x] PhotoMultiplierHitDigi works differently depending on the name of the plugin in #1177
  - [x] TrackPropagation creates lists of surfaces to propagate to #1317
  - [x] ParticlesWithPID #1178
  - [x] IrtCherenkovParticleID #1418 
- [x] 'variadic' factories that accept arbitrary vectors of input tags and collate them before passing to the algorithm,
  - [x] MatchClusters https://github.com/eic/EICrecon/pull/1314
  - [x] MergeTrack https://github.com/eic/EICrecon/pull/1313
  - [x] ReconstructedElectrons #1209
  - [x] TrackerHitCollector #1282
  - [x] MergeCherenkovParticleID #1419

PR #1209, opened 2024-01-08
Variadic OmniFactories

nathanwbrei:

Adds support for variadic PODIO inputs inside JOmniFactories. This lets us fetch an arbitrary number of collections specified simply via input tags. Includes a migrated ReconstructedElectrons factory. 

It also enables us to migrate MatchClusters, MergeTrack, and TrackerHitCollector in the near future.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1176)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -111,6 +111,7 @@ public:
         friend class JOmniFactory;
 
         void GetCollection(const JEvent& event) {
+            m_data.clear();
             for (auto& coll_name : this->collection_names) {
                 m_data.push_back(event.GetCollection<PodioT>(coll_name));
             }
@@ -440,6 +441,7 @@ public:
         }
 
         for (size_t i = 0; auto* input : m_inputs) {
+            input->collection_names.clear();
             if (input->is_variadic) {
                 for (size_t j = 0; j<(vcc/vic); ++j) {
                     input->collection_names.push_back(default_input_collection_names[i++]);
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -129,7 +129,9 @@ public:
 
     struct OutputBase {
         std::string type_name;
-        std::string collection_name;
+        std::vector<std::string> collection_names;
+        bool is_variadic = false;
+
         virtual void CreateHelperFactory(JOmniFactory& fac) = 0;
         virtual void SetCollection(JOmniFactory& fac) = 0;
         virtual void Reset() = 0;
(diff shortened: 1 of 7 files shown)


src/extensions/jana/JOmniFactory.h
```diff
@@ -377,12 +408,48 @@ class JOmniFactory : public JMultifactory {
...
+        size_t tic = m_inputs.size();
+        size_t vic = 0;
+        size_t tcc = default_input_collection_names.size();
+
+        for (auto* input : m_inputs) {
+            if (input->is_variadic) {
+               vic += 1;
+            }
+        }
+        size_t vcc = tcc - (tic - vic);
```
wdconinc, line 420: Use more descriptive names.

src/extensions/jana/JOmniFactory.h
```diff
@@ -34,7 +34,9 @@ class JOmniFactory : public JMultifactory {
 
     struct InputBase {
         std::string type_name;
-        std::string collection_name;
+        std::vector<std::string> collection_names;
+        bool is_variadic = false;
```
wdconinc, line 38: I keep wondering if `is_variadic` can be `collection_names.size() > 1`...

nathanwbrei: In principle, yes. At least until we decide we also need multi-collection inputs with fixed multiplicity (which one person has already asked for!)

wdconinc: If it's multicollection with fixed multiplicity, it can just be repeated. Not sure I agree that's something that needs implementing. Or is there something I'm missing?

nathanwbrei: I'm also leaning against it, just not rejecting it yet. 

The use case is as follows: Imagine a detector that has separate layers. Each layer has its own (subset) collection. The algorithm should be designed for an arbitrary number of layers, but when running EICrecon we know there will be exactly four. Right now we can either have four separate PodioInputs and put their collection pointers in a vector to pass to the algorithm, or we can have one VariadicPodioInput and pass it directly, at the cost of constraining what our other inputs can be. If we could state that there will be exactly four collections in the Input, there would be no constraints on the input and no manual repackaging step.

nathanwbrei: Right now I favor sticking to three flavors of Input: Optional, ExactlyOne, and Variadic. They correspond nicely to the relations in UML: 0..1, 1, *

wdconinc: Yes, and at some point we'll write the UML exporter for the whole set of factories :-)

wdconinc (approved):

PR #1210, opened 2024-01-10
ParticlesWithPID: add associations from ReconstructedParticles to TrackParameters

veprbl:

Allows to relate between physics and tracking objects. Requested by HF group.

Resolves: #725

The current EDM suggests relation goes as

```
 edm4eic::ReconstructedParticle.tracks -> [edm4eic::Track.parameters] -> [edm4eic::TrackParameters]
```

- [x] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


veprbl:
Here is how to use this: https://github.com/eic/snippets/pull/1/files

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -99,8 +99,10 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "CentralTrackSegments",
             "CentralTrackVertices",
             "CentralCKFTrajectories",
+            "CentralCKFTracks",
             "CentralCKFTrackParameters",
             "CentralCKFSeededTrajectories",
+            "CentralCKFSeededTrajectories",
```
wdconinc, line 105: Duplicated?

osbornjd (commented):
What is the long term plan for this (I'm probably missing some discussion points as I am not at the workfests)? In principle the tracking algorithm itself (CKF) should create a track object and this should then be picked up by subsequent algorithms. Is this PR intended be a "stop gap" solution for this until all of the track EDM objects are consolidated into one thing as the output from the CKF?

src/services/io/podio/JEventProcessorPODIO.cc
veprbl: No, just an incomplete edit. Pushed another iteration. Thanks!

veprbl:
> What is the long term plan for this (I'm probably missing some discussion points as I am not at the workfests)?

There is no missing context. Please, let's discuss in this PR if need be.

The long term plan is that we might be splitting TrackParameters -> ReconstructedParticle conversion from the PID (This is to be tracked in #1210), which would be another refactoring. The review of overall soundness of the EIC tracking data model is outside of scope. This PR just implements the existing object relational model into EICrecon.

> In principle the tracking algorithm itself (CKF) should create a track object and this should then be picked up by subsequent algorithms. Is this PR intended be a "stop gap" solution for this until all of the track EDM objects are consolidated into one thing as the output from the CKF?

I suppose it's sort of a stop gap until Track objects are generated, presumably as a result of association of tracks with vertices. At least, that's my interpretation based on the declaration
https://github.com/eic/EDM4eic/blob/6904444649f4753f804f529066fc072db717b871/edm4eic.yaml#L400

osbornjd:
> I suppose it's sort of a stop gap until Track objects are generated, presumably as a result of association of tracks with vertices. At least, that's my interpretation based on the declaration

In an ideal world with everything completed I think we would have the CKF output a `Track` object, which you already linked. Then we would have a vertexing algorithm run and create a `Vertex` object that links vertices to tracks.  So I guess this PR basically does that except in the PID reco algo. 

Right now the CKF outputs 4 (!) containers as seen [here](https://github.com/eic/EICrecon/blob/5c34f24be0f2474867c47874aa8fed67e64872f2/src/algorithms/tracking/CKFTracking.cc#L79-L85). In principle we should just consolidate this into one `TrackCollection` (which if we have updated the default version of edm4eic, not sure if that is the case or not) we should be able to do. So I would say we could merge this PR as is but we should do the longer term solution sooner rather than later as more downstream code will just become dependent on all these containers existing (as is somewhat already the case, which is probably what makes consolidating all of these containers somewhat intrusive)...

veprbl:
Can we approve or reject this?

osbornjd (approved):
I can approve and leave to someone else to merge, in case there are differing opinions

Issue #1220, opened 2024-01-12
Reconstruction for the LYSO + SiPM-on-tile ZDC.  


sebouh137:
**Is your feature request related to a problem? Please describe.**
Yes.  Current version of the ZDC recon doesn't include reconstruction algorithms we've developed for the SiPM-on-tile ZDC.  Also, the dynamic range for the ZDC ECal (which now uses LYSO crystals) is not sufficient.  

**Describe the solution you'd like**
Add new algorithms, such as HEXPLIT and LogWeightReco to reconstruct clusters of hits from the SiPM-on-tile ZDC.  Add factories as appropriate, and modify the ZDC plugin to include these algorithms, and add the banks to the JEventProcessorPODIO.cc.  Also increase dynamic range for ZDC ECAL.   

**Describe alternatives you've considered**
no alternatives have been considered

**Additional context**

sebouh137:
> Hi @sebouh137! Just checking in on this issue: it looks like PR #1221 resolved this. Is there anything else to address here?

Hi @ruse-traveler , I believe you are correct and that this issue was resolved by PR 1221.  Thanks!

ruse-traveler:
Awesome! Thanks for confirming!

PR #1221, opened 2024-01-12
ZDC:  LYSO + SiPM-on-tile

sebouh137:
Adds functionality to run HEXPLIT and other algorithms for the reconstruction of clusters in the SiPM-on-tile ZDC.  Also increases the dynamic range for the ZDC Ecal (LYSO crystal).  

- [ ] Bug fix (issue #__)
- [X] New feature (issue #1220 )
- [ ] Documentation update
- [ ] Other: __

no.  
Changes the values of the parameters for the ZDC ECal digitization.


wdconinc:
I'd already sent to this @Gongzi-Zhang, but looking at https://github.com/eic/EICrecon/blob/sipmzdc/src/algorithms/calorimetry/LogWeightReco.cc that algorithm is already available as https://github.com/eic/EICrecon/blob/sipmzdc/src/algorithms/calorimetry/CalorimeterHitsMerger.h and would be good to reuse.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+  double MIP=m_cfg.MIP/dd4hep::GeV;
+  double Emin=m_cfg.Emin_in_MIPs*MIP/dd4hep::GeV;
+  double tmax=m_cfg.tmax/dd4hep::ns;
+
+
+  double x[nhits];
+  double y[nhits];
+  double z[nhits];
+  double E[nhits];
+  double t[nhits];
```
wdconinc, line 51: This is probably better solved with a std::vector of tuples, which doesn't require using `nhits` to create an array.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+
+  for(int i=0; i<nhits; i++){
```
wdconinc, line 53: Use range-for over `hits`.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+    x[i]=hits[i].getLocal().x;
+    y[i]=hits[i].getLocal().y;
+    z[i]=hits[i].getLocal().z;
```
wdconinc, line 56: Can we store the `Vector3f` instead of pulling apart the x,y,z components?

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+    E[i]=hits[i].getEnergy();
+    t[i]=hits[i].getTime();
+  }
+
+  auto subcellHits = std::make_unique<edm4eic::CalorimeterHitCollection>();
+  double Esum=0;
+  for(int i=0; i<nhits; i++){
```
wdconinc, line 63: Use a range-for loop.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+    //skip hits that do not pass E and t cuts
+    if (E[i]<Emin || t[i]>tmax)
+      continue;
+    //keep track of the energy in each neighboring cell
+    double Eneighbors[SUBCELLS];
```
wdconinc, line 68: Use std::vector with initializer, i.e. constructor (3) at https://en.cppreference.com/w/cpp/container/vector/vector.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+    for (int j=0; j<SUBCELLS; j++)
+      Eneighbors[j]=0;
+
+    for (int j=0; j<nhits; j++){
```
wdconinc, line 72: Use a range-for loop.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+      //only look at hits nearby within two layers of the current layer
+      if (abs(z[i]-z[j])>2.5*layer_spacing || z[i]==z[j])
```
wdconinc, line 74: Since `z` is of type double, `z[i] == z[j]` is not necessarily well defined without using epsilon.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+        continue;
+      //difference in transverse position (in units of side lengths)
+      double dx=(x[j]-x[i])/sl;
+      double dy=(y[j]-y[i])/sl;
+      if (abs(dx)>2 || abs(dy)>sqrt(3))
+        continue;
+
+      //loop over locations of the neighboring cells
+      //and check if the jth hit matches this location
+      double tol=0.01; //tolerance for rounding errors
```
wdconinc, line 86: This is a magical constant that should instead be from numeric_limits, https://en.cppreference.com/w/cpp/types/numeric_limits.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,145 @@
...
+    }
+    double sum_weights=0;
+    for(int k=0; k<SUBCELLS; k++){
+      weights[k]=Eneighbors[neighbor_indices[k][0]]*Eneighbors[neighbor_indices[k][1]]*Eneighbors[neighbor_indices[k][2]];
+      sum_weights+=weights[k];
+    }
+    for(int k=0; k<SUBCELLS;k++){
+
+      //create the subcell hits.  First determine their positions in local coordinates.
+      Eigen::Vector3d local(x[i]+subcell_offsets_x[k]*sl, y[i]+subcell_offsets_y[k]*sl, z[i]);
```
wdconinc, line 106: Since DD4hep is already included and since Eigen is compilation-wise expensive, I'd suggest using DD4hep for coordinate transformations.

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -0,0 +1,68 @@
...
+  private:
+
+
+
+    const int SUBCELLS=12;
+//positions where the overlapping cells are relative to a given cell (in units of hexagon side length)
+    double neighbor_offsets_x[12]={1.5*cos(0), 1.5*cos(M_PI/3), 1.5*cos(2*M_PI/3),1.5*cos(3*M_PI/3), 1.5*cos(4*M_PI/3), 1.5*cos(5*M_PI/3),
+                             -sqrt(3)/2.*sin(0),-sqrt(3)/2.*sin(M_PI/3),-sqrt(3)/2.*sin(2*M_PI/3),-sqrt(3)/2.*sin(3*M_PI/3),-sqrt(3)/2.*sin(4*M_PI/3),-sqrt(3)/2.*sin(5*M_PI/3)};
+    double neighbor_offsets_y[12]={1.5*sin(0), 1.5*sin(M_PI/3), 1.5*sin(2*M_PI/3),1.5*sin(3*M_PI/3), 1.5*sin(4*M_PI/3), 1.5*sin(5*M_PI/3),
+                              sqrt(3)/2.*cos(0), sqrt(3)/2.*cos(M_PI/3), sqrt(3)/2.*cos(2*M_PI/3), sqrt(3)/2.*cos(3*M_PI/3), sqrt(3)/2.*cos(4*M_PI/3), sqrt(3)/2.*cos(5*M_PI/3)};
```
wdconinc, line 50: Fill these in the constructor or with a lambda. Declare as static const.

src/algorithms/calorimetry/HEXPLITConfig.h
```diff
@@ -0,0 +1,24 @@
...
+    double                   MIP{0.000472};
+    double                   Emin_in_MIPs{0.1};
+    double                   tmax{325*dd4hep::ns};
+    //conversion from local to global coordinates.  translate first, then rotate
+    double                   trans_x{0};
+    double                   trans_y{0};
+    double                   trans_z{0};
+    double                   rot_x{0};
+    double                   rot_y{0};
+    double                   rot_z{0};
```
wdconinc, line 21: Define as vectors.

src/detectors/ZDC/ZDC.cc
```diff
@@ -102,5 +106,150 @@ extern "C" {
...
+          "ZDCSubcellHits", {"ZDCRecHits"}, {"ZDCSubcellHits"},
+          {
+            .layer_spacing=25.1*dd4hep::mm,
+            .side_length = side_length,
+            .MIP = 472. * dd4hep::keV,
+            .Emin_in_MIPs=0.1,
+            .tmax=320 * dd4hep::ns,
+            .trans_x=0,
+            .trans_y=0,
+            .trans_z=36601 * dd4hep::mm,
```
wdconinc, line 155: Absolute detector positions should not be in reconstruction configuration.

src/detectors/ZDC/ZDC.cc
```diff
@@ -102,5 +106,150 @@ extern "C" {
...
+            .rot_x=0,
+            .rot_y=-0.025,
```
wdconinc, line 157: Crossing angle should not be in reconstruction configuration.

src/detectors/ZDC/ZDC.cc
```diff
@@ -102,5 +106,150 @@ extern "C" {
             app   // TODO: Remove me once fixed
           )
         );
+
+        //side length of hexagonal cells in SiPM-on-tile part of the ZDC
+        auto side_length = 31.3 * dd4hep::mm;
+
+        app->Add(new JOmniFactoryGeneratorT<CalorimeterHitDigi_factory>(
+          "ZDCRawHits", {"HcalFarForwardZDCHits"}, {"ZDCRawHits"},
```
wdconinc, line 114: Avoid starting collection with Hcal when everything else starts with ZDC.

sebouh137:
> I'd already sent to this @Gongzi-Zhang, but looking at https://github.com/eic/EICrecon/blob/sipmzdc/src/algorithms/calorimetry/LogWeightReco.cc that algorithm is already available as https://github.com/eic/EICrecon/blob/sipmzdc/src/algorithms/calorimetry/CalorimeterHitsMerger.h and would be good to reuse.

@wdconinc  The LogWeightReco takes the hits in a cluster that are from multiple cells (or subcells, if the input to this algorithm comes from HEXPLIT) to reconstruct the position of the center of a shower.  From reading the source code of CalorimeterHitsMergers, it appears to only merge together hits that have the same cell id.  These two algorithms don't do the same thing as one another.

src/detectors/ZDC/ZDC.cc
```diff
@@ -102,5 +106,150 @@ extern "C" {
...
+            .layerField = "layer",
+            .sectorField = "system",
+          },
+          app   // TODO: Remove me once fixed
+        ));
+
+        app->Add(new JOmniFactoryGeneratorT<HEXPLIT_factory>(
+          "ZDCSubcellHits", {"ZDCRecHits"}, {"ZDCSubcellHits"},
+          {
+            .layer_spacing=25.1*dd4hep::mm,
```
wdconinc, line 148: Geometry information should come from geometry, not a parameter in reconstruction.

wdconinc:
> From reading the source code of CalorimeterHitsMergers,...

Sorry, I linked the wrong file :-) I meant https://github.com/eic/EICrecon/blob/sipmzdc/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h which does log-energy weighted cluster centroid determination. All that's needed, I think, may be adding the quadratic term, defaulting to zero for backwards compatibility several other calorimeters use it with just two parameters.

sebouh137:
> > From reading the source code of CalorimeterHitsMergers,...
> 
> Sorry, I linked the wrong file :-) I meant https://github.com/eic/EICrecon/blob/sipmzdc/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h which does log-energy weighted cluster centroid determination. All that's needed, I think, may be adding the quadratic term, defaulting to zero for backwards compatibility several other calorimeters use it with just two parameters.

thanks @wdconinc , I'll look into this and the other things you've posted in here.

src/detectors/ZDC/ZDC.cc
sebouh137: When adding the SiPM-on-tile ZDC to the epic repository (for the geometry), the person reviewing our code (I forget who it was) told us to use this particular name for the hits in this detector.  I can change this back to "ZDCHIts" or "ZDCHcalHits" (which I think sounds more reasonable) if I can get it changed in the epic repository as well, but I think that should be a separate pull request

veprbl:
I'm a bit concerned that the new clustering algorithms largely reproduce the existing algorithms. Ideally we should have common facilities for cell adjacency and for cluster hit combination (CalorimeterClusterRecoCoG), whereas you are adding your own.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -234,7 +234,7 @@ void CalorimeterHitReco::process(
         std::vector<double> cdim;
         // get segmentation dimensions
         auto segmentation_type = m_converter->findReadout(local).segmentation().type();
-        if (segmentation_type == "CartesianGridXY") {
+        if (segmentation_type == "CartesianGridXY" || segmentation_type == "HexGridXY") {
```
wdconinc, line 237: Can we use typeid for this instead of hit-by-hit string comparisons that differ only in the final two letters for some cases? E.g. without the `HexGridXY` yet this looks like: 
```diff
diff --git a/src/algorithms/calorimetry/CalorimeterHitReco.cc b/src/algorithms/calorimetry/CalorimeterHitReco.cc
index 9214b319..7c313c18 100644
--- a/src/algorithms/calorimetry/CalorimeterHitReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitReco.cc
@@ -17,6 +17,9 @@
 #include <DD4hep/Volumes.h>
 #include <DD4hep/config.h>
 #include <DDSegmentation/BitFieldCoder.h>
+#include <DDSegmentation/CartesianGridXY.h>
+#include <DDSegmentation/NoSegmentation.h>
+#include <Evaluator/DD4hepUnits.h>
 #include <Evaluator/DD4hepUnits.h>
 #include <Math/GenVector/Cartesian3D.h>
 #include <Math/GenVector/DisplacementVector3D.h>
@@ -233,16 +236,16 @@ void CalorimeterHitReco::process(
         const auto pos = local.nominal().worldToLocal(gpos);
         std::vector<double> cdim;
         // get segmentation dimensions
-        auto segmentation_type = m_converter->findReadout(local).segmentation().type();
-        if (segmentation_type == "CartesianGridXY") {
+        auto& segmentation_type = typeid(m_converter->findReadout(local).segmentation());
+        if (segmentation_type == typeid(dd4hep::DDSegmentation::CartesianGridXY)) {
             auto cell_dim = m_converter->cellDimensions(cellID);
             cdim.resize(3);
             cdim[0] = cell_dim[0];
             cdim[1] = cell_dim[1];
             m_log->debug("Using segmentation for cell dimensions: {}", fmt::join(cdim, ", "));
         } else {
-            if ((segmentation_type != "NoSegmentation") && (!warned_unsupported_segmentation)) {
-                m_log->warn("Unsupported segmentation type \"{}\"", segmentation_type);
+            if ((segmentation_type != typeid(dd4hep::DDSegmentation::NoSegmentation)) && (!warned_unsupported_segmentation)) {
+                m_log->warn("Unsupported segmentation type \"{}\"", segmentation_type.name());
                 warned_unsupported_segmentation = true;
             }
 
```

src/algorithms/calorimetry/HEXPLIT.cc
sebouh137: Fixed in commit 125559e.

src/algorithms/calorimetry/HEXPLIT.cc
sebouh137: fixed in commit 125559e.

src/algorithms/calorimetry/HEXPLIT.cc
sebouh137: fixed in commit 125559e.

src/algorithms/calorimetry/HEXPLIT.cc
sebouh137: fixed in commit 125559e.

src/detectors/ZDC/ZDC.cc
sebouh137: fixed in commit 22dfc61

src/detectors/ZDC/ZDC.cc
sebouh137: fixed in commit 22dfc61

src/algorithms/calorimetry/HEXPLITConfig.h
sebouh137: outdated, as of commit 22dfc61

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,144 @@
...
+  double E[nhits];
+  double t[nhits];
+
+  for(int i=0; i<nhits; i++){
+    x[i]=hits[i].getLocal().x;
+    y[i]=hits[i].getLocal().y;
+    z[i]=hits[i].getLocal().z;
+    E[i]=hits[i].getEnergy();
+    t[i]=hits[i].getTime();
+    }*/
```
wdconinc, line 59: Please remove all commented blocks.
```suggestion
```

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,144 @@
...
+  auto volman = m_detector->volumeManager();
+  auto subcellHits = std::make_unique<edm4eic::CalorimeterHitCollection>();
+  double Esum=0;
+  for(auto hit : hits){
```
wdconinc, line 63: Please use const references where possible:
```suggestion
  for(const auto& hit : hits){
```

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -0,0 +1,68 @@
...
+
+    //indices of the neighboring cells which overlap to produce a given subcell
+    int neighbor_indices[12][3]={{0, 11,10}, {1, 6, 11},{2, 7, 6}, {3,8,7}, {4,9,8}, {5,10,9},
+                             {6, 11, 7}, {7, 6, 8}, {8, 7, 9}, {9,8,10},{10,9,11},{11,10,6}};
```
wdconinc, line 54: I can see the gist of @veprbl's comment here: this is all implemented in the CalorimeterIslandCluster algorithms where it uses a function to determine adjacency (i.e. which blocks are neighbors). Instead of implementing this here as a new HEXPLIT class, it makes sense to add this in CalorimeterIslandCluster since the changes are actually quite minimal.

src/algorithms/calorimetry/HEXPLIT.cc
sebouh137: fixed, thanks!

src/algorithms/calorimetry/HEXPLIT.cc
sebouh137: fixed, thanks!

src/algorithms/calorimetry/HEXPLIT.cc
sebouh137: done

src/algorithms/calorimetry/CalorimeterHitReco.cc
sebouh137: implemented your proposed changes. thanks.

src/algorithms/calorimetry/HEXPLIT.h
sebouh137: HEXPLIT doesn't just determine if hits are adjacent (a binary condition); it also checks which of twelve different neighboring positions relative to the first hit a second hit is in, and then stores the energies of the neighbors in those 12 different positions in a length-12 array, so that this information can be used to estimate the energy in each of the 12 subcells of the original cell (this is not a one-to-one function, each subcell is the overlap of three neighbor cells.).  So, it's not just determining adjacency, but rather classifying how they are adjacent to one another.

src/detectors/ZDC/ZDC.cc
sebouh137: @wdconinc can you point me to somewhere in the code where I can obtain a named parameter?   I assume that I can have it parse through the detector description xml to get that information; is that correct?

wdconinc: Yes, there are some examples at:
https://github.com/eic/EICrecon/blob/0a27c9a09b52f02e9b6888e9d75003d45243ce6c/src/services/geometry/richgeo/ReadoutGeo.cc#L50-L55

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,130 @@
...
+    //keep track of the energy in each neighboring cell
+    std::vector<double> Eneighbors(SUBCELLS, 0.0);
+
+    for (const auto& other_hit : hits){
+      double tol=0.01; //tolerance for rounding errors
+
+      //only look at hits nearby within two layers of the current layer
+      double dz=abs(hit.getLocal().z-other_hit.getLocal().z);
+      if (dz>2.5*layer_spacing || dz<tol)
+        continue;
```
wdconinc, line 63: If the only reason you use layer_spacing is to check that these are neighbors, then you should probably use the CellID fields. Again, see @veprbl's comments about how this is actually not so different from what other clustering algorithms are doing.

sebouh137: In the most recent commit, I changed this so that it uses the getLayer() function to determine which hits are on nearby layers (ie, if the difference in layer number is -2, -1, 1, or 2).  Doing the same for the transverse position is a little more tricky due to the way the row and column numbers works for staggered geometries, but in principle this is possible.

wdconinc: > Doing the same for the transverse position is a little more tricky

The LFHCAL had a similar issue due to how the stacking is done, I recall.

src/algorithms/calorimetry/LogWeightReco.cc
```diff
@@ -0,0 +1,92 @@
+// Copyright (C) 2023 Sebouh Paul
```
veprbl, line 1: Can we remove this algorithm and add functionality to implement second order terms in the CalorimeterClusterRecoCoG?

src/detectors/ZDC/ZDC.cc
veprbl: If anything, we should rename all output to HcalFarForwardZDC*, EcalFarFowardZDC*

veprbl: (keep ddh4hep inputs starting with ZDC for now)

src/algorithms/calorimetry/LogWeightReco.cc
sebouh137: As per your suggestion, I just removed the LogWeightReco and implemented the second order terms in CalorimeterClusterRecoCoG.

sebouh137: removed!

src/detectors/ZDC/ZDC.cc
sebouh137: thanks.  I have gotten this to work without ever using layer_spacing (by using layer number comparisons) nor side_length (which I have instead used the cell dimensions of the hit to extract).

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
```diff
@@ -27,6 +27,10 @@ class CalorimeterClusterRecoCoG_factory : public JOmniFactory<CalorimeterCluster
     ParameterRef<std::string> m_energyWeight {this, "energyWeight", config().energyWeight};
     ParameterRef<double> m_samplingFraction {this, "samplingFraction", config().sampFrac};
     ParameterRef<double> m_logWeightBase {this, "logWeightBase", config().logWeightBase};
+    ParameterRef<bool> m_variableLogWeightBase {this, "variableLogWeightBase", config().variableLogWeightBase};
```
veprbl, line 30:
```suggestion
```

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -0,0 +1,63 @@
...
+  class HEXPLIT : public WithPodConfig<HEXPLITConfig> {
+
+  public:
+    void init(const dd4hep::Detector* detector, std::shared_ptr<spdlog::logger>& logger);
+    std::unique_ptr<edm4eic::CalorimeterHitCollection> process(const edm4eic::CalorimeterHitCollection &hits) ;
+
+  private:
+
+
+
```
veprbl, line 44:
```suggestion

```

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -0,0 +1,63 @@
...
+#include "algorithms/interfaces/WithPodConfig.h"
+#include "HEXPLITConfig.h"
+
+namespace eicrecon {
+
+  class HEXPLIT : public WithPodConfig<HEXPLITConfig> {
+
+  public:
+    void init(const dd4hep::Detector* detector, std::shared_ptr<spdlog::logger>& logger);
+    std::unique_ptr<edm4eic::CalorimeterHitCollection> process(const edm4eic::CalorimeterHitCollection &hits) ;
```
veprbl, line 39: Please setup your input/output types to use algorithms:: interface (see https://github.com/eic/EICrecon/pull/1238 for a concise example).

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -17,6 +17,9 @@
 #include <DD4hep/Volumes.h>
 #include <DD4hep/config.h>
 #include <DDSegmentation/BitFieldCoder.h>
+#include <DDSegmentation/CartesianGridXY.h>
+#include <DDSegmentation/HexGrid.h>
+#include <DDSegmentation/NoSegmentation.h>
```
veprbl, line 22:
```suggestion
```
Not needed?

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,148 @@
...
+#include <stdexcept>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
+#include <iostream>
+
+#include "HEXPLIT.h"
+
+#include "algorithms/calorimetry/HEXPLITConfig.h"
```
veprbl, line 26:
```suggestion
#include <DD4hep/Alignments.h>
#include <DD4hep/DetElement.h>
#include <DD4hep/Objects.h>
#include <DD4hep/VolumeManager.h>
#include <Evaluator/DD4hepUnits.h>
#include <Math/GenVector/Cartesian3D.h>
#include <Math/GenVector/DisplacementVector3D.h>
#include <edm4hep/Vector3f.h>
#include <stdlib.h>
#include <algorithm>
#include <cmath>
 #include <vector>
 
 #include "HEXPLIT.h"
 #include "algorithms/calorimetry/HEXPLITConfig.h"
```

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -0,0 +1,63 @@
...
+#include <cmath>
+#include <cstddef>
+#include <exception>
+#include <functional>
+#include <memory>
+#include <set>
+#include <vector>
+
+#include "algorithms/interfaces/WithPodConfig.h"
+#include "HEXPLITConfig.h"
```
veprbl, line 31:
```suggestion
#pragma once

#include <algorithms/algorithm.h>
#include <DD4hep/Detector.h>
#include <edm4eic/CalorimeterHitCollection.h>
#include <spdlog/logger.h>
#include <memory>
 
#include "HEXPLITConfig.h"
#include "algorithms/interfaces/WithPodConfig.h"
```

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -234,7 +237,8 @@ void CalorimeterHitReco::process(
         std::vector<double> cdim;
         // get segmentation dimensions
         auto segmentation_type = m_converter->findReadout(local).segmentation().type();
-        if (segmentation_type == "CartesianGridXY") {
+        auto& segmentation_typeid = typeid(m_converter->findReadout(local).segmentation());
```
veprbl, line 240:
```suggestion
```
Not used?

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023 Sebouh Paul
+
+// An algorithm for splitting calorimeter hits in overlapping cells into "subhits" based on the relative
```
veprbl, line 4: Since this is a new algorithm, it would be nice if you could contribute a test for it, see
https://github.com/eic/EICrecon/blob/main/src/tests/algorithms_test/calorimetry_CalorimeterIslandCluster.cc
for an example.

wdconinc:
Don't worry about the `Eigen/Householder` include (only that one). It seems that IWYU has trouble with it. Sometimes it just spends 6 hours going back and forth until the job times out or someone cancels it: https://github.com/eic/EICrecon/actions/runs/7532222215/job/20503403799?pr=1234.

sebouh137:
> Don't worry about the `Eigen/Householder` include (only that one). It seems that IWYU has trouble with it. Sometimes it just spends 6 hours going back and forth until the job times out or someone cancels it: https://github.com/eic/EICrecon/actions/runs/7532222215/job/20503403799?pr=1234.

@wdconinc ,  In this case, there is only one other check that my pull request fails.  Namely, getting the deployment of the docs.  I don't see anything in the logs that is specific to any of the files I have added or changed.   Can this pull request be approved without passing the deploy-docs? 

image

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -0,0 +1,64 @@
...
+             : HEXPLITAlgorithm{name,
+                                   {"inputHits"},
+                                   {"outputSubcellHits"},
+                                   "Split hits into subcell hits"} {}
+
+    void init(const dd4hep::Detector* detector, std::shared_ptr<spdlog::logger>& logger);
+    void process(const Input&, const Output&) const final;
+
+  private:
+      const int SUBCELLS=12;
```
wdconinc, line 48: After defining this suitably, you shouldn't need to repeat it everywhere...

src/algorithms/calorimetry/CalorimeterClusterRecoCoGConfig.h
```diff
@@ -14,6 +14,13 @@ namespace eicrecon {
         double sampFrac = 1.;
         double logWeightBase = 3.6;
 
+        //optional:  have the log weight base depend on the energy
+        // logWeightBase+logWeightBase_lin*l+logWeightBase_quad*l*l
+        // where l = log(cl.getEnergy()/logWeightBase_Eref)
+        double logWeightBase_lin=0;
+        double logWeightBase_quad=0;
+        double logWeightBase_Eref=50;
+
```
wdconinc, line 23: How about `std::vector<double> logWeightCoeffs{3.6}` with index = power of `l`? Fewer variables and combines the info that logically goes together.

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -0,0 +1,153 @@
...
+#include <vector>
+#include "HEXPLIT.h"
+#include "algorithms/calorimetry/HEXPLITConfig.h"
+
+namespace eicrecon {
+
+const double HEXPLIT::neighbor_offsets_x[NEIGHBORS]={1.5*cos(0), 1.5*cos(M_PI/3), 1.5*cos(2*M_PI/3),1.5*cos(3*M_PI/3), 1.5*cos(4*M_PI/3), 1.5*cos(5*M_PI/3),
+    -sqrt(3)/2.*sin(0),-sqrt(3)/2.*sin(M_PI/3),-sqrt(3)/2.*sin(2*M_PI/3),-sqrt(3)/2.*sin(3*M_PI/3),-sqrt(3)/2.*sin(4*M_PI/3),-sqrt(3)/2.*sin(5*M_PI/3)};
+const double HEXPLIT::neighbor_offsets_y[NEIGHBORS]={1.5*sin(0), 1.5*sin(M_PI/3), 1.5*sin(2*M_PI/3),1.5*sin(3*M_PI/3), 1.5*sin(4*M_PI/3), 1.5*sin(5*M_PI/3),
+                          sqrt(3)/2.*cos(0), sqrt(3)/2.*cos(M_PI/3), sqrt(3)/2.*cos(2*M_PI/3), sqrt(3)/2.*cos(3*M_PI/3), sqrt(3)/2.*cos(4*M_PI/3), sqrt(3)/2.*cos(5*M_PI/3)};
```
wdconinc, line 28: You could fill these programmatically with a lambda. Something like:
```suggestion
const double HEXPLIT::neighbor_offsets_x[NEIGHBORS] =[](){
  std::vector<double> x;
  std::vector<double> rs{1.5, -sqrt(3)/2.};
  for (auto& r : rs)
    for (theta = 0; theta < 2 * M_PI; theta += M_PI/3)
      x.push_back(r*cos(theta));
  return x;
}();
```

veprbl:
> > Don't worry about the `Eigen/Householder` include (only that one). It seems that IWYU has trouble with it. Sometimes it just spends 6 hours going back and forth until the job times out or someone cancels it: https://github.com/eic/EICrecon/actions/runs/7532222215/job/20503403799?pr=1234.
> 
> @wdconinc , In this case, there is only one other check that my pull request fails. Namely, getting the deployment of the docs. I don't see anything in the logs that is specific to any of the files I have added or changed. Can this pull request be approved without passing the deploy-docs?


Please ignore that one. I'm working to resolve it in #1245

src/detectors/ZDC/ZDC.cc
```diff
@@ -88,14 +91,154 @@ extern "C" {
...
+            // .transverseEnergyProfileScale = 1.,
+          },
+          app
+        ));
+
+        app->Add(new JOmniFactoryGeneratorT<CalorimeterClusterRecoCoG_factory>(
+           "HcalFarForwardZDC_HEXPLITClusters",
+          {"HcalFarForwardZDCIslandClusterContributions",  // edm4eic::ProtoClusterCollection
+           "HcalFarForwardZDCHits"},                // edm4hep::SimCalorimeterHitCollection
+          {"HcalFarForwardZDC_HEXPLITClusters",             // edm4eic::Cluster
```
veprbl, line 185: If HEXPLIT is better, we use it by default - please, call it just "Clusters".

src/detectors/ZDC/ZDC.cc
```diff
@@ -88,14 +91,154 @@ extern "C" {
...
+                .minClusterHitEdep = 100.0 * dd4hep::keV,
+                .minClusterCenterEdep = 11.0 * dd4hep::MeV,
+                .minClusterEdep = 11.0 * dd4hep::MeV,
+                .minClusterNhits = 10,
+            },
+            app   // TODO: Remove me once fixed
+        ));
+
+        app->Add(new JOmniFactoryGeneratorT<CalorimeterIslandCluster_factory>(
+          "HcalFarForwardZDCIslandClusterContributions", {"HcalFarForwardZDCSubcellHits"}, {"HcalFarForwardZDCIslandClusterContributions"},
```
veprbl, line 168:
```suggestion
          "HcalFarForwardZDCIslandProtoClusters",
          {"HcalFarForwardZDCSubcellHits"},
          {"HcalFarForwardZDCIslandProtoClusters"},
```
IIRC that's the naming convention - `IslandProtoClusters`

src/detectors/ZDC/ZDC.cc
```diff
@@ -88,14 +91,154 @@ extern "C" {
...
+            app   // TODO: Remove me once fixed
+          )
+        );
+
+        app->Add(new JOmniFactoryGeneratorT<CalorimeterClusterRecoCoG_factory>(
+             "HcalFarForwardZDCClusters",
+            {"HcalFarForwardZDCIslandProtoClusters",  // edm4eic::ProtoClusterCollection
+             "HcalFarForwardZDCHits"},                // edm4hep::SimCalorimeterHitCollection
+            {"HcalFarForwardZDCClusters",             // edm4eic::Cluster
+             "HcalFarForwardZDCClusterAssociations"}, // edm4eic::MCRecoClusterParticleAssociation
```
veprbl, line 238:
```suggestion
             "HcalFarForwardZDCNoHEXPLITClusters",
            {"HcalFarForwardZDCIslandProtoClusters",  // edm4eic::ProtoClusterCollection
             "HcalFarForwardZDCHits"},                // edm4hep::SimCalorimeterHitCollection
            {"HcalFarForwardZDCNoHEXPLITClusters",             // edm4eic::Cluster
             "HcalFarForwardZDCNoHEXPLITClusterAssociations"}, // edm4eic::MCRecoClusterParticleAssociation
```

src/detectors/ZDC/ZDC.cc
```diff
@@ -88,14 +91,154 @@ extern "C" {
...
+            .MIP = 472. * dd4hep::keV,
+            .Emin_in_MIPs=0.1,
+            .tmax=320 * dd4hep::ns,
+          },
+          app   // TODO: Remove me once fixed
+        ));
+
+        double side_length=31.3 * dd4hep::mm;
+        app->Add(new JOmniFactoryGeneratorT<ImagingTopoCluster_factory>(
+            "HcalFarForwardZDCImagingClusterContributions", {"HcalFarForwardZDCSubcellHits"}, {"HcalFarForwardZDCImagingClusterContributions"},
```
veprbl, line 153:
```suggestion
            "HcalFarForwardZDCTopoClusterContributions",
            {"HcalFarForwardZDCSubcellHits"},
            {"HcalFarForwardZDCTopoClusterContributions"},
```
Is this even used anywhere? Perhaps, also, `Contributions` -> `ProtoClusters`?

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -176,19 +176,28 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
...
+            "EcalFarForwardZDCRecHits",
+            "EcalFarForwardZDCClusters",
+            "EcalFarForwardZDCClusterAssociations",
+            "HcalFarForwardZDCRawHits",
+            "HcalFarForwardZDCRecHits",
+            "HcalFarForwardZDCClusters",
+            "HcalFarForwardZDCSubcellHits",
+            "HcalFarForwardZDC_HEXPLITClusters",
+            "HcalFarForwardZDCLogWeightClusters",
+            "HcalFarForwardZDCClusterAssociations",
```
veprbl, line 189: Duplicate entry.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -176,19 +176,28 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
...
             "HcalBarrelTruthClusterAssociations",
             "B0ECalRecHits",
             "B0ECalClusters",
             "B0ECalClusterAssociations",
-            "ZDCEcalTruthClusters",
-            "ZDCEcalTruthClusterAssociations",
+            "EcalFarForwardZDCTruthClusters",
+            "EcalFarForwardZDCTruthClusterAssociations",
+            "HcalFarForwardZDCTruthClusters",
+            "HcalFarForwardZDCTruthClusterAssociations",
```
veprbl, line 200: Please move up, for them to be near their respective subsystems.

veprbl (commented):
Need to sort out the naming convention.

sebouh137:
> > > Don't worry about the `Eigen/Householder` include (only that one). It seems that IWYU has trouble with it. Sometimes it just spends 6 hours going back and forth until the job times out or someone cancels it: https://github.com/eic/EICrecon/actions/runs/7532222215/job/20503403799?pr=1234.
> > 
> > 
> > @wdconinc , In this case, there is only one other check that my pull request fails. Namely, getting the deployment of the docs. I don't see anything in the logs that is specific to any of the files I have added or changed. Can this pull request be approved without passing the deploy-docs?
> 
> Please ignore that one. I'm working to resolve it in #1245

In that case, all of the conversations have been resolved, and all of the other checks have been fulfilled.  @wdconinc @veprbl or @nathanwbrei , could you please approve of this pull request?  
Thanks,
Sebouh

nathanwbrei (approved):
Looks good! Since it's been through so many rounds of review at this point I'm just going to approve it

Issue #1223, opened 2024-01-13
No Calorimeter Cluster-Hit Associations in PODIO Output


ruse-traveler:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? No
- Any special versions of ROOT or Geant4? No

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Open a PODIO output file
2. And check the contents of the `_*cal*Clusters_hits` branches or `*cal*Clusters.hits_{begin,end}` leaves.

### Expected Result: (what do you expect when you execute the steps above)

The `*_hits` branches are expected to have something in them, and the `*hits_begin` or `*hits_end` leaves are expected to have values other than 0.

### Actual Result: (what do you get when you execute the steps above)

The `*_hits` branches are empty, and the `*hits_begin` or `*hits_end` leaves only ever have 0.

### Additional Context

Being able to access the hits associated with a cluster from the PODIO output will be extremely useful for visualizing clusters, which will be very useful in debugging our clustering and particle flow algorithms.

PR #1224, opened 2024-01-13
CalorimeterClusterRecoCoG: copy hits references from ProtoCluster

veprbl:
Resolves: #1223

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -175,6 +175,7 @@ std::optional<edm4eic::Cluster> CalorimeterClusterRecoCoG::reconstruct(const edm
     m_log->debug("hit energy = {} hit weight: {}", hit.getEnergy(), weight);
     auto energy = hit.getEnergy() * weight;
     totalE += energy;
+    cl.addToHits(hit);
     if (energy > maxE) {
     }
     const float eta = edm4hep::utils::eta(hit.getPosition());
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -163,7 +163,6 @@ std::optional<edm4eic::Cluster> CalorimeterClusterRecoCoG::reconstruct(const edm
 
   // calculate total energy, find the cell with the maximum energy deposit
   float totalE = 0.;
-  float maxE   = 0.;
   // Used to optionally constrain the cluster eta to those of the contributing hits
   float minHitEta = std::numeric_limits<float>::max();
   float maxHitEta = std::numeric_limits<float>::min();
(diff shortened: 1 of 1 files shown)


src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -175,6 +175,7 @@ std::optional<edm4eic::Cluster> CalorimeterClusterRecoCoG::reconstruct(const edm
     m_log->debug("hit energy = {} hit weight: {}", hit.getEnergy(), weight);
     auto energy = hit.getEnergy() * weight;
     totalE += energy;
+    cl.addToHits(hit);
     if (energy > maxE) {
     }
```
wdconinc, line 180: OT but what's this doing here???

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -175,6 +175,7 @@ std::optional<edm4eic::Cluster> CalorimeterClusterRecoCoG::reconstruct(const edm
     m_log->debug("hit energy = {} hit weight: {}", hit.getEnergy(), weight);
     auto energy = hit.getEnergy() * weight;
     totalE += energy;
+    cl.addToHits(hit);
```
wdconinc, line 178: We also have `hitContributions` in [`edm4eic::Cluster`](https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L258). Could we fill that with `energy` here too?

wdconinc: An interesting question is also whether it makes more sense to fill the `energy` or the weight that goes into the mean position determination... We can't really store both. Are they even different?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
wdconinc: Dropped by @sly2j in https://eicweb.phy.anl.gov/EIC/juggler/-/commit/fabb1ece72b00fbabd8062bf7b775344670abb49#8c0365ec766abf3cb50e95d734f31066c8853df3_173_171 (https://eicweb.phy.anl.gov/EIC/juggler/-/merge_requests/392). This might actually have been a mistake that snuck in via an unrelated MR.

wdconinc: Ok, so this originates in using the block with maximum energy deposition in the cluster to define the `time` of the cluster. At some point (in this MR above), the maximum energy selection was removed, and we ended up writing the `time` based on the first hit in the protocluster. So, right now (and since then) the cluster time is not time of the maximum energy hit anymore.

wdconinc: Deferred to https://github.com/eic/EICrecon/issues/1225.

wdconinc (approved):
Diffs only in hits and hitContributions.

ruse-traveler:
🙌 Very nice!! 🙌

ruse-traveler (approved):

Issue #742, opened 2023-07-05
`eicrecon::JFactoryPodioT::Set` leaks memory


veprbl:
https://github.com/eic/EICrecon/blob/7596c6507e3ba76a0dc6c8620ad4b7457913d848/src/services/io/podio/JFactoryPodioT.h#L189-L197
copies objects and allows to leak memory (violates guarantees given by the `Set` method in JANA2). This causes actual leaks in the calorimetry factories.

See also #731

PR #1227, opened 2024-01-13
rm src/services/io/podio/JFactoryPodioT.h

veprbl:
Resolves: #837


wdconinc:
Also resolves #742.

wdconinc (approved):
If it passes all checks, then sure. Is there a specific jana2 version required for this to work?

veprbl:
This is just unused. Version from JANA replaces this already.

PR #1241, opened 2024-01-20
feat: forward support edm4eic version 5

wdconinc:
This adds forward support for edm4eic version 5, which includes the changes in https://github.com/eic/EDM4eic/pull/58.

- [ ] Bug fix (issue #__)
- [x] New feature (issue, follow-up on https://github.com/eic/EDM4eic/pull/58)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, some fields not written as before since CKFTracking does not yet write edm4eic::Track collection.

### DIFF
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -347,8 +347,10 @@ namespace eicrecon {
 
             // Create trajectory
             auto trajectory = trajectories->create();
-            trajectory.setChi2(trajectoryState.chi2Sum);
-            trajectory.setNdf(trajectoryState.NDF);
+            #if EDM4EIC_VERSION_MAJOR < 5
+              trajectory.setChi2(trajectoryState.chi2Sum);
+              trajectory.setNdf(trajectoryState.NDF);
+            #endif
             trajectory.setNMeasurements(trajectoryState.nMeasurements);
             trajectory.setNStates(trajectoryState.nStates);
             trajectory.setNOutliers(trajectoryState.nOutliers);
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -413,7 +413,6 @@ namespace eicrecon {
               pars.setTimeError(sqrt(static_cast<float>(covariance(Acts::eBoundTime, Acts::eBoundTime))));
             #endif
 
-            track_parameters->push_back(pars);
             trajectory.addToTrackParameters(pars);
 
             // save measurement2d to good measurements or outliers according to srclink index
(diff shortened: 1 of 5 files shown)


src/algorithms/tracking/CKFTracking.cc
```diff
@@ -342,31 +376,42 @@ namespace eicrecon {
...
+              edm4eic::Cov6f cov;
+              for (size_t i = 0; const auto& [a, x] : edm4eic_indexed_units) {
+                for (size_t j = 0; const auto& [b, y] : edm4eic_indexed_units) {
+                  // FIXME why not pars.getCovariance()(i,j) = covariance(a,b) / x / y;
+                  cov(i,j) = covariance(a,b) / x / y;
+                }
+              }
+              pars.setCovariance(cov);
+            #else
+              pars.setCharge(static_cast<float>(boundParam.charge())});
```
wdconinc, line 399:
```suggestion
              pars.setCharge(static_cast<float>(boundParam.charge()));
```

src/algorithms/tracking/CKFTracking.cc
```diff
@@ -342,31 +376,42 @@ namespace eicrecon {
...
                     static_cast<float>(covariance(Acts::eBoundPhi, Acts::eBoundQOverP))
-                },
-                static_cast<float>(parameter[Acts::eBoundTime]),
-                sqrt(static_cast<float>(covariance(Acts::eBoundTime, Acts::eBoundTime))),
-                static_cast<float>(boundParam.charge())};
+                });
+              pars.setTimeError(sqrt(static_cast<float>(covariance(Acts::eBoundTime, Acts::eBoundTime))));
+            #endif
 
             track_parameters->push_back(pars);
```
wdconinc, line 416: If we start from `create()`, we don't need to `push_back()`.
```suggestion
```

src/algorithms/tracking/CKFTracking.cc
```diff
@@ -392,14 +436,22 @@ namespace eicrecon {
                     } else {
                         auto meas2D = meas2Ds[srclink_index];
                         if (typeFlags.test(Acts::TrackStateFlag::MeasurementFlag)) {
+                          #if EDM4EIC_VERSION_MAJOR >= 5
+                            //track.addToMeasurementHits(meas2D);
```
osbornjd, line 440: Don't we still need this in edm4eic v5? Or am I missing something:
https://github.com/eic/EDM4eic/blob/773bf2628711dfed082b12416f9179a0ddf3a67e/edm4eic.yaml#L416

osbornjd (changes requested):

src/algorithms/tracking/CKFTracking.cc
wdconinc: We don't yet fill a edm4eic::Track in CKFTracking.

osbornjd: Ah, you're right I missed that. So I guess that is being left to a future PR then

wdconinc: Yeah. I think that's best.

osbornjd (approved):

PR #1242, opened 2024-01-20
fix: use track direction in localToGlobal in TrackProjector

wdconinc:
While going through this class for other reasons, I noticed the arbitrary direction `{0,0,0}` when we can provide the actual direction. I don't expect any difference due to this, but as we have seen there can be upstream changes that make it such that arbitrarily chosen directions suddenly start to matter. If the interface specifies a direction, then Acts is free to start expecting the direction to make sense and to start producing results that depend on it.

- [x] Bug fix (issue: TrackProjector uses zero-vector as direction in globalToLocal)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @ShujieL also occurs in TrackerMeasurementFromHits in globalToLocal

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackProjector.cc
+++ b/src/algorithms/tracking/TrackProjector.cc
@@ -4,6 +4,7 @@
 #include <Acts/Definitions/TrackParametrization.hpp>
 #include <Acts/EventData/MultiTrajectoryHelpers.hpp>
 #include <Acts/Geometry/GeometryIdentifier.hpp>
+#include <Acts/Utilities/UnitVectors.hpp>
 #include <ActsExamples/EventData/Trajectories.hpp>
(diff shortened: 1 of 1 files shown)


wdconinc:
No diffs, e.g. https://github.com/eic/EICrecon/actions/runs/7596423169/job/20690441799?pr=1242#step:15:82.

bschmookler (approved):
Looks good to me.

I agree this shouldn't make a difference right now, since the ```direction``` seems to only come into play for the ```linesurface``` right now, as detailed here: [https://acts.readthedocs.io/en/latest/core/geometry/surfaces.html](https://acts.readthedocs.io/en/latest/core/geometry/surfaces.html).

PR #1243, opened 2024-01-20
linux-eic-shell.yml: delay posting capybara comment until pages are deployed

veprbl:
This prevents comments from being posted before the links are alive.


wdconinc (approved):
Thanks for fixing this. This has been an annoying issue since you don't actually know when the links do become valid!

PR #1244, opened 2024-01-20
fix: explicitly set off-diagonal elements in the covariance matrices

wdconinc:
Acts uses square matrices for covarianc matrices that are **not** required to be symmetric. This means that we should set **both** off-diagonal elements explicitly (but we can just read one of them). The EDM4eic data model does assume symmetry, so only upper diagonal is stored.

This PR ensures that we explicitly write off-diagonal covariance matrix elements everywhere, even if they are just written as zeros. This makes it clearer that we are not implicitly assuming symmetry.

- [x] Bug fix (issue: symmetry is assumed when not guaranteed)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackerMeasurementFromHits.cc
+++ b/src/algorithms/tracking/TrackerMeasurementFromHits.cc
@@ -57,7 +57,7 @@ namespace eicrecon {
             Acts::SquareMatrix2 cov = Acts::SquareMatrix2::Zero();
             cov(0, 0) = hit.getPositionError().xx * mm_acts * mm_acts; // note mm = 1 (Acts)
             cov(1, 1) = hit.getPositionError().yy * mm_acts * mm_acts;
-            cov(0, 1) = cov(1,0) = 0.0;
+            cov(0, 1) = cov(1, 0) = 0.0;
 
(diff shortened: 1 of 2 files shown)


wdconinc:
No diffs, e.g. https://github.com/eic/EICrecon/actions/runs/7596867924/job/20691396965?pr=1244#step:15:82.

src/algorithms/tracking/TrackerMeasurementFromHits.cc
```diff
@@ -57,8 +57,7 @@ namespace eicrecon {
             Acts::SquareMatrix2 cov = Acts::SquareMatrix2::Zero();
             cov(0, 0) = hit.getPositionError().xx * mm_acts * mm_acts; // note mm = 1 (Acts)
             cov(1, 1) = hit.getPositionError().yy * mm_acts * mm_acts;
-            cov(0, 1) = 0.0;
-
+            cov(0, 1) = cov(1,0) = 0.0;
```
wdconinc, line 60:
```suggestion
            cov(0, 1) = cov(1, 0) = 0.0;
```

osbornjd (approved):

PR #1245, opened 2024-01-21
linux-eic-shell.yml: fix capybara comment regressions

veprbl:


wdconinc (approved):
Didn't try to grok, but if it works :-)

PR #1250, opened 2024-01-22
Fix hit merging and add thresholds for backward HCal

lkosarz:

Implemented 100 ns integration time
Eth=0.25 MIP=0.1875 MeV for backward HCal
Fixed hit merging and plugged into clustering

- [x] Bug fix (fixes no merging)
- [x] New feature (updated thresholds)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/detectors/EHCAL/EHCAL.cc
+++ b/src/detectors/EHCAL/EHCAL.cc
@@ -63,7 +63,7 @@ extern "C" {
           {
             .readout = "HcalEndcapNHits",
             .fields = {"layer", "slice"},
-            .refs = {4, 0},
+            .refs = {4, 0}, // place merged hits at ~1 interaction length deep
           },
           app   // TODO: Remove me once fixed
         ));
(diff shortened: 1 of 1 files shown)


src/detectors/EHCAL/EHCAL.cc
```diff
@@ -62,16 +63,16 @@ extern "C" {
           {
             .readout = "HcalEndcapNHits",
             .fields = {"layer", "slice"},
-            .refs = {1, 0},
+            .refs = {4, 0},
```
wdconinc, line 66: Can you document the reason for `4` here?

lkosarz: This is in order to place the hits in the middle at ~1 interaction length rather than in the front.

wdconinc: I meant add a comment in the source code.

wdconinc:
I have asked @kkauder to review this so he can keep the overview of thresholds and backgrounds in the context of the master spreadsheet. That's all we're waiting for.

kkauder (approved):

PR #1197, opened 2023-12-31
feat: use logger service that comes with algorithms

wdconinc:
### Briefly, what does this PR introduce?
Instead of giving each algorithm an spdlog logger (`m_log`), the algorithms interface provides a centralized logger service with both stream and call interfaces (`error() << msg << endmsg` and `error(msg)`). This directs the CalorimeterClusterRecoCoG and other calorimeter algorithm to use this interface.

TODO:
- [x] depends on https://github.com/eic/algorithms/pull/13
- [x] depends on https://github.com/eic/algorithms/pull/15
- [x] will conflict with https://github.com/eic/EICrecon/pull/1194

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1251, opened 2024-01-23
Introduce factory to subdivide collection

simonge:
Introduces a generic factory which can create several subset collections from a collection based on the cellID of the entry.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No


simonge:
@nathanwbrei is there a cleaner way of handling the templating I could be using between the factories and algorithms.

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -30,6 +32,31 @@ extern "C" {
...
+         {"TaggerTrackerRawHits"},
+         segmentDiv,
+         {
+           .divisions = segmentIDs,
+           .readout   = "TaggerTrackerHits",
+           .division  = {"module","layer"},
+         },
+         app
+      )
+    );
```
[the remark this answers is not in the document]
simonge: I would happily make it more generic but can't currently picture how you're suggesting to go about this.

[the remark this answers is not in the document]
simonge: This will need to be left for to a future PR.

Were you envisioning defining the lambda function in (e.g.) LOWQ2.cc and passing that to a factory constructor? This feels like it's moving part of what should be contained inside an algorithm to the wrong place and I haven't worked out a way of passing an appropriate lambda into the factory.

Options this should presumably be able to cope (amongst others) with:
A collection member potentially being a member of multiple subset collections
Dividing the collection up based on ranges rather than discrete values
The geometry splitting application requires passing the detector segmentation decoder

Anyway got myself a bit tangled up thinking about it.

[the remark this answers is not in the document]
simonge: It's ugly but appears to work by passing the lambda function through the algorithm config. A couple of examples covering my previous use cases here:
[https://github.com/eic/EICrecon/blob/generalise-splitgeometry-to-divide-any-collection-by-lambda/src/detectors/LOWQ2/LOWQ2.cc](https://github.com/eic/EICrecon/blob/generalise-splitgeometry-to-divide-any-collection-by-lambda/src/detectors/LOWQ2/LOWQ2.cc)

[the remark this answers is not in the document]
simonge: Fixed in a way I'm more or less happy with (thanks somewhat to copilot). The SubDivideCollection algorithm now just gets passed a std::function through the config returning indices for which collections an entry should be included in. I've introduced a couple of templated examples as functors into a separate header file.

src/algorithms/meta/SubDivideCollection.h
```diff
@@ -0,0 +1,68 @@
...
+
+          for( auto out : subdivided_entries){
+            out->setSubsetCollection();
+          }
+
+          for (const auto& entry : *entries) {
+
+            auto div_indecies = this->m_cfg.function(entry);
+
+            for (auto index : div_indecies){
```
wdconinc, line 55:
```suggestion
            auto div_indices = this->m_cfg.function(entry);

            for (auto index : div_indices){
```

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -30,6 +32,32 @@ extern "C" {
          app
     ));
 
+    // Divide collection based on geometry segmentation labels
+    // This should really be done before digitization as summing hits in the same cell couldn't evet be mixed between layers. At the moment just prep for clustering.
```
wdconinc, line 36:
```suggestion
    // This should really be done before digitization as summing hits in the same cell couldn't even be mixed between layers. At the moment just prep for clustering.
```

simonge:
@wdconinc Is this in line with what you envisioned? Currently there aren't any outputs which depend on this so it won't be tested by the CI. Should I add a couple of things temporarily until a future PR actually makes proper use of it?

src/algorithms/meta/SubDivideCollection.h
```diff
@@ -0,0 +1,68 @@
...
+                      {"inputCollection"},
+                        {"outputCollection"},
+                          "Sub-Divide collection"
+                      },
+        WithPodConfig<SubDivideCollectionConfig<T>>() {
+        };
+
+        void init(std::shared_ptr<spdlog::logger>& logger){ // set logger
+          m_log      = logger;
+        };
```
wdconinc, line 40: Now that #1197 is merged:
```suggestion
        void init() final { };
```

src/algorithms/meta/SubDivideCollection.h
```diff
@@ -0,0 +1,68 @@
...
+              subdivided_entries[index]->push_back(entry);
+            }
+
+          }
+
+        };
+
+    private:
+      std::shared_ptr<spdlog::logger> m_log;
+
```
wdconinc, line 65: Along with init change above:
```suggestion
```

src/factories/meta/SubDivideCollection_factory.h
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    void Process(int64_t run_number, uint64_t event_number) {
+
+    try {
+      const typename T::collection_type* raw_hits = m_raw_hits_input();
+      std::vector<gsl::not_null<typename T::collection_type*>> split_hits;
+      for (const auto& split_hit : m_split_hits_output()) {
+        split_hits.push_back(gsl::not_null<typename T::collection_type*>(split_hit.get()));
+      }
```
wdconinc, line 49: Why do we need this? Cannot `algo` accept `m_split_hits_output()` as is?

src/factories/meta/SubDivideCollection_factory.h
```diff
@@ -0,0 +1,57 @@
...
+      m_algo->process(raw_hits,split_hits);
+    }
+    catch(std::exception &e) {
+      throw JException(e.what());
```
wdconinc, line 53: In too many cases we end up chasing exceptions that are eaten. So warn with
```suggestion
      logger()->warn(e.what());
      throw JException(e.what());
```

src/factories/meta/SubDivideFunctors.h
```diff
@@ -0,0 +1,98 @@
+// Copyright 2023, Simon Gardner
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
+#pragma once
+
+#include <JANA/JApplication.h>
```
wdconinc, line 6: This file should live in src/algorithms, not in jana land. It can use the algorithms service proxies.

wdconinc (commented):
Looks great! One goal I had in mind when suggesting this templated approach with a function as config is specifically a selector algorithm/factory (think: select all kaons in ReconstructedParticles and store in subset collection). I think that is entirely there now, with a functor that returns an empty vector for false and a single entry vector when true.

src/factories/meta/SubDivideCollection_factory.h
simonge: Looks like it can accept the `m_raw_hits_input()` directly but the compiler complains `m_split_hits_output()` returns a 
`std::vector<std::unique_ptr<edm4eic::RawTrackerHitCollection>>` rather than a `std::tuple<const std::vector<gsl::not_null<edm4eic::RawTrackerHitCollection*>>>`.

wdconinc: I think https://github.com/eic/EICrecon/blob/main/src/global/reco/ReconstructedElectrons_factory.h does something similar, right?

src/factories/meta/SubDivideFunctors.h
wdconinc: An example is in https://github.com/eic/EICrecon/blob/main/src/algorithms/calorimetry/CalorimeterHitDigi.h#L62 and https://github.com/eic/EICrecon/blob/main/src/factories/calorimetry/CalorimeterHitDigi_factory.h#L35.

src/factories/meta/SubDivideCollection_factory.h
simonge: > I think https://github.com/eic/EICrecon/blob/main/src/global/reco/ReconstructedElectrons_factory.h does something similar, right?

I believe the difference will be that this hasn't been converted to an algorithms::Algorithm yet. This was the same for the templated variadic input for the collection collector. Maybe I need to better define the algorithms::Input/Output but so far nothing I've tried has worked

src/factories/meta/SubDivideFunctors.h
simonge: I originally tried this but I'm not sure this how to make this available at the point the lambda/functor is being created

src/factories/meta/SubDivideCollection_factory.h
wdconinc: FYI, https://github.com/eic/EICrecon/pull/1313/files#diff-7f55d43dee21b064cb75128d7539847e30667cd536765794d41bf7abdb32ff5fR43 is to deal with the `gsl::not_null`.

simonge: Is that doing anything fundamentally different or just a bit neater?

wdconinc: No difference. I hadn't looked at your recent changed to the code, and was just commenting on the comment https://github.com/eic/EICrecon/pull/1251#discussion_r1505700489

src/factories/meta/SubDivideFunctors.h
simonge: Coming back to this a few days later hasn't helped, this is probably a lack of the necessary C++ understanding from me. I don't quite understand how giving the factory:
`Service<AlgorithmsInit_service> m_algorithmsInit {this};`

Allows an algorithm access to:
`const algorithms::GeoSvc& m_geo = algorithms::GeoSvc::instance();`

The way the function is passed as a configuration to the factory generator means that the function has to be made before the factory itself, so `AlgorithmsInit_service` can't yet exist.

wdconinc: In short, algorithms are created when the init method of the factory is called, so when the AlgorithmsInit_service has already been instantiated.

simonge: Ok, moving the initialisation of the variables in `GeometrySplit` which rely on the geometry to the first call of the function rather than it's constructor works fine.

src/algorithms/meta/SubDivideFunctors.h
```diff
@@ -0,0 +1,108 @@
...
+// ----------------------------------------------------------------------------
+class GeometrySplit {
+public:
+
+    GeometrySplit(std::vector<std::vector<long int>> ids, std::string readout, std::vector<std::string> divisions)
+    : m_ids(ids), m_readout(readout), m_divisions(divisions){};
+
+    template <typename T>
+    std::vector<int> operator()(T& instance) const {
+        if(!is_init){
```
wdconinc, line 47: Use of `std::call_once` is better practice; https://en.cppreference.com/w/cpp/thread/call_once.

src/algorithms/meta/SubDivideFunctors.h
```diff
@@ -0,0 +1,108 @@
...
+
+    template <typename T>
+    std::vector<int> operator()(T& instance) const {
+        std::vector<int> ids;
+        // Check if requested value matches any configuration combinations
+        std::vector<int> values;
+        (values.push_back((instance.*MemberFunctionPtrs)()), ...);
+        auto index = std::find(m_ids.begin(),m_ids.end(),values);
+        if(index != m_ids.end()){
+            ids.push_back(index-m_ids.begin());
```
wdconinc, line 98: Use `std::distance`; https://en.cppreference.com/w/cpp/iterator/distance.

src/factories/meta/SubDivideCollection_factory.h
```diff
@@ -0,0 +1,51 @@
...
+public:
+    void Configure() {
+      m_algo = std::make_unique<AlgoT>(this->GetPrefix());
+      m_algo->applyConfig(this->config());
+      m_algo->init();
+    }
+
+    void setFunctions(std::function<std::vector<int>(const T&)> f) {
+      m_algo->setFunction(f);
+    }
```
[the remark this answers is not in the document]
wdconinc: It doesn't seem like we need this, right?

src/global/beam/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+# Setting default includes, libraries and installation paths
+plugin_add(${PLUGIN_NAME} )
+
+# The macro grabs sources as *.cc *.cpp *.c and headers as *.h *.hh *.hpp
+# Then correctly sets sources for ${_name}_plugin and ${_name}_library targets
+# Adds headers to the correct installation directory
+plugin_glob_all(${PLUGIN_NAME})
+
+# Find dependencies
+plugin_add_dd4hep(${PLUGIN_NAME})
```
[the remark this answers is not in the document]
wdconinc: Never mind, this may come in through the included SubDivideFunctors.h which uses DD4hep implicitly through the algorithms/geo.h include.

src/factories/meta/SubDivideFunctors.h
wdconinc: Yeah, that's perfect. Only comment I left is to use `call_once` for the flag, which is essentially the same but has more thread safety guarantees.

src/algorithms/meta/SubDivideFunctors.h
wdconinc:
```suggestion
            ids.push_back(std::distance(m_ids.begin(), index));
```

src/algorithms/meta/SubDivideFunctors.h
wdconinc: This is going to need a closing parenthesis and some captured variables, but something like this:
```suggestion
        std::call_once(is_init, [](){
```

simonge: This turned into more of a pain than it should have. Not 100% clear on the details but it seems is_init needs to be a shared_ptr potentially due to a GCC bug.

PR #1256, opened 2024-01-26
treewide: cleanup EDM4eic 2.x deprecation, require at least 3.0

veprbl:


wdconinc (approved):
The force of progress is unstoppable.

PR #1257, opened 2024-01-26
TrackPropagation_factory: resolve an indexing bug by using ACTS surface ID fields

jdbrice:
Potentially fixes the track projection strange behavior and mismatch between detector/surface ID and eta range, etc.

- [x] Bug fix (issue #NA)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: Issue discussed in several meetings (collaboration meeting in Jan 2024)

NO
NO

### DIFF
--- a/src/global/tracking/TrackPropagation_factory.cc
+++ b/src/global/tracking/TrackPropagation_factory.cc
@@ -20,7 +20,7 @@
 #include <spdlog/logger.h>
 #include <Eigen/Geometry>
 #include <algorithm>
-#include <cstddef>
+#include <cstdint>
 #include <gsl/pointers>
 #include <map>
 
--- a/src/global/tracking/TrackPropagation_factory.cc
+++ b/src/global/tracking/TrackPropagation_factory.cc
@@ -60,10 +60,8 @@ void eicrecon::TrackPropagation_factory::Process(const std::shared_ptr<const JEv
         for(auto& surf : m_target_surface_list) {
             auto prop_point = m_track_propagation_algo.propagate(traj, surf);
             if(!prop_point) continue;
-#if EDM4EIC_VERSION_MAJOR >= 3
             prop_point->surface = surf->geometryId().layer();
             prop_point->system  = surf->geometryId().extra();
-#endif
             this_propagated_track.addToPoints(*prop_point);
         }
         propagated_tracks.push_back(this_propagated_track);
--- a/src/global/tracking/TrackPropagation_factory.cc
+++ b/src/global/tracking/TrackPropagation_factory.cc
@@ -58,13 +58,14 @@ void eicrecon::TrackPropagation_factory::Process(const std::shared_ptr<const JEv
 
     for(auto traj: trajectories) {
         edm4eic::MutableTrackSegment this_propagated_track;
-        for(size_t isurf = 0; isurf < m_target_surface_list.size(); isurf++) {
-            auto surf = m_target_surface_list[isurf];
+        for(auto& surf : m_target_surface_list) {
             auto prop_point = m_track_propagation_algo.propagate(traj, surf);
             if(!prop_point) continue;
+            uint64_t sId = surf->geometryId().layer();
+            uint32_t dId = surf->geometryId().extra();
 #if EDM4EIC_VERSION_MAJOR >= 3
-            prop_point->surface = m_target_surface_ID[isurf];
-            prop_point->system = m_target_detector_ID[isurf];
+            prop_point->surface = sId;
+            prop_point->system  = dId;
 #endif
             this_propagated_track.addToPoints(*prop_point);
         }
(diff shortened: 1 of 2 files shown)


src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -58,15 +58,15 @@ void eicrecon::TrackPropagation_factory::Process(const std::shared_ptr<const JEv
 
     for(auto traj: trajectories) {
         edm4eic::MutableTrackSegment this_propagated_track;
-        for(size_t isurf = 0; auto surf: m_target_surface_list) {
+        for(size_t isurf = 0; isurf < m_target_surface_list.size(); isurf++) {
+            auto surf = m_target_surface_list[isurf];
```
wdconinc, line 62: Does this work?
```suggestion
        for(auto& surf : m_target_surface_list) {
```
i.e. can we then get `surface` and `system` from `surf`?

wdconinc: If it is not easy to do that, and it would delay this PR substantially, then we can just file an issue and proceed with the fix as is, though.

jdbrice: Unfortunately I do not think so - or at least I don't know how. We currently get the system id like so:

```
auto BEMC_system_id = m_geoSvc->detector()->constant<uint32_t>("ECalBarrel_ID");
```

so basically it gets the constant by name from `dd4hep::Detector*` - I am at least not familiar enough to know if it is possible to get directly from the surface? Or rather store it on the surface object somehow

wdconinc: It seems we already store them in the surface ID, e.g.
```cpp
    BEMC_prop_surface1->assignGeometryId(Acts::GeometryIdentifier().setExtra(BEMC_system_id).setLayer(1));
    BEMC_prop_surface2->assignGeometryId(Acts::GeometryIdentifier().setExtra(BEMC_system_id).setLayer(2));
```

wdconinc: The geometry ID `extra` field is the 8 lower bits of the unsigned long int, and we store detector ID in it for all 'regular' surfaces defined in the geometry. So we can use them here as well, and they will be propagated to the 8 lower bits of the surface field in the stored track points. (Layer is stored in a different set of bits in the geometry ID. Ref: https://acts.readthedocs.io/en/latest/core/geometry/geometry_id.html)

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -58,15 +58,15 @@ void eicrecon::TrackPropagation_factory::Process(const std::shared_ptr<const JEv
...
             auto prop_point = m_track_propagation_algo.propagate(traj, surf);
             if(!prop_point) continue;
 #if EDM4EIC_VERSION_MAJOR >= 3
             prop_point->surface = m_target_surface_ID[isurf];
             prop_point->system = m_target_detector_ID[isurf];
```
wdconinc, line 67: I think this should work, and avoids the need for the index or even the variables that store the surface and detector ID:
```suggestion
            prop_point->surface = surf.geometryId().layer();
            prop_point->system = surf.geometryId().extra();
```

jdbrice:
https://github.com/eic/EICrecon/pull/1257/commits/5245e369385a5170fd14c9ea2975fcb5ebdae60b
Takes wouter's comment (retrieve the IDs from the surface) and also removes the unneeded vectors for storing the surface and detector ID

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -57,13 +57,14 @@ void eicrecon::TrackPropagation_factory::Process(const std::shared_ptr<const JEv
 
     for(auto traj: trajectories) {
         edm4eic::MutableTrackSegment this_propagated_track;
-        for(size_t isurf = 0; auto surf: m_target_surface_list) {
+        for(auto& surf : m_target_surface_list) {
             auto prop_point = m_track_propagation_algo.propagate(traj, surf);
             if(!prop_point) continue;
-            prop_point->surface = m_target_surface_ID[isurf];
-            prop_point->system = m_target_detector_ID[isurf];
+#if EDM4EIC_VERSION_MAJOR >= 3
```
wdconinc, line 63: Didn't we remove this already in main? @veprbl

veprbl, line 63:
```suggestion
```

src/global/tracking/TrackPropagation_factory.cc
```diff
@@ -57,13 +57,14 @@ void eicrecon::TrackPropagation_factory::Process(const std::shared_ptr<const JEv
...
+            prop_point->surface = surf->geometryId().layer();
+            prop_point->system  = surf->geometryId().extra();
+#endif
```
veprbl, line 66:
```suggestion
```

src/global/tracking/TrackPropagation_factory.cc
veprbl: Yes. And the constant would be undefined because the include is gone too.

veprbl:
Looks like we see a more fair distribution across the systems now:
image

veprbl (approved):

wdconinc:
Once iwyu is applied (I can do this tomorrow on my laptop, not right now on my phone), this looks good.

veprbl (approved):

PR #1258, opened 2024-01-27
fix: avoid CI warnings by upgrading actions; add dependabot

wdconinc:
### Briefly, what does this PR introduce?
We're starting to see a lot of warnings in CI (e.g. https://github.com/eic/EICrecon/actions/runs/7674122445) due to node.js v16 deprecation, ref: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

This PR updates the versions in the CI workflow, and also enabled dependabot to do this automatically from now on.

TODO:
- [x] deal with https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes same name artifacts not allowed anymore 

### What kind of change does this PR introduce?
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1262, opened 2024-01-27
labeler.yml: fix config for v5

veprbl:
Fixes a regression from #1258

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,63 +1,83 @@
 "topic: backward":
-  - "src/detectors/EEMC/**"
-  - "src/detectors/EHCAL/**"
+- changed-files:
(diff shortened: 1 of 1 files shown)


veprbl:
@wdconinc My understanding, is that we can't (easily?) test until this is merged.

wdconinc:
Yes, pull_request_target always uses what's in main.

wdconinc:
If you wanted to, you could submit from the default branch of a fork. But I think this one is just hard to test in any case.

veprbl:
Passes syntax check in https://github.com/veprbl/EICrecon/pull/12

wdconinc (approved):
Looks right to me after reading how the condig was changed. Sorry for the regression.

PR #1263, opened 2024-01-27
linux-eic-shell.yml: use all available cores for cmake build

veprbl:
[Should benefit us since we have 4 cores available](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -88,7 +88,7 @@ jobs:
         run: |
           # install this repo
           CC="${{ matrix.CC }}" CXX="${{ matrix.CXX }}" CXXFLAGS="${{ matrix.CXXFLAGS }}" cmake -B build -S . -DCMAKE_INSTALL_PREFIX=install -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ matrix.CMAKE_BUILD_TYPE }} -DUSE_ASAN=ON -DUSE_TSAN=OFF -DUSE_UBSAN=ON
-          cmake --build build -- -j 2 install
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         run: |
           # install this repo
           CC="${{ matrix.CC }}" CXX="${{ matrix.CXX }}" CXXFLAGS="${{ matrix.CXXFLAGS }}" cmake -B build -S . -DCMAKE_INSTALL_PREFIX=install -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ matrix.CMAKE_BUILD_TYPE }} -DUSE_ASAN=ON -DUSE_TSAN=OFF -DUSE_UBSAN=ON
-          cmake --build build -- -j 2 install
+          cmake --build build -j $(getconf _NPROCESSORS_ONLN) --target install
```
wdconinc, line 91: I was going to suggest `nproc` so I looked up what the benefits and drawbacks are.

`nproc`: Not POSIX and needs coreutils. Returns number of processors available to calling process, regardless of physical processors (I guess this accounts for cgroup restrictions).

`getconf`: POSIX command but non-POSIX config key. More portable since not just Linux. Returns physical cores, not accounting for hyper-threading and (I presume, again) cgroups.

As long as it returns 4, it doesn't matter which one to me.

wdconinc (approved):

.github/workflows/linux-eic-shell.yml
veprbl: From my experiments, _NPROCESSORS_ONLN practically gives logical cores, not physical. And, yes, works on macOS.

PR #1276, opened 2024-02-02
Associate matched calo cluster with reco electrons

jdbrice:
The previously introduced ElectronReconstruction factory outputs a ReconstructedParticleCollection of electron candidates via the E/p ratio. However, the matched cluster (used for the E/p calculation) was not associated. This calls repo_electron.addToClusters(...) to keep the association for use in later electron finder factories that use the electron candidates.

Note:
I also considered adding the energy (from the cluster) to the reco Electron via:
```
reco_electron.setEnergy(cluster.getEnergy());
```
but I am not sure if this is a good idea...

- [ ] Bug fix (issue #__) 
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: __ Something between a bug fix and feature update

NO

NO

### DIFF
--- a/src/algorithms/reco/ElectronReconstruction.cc
+++ b/src/algorithms/reco/ElectronReconstruction.cc
@@ -57,7 +57,7 @@ namespace eicrecon {
               m_log->trace( "ReconstructedParticle: Energy={} GeV, p={} GeV, E/p = {} for PDG (from truth): {}", clu.getEnergy(), edm4hep::utils::magnitude(reco_part.getMomentum()), EoverP, sim.getPDG() );
               auto reco_electron = reco_part.clone();
               reco_electron.addToClusters( clu );
-              
+
               m_log->trace( "ReconstructedElectron: nClusters={}", reco_electron.clusters_size() );
               m_log->trace( "ReconstructedElectron: Energy={} GeV, p={} GeV, E/p = {} for PDG (from truth): {}", reco_electron.getClusters(0).getEnergy(), edm4hep::utils::magnitude(reco_part.getMomentum()), (reco_electron.getClusters(0).getEnergy() / edm4hep::utils::magnitude(reco_part.getMomentum())), sim.getPDG() );
               // Apply the E/p cut here to select electons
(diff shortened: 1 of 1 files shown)


veprbl (approved):

src/algorithms/reco/ElectronReconstruction.cc
```diff
@@ -55,10 +55,14 @@ namespace eicrecon {
               auto reco_part = reco_part_assoc->getRec();
               double EoverP = clu.getEnergy() / edm4hep::utils::magnitude(reco_part.getMomentum());
               m_log->trace( "ReconstructedParticle: Energy={} GeV, p={} GeV, E/p = {} for PDG (from truth): {}", clu.getEnergy(), edm4hep::utils::magnitude(reco_part.getMomentum()), EoverP, sim.getPDG() );
+              auto reco_electron = reco_part.clone();
+              reco_electron.addToClusters( clu );
```
wdconinc, line 59: Two questions:
- We are cloning here instead of reusing the reconstructed electron and storing it in a subset collection. Is that what we want? If we think of the ElectronReconstruction algorithm as following the cluster matching step, then electron finding should not change the electron and a clone should not be necessary, right?
- Relatedly, why isn't `reco_part` not yet filled with clusters? Shouldn't we be using the `ReconstructedParticles`  (tracking + PID + cluster matching) as input here instead of the `ReconstructedChargedParticles` (which are from tracking + PID). We seem to be repeating the cluster matching that is being done in `MatchClusters` here.

jdbrice: Hi wouter, both good points:
- When I first implemented this we did not have the option for subset collections. Is there an example I can follow of a working subset collection? If so I can test that and modify this case.
- I did not realize the difference - I will try modifying this to use `ReconstructedParticles` which should also simplify the code

wdconinc: I think the main active example of the subset collection is the TrackerHitCollector: https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackerHitCollector.h#L36.

veprbl: Should we merge this and file an issue for refactoring?

PR #1278, opened 2024-02-05
refactor: update to edm4eic v5.0.0

osbornjd:
Updates the IVF to edm4eic v5.0.0, whenever that change happens. Tested locally and builds correctly with the updated version

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes, in conjunction with edm4eic v5.0.0

No

### DIFF
--- a/src/algorithms/tracking/IterativeVertexFinder.cc
+++ b/src/algorithms/tracking/IterativeVertexFinder.cc
@@ -118,21 +118,19 @@ std::unique_ptr<edm4eic::VertexCollection> eicrecon::IterativeVertexFinder::prod
   }
 
   for (const auto& vtx : vertices) {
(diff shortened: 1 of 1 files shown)


osbornjd:
This should fail build CI by default since it requires edm4eic v5.0.0, not sure when we update to that but here it is when we do

veprbl:
> This should fail build CI by default since it requires edm4eic v5.0.0, not sure when we update to that but here it is when we do

Please maintain compatibility with edm4eic 3.0.0, like it's usually done
https://github.com/eic/EICrecon/pull/1241/files
By locking EICrecon to 5.0.0, we are hindering our ability to do cross checks bisect for bugs with older environments.

osbornjd:
Okay, I think I added what you requested - let me know if that is not what was intended

veprbl (approved):
builds with main and with v4.0.0

PR #1279, opened 2024-02-05
add algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc

veprbl:
# Things to be tested:

- [ ] Weight functions/Multiple hits
- [ ] Eta bounds (not clear what case this is for)
- [ ] Cluster shapes
- [ ] Null pointers passed for sim hits (algorithm segfaults despite interface suggesting that it should accept those values)


rahmans1:
@veprbl Should this be in February release?

veprbl:
Not relevant to campaigns

veprbl:
This is already added.

PR #1215, opened 2024-01-12
Migrate remaining factories to JOmniFactory

nathanwbrei:
### Briefly, what does this PR introduce?
This PR makes all JANA factories use JOmniFactory as a base class. This: 
1. Reduces the number of things a user has to learn in order to add a new algorithm
2. Simplifies and deduplicates existing code
3. Removes the older abstractions that are more complex and bug-prone 
4. Paves the way for external wiring of factories
5. Paves the way for Generic Algorithms

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: Refactoring (issue #1176)

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators

### Does this PR change default behavior?
- Stops suppressing missing collection exceptions in JTrackerHitCollector_factory

PR #1282, opened 2024-02-07
Generic template collector factory

simonge:
Adds a template omni factory which creates a subset collection, combining the entries from an array of collections of the same type.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


src/algorithms/digi/CollectionCollector.h
```diff
@@ -0,0 +1,55 @@
...
+            const auto [in_collections] = input;
+            auto [out_collection]       = output;
+
+            out_collection->setSubsetCollection();
+
+            for (const auto& collection : in_collections) {
+                 for (const auto& hit : *collection) {
+                     out_collection->push_back(hit);
+                 }
+            //    out_collection->insert(out_collection->end(), collection->begin(), collection->end());
```
wdconinc, line 46:
```suggestion
```

src/factories/digi/CollectionCollector_factory.h
```diff
@@ -0,0 +1,47 @@
...
+
+    void Process(int64_t run_number, uint64_t event_number) {
+
+        try {
+            std::vector<gsl::not_null<const typename T::collection_type*>>  raw_collections;
+            for (const auto& in_collection : m_inputs()) {
+                raw_collections.push_back(gsl::not_null<const typename T::collection_type*>{in_collection});
+            }
+            typename T::collection_type* merged_collection = m_output().get();
+            m_algo->process(raw_collections,merged_collection);
```
wdconinc, line 38:
```suggestion
            std::vector<gsl::not_null<const typename T::collection_type*>>  in_collections;
            for (const auto& in_collection : m_inputs()) {
                in_collections.push_back(gsl::not_null<const typename T::collection_type*>{in_collection});
            }
            typename T::collection_type* merged_collection = m_output().get();
            m_algo->process(in_collections, merged_collection);
```

src/global/tracking/tracking.cc
```diff
@@ -17,7 +17,7 @@
 #include "TrackerMeasurementFromHits_factory.h"
 #include "extensions/jana/JChainMultifactoryGeneratorT.h"
 #include "extensions/jana/JOmniFactoryGeneratorT.h"
-#include "factories/tracking/TrackerHitCollector_factory.h"
+#include "factories/digi/CollectionCollector_factory.h"
```
wdconinc, line 20: Might be useful to create a new type for this, `meta` or `generic`. It is as little a digi factory as a tracking factory.
```suggestion
#include "factories/meta/CollectionCollector_factory.h"
```

wdconinc (commented):
Looks good.

wdconinc:
This does seem to break all tracking, though. E.g. https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/2472823#L1479 for a downstream physics_benchmark. Same issue as in https://github.com/eic/EICrecon/pull/1215 so likely associated with the VariadicPodioInput somehow (@nathanwbrei).

simonge:
> This does seem to break all tracking, though. E.g. https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/2472823#L1479 for a downstream physics_benchmark. Same issue as in #1215 so likely associated with the VariadicPodioInput somehow (@nathanwbrei).

I noticed the Capybara was unhappy and was going to try digging into it further but wasn't quite sure how. Everywhere else I've tested this it appears to work fine but then there's no other previous output to compare my other uses against.

wdconinc:
> but then there's no other previous output to compare my other uses against

I usually download a `sim_....root` and corresponding `rec_....root` artifact from the latest run on main here, use the `sim_....root` as input for a local run of the branch inside eic-shell, then run `capybara bara --serve sim_...new.root sim_....root`.

src/factories/meta/CollectionCollector_factory.h
```diff
@@ -0,0 +1,47 @@
...
+        try {
+            std::vector<gsl::not_null<const typename T::collection_type*>>  in_collections;
+            for (const auto& in_collection : m_inputs()) {
+                in_collections.push_back(gsl::not_null<const typename T::collection_type*>{in_collection});
+            }
+            typename T::collection_type* merged_collection = m_output().get();
+            m_algo->process(in_collections, merged_collection);
+        }
+        catch(std::exception &e) {
+            throw JException(e.what());
```
wdconinc, line 41: For debugging:
```suggestion
            this->logger()->error(e.what());
            throw JException(e.what());
```
I would not be surprised if this was a case of an exception getting silently eaten by JANA2, likely an exception in the `for` loop over `m_inputs()` that then results in not filling `merged_collection` at all.

simonge: On my machine and events, for some currently undiagnosed reason running with this combination works:
SiBarrelTrackerRecHits
SiBarrelVertexRecHits
SiEndcapTrackerRecHits
TOFBarrelRecHit          
MPGDBarrelRecHits
MPGDDIRCRecHits       
B0TrackerRecHits

Adding TOFEndcapRecHits, starts calling the factory Process function then segfaults after a few events

Adding any of the below, the factory Process function is never called.
OuterMPGDBarrelRecHits
BackwardMPGDEndcapRecHits
ForwardMPGDEndcapRecHits

wdconinc: While I hate suggesting it, for now we should probably just try-catch and print `e.what()` as a warning, check if that works in the scope of this PR, and dump the underlying issue on the plate of the TOF and MPGD group as an issue.

simonge: I think this is a case for the optional collections, if any of them are missing the factory doesn't run. `e.what()` was never getting called when missing collections were requested.

It appears I was running on an old simulation file where different MPGD collections existed, looking at a new one the MPGD collections that broke my local run are present whereas MPGDDIRCHits doesn't appear to be. Can you confirm this?

Hopefully the TOF problems are related to the older simulation file too.

veprbl: MPGD DIRC is for epic_arches, we don't need to support it anymore. I would remove `try catch`, if possible.

src/algorithms/meta/CollectionCollector.h
```diff
@@ -0,0 +1,54 @@
...
+#include <algorithms/algorithm.h>
+#include <string>
+#include <string_view>
+
+#include "services/log/Log_service.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+   template<class T>
```
veprbl, line 14: Use two spaces for indent, please.

src/factories/meta/CollectionCollector_factory.h
simonge: If the reconstruction needs to work seamlessly with multiple detector configurations we'll need to allow for these to be optional variadic inputs. That does however allow the input list to get cluttered with unsupported collections. @nathanwbrei is this possible yet?

The list at the moment contains all of the detectors from a/b/c but will only work if they're all there. A separate issue but tangentially related, I hadn't realised the 18x275.xml was still a brycecanyon configuration and all of the CI is done using craterlake.

nathanwbrei: "No hope" is overstating it quite a bit. Go ahead and merge what you have now, when I add optional collection support I'll update this at the same time.

wdconinc (approved):
No diffs. Thanks for this!

veprbl (approved):
Well done!

PR #1283, opened 2024-02-07
Add linear fitting for low-q2 detector

simonge:
Creates a vector from space points provided by layers in the low-q2 tagger.
Needs #1282 #1251 and #921 first

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Adds new collections to the output which contain the reconstructed vectors


veprbl:
Before I post any review, a question: If ACTS already works for B0 tracker, why can't we re-use that work?
cc @rahmans1

simonge:
> Before I post any review, a question: If ACTS already works for B0 tracker, why can't we re-use that work? cc @rahmans1

I have already started working on adding the acts surfaces and based on the discussion with the acts group last week the reconstruction should be possible to some extent in acts if the beamline fields are also included. I'm sure this will come eventually but the acts telescope project sounds still quite preliminary and upcoming changes to the geometry interface that were discussed will make this much easier.

This is not something that I will have time to invest in on the necessary timescales and delay just having first Low-Q2 reconstructed electrons easily available for studies.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,135 @@
...
+          if((*layerHits).size()>m_cfg.layer_hits_max){
+            m_log->info("Too many hits in layer");
+            return;
+          }
+        }
+
+        // Create a matrix to store the hit positions
+        Eigen::MatrixXd hitMatrix(3,m_cfg.n_layer);
+        // Loop over all combinations of hits fitting a track to all layers
+        makeHitCombination(m_cfg.n_layer-1,&hitMatrix,inputhits,outputTracks);
```
veprbl, line 60:
```suggestion
        buildMatrixRecursive(m_cfg.n_layer-1,&hitMatrix,inputhits,outputTracks);
```

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,135 @@
...
+        hitMatrix->col(level) << pos.x, pos.y, pos.z;
+
+        if(level>0){
+          makeHitCombination(level-1,
+                             hitMatrix,
+                             hits,
+                             outputTracks);
+        }
+        else{
+          checkHitCombination(hitMatrix,outputTracks);
```
veprbl, line 82: Move this to process. Does this even need to be recursive?

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,135 @@
...
+      float             time{0};        // Time at this point [ns]
+      float             timeError{0};   // Error on the time at this point
+      float             theta = edm4eic::anglePolar(outVec);          // polar direction of the track at the surface [rad]
+      float             phi   = edm4eic::angleAzimuthal(outVec);         // azimuthal direction of the track at the surface [rad]
+      edm4eic::Cov2f    directionError;  // Error on the polar and azimuthal angles
+      float             pathlength{0};      // Pathlength from the origin to this point
+      float             pathlengthError{0}; // Error on the pathlength
+
+      edm4eic::TrackPoint point({surface,system,position,positionError,momentum,momentumError,time,timeError,theta,phi,directionError,pathlength,pathlengthError});
+      auto segment = (*outputTracks)->create(0,0);
```
veprbl, line 128: Can we, please, define variables with human-readable names for 0 and 0.

src/factories/fardetectors/FarDetectorLinearTracking_factory.h
```diff
@@ -0,0 +1,60 @@
+// Created by Simon Gardner
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
```
veprbl, line 3:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Simon Gardner
```

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,135 @@
...
+
+      segment.addToPoints(point);
+
+
```
veprbl, line 132: It might help with performance to not try to re-use hits that already became a track segment.

simonge: These hits need to be kept around as sometimes a few tracks can be not quite parallel and very close together so a hit might form several of these tracks. There is no ambiguity solver in this reconstruction flow currently. What would probably help more is a very basic filter on the range of positions allowed between subsequent layers. I will look at adding this in the config.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -0,0 +1,135 @@
...
+      edm4hep::Vector3d outVec = V.col(0).data();
+
+      // Make sure fit was pointing in the right direction
+      if(outVec.z>0) outVec = outVec*-1;
+
+      uint64_t          surface{0};     // Surface track was propagated to (possibly multiple per detector)
+      uint32_t          system{0};      // Detector system track was propagated to
+      edm4hep::Vector3f position(outPos.x/dd4hep::mm,outPos.y/dd4hep::mm,outPos.z/dd4hep::mm);        // Position of the trajectory point [mm]
+      edm4eic::Cov3f    positionError;  // Error on the position
+      edm4hep::Vector3f momentum = {(float)outVec.x,(float)outVec.y,(float)outVec.z};       // 3-momentum at the point [GeV]
```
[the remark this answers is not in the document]
simonge: The magnitude is unknown and can't be determined from the track points alone as there is no field and the timing resolution is no where near good enough. All I really want here is a position and direction (and time). Without the time currently included the algorithm itself shouldn't have any idea which the correct sign is even though from our detector setup it should only ever be one way around, this should be included in a later step.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
simonge: This is only called on the inner most recursion, so unless all possible hit matrix combinations are stored, then looped over separately in process. There are probably some nice std functions which would iterate over the combinations but likely have to set all components of the matrix on each iteration rather than just the current recursion step. The reason this is recursive rather than a set number of nested loops was to allow for various numbers of layers to be included. I'm sure acts has nice methods to handle all of this again itself.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
simonge: I've added the new configurations which should stop impossible combinations from being unnecessarily recursed into. Should reduce the combinatorics tested when full background samples are used.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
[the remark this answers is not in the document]
veprbl: I meant that the sign could be flipped from the way linear algebra works out. We might have to enforce z component to always be negative. Also, if momentum is not available, might be a good sense to only fill Theta and Phi.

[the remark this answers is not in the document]
simonge: Ok, good points. If it doesn't technically matter which way around the sign is, it also doesn't matter if you force it to be one way.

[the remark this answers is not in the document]
simonge: That check is actually already there
```
      // Make sure fit was pointing in the right direction
      if(outVec.z>0) outVec = outVec*-1;
```

.github/iwyu.imp
```diff
@@ -16,8 +16,8 @@
   { include: ['<Acts/Geometry/detail/DefaultDetectorElementBase.hpp>', private, '<Acts/Geometry/DetectorElementBase.hpp>', public] },
 
   # Eigen
-  { include: ['@<Eigen/(src/)?(.*?)/.*>', private, '<Eigen/$2>', public] },
-  { include: ['@<Eigen/src/plugins/.*>', private, '<Eigen/Core>', public] },
+  { include: ['@[<"]Eigen/(src/)?(.*?)/.*[>"]', private, '<Eigen/$2>', public] },
+  { include: ['@[<"]Eigen/src/plugins/.*[>"]', private, '<Eigen/Core>', public] },
```
veprbl, line 20: @wdconinc FYI this works

wdconinc: Nice, even more simplification of that file.

veprbl (approved):
Other than wanting to have this implemented in Acts, this looks fine.

simonge:
> Other than wanting to have this implemented in Acts, this looks fine.

Thanks, yeah this can hopefully be replaced in time.

PR #1285, opened 2024-02-08
feat: write edm4eic::Tracks in CKFTracking

wdconinc:
This PR modifies CKFTracking to write edm4eic::Tracks.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl:
Will have to reconcile with #1210

wdconinc:
> Will have to reconcile with #1210

Yes, ParticlesWithPID should take a edm4eic::TrackCollection as input after this, not edm4eic::TrajectoryCollection. It's probably the easiest factory downstream from CKFTracking to be converted.

ShujieL:
A related question: I saw you added Track as an additional output of CKF. Do we still plan to make Track the single output collection for CKF? If yes, how can we store trajectory outliers and acts_trajectories, which are not currently in the Track data structure?

wdconinc:
@osbornjd What was your plan with the following edm4eic::Track fields?
- `type`: unspecified in edm4eic, bitfield in edm4hep (half reserved for internal purposes)
- `position`: clearly different from loc-a, loc-b, and this is before vertexing, so what was the plan for this Vector3? assume 0,0,0 for IP to first order?
- `positionMomentumCovariance`: did you want to marginalize over the other components of the free 8x8 covariance, or just the subset?

osbornjd:
> @osbornjd What was your plan with the following edm4eic::Track fields?
> 
>     * `type`: unspecified in edm4eic, bitfield in edm4hep (half reserved for internal purposes)

I had no plan for this - this was already there and I assumed (perhaps wrongly?) that it served a purpose that was beyond me, either historical or something associated with edm4hep or ...?

>     * `position`: clearly different from loc-a, loc-b, and this is before vertexing, so what was the plan for this Vector3? assume 0,0,0 for IP to first order?

The idea for this was to store the track global position WRT some point - nominally the primary vertex after all of the tracking was complete. After the CKF this would be WRT (0,0,0), then ideally we would find the vertex and then propagate the track to the perigee defined at the vertex position to get the track parameters there for analysis purposes. 

>     * `positionMomentumCovariance`: did you want to marginalize over the other components of the free 8x8 covariance, or just the subset?

The free 8x8 covariance is in basis (x,y,z,t,px_hat, py_hat, pz_hat, q/p). Perhaps we should just store the whole thing for simplicity sake but this has a redundancy in (px_hat, py_hat, pz_hat, q/p). Initially I had thought that we would just store the 6x6 (x,y,z,px,py,pz) since we don't use time anywhere in the track reconstruction, but I could see moving to this at some point in the future in which case it may be easier to just store the whole 8x8 matrix.

osbornjd:
> A related question: I saw you added Track as an additional output of CKF. Do we still plan to make Track the single output collection for CKF? If yes, how can we store trajectory outliers and acts_trajectories, which are not currently in the Track data structure? (edited) @osbornjd do you have any insight on this? or shall we continue this discussion in a different channel?

We wouldn't store the Acts trajectories as it makes us dependent on any change that Acts pushes in between versions. It also wouldn't make sense to store the Acts trajectories in `Track` anyway because they contain basically the same information. So what will need to happen is to have some "conversion" wrapper which allows us to swap in between `edm4eic` and the Acts EDM - that way we are insulated from changes to Acts and have one "access" point by which any of our code that needs a `Acts::Trajectory` or `Acts::TrackParameters` object can call to get that object.

veprbl (approved):
This is an improvement in organization of the code, and the CentralCKFTracks are now fully filled, which wasn't the case in my original implementation where it only had a relation to the trajectory filled.

wdconinc:
> fully filled

Full disclosure: Position, momentum, and their covariance are not filled. Those need more thought and non-trivial code to fill them (we need a jacobian on the perigee surface). I'm punting that for a next PR.

PR #1291, opened 2024-02-11
Included PCA calculations using a straight-line approximation in truth seeded tracking

harsimranhs:

The transverse point of closest approach (PCA) calculated using straight line approximation of truth momentum vector at truth vertex point is given as seed to ACTS.

In the illustration below, the transverse PCA will be the point where the distance of magenta dot-dashed line distance will be minimum from z-axis.
image
  
The `eicrecon` before and after these changes was run on a `ddsim` output file. The specifications of the file are: 5000 events having -ve muons with multiplicity of 10 thrown in a uniform distribution from the vertex point at (x,y,z) = (3,4,5) mm.

The truth (MC) momentum azimuthal angle distribution from `MCParticles` collection is:
phi_truth
 
The track azimuthal angle distribution from `CentralCKFTrackParameters` collection before the changes is:
track_phi

and after the changes is:
track_phi

This shows:
1. The distribution is now flat as it was expected to be. 
2. The number of reconstructed tracks have been increased.

Some other track level information after the changes are:
changed_r 
changed_z

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No


src/global/tracking/TrackParamTruthInit_factory.h
```diff
@@ -11,6 +11,7 @@
 #include <utility>
 #include <vector>
 
+#include "ActsExamples/EventData/Trajectories.hpp"
```
bschmookler, line 14:
```suggestion
#include "services/geometry/acts/ACTSGeo_service.h"
```

bschmookler (commented):
Looks good overall. 

Do we gain any useful information by also writing out the the truth-seeded parameters collection prior to the CKF  (i.e., ```InitTrackParams```)? If we think so, we can add it to the ```JEventProcessorPODIO.cc``` file.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -93,10 +99,35 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
...
+        Acts::Vector3 global(xpca, ypca, zpca);
+
+        // convert from global to local coordinates using the defined line surface
+        Acts::Vector2 localpos;
+        Acts::Vector3 direction(sin(theta)*cos(phi), sin(theta)*sin(phi), cos(theta));
+        auto local = perigee->globalToLocal(m_geoSvc->getActsGeometryContext(), global, direction);
+
+        if(!local.ok())
+        {
+            continue;
```
veprbl, line 122: This just drops a whole track. If this is not supposed to happen, let's throw an exception instead.

osbornjd:
> Looks good overall.
> 
> Do we gain any useful information by also writing out the the truth-seeded parameters collection prior to the CKF (i.e., `InitTrackParams`)? If we think so, we can add it to the `JEventProcessorPODIO.cc` file.

This probably will be useful in the long run, for example to compare truth seeded track parameters and reco seeded track parameters. I assume you must be doing this in some capacity already given you have shown plots of these quantities.

src/algorithms/tracking/TrackParamTruthInit.cc
osbornjd: @wdconinc commented [here](https://github.com/eic/EICrecon/pull/1185#discussion_r1427223257) that we should continue and skip it. This is really a question of how we should handle this, and I can see a few pros/cons:

1. Skipping the track with continue
* We could accidentally and/or unknowingly just not truth seed perfectly good tracks for the event. This would be an "inefficiency" of the truth seeder.
2. Throwing an exception
* Depending on how often this happens, we may never be able to process events. I wouldn't think it would happen frequently, but maybe we should study it since a straight line approximation is used instead of a full helical propagation. 
 
We could alternatively make the [call](https://github.com/acts-project/acts/blob/2cf3fe0254d2bf9434899fdcfbe316366a970956/Core/include/Acts/Surfaces/LineSurface.hpp#L184) with a tolerance that is larger than 0.1 micron which is really small (default tolerance `s_onSurfaceTolerance` is defined [elsewhere](https://github.com/acts-project/acts/blob/2cf3fe0254d2bf9434899fdcfbe316366a970956/Core/include/Acts/Definitions/Tolerance.hpp#L26) in Acts as 0.1 micron). We could make this a few tens of microns (or explicitly determine what value the straight line approximation is good for in our usual simulations) to accommodate this and basically ensure that the `globalToLocal` call never fails by definition. In the case where it is large, we are basically allowing truth seeds with poorly defined track parameters into the CKF which I imagine would then fail and be seen as a different kind of inefficiency.

bschmookler: Based on @harsimranhs definition of (xpca,ypca), we should always have (xpca,ypca) perpendicular to (px,py).

```
auto linesurface_parameter = -(v.x*p.x + v.y*p.y)/(p.x*p.x + p.y*p.y);
 
auto xpca = v.x + linesurface_parameter*p.x;
auto ypca = v.y + linesurface_parameter*p.y;
auto zpca = v.z + linesurface_parameter*p.z;
```

So the dot product between (xpca,ypca) and (px,py) is

```
xpca*px + ypca*py =
vx*px + vy*py + linesurface_parameter * (px**2 + py**2) =
vx*px + vy*py - (vx*px + vy*py) =
0
```

So, I think this should always fulfill the Acts globalToLocal function requirement that the momentum and the pca position are perpendicular in the x,y plane: [https://github.com/acts-project/acts/blob/main/Core/src/Surfaces/LineSurface.cpp#L80-L123](https://github.com/acts-project/acts/blob/main/Core/src/Surfaces/LineSurface.cpp#L80-L123).

bschmookler:
> This probably will be useful in the long run, for example to compare truth seeded track parameters and reco seeded track parameters. I assume you must be doing this in some capacity already given you have shown plots of these quantities.

Yes, we've been doing this for the real seeding and associated tracking. It is probably good to do it here as well as you say.

src/global/tracking/TrackParamTruthInit_factory.h
harsimranhs: I just committed the suggested changes, Barak. Thank you!

src/algorithms/tracking/TrackParamTruthInit.cc
osbornjd: Thanks for the clear explanation @bschmookler , in that case I agree with @veprbl that we should throw an exception because this should really never happen - if it did, that somehow indicates something terribly wrong happened. @wdconinc what do you think?

osbornjd:
> > This probably will be useful in the long run, for example to compare truth seeded track parameters and reco seeded track parameters. I assume you must be doing this in some capacity already given you have shown plots of these quantities.
> 
> Yes, we've been doing this for the real seeding and associated tracking. It is probably good to do it here as well as you say.

Should we just save this for another PR, since it is sort of outside of the scope of this particular development?

src/algorithms/tracking/TrackParamTruthInit.cc
wdconinc: [I was away last week, so only getting back to the mentions now.]

Exceptions are for things that should never happen and where when it happens it is an indication of corruption or incorrect input or something else that can't be recovered from (unless you are going to catch them and handle the issue). I don't think that's quite the case here; we aren't thinking that the rest of the run is bad simply because this issue happens. We don't want to use exceptions as a way to get the simulation production team to let you know when this is starting to bug them enough that you should fix it differently ;-)

I would say that there's a third option here, and that's printing an error message. After the next production campaign, one of you checks if this error appears anywhere in the logs. If not, yay the fix works. If the error is there, then at least it didn't impact the running of the campaign.

harsimranhs: Hi @wdconinc, I have added an error message as you said. Please check and suggest any changes if you want. Thank you!

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -93,10 +99,36 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
...
+        }
+
+        localpos = local.value();
+
         // Insert into edm4eic::TrackParameters, which uses numerical values in its specified units
         auto track_parameter = track_parameters->create();
         track_parameter.setType(-1); // type --> seed(-1)
-        track_parameter.setLoc({static_cast<float>(std::hypot(v.x, v.y)), static_cast<float>(v.z)}); // 2d location on surface [mm]
+        track_parameter.setLoc({(float)localpos(0), (float)localpos(1)}); // 2d location on surface [mm]
```
wdconinc, line 131: Use static_cast.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -93,10 +99,36 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
...
+        auto linesurface_parameter = -(v.x*p.x + v.y*p.y)/(p.x*p.x + p.y*p.y);
+
+        auto xpca = v.x + linesurface_parameter*p.x;
+        auto ypca = v.y + linesurface_parameter*p.y;
+        auto zpca = v.z + linesurface_parameter*p.z;
+
+        Acts::Vector3 global(xpca, ypca, zpca);
+
+        // convert from global to local coordinates using the defined line surface
+        Acts::Vector2 localpos;
```
wdconinc, line 116: Move to first assignment.

harsimranhs: I have done the suggested changes. Please check, thank you!

src/algorithms/tracking/TrackParamTruthInit.cc
harsimranhs: As of now, I am marking this conversation resolved as we have changed the code incorporating the suggestion. In case of any further change, please let us know.

bschmookler (approved):

PR #1293, opened 2024-02-18
dbg: output rate limit info before/after download-artifact actions

wdconinc:
This PR adds a debug output step in the jobs that use dawidd6/action-download-artifact to allow us to see which steps are having most detrimental impact on the rate limits.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -576,6 +576,16 @@ jobs:
           *.dot
           *.svg
         if-no-files-found: error
+    - name: Check rate limit before
+      run: |
+        rate=$(gh api rate_limit | jq '.rate')
+        used=$(echo $rate | jq '.used')
+        remaining=$(echo $rate | jq '.remaining')
+        reset=$(echo $rate | jq '.reset')
+        resetdate=$(TZ="America/New_York" date --date @$reset)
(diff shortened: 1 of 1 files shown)


wdconinc:
The outcome of this is that typically takes about 4 API requests per dawidd6/download-artifact-action. For older PRs we have some pagination in the workflow runs which makes it go up to potentailly 7 or 8 API requests. At ~45 open PRs, that means ~200 API requests per workflow run for the docs collection and ghpages deploy (after including another ~30 for the 7 diffs with capybara). That probably explains the issues. I don't think there's anything really wrong with our approach.

If we want to reduce this substantially, we would need to change the number of API requests needed to find the artifacts. We need:
- one request to get the current workflow (TODO can be avoided by explicitly specifying),
- one request to get the head_sha of the desired PR (TODO can probably be combined with the list-open-prs job),
- at least one request (hopefully, it's paginated) to get the workflow runs for the current workflow on the head_sha,
- at least one request per workflow run to get the desired artifact (if the workflow doesn't have the artifact, we need to try again, so we want to try to guarantee that the artifact is there)

Closing as the purpose of data collection is achieved.

PR #1303, opened 2024-02-25
fix(ci): reduce API requests by specifying workflow name

wdconinc:
Specifying the workflow name with dawidd6/action-download-artifact avoids an API request by bypassing https://github.com/dawidd6/action-download-artifact/blob/master/main.js#L61. This PR uses the current workflow name which will be correct as long as we don't change the workflow name.

- [x] Bug fix (issue: too many API requests)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
This works in e.g. https://github.com/eic/EICrecon/actions/runs/8034301355/job/21947118366#step:4:28, where the workflow is specified so doesn't have to be looked up with an API call.

veprbl (approved):

PR #1304, opened 2024-02-25
fix(ci): store pull_request.head.sha for later use

wdconinc:
### Briefly, what does this PR introduce?
Specifying the commit hash with dawidd6/action-download-artifact avoids an API request by bypassing https://github.com/dawidd6/action-download-artifact/blob/master/main.js#L87. This PR modifies the list-open-prs job to pass the head.sha as commit to the get-docs-from-open-prs matrix.

### What kind of change does this PR introduce?
- [x] Bug fix (issue: too many API requests)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1305, opened 2024-02-25
feat(ci): don't publish docs and capybara for draft PRs

wdconinc:
Draft PRs still get a capybara artifact, but it is not published to ghpages.

TODO:
- [x] propagate draft status to steps that post summary

- [x] Bug fix (issue: too many API requests)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -1029,13 +1029,13 @@ jobs:
           path: publishing_docs/
           retention-days: 7
       - name: Populate capybara summary
-        if: ${{ github.event_name == 'pull_request' }}
+        if: ${{ github.event_name == 'pull_request' && github.event_name.pull_request.draft == 'false' }}
         run: |
          echo "### Capybara summary for PR ${{ github.event.pull_request.number }}" >> capybara_${{ github.event.pull_request.number }}.md
          find publishing_docs/pr/${{ github.event.pull_request.number }}/capybara/ -mindepth 1 -maxdepth 1 -type d -printf \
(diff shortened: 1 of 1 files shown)


veprbl:
What problem is this solving?

wdconinc:
> What problem is this solving?

Reducing API requests by not publishing capybara reports for draft PRs.

wdconinc:
> > What problem is this solving?
> 
> Reducing API requests by not publishing capybara reports for draft PRs.

We only incur API requests through the download action that downloads artifacts from other workflows than the current one. We use that only for capybara diffs and docs publishing. Reducing that to the cases where it is likely to get used will reduce how many API requests we make overall. Arguably, if someone has something in draft, that can be defined to mean that one isn't interested yet in published capybara reports but just in whether it compiles etc. Once it's switched to ready, the next workflow would publish the capybara reports.

The new v4 standard download-artifact action can download from other workflow runs without additional API requests, but it requires a run_id which we would need to obtain through an API request (likely multiple) or otherwise.

veprbl:
I guess the idea here is to give you control whether you want to spend API requests on your PR? I not a fan of putting the burden of such control on ourselves.

wdconinc:
The goal isn't control; the goal is not maxing out the API requests. We have to reduce them somewhere. Maybe we reduced them enough with #1304 and #1305 so this isn't necessary. But if we continue to run out of API requests we need to do something else, certainly as we increase their use in places like epic too. Other options:
- really support PRs from forks and get people to use their own API request quotas (and core hours),
- use non-default tokens which each have a quota.

wdconinc:
We've managed to stay below API limits for a while now, so I'm closing this for now.

PR #1306, opened 2024-02-25
fix(ci): accept only completed workflow runs as source of capybara reference

wdconinc:
In #1299 we started accepting all workflows, not just the ones that succeeded. This had as effect that incomplete workflows (where there is no artifact yet) are used, leading to failures.

This PR requires that pipelines have completed.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
Let's try

PR #1314, opened 2024-02-28
fix: convert MatchClusters to JOmniFactory and algorithms interface

wdconinc:
This PR converts MatchClusters to JOmniFactory and the algorithms interface. Instead of ingesting two variadic cluster collections (not something that works...), we use the collector meta factory to put all clusters and all associations in their separate collections first, and then pass that to MatchClusters.

- [x] Bug fix (issue #1176, in part)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/algorithms/reco/MatchClusters.cc
```diff
@@ -14,197 +15,170 @@
...
 
-            // find associated particle
-            for (const auto &assoc: *inpartsassoc) {
-                if (assoc.getRecID() == inpart.getObjectID().index) {
-                    mcID = assoc.getSimID();
-                    break;
-                }
+        // find associated particle
+        for (const auto &assoc: *inpartsassoc) {
+            if (assoc.getRecID() == inpart.getObjectID().index) {
```
simonge, line 55:
```suggestion
            if (assoc.getRec() == inpart) {
```

src/algorithms/reco/MatchClusters.cc
```diff
@@ -14,197 +15,170 @@
...
-            assoc.setWeight(1.0);
-            assoc.setRec(outpart);
-            assoc.setSim((*mcparticles)[mcID]);
+        if (clusterMap.count(mcID)) {
+            const auto &clus = clusterMap[mcID];
+            debug("    --> found matching cluster with energy: {}", clus.getEnergy());
+            debug("    --> adding cluster to reconstructed particle");
+            outpart.addToClusters(clus);
+            clusterMap.erase(mcID);
         }
```
simonge, line 74: Might there not be more than one inpart with the same mcID? Similar to how you have delt with the largest cluster energy, is there anything which might be more appropriate than coming from the order in the collection for the tracks.

src/algorithms/reco/MatchClusters.cc
```diff
@@ -14,197 +15,170 @@
...
+        assoc.setRec(outpart);
+        assoc.setSim((*mcparticles)[mcID]);
+    }
+}
 
-            // loop over clusters
-            for (const auto cluster: *clusters) {
+// get a map of mcID --> cluster
+// input: clusters --> all clusters
+std::map<int, edm4eic::Cluster> MatchClusters::indexedClusters(
```
simonge, line 124:
```suggestion
std::map<edm4eic::MCParticle, edm4eic::Cluster> MatchClusters::indexedClusters(
```

Not sure on this one but it might be nicer to pass around the reference to the MCParticle directly rather than the ID within the collection. Would need to be altered in a few placed.

src/algorithms/reco/MatchClusters.cc
```diff
@@ -14,197 +15,170 @@
...
+    part.setMomentum(momentum);
+    part.setPDG(pdg);
+    part.setCharge(0);
+    part.setEnergy(energy);
+    part.setMass(mass);
+    part.addToClusters(*cluster);
+    return part;
+}
 
 } // namespace Jug::Fast
```
simonge, line 184:
```suggestion
} // namespace eicrecon
```

src/algorithms/reco/MatchClusters.cc
wdconinc: The algorithm can certainly be improved, and this may indeed be one area for improvement :-)

This PR is just an interface change, and assessing the correctness of the interface change by verifying that we get identical results relies on not changing how the algorithm works. So, for changes to the algorithm itself I'd prefer to postpone that to a different PR.

simonge: Capybara looked fine then to best of my knowledge and the interface changes looked sensible.

simonge (approved):
Good and consistent with previous output.

PR #1321, opened 2024-03-05
fix: use proper SPDX headers to clarify LGPL3 vs JSA licensing

wdconinc:
This modifies the headers to be clear about which license applies. All unclear licensing is deferred to JSA especially when it was still referring to the "top-level LICENSE file" which at the time was JSA.

- [x] Bug fix (issue #2)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/detectors/BEMC/BEMC.cc
+++ b/src/detectors/BEMC/BEMC.cc
@@ -1,7 +1,5 @@
-// Copyright 2022, David Lawrence
-// Subject to the terms in the LICENSE file found in the top-level directory.
-//
-//
+// SPDX-License-Identifier: JSA
+// Copyright (C) 2022 David Lawrence
 
 #include <Evaluator/DD4hepUnits.h>
 #include <JANA/JApplication.h>
--- a/LICENSE.JSA
+++ b/LICENSE.JSA
@@ -1,0 +1,33 @@
+
+Copyright (c) 2020, Jefferson Science Associates, LLC. All Rights Reserved. Redistribution
+and use in source and binary forms, with or without modification, are permitted as a
+licensed user provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice, this
+   list of conditions and the following disclaimer in the documentation and/or other
+   materials provided with the distribution.
+3. The name of the author may not be used to endorse or promote products derived
+   from this software without specific prior written permission.
+
+This material resulted from work developed under a United States Government Contract.
+The Government retains a paid-up, nonexclusive, irrevocable worldwide license in such
+copyrighted data to reproduce, distribute copies to the public, prepare derivative works,
+perform publicly and display publicly and to permit others to do so.
+
+THIS SOFTWARE IS PROVIDED BY JEFFERSON SCIENCE ASSOCIATES LLC "AS IS" AND ANY EXPRESS
+OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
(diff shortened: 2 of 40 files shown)


wdconinc:
@c-dilks I'm going through and clarifying the licensing of EICrecon source in some cases. Many cases of code added by you (before you were an employee of Jefferson Lab) was made with the default language about top level LICENSE file, which didn't exist and could have been assumed to refer to the JSA public domain language in the NOTICE.md file. Since you retain the copyright, would you be willing to license this code under the LGPL?

src/detectors/BEMC/BEMC.cc
```diff
@@ -1,7 +1,5 @@
-// Copyright 2022, David Lawrence
-// Subject to the terms in the LICENSE file found in the top-level directory.
-//
-//
+// SPDX-License-Identifier: JSA
+// Copyright (C) 2022 David Lawrence
```
veprbl, line 2: Almost none of these attributions in src/detectors are correct. You, @wdconinc , are the primary author after all the conversions. The original configuration variables were derived from the LGPL3 code. Single item from API like `void InitPlugin(JApplication *app) {` is not subject to copyright.

wdconinc: You may be right about this. I was going to keep the first batch of commits limited to making a literal translation from the various headers we have to the recommended spdx headers. We can handle the changes in licensing through code review (for some files where I am the sole copyright holder, I have changed this from JSA license to LGPL3).

c-dilks:
> @c-dilks I'm going through and clarifying the licensing of EICrecon source in some cases. Many cases of code added by you (before you were an employee of Jefferson Lab) was made with the default language about top level LICENSE file, which didn't exist and could have been assumed to refer to the JSA public domain language in the NOTICE.md file. Since you retain the copyright, would you be willing to license this code under the LGPL?

I'm fine with LGPLv3. All of my contributions in any EIC organization-owned repository were before my employment at Jefferson Lab, except for small bug fixes.

Luigi and Alexander may not want their contributions under JSA, since I think we've been assuming we are under LGPLv3 and they are not Jefferson Lab employees.

src/detectors/BEMC/BEMC.cc
wdconinc: I'm asking and will be asking others (see https://github.com/eic/EICrecon/pull/1321#issuecomment-1977818095).

When these contributions were made, even by people who may be perfectly fine with LGPL3, the licensing language in the repository was (at best) ambiguous, and definitely not as unambiguously LGPL3 as it is now. Until Jan 15, 2024, https://github.com/eic/EICrecon/pull/1237, only the NOTICE.md was include at top-level, and it indicated that contributions were under the JSA public domain clause, unless otherwise indicated (such as for src/algorithms, since https://github.com/eic/EICrecon/pull/185, or for files with explicit LGPL-3.0-or-later SPDX header, such as most of src/factories). Any other code from before Jan 15 should be assumed to have been contributed under the conditions of the NOTICE.md file, and we should secure permissions from the copyright holders to license it under LGPL3. I think in that sense the only approach is to assume the contributions were made in line with the JSA notice, and ask all copyright holders to license them as LGPL3.

Some clear cases that can be resolved right away, however (and this isn't exhaustive):
- under src/algorithms, src/factories: this is all LGPL-3.0-or-later (possibly there are others), and it should be labeled like that,
- some code has been mistakenly labeled as JSA in the SPDX header (e.g. src/algorithms/tracking/TrackParamTruthInit.cc which came from LGPL-3.0-or-later juggler), and this should be fixed,
- some code has mistakenly not been labeled as LGPL-3.0-or-later (i.e. missing SPDX header), and this should be fixed.

I don't think there will be many issues getting people to agree to license their changes to LGPL3, though we may get one or two holdouts.

PR #1322, opened 2024-03-05
fix: don't trigger container build with VERSION

wdconinc:
This PR removes the VERSION variable from the trigger of the container build.

We specified the VERSION so the published container is identifiable. This also means that every trigger with a VERSION ends up getting published to docker hub, ghcr.io and the eicweb internal registry. This was fine for an occasional trigger, but it's overwhelming the registry storage (in particular eicweb) when we trigger a container for every commit on every branch, even with automatic cleanup rules.

So, with this PR we stop deploying the containers to tags with name eicrecon-$SHA.

- [x] Bug fix (issue: eicweb container registry instability)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -803,7 +803,6 @@ jobs:
           GITHUB_REPOSITORY=${{ github.repository }}
           GITHUB_SHA=${{ github.event.pull_request.head.sha || github.sha }}
           GITHUB_PR=${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


veprbl (approved):

veprbl:
This also means that we can't run benchmarks, since containers are not published?

wdconinc:
> This also means that we can't run benchmarks, since containers are not published?

They receive an internal tag (which includes the pipeline ID) that is cleaned up at the end of the pipeline (and all its sub-pipelines). All this PR does is not push the image to docker hub, ghcr.io, and eicweb under the official export tag.

PR #1323, opened 2024-03-06
fix: rework podio options to allow specifying list, includes, and excludes

### LINKED ISSUE #300: PODIO, add flag to save additional collections to the output file
Now there is `-Ppodio:output_include_collections` and if any value is written there it overrides full default list which now has ~60 names in it. In some cases it is very desired behavior and this should not be changed. But there should be a way to save additional data collections, which are not saved by default *in addition* to the default list. 

Something like this: 

```
# no flag means all 60 default collections will be written (if exists)
eicrecon

# only MCParticles and ReconstructedChargedParticles are saved (no other 60 default values)
eicrecon -Ppodio:output_include_collections=MCParticles,ReconstructedChargedParticles

# new proposed flag
# now MyCustomCollection and all 60 default values are saved
eicrecon -Ppodio:add_output_collections=MyCustomCollection
```

wdconinc:
This PR somewhat redefines the behavior of the option `-Ppodio:output_include_collections` in order to fix #300. There are now three similar options:
- `-Ppodio:output_collections` specifies which output collections to write,
- `-Ppodio:output_include_collections` specifies which output collections to include in addition to the default set,
- `-Ppodio:output_exclude_collections` specifies which output collections to exclude from the default set.

The behavior is technically backwards compatible. Where previously `-Ppodio:output_include_collections` resulted in only those collections being added to the output, now those collections will be added to the default set.

- [x] Bug fix (issue #300)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No. More collections will be written when using `-Ppodio:output_include_collections`.

### DIFF
--- a/.github/lsan.supp
+++ b/.github/lsan.supp
@@ -1,6 +1,7 @@
 leak:CherenkovPID::AddMassHypothesis
 leak:dd4hep::DetElement::DetElement
 leak:dd4hep::Header::Header
+leak:edm4eic::*::createBuffers
 leak:edm4hep::*::createBuffers
 leak:libActsCore.so
 leak:libCling.so
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -401,7 +401,7 @@ jobs:
           export DETECTOR_CONFIG=${DETECTOR}_${{ matrix.detector_config }}
           export LD_LIBRARY_PATH=$PWD/install/lib:$LD_LIBRARY_PATH
           export JANA_PLUGIN_PATH=$PWD/install/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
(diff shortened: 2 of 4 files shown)


veprbl:
I've been looking at this for a while now. I guess I have two concerns regarding this:

1. This continues with the current practice of special treatment of empty value in `podio:output_collections` to instead initialize it with the default list. In practice, there is a case for not outputting any collections, when a plugin with custom JANA processor is loaded (this is done, for example, in EICrecon "benchmarks").
2. This changes the meaning of `podio:output_include_collections`. The current user workflows are already using this parameter to improve performance of the reconstruction, or for other reasons. I think we could achieve your proposed naming, but, perhaps, it would be better to have some migration story first (e.g. deprecate the `podio:output_include_collections` before reintroducing it with a new meaning).

wdconinc:
Both reasonable concerns.

1. Happy to remove the special condition. Do we need a boolean `podio:output_copy_input_collections` or so?
2. We can deprecate it first. We'd need a command line argument deprecation strategy first: we could just print a warning (which won't be seen), or fail but accept a `allow_deprecated_options` for short term fixes. We can deprecate in 24.04, remove in 24.05, new syntax in 24.06.

PR #1324, opened 2024-03-09
remove JEventSourcePODIOLegacy

veprbl:
EventStore is removed as of PODIO 00-17-04


wdconinc (approved):
Should we update CMakeLists.txt to require a specific version of podio?

veprbl:
> Should we update CMakeLists.txt to require a specific version of podio?

No - this is backwards-compatible.

Issue #1264, opened 2024-01-29
linux-eic-shell.yml: No input available for capybara for push trigger


veprbl:
"Error: No artifacts found matching pattern '*.capy'" on `main` CI

https://github.com/eic/EICrecon/actions/runs/7700786196/job/20985889454#step:8:20

https://github.com/eic/EICrecon/blob/cb81e0a4c5185de21bd20bf9132bffee1e9dc5e8/.github/workflows/linux-eic-shell.yml#L846-L851

Looks like [action-download-artifact](https://github.com/dawidd6/action-download-artifact) doesn't work there.

https://github.com/eic/EICrecon/blob/cb81e0a4c5185de21bd20bf9132bffee1e9dc5e8/.github/workflows/linux-eic-shell.yml#L583C59-L583C74

Perhaps, this needs to use a commit of `github.ref_name + "~"`?

wdconinc:
https://github.com/eic/EICrecon/actions/runs/7838582681 @rahmans1

PR #1326, opened 2024-03-14
Add filter by association templated factory

simonge:
Adds a factory which creates a subset collection containing only members which are associated to the contents of another collection.

The motivation and example is in finding the reconstructed electron which is associated with the scattered electron. This is currently conducted independently in several of the reconstruction algorithms using copied code. Having a common collection will stop this being run several times. An alternative would be providing a single function which would clean up the code but still require running multiple times for the same result.

- [ ] Bug fix (issue #__)
- [x] New feature
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc:
Rebased to pick up hotfix #1332 for CI pipelines.

src/algorithms/meta/FilterByAssociations.h
```diff
@@ -0,0 +1,76 @@
...
+
+
+  template<typename T,typename TA, auto MemberFunctionPtr>
+  class FilterByAssociations : public FilterByAssociationsAlgorithm<T,TA> {
+
+    public:
+    FilterByAssociations(std::string_view name)
+      : FilterByAssociationsAlgorithm<T,TA>{name,
+                        {"inputCollection","associatedCollection"},
+                        {"isAssociatedCollection","isNotAssociatedCollection"},
```
veprbl, line 36: I don't think these values matter that much, but for the sake of documentation:
```suggestion
                        {"inputCollection", "inputAssociationsCollection"},
                        {"outputMatchedAssociations", "outputUnmatchedAssociations"},
```

src/algorithms/meta/FilterByAssociations.h
```diff
@@ -0,0 +1,76 @@
...
+#include <algorithms/algorithm.h>
+#include <string>
+#include <string_view>
+
+#include "services/log/Log_service.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+   template<class T, class TA>
```
veprbl, line 16: Can we improve code reader's experience by making a rename?
```suggestion
   template<typename ObjectT, typename AssociationObjectT>
```

src/algorithms/meta/FilterByAssociations.h
```diff
@@ -0,0 +1,76 @@
...
+          for (const auto& associatedEntry : *associatedEntries){
+
+            auto associationID = (associatedEntry.*MemberFunctionPtr)().getObjectID();
+
+            bool foundAssociation = false;
+
+            // Tries to find the association in the entries
+            for(const auto& entry : *entries){
+              if(entry.getObjectID() == associationID){
+                is_associated->push_back(associatedEntry);
```
veprbl, line 60: Any reason why we prefer to output associations and not the objects themselves?

src/global/beam/beam.cc
```diff
@@ -22,8 +22,8 @@ extern "C" {
     using namespace eicrecon;
 
     // Divide MCParticle collection based on generator status and PDG
-    std::vector<std::string> outCollections{"BeamElectrons","BeamProtons"};
-    std::vector<std::vector<int>> values{{4,11},{4,2212}};
+    std::vector<std::string> outCollections{"MCBeamElectrons","MCBeamProtons","MCScatteredElectrons","MCScatteredProtons"};
```
veprbl, line 25: Should we review addition of MC prefix separately? Do we expect an impact on the existing users?

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -124,6 +126,8 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "ReconstructedChargedParticleAssociations",
             "ReconstructedSeededChargedParticles",
             "ReconstructedSeededChargedParticleAssociations",
+            "MCScatteredElectronAssociations", // Remove if/when used internally
+            "MCNotScatteredElectronAssociations", // Remove if/when used internally
```
veprbl, line 130:
```suggestion
            "MCNonScatteredElectronAssociations", // Remove if/when used internally
```

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -47,8 +47,10 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
 
             // Truth record
             "MCParticles",
-            "BeamElectrons",
-            "BeamProtons",
+            "MCBeamElectrons", // Remove if/when used internally
+            "MCBeamProtons", // Remove if/when used internally
```
veprbl, line 51: Does the comment really apply here?

src/algorithms/meta/FilterByAssociations.h
```diff
@@ -0,0 +1,76 @@
...
+#include <spdlog/spdlog.h>
+#include <algorithms/algorithm.h>
+#include <string>
+#include <string_view>
+
+#include "services/log/Log_service.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
```
veprbl, line 15: Can we add a short comment explaining what this does in English?

veprbl (commented):
This makes sense to me. I have some suggestions, mostly regarding naming of things. Let me know what you think.

src/global/beam/beam.cc
simonge: This can be done separately, although there are no existing users as far as I'm aware. The beam directory and beam.cc should probably be renamed to something that encompasses more than just the beam anyway.

src/algorithms/meta/FilterByAssociations.h
simonge: I've been trying to work out if this is more generalisable. The motivation here is that the collection was already filtered e.g. into beam particles and the equivalent filter needed to be applied to the associations. If member functions of both the object and associated object are passed which need to match the inverse filter could be available by reversing the arguments. This could also allow one set of associations to filter by another without needing the object itself.

src/algorithms/meta/FilterByAssociations.h
simonge: Is the comment above sufficient and not too confusing?

veprbl: Yes. That's good now.

veprbl: Thank you

src/algorithms/meta/FilterByAssociations.h
```diff
@@ -0,0 +1,84 @@
...
+          for (const auto& associatedEntry : *associatedEntries){
+
+            auto associationID = ToFilterFunction(&associatedEntry);
+
+            bool foundAssociation = false;
+
+            // Tries to find the association in the entries
+            for(const auto& entry : *entries){
+              auto objectID = FilterByFunction(&entry);
+              if(entry.getObjectID() == associationID){
```
veprbl, line 67:
```suggestion
              if(objectID == associationID){
```

src/algorithms/meta/FilterByAssociations.h
```diff
@@ -0,0 +1,84 @@
...
+        >,
+       typename algorithms::Output<
+        typename ToFilterObjectT::collection_type,
+        typename ToFilterObjectT::collection_type
+        >
+     >;
+
+
+  template<typename FilterByObjectT,auto FilterByFunction,typename ToFilterObjectT,auto ToFilterFunction>
+  class FilterByAssociations : public FilterByAssociationsAlgorithm<FilterByObjectT,ToFilterObjectT> {
```
veprbl, line 36: At this point, there is nothing in this that requires ToFilterObjectT to be an association. We could come up with a simpler name
```suggestion
  class FilterMatching : public FilterMatchingAlgorithm<FilterByObjectT,ToFilterObjectT> {
```
Also I'd swap the arguments. Perhaps, it's more natural to expect the first argument to be propagated, even if in some case it will be an association.

simonge: The argument name and ordering had evolved to be a bit unnatural, thanks. This should now be fixed.

veprbl (approved):

wdconinc:
tl;dr The second auto-merge will succeed.

Huh... this failure to auto-merge one is funny:
- build_docs failed at https://github.com/eic/EICrecon/actions/runs/8806822970/job/24173020184 due to no *.capy for actions/upload-artifact/merge,
- there are no *.capy artifacts even though the eicrecon-dis jobs ran fine at https://github.com/eic/EICrecon/actions/runs/8806822970/job/24172669022,
- but the eicrecon-dis jobs didn't find a previous workflow because it is looking for `branch: gh-readonly-queue/main/pr-1326-14bc49be8ad313ab2311fbbc08163d2cf81dfab7` at https://github.com/eic/EICrecon/actions/runs/8806822970/job/24172669022#step:13:31,
- we tell the download_previous_artifacts step to download for `github.event.pull_request.base.ref || github.ref_name` at https://github.com/eic/EICrecon/blob/14bc49be8ad313ab2311fbbc08163d2cf81dfab7/.github/workflows/linux-eic-shell.yml#L831,
- since this is a merge_group event, we end up with `github.ref_name`, which is the branch name `gh-readonly-queue/main/pr-1326-14bc49be8ad313ab2311fbbc08163d2cf81dfab7`, where the hash is the current base ref in main,
- there is no previous benchmark for this branch name since it is the first time a merge_group runs for this base ref,
- running a second auto-merge seems like it will succeed since it will find the previous artifact of this first auto-merge.

All this seems to indicate we should change something, or just disable capybara when in merge_groups (including the merge).

veprbl:
@wdconinc I think, what you describe here is #1264. But that is probably not the problem. We need this instead #1397.

Issue #1254, opened 2024-01-25
Specific Jet Algorithms are Initialized Incorrectly


ruse-traveler:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): `main`
- Which version (or `HEAD` for the most recent on git): `HEAD`
- Any specific OS or system where the issue occurs? No
- Any special versions of ROOT or Geant4? No

Not all jet algorithms in `fastjet` are initialized the same way as anti-kt. So certain jet algorithms cause `fastjet` to throw an error when they're initialized with their name and the `rJet` parameter.

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Set the jet reconstruction algorithm to an algorithm with no parameters, e.g. `ee_kt_algorithm`

### Expected Result: (what do you expect when you execute the steps above)

The jet reconstruction algorithm should initialize `fastjet` appropriately for the algorithm.

### Actual Result: (what do you get when you execute the steps above)

`fastjet` will halt execution with an error similar to this one:

```
fastjet::error:  The jet algorithm you requested (50) should be constructed with 0 parameter(s) but was called with 1 parameter(s)
```

(In particular, this was for `ee_kt_algorithm`.)

PR #1331, opened 2024-03-15
Fix Jet Algorithm Initialization Bug

ruse-traveler:

This PR resolves issue #1254.  The `JetReconstruction` algorithm now determines the number of parameters needed for the selected fastjet algorithm and calls the relevant `JetDefinition` ctor.

In the process of resolving the issue, a new user-configurable parameter has been introduced into the algorithm's configuration: `pJet`, which sets the exponent on the energy weights of the distance measure in the case of the generalized kt algorithms.

- [X] Bug fix (issue #1254 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Rebased to pick up hotfix #1332 for CI pipelines.

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -83,16 +82,33 @@ namespace eicrecon {
...
+        break;
+
+      default:
+        m_jet_def = std::make_unique<JetDefinition>( m_mapJetAlgo[m_cfg.jetAlgo], m_cfg.rJet, m_mapRecombScheme[m_cfg.recombScheme] );
+        break;
+
+    }  // end switch (jet algorithm)
+
+    // Define jet area
+    m_area_def = std::make_unique<AreaDefinition>( m_mapAreaType[m_cfg.areaType], GhostedAreaSpec(m_cfg.ghostMaxRap, m_cfg.numGhostRepeat, m_cfg.ghostArea) );
```
simonge, line 104: Can this configuration not be called in the initialization?

johnlajoie: I agree with @simonge that it would seem to make more sense to move creating the jet and area definitions to the initialization as in principle this only needs to be done once.

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -83,16 +82,33 @@ namespace eicrecon {
...
 
     // Run the clustering, extract the jets
-    ClusterSequenceArea clus_seq(particles, jet_def, area_def);
-    std::vector<PseudoJet> jets = sorted_by_pt(clus_seq.inclusive_jets(m_cfg.minJetPt));
+    m_clus_seq = std::make_unique<ClusterSequenceArea>( particles, *m_jet_def, *m_area_def );
+    std::vector<PseudoJet> jets = sorted_by_pt( m_clus_seq -> inclusive_jets(m_cfg.minJetPt) );
```
veprbl, line 108:
```suggestion
    std::vector<PseudoJet> jets = sorted_by_pt(m_clus_seq->inclusive_jets(m_cfg.minJetPt));
```

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -83,16 +82,33 @@ namespace eicrecon {
...
 
     // Print out some infos
-    m_log->trace("  Clustering with : {}", jet_def.description());
+    m_log->trace("  Clustering with : {}", m_jet_def -> description());
```
veprbl, line 111:
```suggestion
    m_log->trace("  Clustering with : {}", m_jet_def->description());
```

src/algorithms/reco/JetReconstruction.cc
ruse-traveler: Oh! Good point! I'll definitely move this into `init()`...

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -50,6 +49,28 @@ namespace eicrecon {
       m_log->error(" Unknown area type \"{}\" specified!", m_cfg.areaType);
       throw JException(out.what());
     }
+
+    // Choose jet definition based on no. of parameters
+    const int nParams = get_num_params();
```
veprbl, line 54: Not clear why we need to have two separate switches to exchange nParams value when you can just check for algorithm type right here.

bspage912 (dismissed):
Modulo others' comments, changes seem fine.

src/algorithms/reco/JetReconstruction.cc
ruse-traveler: Ah! You're right! When I first coded it up I wanted to be _really_ explicit about what we were switching on, but just checking the algorithm type & some judicious comments will accomplish the same thing

veprbl (approved):
LGTM!

src/algorithms/reco/JetReconstruction.cc
veprbl: Looks much clearer now, thanks!

ruse-traveler:
Thanks for the feedback, all! Looks like discussion's done, so I'll go ahead and merge...

PR #1335, opened 2024-03-18
1300 putting real seeding covariance matrix error in a configuration file

Jeet-bhai:
This PR is about changing the initial covariance error matrix (loc a, loc b, q/p, theta, phi). In the default setting, these error values are hard coded and there has not been any study on them. We are proposing to put them in a config files, so that it's easier to study and change. Also, we are proposing a different values based on our study. The documents are attached here.

https://indico.bnl.gov/event/22499/contributions/88513/attachments/53179/90916/Covariance%20error%20matrix-3.pdf

[theta error, phi error.pdf](https://github.com/eic/EICrecon/files/14640624/theta.error.phi.error.pdf)

Screenshot 2024-03-18 at 2 23 30 PM

Screenshot 2024-03-18 at 2 23 51 PM

Screenshot 2024-03-18 at 2 24 07 PM

In the above table, we observe an improvement in all the measurements. 
Here "default" mean the current covariance matrix
"new" refers to what we are proposing

- [ x] Bug fix (issue #__)
- [ x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ x] Tests for the changes have been added
- [ x] Documentation has been added / updated
- [ x] Changes have been communicated to collaborators


bschmookler:
@ShujieL @osbornjd What do you think about these updated values for the initial covariance matrix?

The proposed new values are guided by the seed resolutions, and they improve the real-seeded tracking results with single-particle simulations, as @Jeet-bhai showed above.

Some additional tuning will probably be required using the DIS and DIS+background events, but I think the proposed values are a much better starting point for this compared to the current defaults.

ShujieL:
The change on location and q/p looks reasonable. But the theta error became a lot smaller. Did you do any check on that?

Jeet-bhai:
Hi Shujie,

Yup, the theta resolution for real seeding is very small (1-2 mrad). I have
attached a pdf titled "theta error, phi error.pdf" in the PR. It shows the
resolution plots and the data for various theta errors.

ShujieL (dismissed):

ShujieL:
> Hi Shujie, Yup, the theta resolution for real seeding is very small (1-2 mrad). I have attached a pdf titled "theta error, phi error.pdf" in the PR. It shows the resolution plots and the data for various theta errors.
> […](#)

Thanks it looks reasonable. @wdconinc I see one failed check with eic-container. Is that a real issue that will stoop us from merging?

osbornjd:
Only comment on this is that the angular resolutions look small - these will probably need to be revisited when thinking about performance in backgrounds etc. But I think you showed in some slides there was not a dramatic change when opening the angular covariance windows up, so that is probably enough to motivate these choices for now.

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
```diff
@@ -78,5 +78,13 @@ namespace eicrecon {
     float  m_seedConfMaxZOrigin_forw = 150.0 * Acts::UnitConstants::mm;
     float  m_minImpactSeedConf_forw = 1.0 * Acts::UnitConstants::mm;
 
+    //////////////////////////////////////
+    ///Seed Covariance Error Matrix
+    float m_Loc_a_Error = 1.5 * Acts::UnitConstants::mm;    //Error on Loc a
+    float m_Loc_b_Error = 1.5 * Acts::UnitConstants::mm;    //Error on Loc b
```
wdconinc, line 84: Comment here and throught out the use of this parameter is to use lower case for consistency in the naming scheme (start lower case, capital for new words):
```suggestion
    float m_loc_a_Error = 1.5 * Acts::UnitConstants::mm;    //Error on Loc a
    float m_loc_b_Error = 1.5 * Acts::UnitConstants::mm;    //Error on Loc b
```

src/algorithms/tracking/TrackSeeding.cc
```diff
@@ -239,18 +239,18 @@ std::unique_ptr<edm4eic::TrackParametersCollection> eicrecon::TrackSeeding::make
...
-        cov(0,0) = 0.1; // loc0
-        cov(1,1) = 0.1; // loc1
-        cov(2,2) = 0.05; // theta
-        cov(3,3) = 0.05; // phi
-        cov(4,4) = 0.05; // qOverP
-        cov(5,5) = 0.1; // time
+        cov(0,0) = m_cfg.m_Loc_a_Error; // loc0
+        cov(1,1) = m_cfg.m_Loc_b_Error; // loc1
+        cov(2,2) = m_cfg.m_theta_Error; // theta
+        cov(3,3) = m_cfg.m_phi_Error; // phi
```
wdconinc, line 245: We should use the same ordering as in Acts here: https://github.com/acts-project/acts/blob/3eb0b4febd16ad8cee5664302e3542044ab17d35/Core/include/Acts/Definitions/TrackParametrization.hpp#L37-L56

```suggestion
        cov(2,2) = m_cfg.m_phi_Error; // phi
        cov(3,3) = m_cfg.m_theta_Error; // theta
```

And you may want to double check that this is then read in correctly where the seeds are used too.

wdconinc (changes requested):

src/algorithms/tracking/TrackSeeding.cc
bschmookler: I guess this should be adjusted for the truth-seeding as well then:

[https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackParamTruthInit.cc#L139-L140](https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackParamTruthInit.cc#L139-L140)

bschmookler: Based on this code in CKFTracking.cc:

```
Acts::BoundSquareMatrix cov = Acts::BoundSquareMatrix::Zero();
#if EDM4EIC_VERSION_MAJOR >= 5
  for (size_t i = 0; const auto& [a, x] : edm4eic_indexed_units) {
   for (size_t j = 0; const auto& [b, y] : edm4eic_indexed_units) {
      cov(a, b) = track_parameter.getCovariance()(i,j) * x * y;
      ++j;
   }
   ++i;
  }
```

I think that, for example, the `cov(2,2)` value set in TrackSeeding.cc will be treated as initial value of `cov(2,2)` in CKFTracking.cc. Is that correct @wdconinc?

But this means that Jeet's studies of the impact of the initial theta (phi) error might have actually been studies of the initial phi (theta) error. So, we should probably check this next week with the correct indexing.

bschmookler: Sure. I can do that. 

But why do you think something needs to be changed in CKFTracking? It looks like it is reading the correct indices for the initial covariance matrix. It is just that cov(2,2) corresponds to phi and cov(3,3) corresponds to theta.

So, in the truth seeding code (lines 139-140), where we currently do have different initial errors for theta and phi, we should change

 ```
cov(2,2) = 0.01; // theta
cov(3,3) = 0.05; // phi
```

to

 ```
cov(2,2) = 0.05; // phi
cov(3,3) = 0.01; // theta
```

And that would correct it (for the truth-seeding case)?

bschmookler: I guess you mean these lines should change:

[https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/CKFTracking.cc#L72-L73](https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/CKFTracking.cc#L72-L73)

bschmookler: I think I'll retract my previous statement that we might have been previously been studying the effect of the initial theta (phi) when we thought we were looking at the initial phi (theta) error. If the loop was reading in from `edm4eic_indexed_units` in order, then it was previously doing something like:

 `cov(Acts::eBoundTheta, Acts::eBoundTheta) =  track_parameter.getCovariance()(2,2);`

with `Acts::eBoundTheta==3`.  So, previously the CKF was being passed the correct error for theta and phi, although in a confusing way.

If that is the case, there should not be any effect from this new indexing. But, @Jeet-bhai and I can double check next week.

bschmookler:
I updated the capitalization as Wouter suggested. 

I also removed the units from the errors in the configuration file, since they are applied in the CKFTracking. (All the units used were unity; so, it won't have any effect.)

src/algorithms/tracking/TrackSeeding.cc
wdconinc: @bschmookler I'm ok with resolving this conversation (i.e. no outstanding concerns). Are you too? If so, please click resolve.

Then we can get back to overall review and approval.

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
```diff
@@ -78,5 +78,13 @@ namespace eicrecon {
...
     float  m_minImpactSeedConf_forw = 1.0 * Acts::UnitConstants::mm;
 
+    //////////////////////////////////////
+    ///Seed Covariance Error Matrix
+    float m_loc_a_Error = 1.5;     //Error on Loc a
+    float m_loc_b_Error = 1.5;     //Error on Loc b
+    float m_phi_Error = 0.02;      //Error on phi
+    float m_theta_Error = 0.002;   //Error on theta
+    float m_qOverP_Error = 0.025;  //Error on q over p
+    float m_time_Error = 0.1;      //Error on time
```
wdconinc, line 88: Physical quantities should have units.
```suggestion
    float m_loc_a_Error = 1.5 * Acts::UnitConstants::mm;     //Error on Loc a
    float m_loc_b_Error = 1.5 * Acts::UnitConstants::mm;     //Error on Loc b
    float m_phi_Error = 0.02 * Acts::UnitConstants::rad;     //Error on phi
    float m_theta_Error = 0.002 * Acts::UnitConstants::rad;  //Error on theta
    float m_qOverP_Error = 0.025 / Acts::UnitConstants::GeV; //Error on q over p
    float m_time_Error = 0.1 * Acts::UnitConstants::mm;      //Error on time
```
These units are such that they are multiplying by 1 (Acts uses units of mm, rad, GeV, and time is defined in terms of distance, https://acts.readthedocs.io/en/latest/core/definitions/units.html).

We may want to consider (in the future) changing the last line to:
```
    float m_time_Error = 0.333 * Acts::UnitConstants::ps;      //Error on time
```
This is what you have there now (but someone should double-check my algebra).

bschmookler: I understand wanting to put units here. But aren't we in principle 'double applying' the units since they are used here as well:

https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/CKFTracking.cc#L70-L75

Since `Acts::UnitConstants::mm`, 'Acts::UnitConstants::rad', and 'Acts::UnitConstants::GeV' are all equal to 1, this won't have any practical effect.

This may be important to consider for the time in the future, since we currently mulitply by 'Acts::UnitConstants::ns', in CKFTracking which is not equal to 1. Right now, we have

`float m_time_Error = 0.1`

which I guess is getting converted in the CKF tracking to `0.1ns = 0.1 * 1e-9 * 3e11 = 30mm`.

bschmookler:
@Jeet-bhai repeated his studies and the results look consistent with the updated branch (updated theta/phi indices). So, I think this can be merged once the unit issue question above is resolved. 

As @osbornjd mention, as a next step we can do some tests using DIS events. Having the parameters in the configuration file is useful for this, since we can adjust them on the command line.

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
bschmookler: @wdconinc. What do you think about my comment yesterday? If you think it is best to add the mm, rad, GeV units to the values in the configuration file, I'll do that. Thanks.

wdconinc: We should apply units where the numerical dimensioned quantities are specified. The units in `edm4eic_indexed_units` are not units but unit **conversion** ratio when going from EDM4eic quantities to Acts quantities. The rules are:
- when you write a number, you need to multiply with a unit,
- when you change from one unit domain (EDM4eic, Acts, DD4hep) to another, you need a unit conversion ratio.

In the following line:
https://github.com/eic/EICrecon/blob/ac3821d47eb4cbd6f39bddaac19710a7840104e0/src/algorithms/tracking/CKFTracking.cc#L169
we change from EDM4eic to Acts unit domains with the unit conversion ratios.

In e.g. the following line:
https://github.com/eic/EICrecon/blob/ac3821d47eb4cbd6f39bddaac19710a7840104e0/src/algorithms/tracking/OrthogonalTrackSeedingConfig.h#L79
an explicit quantity is defined in the Acts unit domain. This is the same file you want to add other quantities to in this PR, so it should, for consistency sake, be in the same Acts unit domain, which is manifested by using the Acts::UnitConstants.

In the following lines:
https://github.com/eic/EICrecon/blob/ac3821d47eb4cbd6f39bddaac19710a7840104e0/src/algorithms/tracking/TrackSeeding.cc#L241-L248
this is where the quantities determined in the Acts unit domain are written into EDM4hep unit domain objects. There should be a unit conversion here. In many cases, we have an explicit unit division, so instead of
https://github.com/eic/EICrecon/blob/8469226a42ee9a39170e5ec102913b0020455282/src/algorithms/tracking/TrackSeeding.cc#L242-L247
here we would need:
```
 cov(0,0) = m_cfg.m_loc_a_Error / Acts::UnitConstants::mm; // loc0 
 cov(1,1) = m_cfg.m_loc_b_Error / Acts::UnitConstants::mm; // loc1 
 cov(2,2) = m_cfg.m_phi_Error / Acts::UnitConstants::rad; // phi 
 cov(3,3) = m_cfg.m_theta_Error / Acts::UnitConstants::rad; // theta 
 cov(4,4) = m_cfg.m_qOverP_Error * Acts::UnitConstants::GeV; // qOverP 
 cov(5,5) = m_cfg.m_time_Error / Acts::UnitConstants::ns; // time
```
following the EDM4eic units in https://github.com/eic/EDM4eic/blob/0bd6edd97d0b28b30605ab90dcf338a3244d16dd/edm4eic.yaml#L409-L413.

wdconinc: But I will agree that we do need more explicit treatment of EDM4eic units, see https://github.com/eic/EDM4eic/issues/33, https://github.com/eic/EICrecon/issues/925, https://github.com/AIDASoft/podio/pull/445.

bschmookler: Thanks for the detailed explanation. I will add the units to both the configuration file and the trackseeding code.

src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
```diff
@@ -78,5 +78,13 @@ namespace eicrecon {
...
     float  m_minImpactSeedConf_forw = 1.0 * Acts::UnitConstants::mm;
 
+    //////////////////////////////////////
+    ///Seed Covariance Error Matrix
+    float m_loc_a_Error = 1.5 * Acts::UnitConstants::mm;     //Error on Loc a
+    float m_loc_b_Error = 1.5 * Acts::UnitConstants::mm;     //Error on Loc b
+    float m_phi_Error = 0.02 * Acts::UnitConstants::rad;     //Error on phi
+    float m_theta_Error = 0.002 * Acts::UnitConstants::rad;  //Error on theta
+    float m_qOverP_Error = 0.025 / Acts::UnitConstants::GeV; //Error on q over p
+    float m_time_Error = 0.1 * Acts::UnitConstants::mm;      //Error on time
```
wdconinc, line 88:
```suggestion
    float m_time_Error = 0.1 * Acts::UnitConstants::mm;      //Error on time
    // Note: Acts native time units are mm: https://acts.readthedocs.io/en/latest/core/definitions/units.html
```

wdconinc (approved):

wdconinc (approved):

PR #1344, opened 2024-03-21
Update JetReconstruction to algorithm interface

johnlajoie:

Converts the JetReconstruction module to the new algorithms interface

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, output from the JetReconstruction algorithm is the same as before. 

No


veprbl (dismissed):

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -79,11 +80,13 @@ namespace eicrecon {
...
 
-
-  template <typename T> std::unique_ptr<edm4eic::ReconstructedParticleCollection> JetReconstruction::process(const T* input_collection) {
-
-    // Store the jets
-    std::unique_ptr<edm4eic::ReconstructedParticleCollection> jet_collection { std::make_unique<edm4eic::ReconstructedParticleCollection>() };
+  void JetReconstruction::process(
+                                  const JetReconstruction::Input& input,
+                                  const JetReconstruction::Output& output
+                                 ) const {
```
ruse-traveler, line 86: I touched on this offline, @johnlajoie , but one thing to consider is how to handle the templating here. I suppose a question for @veprbl , @wdconinc , and @nathanwbrei: can the algorithms interface handle templated Inputs? e.g.
```
template <typename T>
using JetReconstructionAlgorithm = algorithms::Algorithm<
      algorithms::Input<T>,
      algorithms::Output<edm4eic::ReconstructedParticle>
 >;
```

Because if possible, I'd prefer to keep the templating in place, especially with formal tracks coming in the near future...

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -105,13 +108,13 @@ namespace eicrecon {
...
-      return jet_collection;
+      return;
     }
     m_log->trace("  Number of particles: {}", particles.size());
 
     // Run the clustering, extract the jets
-    m_clus_seq = std::make_unique<ClusterSequenceArea>(particles, *m_jet_def, *m_area_def);
-    std::vector<PseudoJet> jets = sorted_by_pt(m_clus_seq->inclusive_jets(m_cfg.minJetPt));
+    fastjet::ClusterSequenceArea m_clus_seq(particles, *m_jet_def, *m_area_def);
+    std::vector<PseudoJet> jets = sorted_by_pt(m_clus_seq.inclusive_jets(m_cfg.minJetPt));
```
ruse-traveler, line 117: I'm still mystified that the pointer version doesn't compile with the changes... 🤔 

Either way, if we do implement this change, please don't forget to remove the corresponding `std::unique_ptr<etc> m_clus_seq` from the header!

src/algorithms/reco/JetReconstruction.h
```diff
@@ -11,22 +12,44 @@
...
+        public WithPodConfig<JetReconstructionConfig> {
+
+    public:
+
+    JetReconstruction(std::string_view name) :
+      JetReconstructionAlgorithm {
+        name,
+        {"inputReconstructedParticles"},
+        {"outputReconstructedParticles"},
+        "Performs jet reconstruction using a configured FastJet algorithm."
```
ruse-traveler, line 43:
```suggestion
        "Performs jet reconstruction using a FastJet algorithm."
```

ruse-traveler (changes requested):

src/algorithms/reco/JetReconstruction.cc
veprbl: Alias templates are a thing https://en.cppreference.com/w/cpp/language/type_alias apparently

wdconinc: I don't see why this shouldn't work. It's exactly the same as passing a `std::vector<T>` in the variadic input factories, https://github.com/eic/EICrecon/blob/main/src/algorithms/meta/CollectionCollector.h#L14.

veprbl: Implemented

src/algorithms/reco/JetReconstruction.h
veprbl: Done

johnlajoie:
Thanks for the help with the template conversion @veprbl !

veprbl (approved):

ruse-traveler (approved):
LGTM!

PR #1346, opened 2024-03-24
treewide: use PODIO ObjectID instead SimID/RecID indices

veprbl:

This is a subset of #970, that removes use of SimID and RecID fields (while switching to comparisons of full object id's instead of just the indices).
In principle, we could even drop `.getObjectID()` in comparisons and just use `Object::operator==` https://github.com/AIDASoft/podio/blob/400f4f06bef3e7666c71e31be579ec715c412b8a/python/templates/macros/implementations.jinja2#L160, which relies on pointer comparison instead.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/pid/ParticlesWithPID.cc
+++ b/src/algorithms/pid/ParticlesWithPID.cc
@@ -198,7 +198,7 @@ namespace eicrecon {
                                  mcpart.getPDG());
 
                     m_log->debug(" Assoc: id={} SimId={} RecId={}",
-                                 rec_assoc.getObjectID().index, rec_assoc.getSimID(), rec_assoc.getSimID());
+                                 rec_assoc.getObjectID().index, rec_assoc.getSim().getObjectID().index, rec_assoc.getSim().getObjectID().index);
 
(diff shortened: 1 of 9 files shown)


simonge:
If this is just a purge of using SimID and RecID then looks good. Otherwise it might be a good idea to remove all calls to `.index` other than in log outputs so the check that it really is the same object from the correct collection is in place and not something from another collection with the same index.

What is the argument against comparing the objects as opposed to using `getObjectID`?

veprbl:
> If this is just a purge of using SimID and RecID then looks good. Otherwise it might be a good idea to remove all calls to `.index` other than in log outputs so the check that it really is the same object from the correct collection is in place and not something from another collection with the same index.

Note, this is already removing calls to `.index` in comparisons, so both index and collection ID are being compared.

> 
> What is the argument against comparing the objects as opposed to using `getObjectID`?

Not really an argument. I'm just having hard time convincing myself that pointer comparison will always do what we want. You think it's fine?

simonge (approved):
MatchClusters wouldn't be a simple change so happy within the PR scope

PR #1347, opened 2024-03-24
linux-eic-shell.yml: run eicrecon with janatop

veprbl:
This will show us what slows down the reconstruction.


wdconinc (approved):
Good idea! (spoiler: it's the track projector)

Issue #1337, opened 2024-03-19
Channel Masking in Digitization


ruse-traveler:
The ability to mask specific channels/apply a dead map will be crucial to assessing the performance of calorimeters such as the LFHCAL. Currently, there is no way to handle this in EICrecon, and implementing this post-facto in afterburners will be immensely challenging.

**Describe the solution you'd like**
This functionality could be implemented in the digitization algorithm in a manner similar to the adjacency matrix, where the user provides an expression via a string which identifies which channels to turn off.

**Additional context**
This pertains to one of the identified high-priority TDR needs from the calorimetry side. Namely, a more robust digitization + noise model.

Issue #799, opened 2023-07-26
CalorimeterHitDigi,CalorimeterHitReco: Calibration constant to be computed from an expression string


veprbl:
Using an expression evaluator (similar to an adjacencyMatrix) would allow to define non-linear calibrations, calibrations dependent on location and layer-dependent calibrations like those needed to resolve #796.

veprbl:
This helps with leet issue #1337

PR #1349, opened 2024-03-25
CalorimeterHitReco: use TCling to define SF via expressions

veprbl:

This introduces an ability to define sampling fraction that varies across the calorimeter cells.
This should enable following kind of studies for TDR:
 - Non-uniform tower gains (including dead channels)
 Example: `-PEEMC:EcalEndcapNRecHits:samplingFraction='[row, column](){ static map<std::tuple<double, double>, double> sf; auto index = std::make_tuple(row, column); if (!sf.count(index)) { std::default_random_engine gen(row + 100 * column); std::normal_distribution R{1.0, 0.1}; sf[index] = R(gen); } return sf[index]; }()'` (yes, the static variables in lambda are actually static in the scope it was defined in, so this provides caching)
 - Rapidity-dependent calibrations

Use of TCling is very ugly and inefficient, this is not mean to be permanent (... they always say). We should be replacing this with some proper database, or a better scripting language.

Closes: #799
Resolves: #1337

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -39,7 +38,7 @@ using namespace dd4hep;
 
 namespace eicrecon {
 
-void CalorimeterHitReco::init() {
+void CalorimeterHitReco::init(std::shared_ptr<Evaluator_service> evaluator) {
```
wdconinc, line 41: Can we avoid this? We are (were) moving towards a uniform interface for all algorithms: `void init() final` and `void process(const Input&, const Output&) const final`, so we don't need to rewrite all these factories all the time. Is there a way that we can pass this evaluator service through other means?

@sly2j for potential `algorithms::` interface ideas.

veprbl: Can we register JANA service as algorithms service?

src/algorithms/calorimetry/CalorimeterHitReco.h
```diff
@@ -47,15 +51,20 @@ namespace eicrecon {
                             {"outputRecHitCollection"},
                             "Reconstruct hit from digitized input."} {}
 
-    void init() final;
+    void init();
     void process(const Input&, const Output&) const final;
 
   private:
 
+    algorithms::Service<EvaluatorSvc> m_evaluatorSvc {""};
```
veprbl, line 59: This is probably incorrect.

src/algorithms/calorimetry/CalorimeterHitRecoConfig.h
```diff
@@ -23,8 +23,7 @@ namespace eicrecon {
     double                   thresholdValue{0};
 
     // sampling fraction
-    double                   sampFrac{1.0};
-    std::vector<double>      sampFracLayer{};
+    std::string              sampFrac{"1.0"};
```
wdconinc, line 26: Can we handle this as follows?
```suggestion
    std::variant<double,std::string> sampFrac{1.0};
```

Main benefit would actually be for expansion of this pattern. If we start to roll this out as everywhere, but it it requires us to change defaults to runtime-evaluated strings, we lose the compile-time validity checks on numerical values.

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -156,23 +156,7 @@ extern "C" {
...
-              0.037, //  6
-              0.037, //  7
-              0.037, //  8
-              0.037, //  9
-              0.037, // 10
-              0.037, // 11
-              0.037, // 12
-              0.037, // 13
-            },
+            .sampFrac = "(rlayerz == 0) ? 0.019 : 0.037", // 0.019 only in the 0-th tile
```
wdconinc, line 159: Nice :-)

src/factories/calorimetry/CalorimeterHitReco_factory.h
```diff
@@ -28,7 +28,7 @@ class CalorimeterHitReco_factory : public JOmniFactory<CalorimeterHitReco_factor
     ParameterRef<double> m_resolutionTDC {this, "resolutionTDC", config().resolutionTDC};
     ParameterRef<double> m_thresholdFactor {this, "thresholdFactor", config().thresholdFactor};
     ParameterRef<double> m_thresholdValue {this, "thresholdValue", config().thresholdValue};
-    ParameterRef<double> m_samplingFraction {this, "samplingFraction", config().sampFrac};
+    ParameterRef<std::string> m_samplingFraction {this, "samplingFraction", config().sampFrac};
```
wdconinc, line 31: Though this probably doesn't handle variants... Maybe we can somehow 'pick one' here, naturally the string. Maybe:
```
    ParameterRef<std::string> m_samplingFraction {this, "samplingFraction", std::get<std::string>(config().sampFrac)};
```

src/algorithms/calorimetry/CalorimeterHitReco.h
```diff
@@ -47,7 +49,7 @@ namespace eicrecon {
                             {"outputRecHitCollection"},
                             "Reconstruct hit from digitized input."} {}
 
-    void init() final;
+    void init();
```
wdconinc, line 52: Can still be `final`.

src/services/evaluator/EvaluatorSvc.cc
```diff
@@ -0,0 +1,46 @@
...
+#include "EvaluatorSvc.h"
+
+namespace eicrecon {
+
+void EvaluatorSvc::init() {
+}
+
+std::function<double(const std::unordered_map<std::string, double>&)>
+EvaluatorSvc::_compile(const std::string& expr, std::vector<std::string> params) {
+  std::string func_name = fmt::format("_eicrecon_{}", function_id++);
```
wdconinc, line 20:
```
std::lock_guard<std::mutex> lock(m_mutex);
```

src/algorithms/calorimetry/CalorimeterHitReco.h
veprbl: Nice catch

src/services/evaluator/EvaluatorSvc.cc
veprbl: Done

veprbl:
@ruse-traveler Check this out. This is like adjacency matrix but for your calibration constants.

ruse-traveler:
> @ruse-traveler Check this out. This is like adjacency matrix but for your calibration constants.

Nice! This is very cool! So to check my understanding: if I wanted to 0 out some channels, I would do something like the following, right?
```
.sampFrac = "(eta == phi) ? 0.0 : 0.033"  // unalive all channels on a diagonal, otherwise use default
```

(c.c. @FriederikeBock )

veprbl:
Yes. That's the idea. The plan is to apply the same for CalorimeterHitDigi, and add additional parameters that can be used, including energy for a non-linear response.

src/algorithms/calorimetry/CalorimeterHitRecoConfig.h
wdconinc: Just to make sure you don't think you need to keep this open on my behalf, I'm ok moving away from the variant approach if it doesn't work technically, but I am worried about what we lose when we stick everything in strings including constants.

ruse-traveler:
> Yes. That's the idea. The plan is to apply the same for CalorimeterHitDigi, and add additional parameters that can be used, including energy for a non-linear response.

Gotcha! So then this is also where we would stick additional "noise" if we wanted something other than what's in the vanilla algorithm...

ruse-traveler (approved):
Though I do share Wouter's concern about defining constants in strings, I would be okay with integrating these changes as-is if the variant approach doesn't work. This Should address the two primary digitization needs from the perspective of the LFHCAL..

ruse-traveler:
Checking in on this: aside from the conflicts, are you ready to merge this @veprbl ?

veprbl:
> Checking in on this: aside from the conflicts, are you ready to merge this @veprbl ?

Yes. This needs another approve.

ruse-traveler (dismissed):
👍

veprbl:
@ruse-traveler One more time, please :)

ruse-traveler (approved):
👍 👍

PR #1311, opened 2024-02-27
feat: propagate AlgorithmsInit_service use for algorithms::GeoSvc

wdconinc:
### Briefly, what does this PR introduce?
This propagates #1195 to the other geometry users that already use the algorithms interfaces.

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue: roll out proof of concept and initial implementation of #1195 to others)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1351, opened 2024-03-26
CalorimeterHitDigi: require DD4hep readout to be present

veprbl:
This will be later relied on, when we introduce ability to define corrMeanScale that is dependent on cellIDs and such.

This is a prerequisite of implementing #1349 for CalorimeterHitDigi.


src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -74,31 +74,33 @@ void CalorimeterHitDigi::init() {
     tRes       = m_cfg.tRes / dd4hep::ns;
     stepTDC    = dd4hep::ns / m_cfg.resolutionTDC;
 
+    // sanity checks
+    if (!m_geo.detector()) {
+        error("Unable to locate geometry");
+        throw std::runtime_error("Unable to locate Geometry Service");
+    }
```
wdconinc, line 81: I know we probably shouldn't rely on the interface remaining constant in this regards, but this is currently defined as:
```
gsl::not_null<const dd4hep::Detector*> detector() const { return m_detector; }
```
so this cannot ever happen.

wdconinc (dismissed):
Looks good.

ruse-traveler (dismissed):
Nice! Looks good!

veprbl:
This needed a rebase after the #1311. Can I get a reapproval, please?

wdconinc (approved):
Congratulations! You have been pre-approved!

Issue #1320, opened 2024-03-04
Clusters should be associated to ReconstructedParticles before they go into electron finder


veprbl:
### Two questions:
- We are cloning here instead of reusing the reconstructed electron and storing it in a subset collection. Is that what we want? If we think of the ElectronReconstruction algorithm as following the cluster matching step, then electron finding should not change the electron and a clone should not be necessary, right?
- Relatedly, why isn't `reco_part` not yet filled with clusters? Shouldn't we be using the `ReconstructedParticles`  (tracking + PID + cluster matching) as input here instead of the `ReconstructedChargedParticles` (which are from tracking + PID). We seem to be repeating the cluster matching that is being done in `MatchClusters` here.

_Originally posted by @wdconinc in https://github.com/eic/EICrecon/pull/1276#discussion_r1477107306_

PR #1277, opened 2024-02-03
Algorithms and factories for scattered electrons

jdbrice:
### Briefly, what does this PR introduce?
This PR introduces a key part of the electron finder reco chain. It provides a ranked list of ReconstructedParticles believed to be the scattered electrons. Two different approaches are included for now.

- ScatteredElectronsTruth (algorithms / factory) find the scattered electron ( as a ReconstructedParticle) utilizing truth information. The MC particle collection is used to identify the scattered electron and McReco association are used to determine the corresponding reconstructed particle. 
- ScatteredElectronsEMinusPz (algorithms / factory) computes E-Pz for each electron candidate identified by the ReconstructedElectron_factory. The reconstructed particle candidates are sorted into descending order (maximum E-Pz first) and saved if they pass the configured min/max E-Pz cut.

Add 2 new podio output collections:
"ScatteredElectronsTruth"
"ScatteredElectronsEMinusPz"

The idea has been discussed in various meeting including the jan collaboration meeting. 

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x ] New feature (issue #__) Electron Finder
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
NO

### Does this PR change default behavior?
YES
- adds the above two collections to polio output by default
- Also, I found that the E-Pz algorithm usually gets the correct electron even without strict electron PID cuts (E/p) - so in order to not lose candidates (as was happening with existing cuts) I made the default E/p cuts for the ElectronReconstruction_factory wider (0.7 < E/p < 1.3) for electron candidates

PR #1355, opened 2024-03-28
Adapt ElectronReconstruction to use MatchClusters

tlprotzman:
The ReconstructedElectron algorithm now uses the particle-cluster associations created by MatchClusters_factory rather than doing it itself.  It makes the collection ReconstructedElectrons a subset collection of ReconstructedParticles, so the association is kept between them.  

I will leave this as a draft until #1277 is merged, then update that code as well to take advantage of it.

- [x] Bug fix (issue #1320 )
- [ ] New feature (issue #__ )
- [ ] Documentation update
- [ ] Other: __

ReconstructedElectrons is now a subset collection, but I don't believe that should break anything.

No

### DIFF
--- a/src/algorithms/reco/ElectronReconstruction.cc
+++ b/src/algorithms/reco/ElectronReconstruction.cc
@@ -3,6 +3,7 @@
 #include "ElectronReconstruction.h"
 
 #include <edm4eic/ClusterCollection.h>
+#include <edm4eic/ReconstructedParticleCollection.h>
 #include <edm4hep/utils/vector_utils.h>
 #include <fmt/core.h>
 #include <podio/ObjectID.h>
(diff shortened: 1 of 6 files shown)


tlprotzman:
Hi @simonge, @wdconinc, while I was working on this it occurred to me that the output of the scattered electron algorithms should be a subset collection of the `ReconstructedElectronsForDIS` collection as well.  Before I do that, I wanted to check that it makes sense to you as well?  A subset of a subset would maintain the relation all the way up to the `ReconstructedParticles` collection, right?

simonge:
> > A subset of a subset would maintain the relation all the way up to the ReconstructedParticles collection, right?
> 
> I haven't used a subset of a subset, but these are really just reference-storing containers, so expected behavior should just store a reference to the original object. If it doesn't, I think podio devs would probably agree that that is a bug and fix it there. @veprbl have you done this?

I believe I've done this before and the behaviour was as expected, it just refers to the collection that owns the objects. It would be easy enough to test again.

veprbl:
I'm uploading capybara report of this for posterity
https://veprbl.github.io/capybara-reports/1fe13e24b65152347eca5828ffcffd80/#ReconstructedElectrons

There is a very minor efficiency loss, and it shows some electrons have charge +1.

veprbl (approved):
Looks good, and addresses the issue.

Issue #1330, opened 2024-03-15
Add support for ACTS::CylindricalSurfaces in obj output from `draw_surfaces`


mposik1983:
**Is your feature request related to a problem? Please describe.**
EICrecon takes DD4HEP surfaces and casts them to ACTS::PlanarSurfaces, however we have DD4HEP geometries which are curved and can not be cast to PlanarSurfaces. Extend EICrecon to use correct ACTS surface based on surface geometry. 

`acts_init] [error] Warning: Attempting cast a Acts::CylinderSurface to Acts::PlaneSurface returns nullptr. This surface will not be added to the .obj output.`

**Describe the solution you'd like**
Allow for curved surfaces to be cast to ACTS::CylindricalSurface

**Describe alternatives you've considered**
none.

**Additional context**

wdconinc:
The error is produced by the following line:
https://github.com/eic/EICrecon/blob/ac3821d47eb4cbd6f39bddaac19710a7840104e0/src/algorithms/tracking/ActsGeometryProvider.cc#L72

That code only(!) creates an obj file with the tracking geometry. The obj format only (easily) supports rectangular surfaces (through simple meshes), so we don't currently export cylinders.

wdconinc:
I've added #1357 which exports full tracking geometry as obj + mtl, and as ply files. You can open those with [meshlab](https://www.meshlab.net) (just open the whole set at once). It approximates the cylinder surfaces through triangulations.

PR #1357, opened 2024-04-01
feat: replace draw_surfaces with Acts-based obj and ply export

wdconinc:
Instead of writing our own obj files in `draw_surfaces` (with only support for planar surfaces), this PR converts it to use the Acts support for export to obj and ply files, optionally and with parameter support for coloring etc. Syntax is:
```
eicrecon -Pacts:WriteObj=true -Pacts:WritePly=true -Pacts:OutputDir=$PWD/tracking_geometry
```
where `OutputDir` must exist and defaults to current working directory (because it exists; upstream choice).

To open the obj and ply models, you can use e.g. [meshlab](https://www.meshlab.net). Just import all files at once.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1330)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, by default `tracking_geometry.obj` will not be written anymore.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -740,7 +740,7 @@ jobs:
           export DETECTOR_CONFIG=${DETECTOR}_${{ matrix.detector_config }}
           export LD_LIBRARY_PATH=$PWD/install/lib:$LD_LIBRARY_PATH
           export JANA_PLUGIN_PATH=$PWD/install/lib/EICrecon/plugins:/usr/local/plugins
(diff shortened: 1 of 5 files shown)


wdconinc:
This is probably going to look mostly like modern art, but here is what all these layers look like in meshlab (default color assignments):

image

mposik1983 (approved):

wdconinc:
I will go ahead and merge this already. The reason why this is exported for each DIS test sample (even though this only depends on geometry currently) is because I am working on exports of event data as well (covariance ellipses for measurements, etc), see https://acts.readthedocs.io/en/latest/core/visualization/3d.html. That's likely something that should be done in a next PR.

PR #1359, opened 2024-04-02
feat: Bucket silicon raw hits in time

kkauder:
For longer time slices, hits for a cellId cannot be integrated throughout the "event". This PR creates a new hit when the time stamp falls out of all existing hits' `[time, time+integration window]`.

* Are the time resolution (or integration) values correct?

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes.


src/algorithms/digi/SiliconTrackerDigi.cc
```diff
@@ -66,31 +66,43 @@ void SiliconTrackerDigi::process(
...
-            // keep earliest time for hit
-            auto time_stamp = hit.getTimeStamp();
-            hit.setTimeStamp(std::min(hit_time_stamp, hit.getTimeStamp()));
-
-            // sum deposited energy
-            auto charge = hit.getCharge();
-            hit.setCharge(charge + (std::int32_t) std::llround(sim_hit.getEDep() * 1e6));
-        }
-    }
+                hit_time_stamp  // ns->ps <-- TODO: what does this comment mean?!
```
veprbl, line 96:
```suggestion
                hit_time_stamp
```
Should have been removed in
https://github.com/eic/EICrecon/commit/3a1bf554d0e8ea94ae96cd05068da98af819de22

src/algorithms/digi/SiliconTrackerDigiConfig.h
```diff
@@ -11,7 +11,7 @@ namespace eicrecon {
     // sub-systems should overwrite their own
     // NB: be aware of thresholds in npsim! E.g. https://github.com/eic/npsim/pull/9/files
     double threshold  = 0 * dd4hep::keV;
-    double timeResolution = 8;   /// TODO 8 of what units??? Same TODO in juggler. Probably [ns]
+    double timeResolution = 8;   /// same unit as sim_hit.getTime, probably [ns]
```
veprbl, line 14: The plan is to have eventually to everywhere use
```c++
#include <edm4eic/unit_system.h>
```
And it is already appropriate here:
```suggestion
    double timeResolution = 8 * edm4eic::unit::ns;
```
https://github.com/eic/EDM4eic/blob/0bd6edd97d0b28b30605ab90dcf338a3244d16dd/utils/include/edm4eic/unit_system.h#L32

src/algorithms/digi/SiliconTrackerDigi.cc
```diff
@@ -66,31 +66,43 @@ void SiliconTrackerDigi::process(
             continue;
         }
 
-        if (cell_hit_map.count(sim_hit.getCellID()) == 0) {
-            // This cell doesn't have hits
-            cell_hit_map[sim_hit.getCellID()] = {
+        bool bucket_found = false;
+        if (cell_hit_map.count(sim_hit.getCellID()) == 1) {
```
veprbl, line 70: This should have been
```с++
        if (cell_hit_map.count(sim_hit.getCellID()) >= 1) {
```
just drop this condition and keep the inner loop.

veprbl (commented):
I was thinking about a similar thing, but for calorimetry. I suppose, for MAPS tracking this would look more like histogramming in time, rather than this. Otherwise looks like a fine first prototype.

src/algorithms/digi/SiliconTrackerDigi.cc
veprbl: Actually, protecting against an implicit insertion in `operator[]` is kind of a good manner, even if it doesn’t matter yet. This is fine to keep, to stay explicit.

kkauder: I actually didn't think about it much, I just kept (well, inverted) the pre-existing `if (cell_hit_map.count(sim_hit.getCellID()) == 0)` conditional

kkauder:
> I was thinking about a similar thing, but for calorimetry. I suppose, for MAPS tracking this would look more like histogramming in time, rather than this. Otherwise looks like a fine first prototype.

I'm planning to do a similar thing for all hits, using this as a template as far as it's appropriate.

kkauder:
This PR should be ready to merge. Liaisons, please check I used the right values for resolution and integration window, and that the implementation makes some sense.

nschmidtALICE (dismissed):

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -35,18 +35,21 @@ extern "C" {
     std::string tracker_readout = "TaggerTrackerHits";
 
     // Digitization of silicon hits
+    auto TaggerTrackerTimeResolution    = 0.195 * dd4hep::ns;
+    auto TaggerTrackerIntegrationWindow = 5.0   * dd4hep::ns;
```
simonge, line 39:
```suggestion
    auto TaggerTrackerIntegrationWindow = 25.0   * dd4hep::ns;
```
Upon further thought I think I've had this wrong, 5 ns is about the analogue response time for a MIP but the minimum digital time over threshold is 25 ns.

simonge: Please can you remind me where the spreadsheet is so I can amend it there too

ruse-traveler:
Closing out after discussion in [04.27.2026 Reco WG meeting](https://indico.bnl.gov/event/32306/).

PR #1365, opened 2024-04-07
Implement PID lookup table

nathanwbrei:

Implement PID lookup table as per:

- https://docs.google.com/document/d/1ViIBV0448Vnob0zE_Sm7kQsggsKOJZ9oYdmPp--QPvE/edit#heading=h.tf2uykgho79e
- https://github.com/rdom/fastpid/

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No
However, it does download a ~100MB file to the working directory.


src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+    void ChangeRun(int64_t run_number) {
+    }
+
+    void Process(int64_t run_number, uint64_t event_number) {
+
+        // TODO: This is all very handwavy because I haven't been attending the PID datamodel discussions.
+        // Please look over this carefully and correct as needed!
+
+        for (const auto& mcparticle : *m_particles_in()) {
+
```
wdconinc, line 50: Creating ReconstructedParticles from MCParticles is not exactly what we want. This only gives an empty shell of a ReconstructedParticle without any tracking or energy information, no associations, etc. We need to 'inject' the PDG information in the ReconstructedParticles that are already created from clustering and tracking, in or immediately after the MatchClusters algorithm, most likely.

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+            // Unpack lookup values from input
+
+            int pdg = mcparticle.getPDG();
+            int charge = mcparticle.getCharge();
+            double momentum = edm4hep::utils::magnitude(mcparticle.getMomentum());
+
+            // TODO: I'm still confused as to whether our lookup table actually contains eta vs theta.
+            double eta = edm4hep::utils::eta(mcparticle.getVertex());
+            double theta = edm4hep::utils::anglePolar(mcparticle.getVertex());
+            double phi = edm4hep::utils::angleAzimuthal(mcparticle.getVertex());
```
wdconinc, line 60: This gives us the eta, theta, phi of the production vertex, not the momentum direction.

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+            if (entry != nullptr) {
+
+                m_gen.seed(event_number);
+                double random_unit_interval = m_dist(m_gen);
+
+                if (random_unit_interval < entry->prob_electron) {
+                    identified_pdg = 11; // electron
+                }
+                else if (random_unit_interval < (entry->prob_electron + entry->prob_pion)) {
+                    identified_pdg = 211; // pion
```
wdconinc, line 76: This (and others) ignores sign. We should assign the correct sign to the PDG code since analyzers will expect this.

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+            double eta = edm4hep::utils::eta(mcparticle.getVertex());
+            double theta = edm4hep::utils::anglePolar(mcparticle.getVertex());
+            double phi = edm4hep::utils::angleAzimuthal(mcparticle.getVertex());
+
+            auto entry = m_lut->Lookup(pdg, charge, momentum, eta, phi);
+
+            int identified_pdg = -1; // unknown
+                                     // TODO: What is the PDG for 'unknown', actually?
+
+            if (entry != nullptr) {
```
wdconinc, line 67: We should write the probabilities as [ParticleID](https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L217) entries, update the [ParticleIDused](https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L212) entry with the selected one, and update the PDG field accordingly.

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+                if (random_unit_interval < entry->prob_electron) {
+                    identified_pdg = 11; // electron
+                }
+                else if (random_unit_interval < (entry->prob_electron + entry->prob_pion)) {
+                    identified_pdg = 211; // pion
+                }
+                else if (random_unit_interval < (entry->prob_electron + entry->prob_pion + entry->prob_kaon)) {
+                    identified_pdg = 321; // kaon
+                }
+                else if (random_unit_interval < (entry->prob_electron + entry->prob_pion + entry->prob_kaon + entry->prob_electron)) {
```
wdconinc, line 81: The approach of throwing one uniform value and comparing to the sum of the N probabilities is not exactly what we want and can do with the information provided. We should produce N random variables, which sum to 1, and where the expectation value of each is given by the entry in the table. Those N random variables are to represent the probability of each individual hypothesis.

The LUT doesn't have sufficient information to do this correctly, of course. We don't have the multivariate distributions for this random variable problem, only their summary statistic of the mean. But if this is intended to allow for a prototype that will get more precise, we should adhere to an interface that is what will be present in the future for full PID reconstruction.

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+                    // These end up here
+                }
+            }
+
+            m_recoparticles_out() = std::make_unique<edm4eic::ReconstructedParticleCollection>();
+            auto recopart = m_recoparticles_out()->create();
+
+            recopart.setPDG(identified_pdg);
+            // TODO: Set other fields?
+            // TODO: Association with MCParticle?
```
wdconinc, line 96: ReconstructedParticles will already have an MCParticle, which is the one that should be used to begin with. This algorithm should ingest ReconstructedParticles and ReconstructedParticleAssociations, and emit ReconstructedParticles and ReconstructedParticleAssociations.

src/services/pid_lut/PIDLookupTable_service.h
```diff
@@ -0,0 +1,46 @@
...
+        auto pair = m_cache.find(filename);
+        if (pair == m_cache.end()) {
+            auto lut = std::make_unique<PIDLookupTable>();
+            LOG << "Loading PID lookup table: " << filename << LOG_END;
+
+            if (!std::filesystem::exists(filename)) {
+                LOG << "PID lookup table '" << filename << "' not found. Attempting to download..." << LOG_END;
+                std::ostringstream oss;
+                oss << "curl -k -o " << filename << " " << url;
+                system(oss.str().c_str());
```
wdconinc, line 34: Algorithms and services should not run curl, but assume the lookup tables are present. We have artifact distribution mechanisms (however imperfect) through the geometry, and these should be reused instead of adding new interfaces. If the filename does not exist locally, the service should fail.

src/factories/pid/PIDLookupTable_factory.h
nathanwbrei: Do you know what the collection names are for the ReconstructedParticles coming off of the DIRC, pfRICH, and TOF? Or do such collections even exist? It looks like MatchClusters_factory is only matching against ECalClusters at the moment.  MatchCluster's input is "ReconstructedChargedParticles" (which looks like it includes everything from all detectors) and it's output is "ReconstructedParticles" (which isn't named as if it is meant to be ECal-specific)

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
+// Copyright 2024, Nathan Brei
+// Subject to the terms in the LICENSE file found in the top-level directory.
```
veprbl, line 2:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Nathan Brei
```

src/services/pid_lut/PIDLookupTable.cc
```diff
@@ -0,0 +1,162 @@
+// Copyright 2024, Nathan Brei
+// Subject to the terms in the LICENSE file found in the top-level directory.
```
veprbl, line 2:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Nathan Brei
```

src/services/pid_lut/PIDLookupTable.h
```diff
@@ -0,0 +1,58 @@
+// Copyright 2024, Nathan Brei
+// Subject to the terms in the LICENSE file found in the top-level directory.
```
veprbl, line 2:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Nathan Brei
```

src/services/pid_lut/PIDLookupTable_service.h
```diff
@@ -0,0 +1,46 @@
+// Copyright 2024, Nathan Brei
+// Subject to the terms in the LICENSE file found in the top-level directory.
```
veprbl, line 2:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Nathan Brei
```

src/tests/pid_lut_test/PIDLookupTable_test.cc
```diff
@@ -0,0 +1,162 @@
+
```
veprbl, line 1:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Nathan Brei

```

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+
+            // TODO: I'm still confused as to whether our lookup table actually contains eta vs theta.
+            double eta = edm4hep::utils::eta(mcparticle.getVertex());
+            double theta = edm4hep::utils::anglePolar(mcparticle.getVertex());
+            double phi = edm4hep::utils::angleAzimuthal(mcparticle.getVertex());
+
+            auto entry = m_lut->Lookup(pdg, charge, momentum, eta, phi);
+
+            int identified_pdg = -1; // unknown
+                                     // TODO: What is the PDG for 'unknown', actually?
```
veprbl, line 65:
```suggestion
            int identified_pdg = 0; // unknown
```
I think `0` is not occupied. The `-1`, at least formally, is a anti-d quark.

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+                    identified_pdg = 211; // pion
+                }
+                else if (random_unit_interval < (entry->prob_electron + entry->prob_pion + entry->prob_kaon)) {
+                    identified_pdg = 321; // kaon
+                }
+                else if (random_unit_interval < (entry->prob_electron + entry->prob_pion + entry->prob_kaon + entry->prob_electron)) {
+                    identified_pdg = 2212; // proton
+                }
+                else {
+                    identified_pdg = -1; // unknown
```
veprbl, line 85:
```suggestion
                    identified_pdg = 0; // unknown
```

src/factories/pid/PIDLookupTable_factory.h
wdconinc: 0 used to be optical photons in geant4, but they shifted those to -22, I think.

src/factories/pid/PIDLookupTable_factory.h
veprbl: I don't agree. The statistical fluctuations of the original simulation define a systematic uncertatinty on the purities/backgrounds, the statistical fluctuations in the data define statistical uncertainty. Smearing, like you suggest, the fractions that we plan to use for the analysis is not a correct approach.

src/factories/pid/PIDLookupTable_factory.h
veprbl: The sign used in the analysis should come from the original reconstructed particle. I don't think there is an actual expectation that PDG code is provided with the sign here.

src/factories/pid/PIDLookupTable_factory.h
veprbl: I wonder, if it would be to just not provide PID hypothesis at all in this case.

src/factories/pid/PIDLookupTable_factory.h
wdconinc: I meant that the vertex field is a position. https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L194

src/services/pid_lut/PIDLookupTable_service.h
wdconinc: If every service or every algorithm is going to implement its own way to get artifacts, it increases the possibility for errors and our support burden. For operational reasons this is also not a workable solution:
- How will you deal with updates to the LUT? This only checks that the file exists, not that it is the latest version. Anyone running this once in their local development directory will forever be stuck on the first downloaded version.
- How do you deal with this when we run 10k jobs in production? All of them are going to be running in a job-specific directory where the LUT file doesn't exist yet. Are they all going to download the file?

These are not new considerations, but we have solved them for files like Acts material maps.

src/factories/pid/PIDLookupTable_factory.h
wdconinc: I agree that the charge should come from the original reconstructed particle (in the `charge` field), but I think we should ensure that what is stored in the PDG field for the reconstructed particles that analyzers use agrees with that charge, i.e. charge = -1 should then have -211 IFF the dice is rolled and comes up with 211. If we don't do this, people will complain about it and make mistakes.

src/factories/pid/PIDLookupTable_factory.h
wdconinc: > It is sufficient to not set the PDG field in ReconstructedParticle in this case?

That gives you zero.

veprbl: I think, we would need to use ReconstructedParticle -> ParticleID relation [1] to support multiple detectors anyway. Then, when PID is not available from a given detector, the entry is not populated. The `ReconstructedParticle::PDG` field comes without any field to describe the likelihood, we shouldn't use that, perhaps.

[1] https://github.com/key4hep/EDM4hep/blob/792cefc3a0f7cb3dfa12aba50a24a6f37089e84e/edm4hep.yaml#L475

src/services/pid_lut/PIDLookupTable_service.h
veprbl: Submitted to https://github.com/eic/epic/pull/711

veprbl: Once that's in nightly, you can expect the `calibrations/hpdirc_positive.lut` to just exist when you run your algorithm.

src/services/pid_lut/pid_lut.cc
```diff
@@ -0,0 +1,15 @@
+// Copyright 2024, Nathan Brei
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
```
veprbl, line 3:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Nathan Brei
```

src/factories/pid/PIDLookupTable_factory.h
```diff
@@ -0,0 +1,101 @@
...
+private:
+    PodioInput<edm4hep::MCParticle> m_particles_in {this};
+    PodioOutput<edm4eic::ReconstructedParticle> m_recoparticles_out {this};
+
+    ParameterRef<std::string> m_filename {this, "filename", config().filename, "Relative to current working directory"};
+    ParameterRef<std::string> m_url {this, "url", config().url, "Only used if not found via filename"};
+    Service<PIDLookupTable_service> m_lut_svc {this};
+
+    std::mt19937 m_gen;
+    std::uniform_real_distribution<double> m_dist {0, 1};
```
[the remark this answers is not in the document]
veprbl: This advice didn't age well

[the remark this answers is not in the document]
wdconinc: RandomSvc is still more random than `const` :-)

veprbl:
@mdiefent @nathanwbrei  @wdconinc This is ready for some preliminary review. Some known issues:
 - [ ] Need to wire factories sequentially and into the default particle stream (requires other WIP changes)
 - [ ] The binning was largely guessed, so may be wrong with up to whether the bin centers provided / lower edges provided
 - [ ] Unit tests need to be fixed
 - [ ] Need to merge https://github.com/eic/epic/pull/719
 - [ ] This functionality could use a benchmark
 - [ ] Need to upload the latest DIRC table somewhere (GitHub won't let you put 100MB+ files)

src/algorithms/pid_lut/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.16)
+
```
wdconinc, line 2:
```suggestion
```

src/algorithms/pid_lut/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+# Setting default includes, libraries and installation paths
+plugin_add(${PLUGIN_NAME} WITH_SHARED_LIBRARY WITHOUT_PLUGIN)
+
+# The macro grabs sources as *.cc *.cpp *.c and headers as *.h *.hh *.hpp Then
+# correctly sets sources for ${_name}_plugin and ${_name}_library targets Adds
+# headers to the correct installation directory
+plugin_glob_all(${PLUGIN_NAME})
+
+# Find dependencies
+plugin_add_algorithms(${PLUGIN_NAME} algopid_lut)
```
wdconinc, line 15: Do we need the (non-existent) algopid_lut in algorithms?
```suggestion
plugin_add_algorithms(${PLUGIN_NAME})
```

src/algorithms/pid_lut/PIDLookup.cc
```diff
@@ -0,0 +1,129 @@
...
+    .charge_values=m_cfg.charge_values,
+    .momentum_edges=m_cfg.momentum_edges,
+    .polar_edges=m_cfg.polar_edges,
+    .azimuthal_binning=m_cfg.azimuthal_binning,
+    .polar_bin_centers_in_lut=m_cfg.polar_bin_centers_in_lut,
+    .skip_legacy_header=m_cfg.skip_legacy_header,
+    .use_radians=m_cfg.use_radians,
+    .missing_electron_prob=m_cfg.missing_electron_prob,
+  });
+  if (!m_lut) {
```
wdconinc, line 29:
```suggestion
  if (m_lut == nullptr) {
```

src/detectors/DIRC/CMakeLists.txt
```diff
@@ -15,3 +15,8 @@ plugin_glob_all(${PLUGIN_NAME})
 # Find dependencies
 plugin_add_event_model(${PLUGIN_NAME})
 plugin_add_dd4hep(${PLUGIN_NAME})
+
+# Add libraries (same as target_include_directories but for both plugin and
+# library)
+plugin_link_libraries(${PLUGIN_NAME} algorithms_digi_library
```
wdconinc, line 21: How does algorithms_digi_library get pulled in here?

src/services/pid_lut/PIDLookupTableSvc.h
```diff
@@ -0,0 +1,49 @@
...
+
+#pragma once
+
+#include <algorithms/logger.h>
+#include "PIDLookupTable.h"
+#include <JANA/Services/JServiceLocator.h>
+#include <JANA/JLogger.h>
+#include <mutex>
+#include <filesystem>
+#include <cstdlib>
```
wdconinc, line 12: Curious why this gets pulled in.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -89,17 +89,24 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "TOFBarrelHitAssociations",
             "TOFEndcapHitAssociations",
 
+            "CombinedTOFParticleIDs"
```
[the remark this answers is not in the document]
wdconinc: It paid off!!!

.github/workflows/linux-eic-shell.yml
```diff
@@ -266,7 +267,7 @@ jobs:
         llvm-profdata-15 merge -sparse src/tests/algorithms_test/algorithms_test.profraw \
           -o src/tests/algorithms_test/algorithms_test.profdata
         LIB_PATHS=()
-        for LIB_NAME in algorithms_calorimetry algorithms_digi algorithms_pid algorithms_reco algorithms_tracking; do
+        for LIB_NAME in algorithms_calorimetry algorithms_digi algorithms_pid algorithms_pid_lut algorithms_reco algorithms_tracking; do
```
wdconinc, line 270: Fragile and probably should be programmatically derived.

src/global/pid_lut/PIDLookup_factory.h
```diff
@@ -0,0 +1,49 @@
...
+  PodioOutput<edm4hep::ParticleID> m_particleids_output{this};
+
+  ParameterRef<std::string> m_filename{this, "filename", config().filename,
+                                       "Relative to current working directory"};
+  ParameterRef<int> m_system{this, "system", config().system, "For the ParticleID record"};
+
+public:
+  void Configure() {
+    m_algo = std::make_unique<AlgoT>(this->GetPrefix());
+    m_algo->level((algorithms::LogLevel)logger()->level());
```
wdconinc, line 36: static_cast?

wdconinc (dismissed):
Looks great to me.

.github/workflows/linux-eic-shell.yml
veprbl: We have src/algorithms/interfaces and src/algorithms/meta that are not even libraries. If you know of a clean way, let me know.

wdconinc: Would something like this work, or is it too broad?
```suggestion
        for LIB_NAME in $(find src/ -type f -name "libalgorithms_*.so"); do
```
Also makes the next line obsolete.

src/algorithms/pid_lut/PIDLookup.cc
```diff
@@ -0,0 +1,142 @@
...
+#include <edm4eic/ReconstructedParticleCollection.h>
+#include <edm4hep/MCParticleCollection.h>
+#include <edm4hep/utils/vector_utils.h>
+#include <fmt/core.h>
+#include <podio/ObjectID.h>
+#include <cmath>
+#include <gsl/pointers>
+#include <stdexcept>
+
+#include "PIDLookup.h"
```
wdconinc, line 15: Use full path under src/ to allow for easier future separation of headers from implementation.

src/algorithms/pid_lut/PIDLookup.cc
```diff
@@ -0,0 +1,142 @@
...
+      continue;
+    }
+
+    int true_pdg    = mcpart.getPDG();
+    int true_charge = mcpart.getCharge();
+    int charge      = recopart.getCharge();
+    double momentum = edm4hep::utils::magnitude(recopart.getMomentum());
+
+    double theta = edm4hep::utils::anglePolar(recopart.getMomentum()) / M_PI * 180.;
+    double phi   = edm4hep::utils::angleAzimuthal(recopart.getMomentum()) / M_PI * 180.;
```
wdconinc, line 77: In the next iteration we should push for storing angles in radians in the LUT itself.

src/detectors/BTOF/BTOF.cc
```diff
@@ -46,5 +57,38 @@ void InitPlugin(JApplication *app) {
...
 
+    int BarrelTOF_ID = 0;
+    try {
+        auto detector = app->GetService<DD4hep_service>()->detector();
+        BarrelTOF_ID = detector->constant<int>("BarrelTOF_ID");
+    } catch(const std::runtime_error&) {
+        // Nothing
+    }
+    for (auto qualifier : std::vector<std::string>({"", "Seeded"}))
+    app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
```
wdconinc, line 68: Indentation and braces.

src/detectors/DIRC/DIRC.cc
```diff
@@ -73,5 +84,36 @@ extern "C" {
...
 
+    int BarrelDIRC_ID = 0;
+    try {
+        auto detector = app->GetService<DD4hep_service>()->detector();
+        BarrelDIRC_ID = detector->constant<int>("BarrelDIRC_ID");
+    } catch(const std::runtime_error&) {
+        // Nothing
+    }
+    for (auto qualifier : std::vector<std::string>({"", "Seeded"}))
+    app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
```
wdconinc, line 95: Indentation and braces.

src/detectors/DRICH/DRICH.cc
```diff
@@ -180,6 +189,39 @@ extern "C" {
...
 
+    int ForwardRICH_ID = 0;
+    try {
+        auto detector = app->GetService<DD4hep_service>()->detector();
+        ForwardRICH_ID = detector->constant<int>("ForwardRICH_ID");
+    } catch(const std::runtime_error&) {
+        // Nothing
+    }
+    for (auto qualifier : std::vector<std::string>({"", "Seeded"}))
+    app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
```
wdconinc, line 200: Indentation and braces.

src/detectors/PFRICH/PFRICH.cc
```diff
@@ -66,5 +78,39 @@ extern "C" {
...
 
+    int BackwardRICH_ID = 0;
+    try {
+        auto detector = app->GetService<DD4hep_service>()->detector();
+        BackwardRICH_ID = detector->constant<int>("BackwardRICH_ID");
+    } catch(const std::runtime_error&) {
+        // Nothing
+    }
+    for (auto qualifier : std::vector<std::string>({"", "Seeded"}))
+    app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
```
wdconinc, line 89: You guessed it.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -277,9 +284,11 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "HcalFarForwardZDCTruthClusters",
             "HcalFarForwardZDCTruthClusterAssociations",
 
-
             // DIRC
-            "DIRCRawHits"
+            "DIRCRawHits",
+            "DIRCPID",
+            "DIRCParticleIDs",
+            "DIRCSeededParticleIDs",
```
wdconinc, line 291: Is a final comma ok here?

veprbl: I see how it harms anything

veprbl: I don't see how it harms anything

src/algorithms/pid_lut/PIDLookup.cc
veprbl: We can. I think the degree interface is based on the original drich luts.

.github/workflows/linux-eic-shell.yml
veprbl: I think this can work, as long as libraries don't disappear from the build. I've implemented.

wdconinc (approved):

veprbl (dismissed):

veprbl (approved):
based on previous reviews

Issue #1249, opened 2024-01-22
iwyu tool is more trouble than it is worth


nathanwbrei:
I'm beginning to believe that the 'iwyu' tool is more trouble than it is worth and should be removed from our CI pipeline.

- Here, the tool is getting caught in a loop and running for 3+ hours: https://github.com/eic/EICrecon/actions/runs/7612607107/job/20734769644?pr=1221

- Even when functioning, it is sensitive to extremely minor changes in code. Practically every push triggers an iwyu failure, leading to pull requests where every other commit is an iwyu patch. This obscures the history of the pull request, unless the developer aggressively force-pushes.

- Because iwyu runs late in the pipeline and almost always fails, it lengthens the feedback loop between the developer and the CI. 

### Environment: (where does this bug occur, have you tried other environments)
All

### Steps to reproduce: (give a step by step account of how to trigger the bug)
Any

### Expected Result: (what do you expect when you execute the steps above)
Being able to test and merge PRs in a bounded amount of time.

### Actual Result: (what do you get when you execute the steps above)
After any push, we have to wait around for >30mins until the CI returns a meaningless failure.

nathanwbrei:
Tagging @veprbl and @wdconinc because I'm curious to hear their thoughts

wdconinc:
Nothing in the ci pipelines depends on iwyu successful completion, so there is no requirement to fix the issues until the very end of a development cycle (or am I missing something).

The reason we introduced this is because of some over-eager inclusion of expensive templated headers (JEvent.h is one of them, but mostly Eigen is the culprit through Acts).

The loop with Eigen/Householder is annoying but just needs a keep pragma to fix it.

But we don't want this to be an annoyance so how do we make it less so?

nathanwbrei:
```
(/home/runner/work/EICrecon/EICrecon/src/services/io/podio/JEventProcessorPODIO.h has correct #includes/fwd-decls)

(/home/runner/work/EICrecon/EICrecon/src/services/io/podio/JEventProcessorPODIO.cc has correct #includes/fwd-decls)

/home/runner/work/EICrecon/EICrecon/src/services/pid_lut/PIDLookupTable.cc:51:19: error: implicit instantiation of undefined template 'std::basic_ifstream<char>'
    std::ifstream file(filename);
                  ^
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/iosfwd:116:11: note: template is declared here
    class basic_ifstream;
          ^
[...]

PR #1368, opened 2024-04-08
linux-eic-shell.yml: run newer version of IWYU from Ubuntu instead of eic-shell

veprbl:
We already have to do this for llvm-cov, and upgrading LLVM in eic-shell is undesirable.

Resolves: #1249

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -132,7 +132,59 @@ jobs:
         path: build.tar.zst
         if-no-files-found: error
 
-  clang-tidy-iwyu:
+  include-what-you-use:
+    name: include-what-you-use (IWYU)
+    runs-on: ubuntu-latest
+    needs: build
+    container:
+      # provides IWYU 8.21
+      image: ubuntu:24.04
+      volumes:
+      - /home/runner:/home/runner
+    steps:
+    - run: apt update && apt install -y clang-tidy git iwyu zstd
+    - run: echo "FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> "${GITHUB_ENV}"
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: ${{ env.FETCH_DEPTH }}
+    - name: Workaround file ownership guard
+      run: git config --global --add safe.directory "$PWD"
+    - name: Download build artifact
+      uses: actions/download-artifact@v4
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -135,12 +135,15 @@ jobs:
...
     needs: build
+    container:
+      # provides IWYU 8.21
+      image: ubuntu:24.04
     steps:
+    - run: apt update && apt install -y clang-tidy git iwyu zstd
     - run: echo "FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> "${GITHUB_ENV}"
     - uses: actions/checkout@v4
       with:
         fetch-depth: ${{ env.FETCH_DEPTH }}
```
wdconinc, line 146: OT but at some point we should just switch to  `actions/checkout` with `filter: "tree:0"` instead of determining FETCH_DEPTH.

.github/workflows/linux-eic-shell.yml
```diff
@@ -149,57 +156,40 @@ jobs:
...
       run: tar -xaf build.tar.zst
     - name: Run clang-tidy on changed files
-      uses: eic/run-cvmfs-osg-eic-shell@main
       if: ${{ github.event_name == 'pull_request' }}
-      with:
-        platform-release: "${{ env.platform-release }}"
-        run: |
-          git diff ${{github.event.pull_request.base.sha}} ${{ github.event.pull_request.base.sha }} | clang-tidy-diff -p 1 -path build -quiet -export-fixes clang_tidy_fixes.yml -extra-arg='-std=c++20' -checks='-*,bugprone-*,-bugprone-narrowing-conversions' -clang-tidy-binary run-clang-tidy
+      run: |
+        git diff ${{github.event.pull_request.base.sha}} | clang-tidy-diff -p 1 -path build -quiet -export-fixes clang_tidy_fixes.yml -extra-arg='-std=c++20' -checks='-*,bugprone-*,-bugprone-narrowing-conversions' -clang-tidy-binary run-clang-tidy
```
veprbl, line 161: `${{github.event.pull_request.base.sha}}` entered twice here. I assume, it was a bug.

veprbl: Introduced 6 months ago in 17c4137878de4695286caad7e1a26982402fe626
I was wondering why we haven't had any clang-tidy reports in a long while.

veprbl:
This now runs an updated IWYU. However, I'm now realizing that it should have no way to access dependency headers, so it's not possible to meaningfully handle those.

Issue #199, opened 2022-10-07
Doxygen produces boxes instead of text in diagrams


DraTeots:
<img width="1359" alt="image" src="https://user-images.githubusercontent.com/4732545/194472873-b4c080f9-788d-4552-8d04-0679ac5d13a0.png">

https://eic.github.io/EICrecon/doxygen/classCalorimeterIslandCluster.html

https://eic.github.io/EICrecon/doxygen/

Related with probable fix
https://stackoverflow.com/questions/21587974/graphviz-plugin-in-confluence-showing-text-as-boxes

PR #1369, opened 2024-04-08
linux-eic-shell.yml: drop use of alpine

veprbl:
Doesn't seem to be needed anymore.


wdconinc (approved):
LGTM. Let's hope we don't end up with boxes in doxygen output again (https://github.com/eic/EICrecon/issues/199)...

PR #1378, opened 2024-04-11
Add FastJet plugin algorithms to jet reconstruction

johnlajoie:

This PR adds Centauro FastJet contributed algorithm

- [ ] Bug fix (issue #__)
- [ X ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No 

It does change the default behavior of the reconstruction in the generated and reconstructed Centauro jets will be created.


johnlajoie:
Not yet working - the plumbing has been added to allow cmake to find the fjcontrib library, but EICRecon seems to load it before the fastjet library, which leads to an abort due to a missing symbol: 

`
[DEBUG] JPluginLoader: /usr/local/lib/libfastjetcontribfragile.so: undefined symbol: _ZTIN7fastjet28JetMedianBackgroundEstimatorE
[WARN] JPluginLoader: Loading failure: [DEBUG] JPluginLoader: Looking for '/usr/local/lib/EICrecon/plugins/reco.so' ....
`
`
nightly> lajoie@lap140311:~/ePIC$ c++filt _ZTIN7fastjet28JetMedianBackgroundEstimatorE
typeinfo for fastjet::JetMedianBackgroundEstimator
`
Is there a way to control the load order in the way we add libraries to JANA2 plugins?

veprbl:
Just add libfastjettools
```
# nm /usr/local/lib/libfastjetcontribfragile.so | grep _ZTIN7fastjet28JetMedianBackgroundEstimatorE
                 U _ZTIN7fastjet28JetMedianBackgroundEstimatorE
# nm /usr/local/lib/libfastjettools.so | grep _ZTIN7fastjet28JetMedianBackgroundEstimatorE
0000000000049a68 V _ZTIN7fastjet28JetMedianBackgroundEstimatorE
```

cmake/FindFJContrib.cmake
```diff
@@ -0,0 +1,25 @@
+# * Locate FastJetContrib library Defines:
```
wdconinc, line 1: It would be nice if we could upstream the cmake support so it's in the next fastjet and fjcontrib releases. Then we don't have to maintain FindFastJet.cmake and FindFJContrib.cmake.

johnlajoie:
This now works in that it compiles and links both the FastJet tools and contrib libraries correctly, the Centauro algorithm is found, runs, and produces output that looks reasonable. I still need to do some work to clean up how that plugin algorithms are configured and I would also like to add the energy correlations plugin as well (but I need to see what sort of output it produces).

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -55,6 +56,19 @@ namespace eicrecon {
...
+      case JetAlgorithm::plugin_algorithm:
+
+        // expand to other algorithms as required
+        if(m_cfg.jetContribAlgo != "Centauro"){
+          m_log->error(" Unknown contributed FastJet algorithm \"{}\" specified!", m_cfg.jetContribAlgo);
+          throw JException("Invalid contributed FastJet algorithm");
+        }
+        else{
+          m_jet_def = std::make_unique<JetDefinition>(new contrib::CentauroPlugin(m_cfg.rJet));
+        }
```
ruse-traveler, line 69:
```suggestion
        // expand to other algorithms as required
        if(m_cfg.jetContribAlgo == "Centauro"){
          m_jet_def = std::make_unique<JetDefinition>(new contrib::CentauroPlugin(m_cfg.rJet));
        }
        else {
          m_log->error(" Unknown contributed FastJet algorithm \"{}\" specified!", m_cfg.jetContribAlgo);
          throw JException("Invalid contributed FastJet algorithm");
        }
```

This makes it a little easier to extend if we want to add another contributed algorithm!

ruse-traveler (changes requested):
Very nice! Thank you for the development! I have one minor suggestion, but am otherwise happy with this.

ruse-traveler:
Whoops! And one last suggestion that slipped my mind: don't forget to add your name to the list of contributors in `JetReconstruction.{cc,h}`, @johnlajoie!

src/algorithms/reco/JetReconstruction.cc
```diff
@@ -55,6 +56,19 @@ namespace eicrecon {
     // Choose jet definition based on no. of parameters
     switch (m_mapJetAlgo[m_cfg.jetAlgo]) {
 
+      // contributed algorithms
+      case JetAlgorithm::plugin_algorithm:
+
+        // expand to other algorithms as required
+        if(m_cfg.jetContribAlgo == "Centauro"){
+          m_jet_def = std::make_unique<JetDefinition>(new contrib::CentauroPlugin(m_cfg.rJet));
```
veprbl, line 64: Somehow this is detected as a memory leak:
```
Direct leak of 64 byte(s) in 2 object(s) allocated from:
    #0 0x55f986acb1fd in operator new(unsigned long) (/home/runner/work/EICrecon/EICrecon/install/bin/eicrecon+0x10e1fd) (BuildId: b5a759ebf1114ef904000411f83b34f5cafce8b5)
    #1 0x7f5868a08f7c in eicrecon::JetReconstruction<edm4eic::ReconstructedParticle>::init(std::shared_ptr<spdlog::logger>) /home/runner/work/EICrecon/EICrecon/src/algorithms/reco/JetReconstruction.cc:64:55
    #2 0x7f58691db696 in eicrecon::JetReconstruction_factory<edm4eic::ReconstructedParticle>::Configure() /home/runner/work/EICrecon/EICrecon/src/factories/reco/JetReconstruction_factory.h:49:17
    #3 0x7f58703cdf96 in __pthread_once_slow nptl/pthread_once.c:116:7
```

veprbl: I don't see anything suspicious in the `include/fastjet/contrib/Centauro.hh`:
```c++
    CentauroPlugin (double R) : _R(R), _gammaE(0), _gammaPz(0){}
```

veprbl: I see that `JetDefinition::Plugin` has a virtual destructor and none in `class CentauroPlugin : public JetDefinition::Plugin {`. I wonder, if this can be the cause?

wdconinc: Can you store the plugin in `std::unique_ptr m_jet_plugin` so its ownership is clear?

veprbl: Right. The JetDefinition is constructed from it, but we don't store the pointer to the plugin.
```c++
  /// constructor based on a pointer to a user's plugin; the object
  /// pointed to must remain valid for the whole duration of existence
  /// of the JetDefinition and any related ClusterSequences
  JetDefinition(const Plugin * plugin_in) {
```

johnlajoie: Thanks guys - I had a similar problem trying to do this before, it yields error messages I don't understand.  

In JetReconstruction.h I declared: 

`      
std::unique_ptr<fastjet::JetDefinition::Plugin> m_jet_plugin;
`
and then in the code I use this as: 

`
        if(m_cfg.jetContribAlgo == "Centauro"){
	  m_jet_plugin = std::make_unique<contrib::CentauroPlugin>(m_cfg.rJet); 
          m_jet_def = std::make_unique<JetDefinition>(m_jet_plugin);
        }
`

This yields the error at compilation: 

`
/home/lajoie/ePIC/EICrecon/src/algorithms/reco/JetReconstruction.cc: In member function ‘void eicrecon::JetReconstruction<InputT>::init(std::shared_ptr<spdlog::logger>)’:
/home/lajoie/ePIC/EICrecon/src/algorithms/reco/JetReconstruction.cc:64:26: error: expected primary-expression before ‘const’
   64 |           m_jet_plugin = const std::make_unique<contrib::CentauroPlugin>(m_cfg.rJet);
      |                          ^~~~~
In file included from /usr/include/c++/12/memory:76,
                 from /usr/local/include/fmt/format.h:40,
                 from /usr/local/include/algorithms/detail/demangle.h:9,
                 from /usr/local/include/algorithms/algorithm.h:39,
                 from /home/lajoie/ePIC/EICrecon/src/algorithms/reco/JetReconstruction.h:6,
                 from /home/lajoie/ePIC/EICrecon/src/algorithms/reco/JetReconstruction.cc:5:
/usr/include/c++/12/bits/unique_ptr.h: In instantiation of ‘std::__detail::__unique_ptr_t<_Tp> std::make_unique(_Args&& ...) [with _Tp = fastjet::JetDefinition; _Args = {const unique_ptr<fastjet::JetDefinition::Plugin, default_delete<fastjet::JetDefinition::Plugin> >&}; __detail::__unique_ptr_t<_Tp> = __detail::__unique_ptr_t<fastjet::JetDefinition>]’:
/home/lajoie/ePIC/EICrecon/src/algorithms/reco/JetReconstruction.cc:65:54:   required from ‘void eicrecon::JetReconstruction<InputT>::init(std::shared_ptr<spdlog::logger>) [with InputT = edm4eic::ReconstructedParticle]’
/home/lajoie/ePIC/EICrecon/src/algorithms/reco/JetReconstruction.cc:160:18:   required from here
/usr/include/c++/12/bits/unique_ptr.h:1065:30: error: no matching function for call to ‘fastjet::JetDefinition::JetDefinition(const std::unique_ptr<fastjet::JetDefinition::Plugin>&)’
 1065 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
`
which I think is telling me I'm using the pointer wrong? 

Any pointers? (pun intended)

wdconinc: Try with
```
m_jet_def = std::make_unique<JetDefinition>(m_jet_plugin.get())
```

wdconinc: The `get()` gives you a regular pointer from the smart pointer.

johnlajoie: Well that seems to be the secret sauce!

src/global/reco/reco.cc
```diff
@@ -305,5 +305,37 @@ void InitPlugin(JApplication *app) {
...
+            "GeneratedBreitFrameParticles",
+            {"MCParticles","InclusiveKinematicsElectron","GeneratedParticles"},
+            {"GeneratedBreitFrameParticles"},
+            {},
+            app
+    ));
+
+    app->Add(new JOmniFactoryGeneratorT<JetReconstruction_factory<edm4eic::ReconstructedParticle>>(
+            "GeneratedCentauroJets",
+            {"GeneratedBreitParticles"},
```
veprbl, line 318:
```suggestion
            {"GeneratedBreitFrameParticles"},
```

johnlajoie: Dammit. Spent 20 minutes debugging this when I should have just read @veprbl 's review. Good catch!

cmake/FindFJContrib.cmake
ruse-traveler: For my own understanding (since I'm very unfamiliar with cmake), would this entail adding FastJet and FJContrib to the top-level CMakeLists?

ruse-traveler (approved):
Nice! I'm now quite happy with this!

cmake/FindFJContrib.cmake
wdconinc: No need to add to the top-level CMakeLists. I meant contributing FindFastJet.cmake and FindFJContrib.cmake as PRs to the fastjet project. These are files that not just us are using, but likely many others too. I suspect you copied them from another project in the first place. Having fastjet provide the reference implementation will ensure uniformity.

ruse-traveler: Ah! I see! I have to admit, I haven't ever used them elsewhere since all other projects I've worked on didn't use cmake 😅 

But I agree! That would be a better approach long-term!

veprbl: I'm not sure that we can contribute cmake support directly. The fastjet.fr website doesn't mention any route to do that. We can contact the developers to see if they are willing to accept external contributions.

Issue #1366, opened 2024-04-08
Some EICrecon algorithms are broken with PODIO 00-17-04


veprbl:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): any
- Which version (or `HEAD` for the most recent on git): any
- Any specific OS or system where the issue occurs? No
- Any special versions of ROOT or Geant4? No

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. compile and run with PODIO 00-17-04
2. check empty collections in the output

### Expected Result: (what do you expect when you execute the steps above)
Everything works as well as with 00-17-03

### Actual Result: (what do you get when you execute the steps above)

At least few calorimetry factories are broken
https://github.com/eic/EICrecon/blob/d9b185cac2b6284444f67d31d13aabee73a50f4d/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc#L76
https://github.com/eic/EICrecon/blob/d9b185cac2b6284444f67d31d13aabee73a50f4d/src/algorithms/calorimetry/ImagingClusterReco.h#L92
https://github.com/eic/EICrecon/blob/d9b185cac2b6284444f67d31d13aabee73a50f4d/src/algorithms/calorimetry/ImagingClusterReco.h#L95
Possibly others, hard to tell until we fix the error reporting via #1069 or something else.

Adding a `try {} catch(...) { std::terminate(); }` block detects an `std::invalid_argument` "Can only add immutable objects to subset collections". A workaround is to add some `clone()` calls.

PR #1379, opened 2024-04-13
treewide: fix "Immutable objects can only be added to subset collections" exception

veprbl:
This fixes compatibility with PODIO 00-17-04+ where check

https://github.com/AIDASoft/podio/blob/5964958a9b151ea3637342e1e9aad1629f6de7a6/python/templates/Collection.cc.jinja2#L161C34-L161C91

was implemented as a part of the

https://github.com/AIDASoft/podio/commit/e5a13c2b8382c6969c6b3a6092de7adda0fd33f0

Fixes: #1366


wdconinc (approved):
Looks good to me.

Issue #1387, opened 2024-04-17
Docs: Outdated Comment Left in MC-Cluster Association Code


ruse-traveler:
There is a comment left in the MCParticle-Cluster association code stating that the subsequent lines don't work for `HcalEndcapPClusters`. This no longer seems to be case, though. Attached is (1) a screenshot of the code as written, and (2) a text file of EICrecon output running over 100 events of single pi- pointed at the forward region

*Note:* only `MCParticles`, `GeneratedParticles`, `LFHCALHits`, `LFHCalRecHits`, `LFHCalClusters`, and `LFHCALTruthClusters` were included in the output collections.

![TruthClusterAssocationCodeAsWritten d17m4y2024](https://github.com/eic/EICrecon/assets/35080855/16a2ec31-84bd-4fbd-81fc-ddd077b091d9)
[forOldAssocCodeTest.withDebugMessagesAndPlugin.e10tht75335n1pim.log](https://github.com/eic/EICrecon/files/15015237/forOldAssocCodeTest.withDebugMessagesAndPlugin.e10tht75335n1pim.log)

PR #1388, opened 2024-04-17
Remove outdated comment on mc-cluster associations

ruse-traveler:

This PR removes an old comment left in the MCParticle-Cluster association code. The comment stated that the code doesn't work for HcalEndcapPClusters, However, this seems to be no longer the case.

- [X] Bug fix (issue #1387 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -91,14 +91,6 @@ namespace eicrecon {
           }
         );
 
-        // FIXME: The code below fails for HcalEndcapPClusters. This does not happen for
-        // FIXME: all calorimeters. A brief scan of the code suggests this could be caused
-        // FIXME: by the CalorimeterHitDigi algorithm modifying the cellID for the raw hits.
-        // FIXME: Thus, the cellID values passed on through to here no longer match those
-        // FIXME: in the low-level truth hits. It likely works for other detectors because
-        // FIXME: their u_fields and u_refs members are left empty which effectively results
-        // FIXME: in the cellID being unchanged.
-
         // 2. find mchit with same CellID
         // find_if not working, https://github.com/AIDASoft/podio/pull/273
         //auto mchit = std::find_if(


src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -91,14 +91,6 @@ namespace eicrecon {
...
-        // FIXME: all calorimeters. A brief scan of the code suggests this could be caused
-        // FIXME: by the CalorimeterHitDigi algorithm modifying the cellID for the raw hits.
-        // FIXME: Thus, the cellID values passed on through to here no longer match those
-        // FIXME: in the low-level truth hits. It likely works for other detectors because
-        // FIXME: their u_fields and u_refs members are left empty which effectively results
-        // FIXME: in the cellID being unchanged.
-
         // 2. find mchit with same CellID
         // find_if not working, https://github.com/AIDASoft/podio/pull/273
         //auto mchit = std::find_if(
```
veprbl, line 96: How about we uncomment the code and remove the loop version to show that this is indeed would be a no-op change?

ruse-traveler: Well, std::find_if still isn't working (the PODIO issue linked to still isn't resolved), but the loop version is...

veprbl: Okay, I see now. The comment was added in 46b23c42efc5fb3f977eec5e501c173aa88d7dd7 during the porting effort. There are some tricks that Digi/Reco can do to the cellID's, so we are not completely out of the water, but storing issues in the FIXME's is not the way in any case.

veprbl (approved):
Let's remove the FIXME comments. Issues need to be filed, if bugs are reappear in interplay of Digi and clustering.

ruse-traveler:
> Let's remove the FIXME comments. Issues need to be filed, if bugs are reappear in interplay of Digi and clustering.

Sounds good!

PR #1391, opened 2024-04-18
Cluster axis direction

sebouh137:
Added code to produce the direction of clusters in CalorimeterClusterRecoCoG

adds the cluster direction (i.e., the eigenvector corresponding to the largest eigenvector of the matrix corresponding to the inertia matrix of the cluster)

- [ ] Bug fix (issue #__)
- [X] New feature (issue #1462 )
- [ ] Documentation update
- [ ] Other: __

no
Yes.  While calculating the shape parameters, the hits are assigned weights, in a similar manner to how they are assigned weights when determining the cluster position.  Previously, there were some inconsistencies in how these weights were determined when using certain options in the cluster fitting, which were fixed in order to make this feature (the cluster axis direction) work properly.


src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -289,11 +290,32 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+      auto eigenvectors= es_3D.eigenvectors();
+      double maxEigenvalue=0;
+      int indexOfMaxEigenvalue=0;
+      for (int i_eigen=0;  i_eigen<3; i_eigen++){
+        auto eigenvalue=eigenValues_3D(i_eigen,0);
+        if(eigenvalue.real()>maxEigenvalue){
+          maxEigenvalue=eigenvalue.real();
+          indexOfMaxEigenvalue=i_eigen;
+        }
+      }
```
wdconinc, line 308: Use std::max on `es_3D` with compare on the eigenvalues.

sebouh137: Hi @wdconinc , I'm trying to get the _index_ of the maximum eigenvalue, not the maximum eigenvalue itself.  Does  std::max return the maximum value or the index of the maximum value?

wdconinc: `std::max_element`

wdconinc: https://en.cppreference.com/w/cpp/algorithm/max_element

sebouh137: @wdconinc 
max_element does not work unless the object you are iterating over is an iterator.  I don't see any way to convert this object, eigenValues_3D, which is type ‘Eigen::Vector3cf’ {aka ‘Eigen::Matrix<std::complex<float>, 3, 1>’} to an iterator type (such as std::vector) without using a similar amount of code to what I used.  So I think this code should remain the way it is.  
Best regards,
Sebouh

sebouh137:
I am not sure why the tests is failing.  It does not appear to be anything on my end.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
wdconinc: `eigenValues_3D.begin()` is an iterator. See https://godbolt.org/z/vbzGY3P6r for a working example.

wdconinc: The gist is
```cpp
    auto max_eigenvalue_it = std::max_element(
        eigenvalues.begin(),
        eigenvalues.end(),
        [](auto a, auto b) {
            return std::norm(a) < std::norm(b);
        }
    );
    auto max_eigenvector = eigenvectors.col(
        std::distance(
            eigenvalues.begin(),
            max_eigenvalue_it
        )
    );
```

sebouh137: thanks.  After a bit of work, I got it to work.

sebouh137:
> @wdconinc I am not sure why the tests is failing. It does not appear to be anything on my end. Could you check if this is something on my end or something outside the scope of this pull request? Best regards, Sebouh

 I got this to pass the checks that were failing, with @nathanwbrei 's help.  Now it is passing all checks.

veprbl (approved):

veprbl:
@sebouh137 I have some second thoughts about this. First of all, could we have used intrinsicEta inrinsicPhi fields? Second, could you, please, update the documentation https://github.com/eic/EDM4eic/blob/3b4975588356d81321b396490fe67fe31e438ac1/edm4eic.yaml#L292 to reflect new status-quo.

PR #1395, opened 2024-04-23
feat: convert RichTrack factory to JOmniFactory

wdconinc:
This PR converts the RichTrack_factory to JOmniFactory. This is one of the holdouts for this conversion since it does so much in terms of setting up planes, including parsing of collection names. Well, that's a thing of the past now.
- RichTrack_factory is actually a TrackPropagation_factory and could likely be the same factory (descoped).
- The configuration (tracking and filter surfaces) is now created in the plugin.
- TrackPropagation algorithm now understands about filter surfaces in its configuration.
- RichGeo_service's ActsGeo now returns tracking and filter surfaces instead of Acts surfaces.

What is lost compared to previously?
- `eicrecon -PDRICH:gas:numPlanes=10` won't work anymore and this needs to be provided through the plugin. If really required, this can be brought back since `DRICH::InitPlugin` has access to the `JApplication*` pointer, but I'd rather not set a precedent that it's ok to define command line arguments there...
- This now runs track propagation for DRICH tracks twice: once in the context of aerogel filtering and once for gas filtering. This all needs a more comprehensive re-evaluation that we are currently not ready for, though. Ultimately we need 1 track propagation to all surfaces, and then a distribution of those results to algorithms. Acts is not ready for that, nor are we.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes, if you were relying on the command line parameter `DRICH:gas:numPlanes=10`.

Yes, if you were relying on the command line parameter `DRICH:gas:numPlanes=10`.


wdconinc:
Asked for early reviews, even though this is currently killing the DRICH reconstruction, but that's just a temporary issue.

wdconinc:
Only diff now:
image

veprbl (dismissed):
LGTM

veprbl:
https://github.com/eic/EICrecon/actions/runs/8855484300/job/24333764925#step:8:225
Must be due to some exception thrown.

veprbl:
https://github.com/JeffersonLab/JANA2/issues/274

src/detectors/DRICH/DRICH.cc
```diff
@@ -67,10 +71,30 @@ extern "C" {
...
-    track_cfg.numPlanes.insert({ "Aerogel", 5  });
-    track_cfg.numPlanes.insert({ "Gas",     10 });
+    // Track propagation
+    TrackPropagationConfig aerogel_track_cfg;
+    TrackPropagationConfig gas_track_cfg;
+
+    // get RICH geo service
+    auto richGeoSvc = app->GetService<RichGeo_service>();
+    auto dd4hepGeo = richGeoSvc->GetDD4hepGeo();
+    try {
```
wdconinc, line 81: This needs to be an `if` with a `dd4hepGeo->detector("DRICH")` or similar call.

veprbl (approved):

Issue #898, opened 2023-09-05
Reconstruction: **[Particle Flow]** Validation and enhancement of existing MC-cluster associations


ruse-traveler:
Currently, EICrecon constructs associations between calorimeter clusters and the highest energy contributing MCParticle. A more judicious design choice -- e.g. associating each cluster to all contributing particles which initiate a shower in in a given calorimeter -- would make analysis easier downstream, and would substantially simplify debugging more complex reconstruction processes (like particle flow).

ruse-traveler:
Comment for posterity: I've added a flag indicating that this was identified as high-priority TDR need from the calorimetry side.

PR #1396, opened 2024-04-23
Create Multiple Primary MC-Cluster Associations

ruse-traveler:

This PR adjusts the logic used in establishing the truth-cluster associations in `CalorimeterClusterRecoCoG`. Previously, the truth in the truth-cluster associations was defined to be the MCParticle of the 1st contribution stored in the list of contributions to the SimCalorimeterHit corresponding to the highest energy hit in the cluster. This PR changes the logic such that now

1. Multiple MC-Cluster associations are created,
2. Associations link back to primary (`generator status == 1`) MC particles, and
3. Each association carries a weight of `E_{contributed} / E_{cluster}`.

This PR addresses #1475 and partially addresses #898. The `generator status == 1` condition should be revisited in a subsequent PR: this definition of primary may miss cases, particularly for longer lived neutral resonances such as the  `K^{0}_{L}`, which would be useful information to retain.

### What kind of change does this PR introduce?a
- [X] Bug fix (issues #898, #1475)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

While this PR doesn't introduce breaking changes, it does introduce substantial changes that users might need to account for in downstream analysis code (see below).

**Yes:** MC-Cluster associations (`edm4eic::MCRecoClusterParticleAssociations`) will now contain _multiple_ associations per clusters, and will link back to primary (`status == 1`) MC particles.


ruse-traveler:
@wdconinc I remember you pointing out that this could be an issue for systems with large numbers of contributions (specifically the LFHCal) in the chat today. I'll check to make sure this doesn't cause a drastic slowdown...

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -121,8 +121,19 @@ namespace eicrecon {
...
 
-        // 3. find mchit's MCParticle
-        const auto& mcp = mchit->getContributions(0).getParticle();
+        // 3. find mchit's highest energy MCParticle
+        double eMcPar = 0.;
+        double eMcMax = 0.;
+        uint32_t iMcMax = 0;
+        for (uint32_t iContrib = 0; const auto& contrib : mchit->getContributions()) {
+          eMcPar = contrib.getParticle().getEnergy();
+          if (eMcPar > eMcMax) {
```
wdconinc, line 130: You have more info that you can use here: you know that the total energy of all contributions is the total cluster energy. As you loop over the contributions, you can add up how much of the energy you have already accounted for. E.g. if your cluster has total energy of 5 GeV, your current maximum is 1 GeV, and your running sum is 4.2 GeV (which includes the 1 GeV), then you can stop evaluating the rest of the array. Requires non-log-weighted sum. This may be particularly useful if the most energetic contributions happen to appear early.

veprbl:
I don't think this is correct. If you have a low-energy electron and a high-energy pion in ECal, your pion can leave a MIP response, while electron will actually contribute most of the energy in the cluster. The code tries to answer the question: which particle is measured by the cluster?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: Oh! Very good point!!

ruse-traveler:
> I don't think this is correct. If you have a low-energy electron and a high-energy pion in ECal, your pion can leave a MIP response, while electron will actually contribute most of the energy in the cluster. The code tries to answer the question: which particle is measured by the cluster?

Ah! I see what you're saying. I was thinking about this in a very "HCal" sort of way... So maybe rather than selecting the highest _energy_ particle, I select the one associated with the largest contribution? That would dovetail nicely with what Wouter suggested above!

veprbl:
I think the current implication is that the highest energy contribution is at the first index. We should, of course, check that. My personal suspicion is that we might need to calculate total primary particles contributions including their downstream contributions in the shower, and then we pick the maximal one. Basically, I think we need to traverse the tree to the nodes with status=1.

veprbl:
And yes, it kind of makes sense for HCals, I agree.

ruse-traveler:
> I think the current implication is that the highest energy contribution is at the first index. We should, of course, check that. My personal suspicion is that we might need to calculate total primary particles contributions including their downstream contributions in the shower, and then we pick the maximal one. Basically, I think we need to traverse the tree to the nodes with status=1.

That lines up with what I've been seeing: I don't think I've ever seen a 2nd particle be the highest energy (granted this is for single particle events and we're only considering the highest energy hit in a cluster). But I agree: it would be good to confirm this...

On the status=1, I'm curious: what about situations where a shower particle from an ECal (or possibly from a support structure or the beampipe) creates a cluster in an HCal? My concern would be that the HCal isn't necessarily measuring the primary in this case...

veprbl:
According to capybara, this doesn't do much for any calorimeter. Maybe, B0 is slightly modified, but we don't even know why.

> On the status=1, I'm curious: what about situations where a shower particle from an ECal (or possibly from a support structure or the beampipe) creates a cluster in an HCal? My concern would be that the HCal isn't necessarily measuring the primary in this case...

It depends on what the purpose is. We don't build HCals to look good on beam tests, we need to do physics with those. Matching with primaries is a way to study how well we are doing that. Allegedly we don't store MCParticles outside of the tracking volume, so what would it even associate with.

ruse-traveler:
> Allegedly we don't store MCParticles outside of the tracking volume, so what would it even associate with.

AH! If that's the case then agreed: we should just stick with primaries.

ruse-traveler:
To see if these changes introduce a dramatic reduction in performance, I tried out these changes on 1000 18x275, Q^2>100 NC DIS events and compared against main run on the same events. The average throughput doesn't seem to be too different (0.201 for this PR vs. 0.196 for main)...
[main_nEvt1K.log](https://github.com/eic/EICrecon/files/15223106/main_nEvt1K.log)
[pr1396_withChanges_nEvt1K.log](https://github.com/eic/EICrecon/files/15223107/pr1396_withChanges_nEvt1K.log)

[BTW I ran this on SDCC]

veprbl:
This is interesting. Looking at capybara reports, I can't quite see a big change for HCals. Has anyone looked at changes for $E/p$ before and after this?

ruse-traveler:
> This is interesting. Looking at capybara reports, I can't quite see a big change for HCals. Has anyone looked at changes for E/p before and after this?

I did not! But I can quickly check that...

ruse-traveler:
Here is the E/p (cluster energy / association momentum) for several of the calorimeters in the central detector. In general, there's not much of a difference between main and this PR...

PECal_clustParEP mainVsPR1396 py8s18x275minq100ang0025div1NEvt1K d7m5y2024
PHCal_clustParEP mainVsPR1396 py8s18x275minq100ang0025div1NEvt1K d7m5y2024
BECal_clustParEP mainVsPR1396 py8s18x275minq100ang0025div1NEvt1K d7m5y2024
BHCal_clustParEP mainVsPR1396 py8s18x275minq100ang0025div1NEvt1K d7m5y2024
NECal_clustParEP mainVsPR1396 py8s18x275minq100ang0025div1NEvt1K d7m5y2024
NHCal_clustParEP mainVsPR1396 py8s18x275minq100ang0025div1NEvt1K d7m5y2024

veprbl:
So, if this doesn't do anything, we don't need to try to rush this into the release?

ruse-traveler:
> So, if this doesn't do anything, we don't need to try to rush this into the release?

I'd say so! And I think I'd appreciate a little more time to let this stew...

ruse-traveler:
Putting this back in draft while I implement changes...

ruse-traveler:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=eic_EICrecon&pullRequest=1396) **Quality Gate failed**
> Failed conditions  [15.4% Duplication on New Code](https://sonarcloud.io/component_measures?id=eic_EICrecon&pullRequest=1396&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=eic_EICrecon&pullRequest=1396)

Hmmmm... It looks like all of the duplication warnings are for the plugins where all of the changes are just adding preprocessor commands and swapping out collections where needed 😕

wdconinc:
> It looks like all of the duplication warnings are for the plugins

Because we treat our code as something to store data :-)

ruse-traveler:
> Because we treat our code as something to store data :-)

I'm not sure I follow 😅 
Like metadata?

ruse-traveler:
All right, I think this PR is in a state ready for another review!

If folks are okay with it, I've opted to hold off further investigation into the primary definition until next month. We can follow up in a subsequent PR if we decide that the definition here (status = 1) is insufficient.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -77,71 +81,24 @@ namespace eicrecon {
...
-        clusterassoc.setRecID(cl.getObjectID().index); // if not using collection, this is always set to -1
-        clusterassoc.setSimID(mcp.getObjectID().index);
-        clusterassoc.setWeight(1.0);
-        clusterassoc.setRec(cl);
-        clusterassoc.setSim(mcp);
-      } else {
+      // If sim hits are available, associate cluster with MCParticle
+#if EDM4EIC_VERSION_MAJOR >= 7
+      if (mchitassociations->size() == 0) {
+        debug("No mcHitAssociationCollection was provided, so no truth association will be performed.");
```
wdconinc, line 87: A collection was provided but was empty. That's different from what the debug message says.

src/detectors/BEMC/BEMC.cc
```diff
@@ -95,7 +95,11 @@ extern "C" {
           new JOmniFactoryGeneratorT<CalorimeterClusterRecoCoG_factory>(
              "EcalBarrelScFiClusters",
             {"EcalBarrelScFiProtoClusters",        // edm4eic::ProtoClusterCollection
+#if EDM4EIC_VERSION_MAJOR >= 7
+             "EcalBarrelScFiRawHitAssociations"},  // edm4eic::MCRecoCalorimeterHitAssociation
```
wdconinc, line 99: Can we already put this in for the EcalBarrelImagingClusters, even though I suspect EcalBarrelImagingRawHitAssociations is missing?

src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc
wdconinc, line 1: Can we put in a unit test for the new functionality?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,170 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+  std::vector<int> vecAssocSimHits;
+
+  // --------------------------------------------------------------------------
+  // 1. get associated sim hits and sum energy
+  // --------------------------------------------------------------------------
+  double eSimHitSum = 0.;
+  for (std::size_t iHit = 0; auto clhit : cl.getHits()) {
+
+    // make sure list of associated sim hits is empty
+    vecAssocSimHits.clear();
```
wdconinc, line 332: Instead of doing a clear, put `vecAssocSimHits` in this scope only.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,170 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+   *     the energy of the sim hits.
+   *  2. for each sim hit, identify all contributing primary particles, and sum
+   *     their contributed energy.
+   *  3. create an association for each contributiong primary with a weight
+   *     of contributed energy over total energy.
+   */
+
+  // bookkeeping containers
+  std::map<int, std::pair<int, int>> mapMCParToSimIndices;
+  std::map<int, double> mapMCParToContrib;
```
wdconinc, line 322: Maps from indices should use unsigned int, i.e. `std::size_t`.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,170 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+    // make sure list of associated sim hits is empty
+    vecAssocSimHits.clear();
+
+#if EDM4EIC_VERSION_MAJOR >= 7
+    std::size_t iAssociations = mchitassociations->size();
+    for (std::size_t iAssoc = 0; iAssoc < mchitassociations->size(); ++iAssoc) {
+
+      // if raw hit has same cell ID as reco hit, add sim hit to vector
+      // and increment energy sum
+      if (clhit.getCellID() == (*mchitassociations)[iAssoc].getRawHit().getCellID()) {
```
wdconinc, line 340: Why do we still need to do matching on CellID? Wasn't the point of the associations that you can now do matching on the actual hit?

wdconinc (commented):
I thought with the associations (EDM4eic >= 7) we would not need to do any CellID comparison. Why are they still done?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: Ah! That was just for matching rec-to-raw hits. I forgot at the time I could also just match based on ObjectID's (which would probably be preferable):
```
if (clhit.getRawHit().getObjectID() == (*mchitassociations)[iAssoc].getRawHit().getObjectID()) {
```

wdconinc: I think it should be just a regular equality even:
```suggestion
      if (clhit == (*mchitassociations)[iAssoc].getRawHit()) {
```
(which is probably exactly the same when compiled, but is neater in code)

src/detectors/BEMC/BEMC.cc
ruse-traveler: Hmmm... Good question 😕 

But that's correct: right this PR doesn't update the ImagingClusterReco algorithm, and there's no PR open (right now) to update it. So would it be better to keep this focused on ClusterRecoCoG and open a parallel PR for ImagingClusterReco, or make the change in this one?

src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc
ruse-traveler: Sure! I have to admit, I haven't put together a unit test yet, so it's completely clear to me how to go about that... The changes I made to `[calorimetry_CalorimeterClusterRecoCoG.cc](https://github.com/eic/EICrecon/blob/main/src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc)` probably aren't what we would be looking for, right?

src/detectors/BEMC/BEMC.cc
wdconinc: The imaging calorimeter also uses ClusterRecoCoG. The differences are only upstream, in how proto clusters are found. In that sense this is independent of which kind of clustering is used.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: Ah! I didn't realize you could do that!

src/detectors/BEMC/BEMC.cc
ruse-traveler: Ah! Do the Imaging Clusters also use RecoCoG? I could've sworn they used ImagingClusterReco...

wdconinc: Yeah, you're right.

ruse-traveler: In that case, it looks like the logic is identical to ClusterRecoCoG. So it should be pretty straightforward to swap out what's in ImagingClusterReco with what I have in this PR!

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+  // 1. get associated sim hits and sum energy
+  // --------------------------------------------------------------------------
+  double eSimHitSum = 0.;
+  for (std::size_t iHit = 0; auto clhit : cl.getHits()) {
+
+    // vector to hold associated sim hits
+    std::vector<std::size_t> vecAssocSimHits;
+
+#if EDM4EIC_VERSION_MAJOR >= 7
+    std::size_t iAssociations = mchitassociations->size();
```
veprbl, line 334:
```suggestion
```
unused

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+  std::map<std::size_t, double> mapMCParToContrib;
+
+  // --------------------------------------------------------------------------
+  // 1. get associated sim hits and sum energy
+  // --------------------------------------------------------------------------
+  double eSimHitSum = 0.;
+  for (std::size_t iHit = 0; auto clhit : cl.getHits()) {
+
+    // vector to hold associated sim hits
+    std::vector<std::size_t> vecAssocSimHits;
```
veprbl, line 331: Code with indices is incomprehensible. Just make a vector of those proxy objects
```suggestion
    std::vector<edm4eic::MCRecoCalorimeterHitAssociation> vecAssocSimHits;
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+    // otherwise increment sum and add to vector
+    if (iSimMatch == mchits->size()) {
+      debug("No matching SimHit for hit {}", clhit.getCellID());
+      continue;
+    } else {
+      vecAssocSimHits.push_back(iSimMatch);
+      eSimHitSum += (*mchits)[iSimMatch].getEnergy();
+    }
+#endif
+    trace("{} associated sim hits found for reco hit (cell ID = {})", vecAssocSimHits.size(), clhit.getCellID());
```
veprbl, line 364:
```suggestion
    debug("{} associated sim hits found for reco hit (cell ID = {})", vecAssocSimHits.size(), clhit.getCellID());
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+        if (mapMCParToContrib.find(idPrim) == mapMCParToContrib.end()) {
+          mapMCParToSimIndices.insert(
+            {idPrim, std::make_pair(iAssocSimHit, iContrib)}
+          );
+          mapMCParToContrib.insert(
+            {idPrim, contrib.getEnergy()}
+          );
+        } else {
+          mapMCParToContrib[idPrim] += contrib.getEnergy();
+        }
```
veprbl, line 390:
```suggestion
        if (!mapMCParToContrib.count(idPrim)) {
          mapMCParToSimIndices.insert(
            {idPrim, std::make_pair(iAssocSimHit, iContrib)}
          );
        }
        mapMCParToContrib[idPrim] += contrib.getEnergy();
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+      cl.getObjectID().index,
+      primary.getObjectID().index,
+      primary.getPDG(),
+      primary.getGeneratorStatus(),
+      primary.getEnergy(),
+      weight
+    );
+
+  }  // end contributor loop
+  return;
```
veprbl, line 438:
```suggestion
```
This is not Fortran.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+
+  // walk back through parents to find primary
+  //   - TODO finalize primary selection. This
+  //     can be improved!!
+  primary = contributor;
+  while (primary.parents_size() > 0) {
+    primary = primary.getParents(0);
+    if (primary.getGeneratorStatus() == 1) break;
+  }
+  return;
```
veprbl, line 459:
```suggestion
  return primary;
```
Return the proxy object, don't need to update it by reference.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+    const uint32_t iContrib = parAndSimIndices.second.second;
+    const double weight = mapMCParToContrib[parAndSimIndices.first] / eSimHitSum;
+
+    // get relevant MCParticle
+    edm4hep::MCParticle primary;
+#if EDM4EIC_VERSION_MAJOR >= 7
+    get_primary((*mchitassociations)[iAssocSimHit].getSimHit().getContributions(iContrib), primary);
+#else
+    get_primary((*mchits)[iAssocSimHit].getContributions(iContrib), primary);
+#endif
```
veprbl, line 419: Why do you need to do another lookup, if you have `parAndSimIndices.first`?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -77,71 +81,24 @@ namespace eicrecon {
...
-        clusterassoc.setWeight(1.0);
-        clusterassoc.setRec(cl);
-        clusterassoc.setSim(mcp);
+      // If sim hits are available, associate cluster with MCParticle
+#if EDM4EIC_VERSION_MAJOR >= 7
+      if (mchitassociations->size() == 0) {
+        debug("Provided MCRecoCalorimeterHitAssociation collection is empty! No truth associations will be performed.");
+#else
+      if (mchits->size() == 0) {
+        debug("Provided SimCalorimeterHitCollection is empty! No truth association will be performed.");
```
veprbl, line 90:
```suggestion
        debug("Provided SimCalorimeterHitCollection is empty.");
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -77,71 +81,24 @@ namespace eicrecon {
...
-        auto clusterassoc = associations->create();
-        clusterassoc.setRecID(cl.getObjectID().index); // if not using collection, this is always set to -1
-        clusterassoc.setSimID(mcp.getObjectID().index);
-        clusterassoc.setWeight(1.0);
-        clusterassoc.setRec(cl);
-        clusterassoc.setSim(mcp);
+      // If sim hits are available, associate cluster with MCParticle
+#if EDM4EIC_VERSION_MAJOR >= 7
+      if (mchitassociations->size() == 0) {
+        debug("Provided MCRecoCalorimeterHitAssociation collection is empty! No truth associations will be performed.");
```
veprbl, line 87:
```suggestion
        debug("Provided MCRecoCalorimeterHitAssociation collection is empty. No truth associations will be performed.");
```

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -338,4 +295,169 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
+#endif
+  edm4eic::MCRecoClusterParticleAssociationCollection* assocs
+) const {
+
+  // --------------------------------------------------------------------------
+  // Association Logic
+  // --------------------------------------------------------------------------
+  /*  1. identify all sim hits associated with a given protocluster, and sum
+   *     the energy of the sim hits.
+   *  2. for each sim hit, identify all contributing primary particles, and sum
```
veprbl, line 314:
```suggestion
   *  2. for each sim hit, identify the parent of the contributing particle, and sum
```

veprbl (commented):
This looks like a good change. In ideal world, we'd switch to using associations in a separate PR, but it's okay to do it together too, probably.

src/detectors/BEMC/BEMC.cc
veprbl: Marking resolved for this is out of scope.

ruse-traveler: Sounds good. I'll put in a parallel PR for the ImagingClusterReco.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: Fair! 😝

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: I'm not opposed to going about it this way (it would definitely simplify the bookkeeping), but I think this actually touches on the discussion in #1600 a little:
https://github.com/eic/EICrecon/pull/1600#discussion_r1733984818

Would there be an issue with working with the user-layer objects in this case?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: This was mostly a consequence of not providing `MCParticles` as an input to the algorithm (which -- TBH -- I didn't think to do just because the original code didn't do so). Since I couldn't do something like `(*mcpars)[ parAndSimIndices.first ]`, I instead just had it look the primary up again based on its associated sim hit and contribution indices.

One way this could be simplified might be to key these maps on the proxy objects themselves, which would eliminate the need for the lookup. Or I could provide `MCParticles` as an argument and look it up via the `ObjectID().index`.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
veprbl: Yes, this is the intended way.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
veprbl: Technically, MCParticles should be provided, but that's not the point. If you just saved `edm4eic::MCParticles` to the key, that would allow you to access it here.

ruse-traveler: Gotcha! Makes sense!

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: I cleaned up how the preprocessor if-else's around these lines: now there's just one if-else statement, and hopefully this block is a lot more readable. Now it should be clear that if there are no hit associations or mc hits, no attempt at cluster association will be done.

So in the end, I replaced the "!" but left the "No truth association will be preformed". This is partly for parallelism the hit association case, but also just to make the debugging output more explicit.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
ruse-traveler: I reworked the description of the logic here: I tried to point out that we do look through the parents of each contributing particle for each sim hit, and that we only add it to the list of associations to make if the parent is a primary.

Let me know if there's a more clear way to phrase things...

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
veprbl: Yes, you are right, we should keep second part. I must have misread the code.

ruse-traveler:
Hi all, just wanted to check in on this: does the recent round of commits resolve the open issues, or is there more to discuss? Thanks!

veprbl:
[@ruse-traveler](https://github.com/ruse-traveler) I'd like to spend some more time checking the actual change. Have you done any benchmarking on your end?

ruse-traveler:
Same situation as #1406: mostly just spot-checks. Like with the that one, I can run some more detailed checks...

veprbl:
I'm experiencing some issues with this while processing DIS CC (simulated from `EPIC/EVGEN/DIS/CC/10x100/minQ2=1000/pythia8CCDIS_10x100_minQ2=1000_beamEffects_xAngle=-0.025_hiDiv_1.hepmc3.tree.root`).
```
reco> [EEMC:EcalEndcapNClusters] [debug] Associated cluster #5 to MC Particle #0 (pid = 2212, status = 4, energy = 99.94178715303674) with weight (0.9867191517808613)
reco> [EEMC:EcalEndcapNClusters] [debug] Associated cluster #5 to MC Particle #39 (pid = -211, status = 1, energy = 0.3006170675754705) with weight (0.01328087107687527)
```
```diff
diff --git a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
index ff75e3aa..2cffdd50 100644
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -419,6 +419,12 @@ edm4hep::MCParticle CalorimeterClusterRecoCoG::get_primary(const edm4hep::CaloHi
   while (primary.parents_size() > 0) {
     primary = primary.getParents(0);
     if (primary.getGeneratorStatus() == 1) break;
+    trace("descending MC Particle #{} (pid = {}, status = {}, energy = {})",
+      primary.getObjectID().index,
+      primary.getPDG(),
+      primary.getGeneratorStatus(),
+      primary.getEnergy()
+    );
   }
   return primary;
 }
```
yields outputs like
```
reco> [EEMC:EcalEndcapNClusters] [trace] descending MC Particle #57 (pid = 310, status = 2, energy = 1.527582375052151)
reco> [EEMC:EcalEndcapNClusters] [trace] descending MC Particle #40 (pid = 311, status = 2, energy = 1.5275850283377552)
reco> [EEMC:EcalEndcapNClusters] [trace] descending MC Particle #6 (pid = 2101, status = 71, energy = 39.28357352176002)
reco> [EEMC:EcalEndcapNClusters] [trace] descending MC Particle #2 (pid = 2101, status = 63, energy = 39.28357352176002)
reco> [EEMC:EcalEndcapNClusters] [trace] descending MC Particle #0 (pid = 2212, status = 4, energy = 99.92043455278755)
reco> [EEMC:EcalEndcapNClusters] [trace] Identified primary: id = 0, pid = 2212, total energy = 99.92043455278755, contributed = 0.033173649972013664
```
Not sure if this is only effect for beam remnants, or status codes are overwritten somehow else (and then it's a bug somewhere). Workaround for this is to stop at the generated event boundary.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -417,8 +417,8 @@ edm4hep::MCParticle CalorimeterClusterRecoCoG::get_primary(const edm4hep::CaloHi
   //     can be improved!!
   edm4hep::MCParticle primary = contributor;
   while (primary.parents_size() > 0) {
+    if (primary.getGeneratorStatus() != 0) break;
     primary = primary.getParents(0);
-    if (primary.getGeneratorStatus() == 1) break;
```
veprbl: Oh, and the original ordering must have been the logical error. However even after this reordering I see status code 2, so I'd keep the new condition too.

image
(red: main, blue: this commit).

veprbl (approved):
Tested in a private benchmark. Doesn't seem to break it anymore.

veprbl:
Another fun plot:
image

ruse-traveler:
> Another fun plot: image (red: main, blue: this commit).

Nice!! Looks very good! Thanks for catching that bug with the event boundary! 🙌 
(Also apologies for the delayed response!)

PR #1397, opened 2024-04-24
linux-eic-shell.yml: run trigger-container for when merge_group is tested

veprbl:
Since we now require container build to pass as a requirement for successful merge to main, we need to actually attempt to build it.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -886,11 +886,12 @@ jobs:
     runs-on: ubuntu-latest
     env:
       secret: ${{ secrets.EICWEB_CONTAINER_TRIGGER }}
-    if: ${{ github.actor != 'dependabot[bot]' && secret != '' }}
+    if: ${{ github.actor != 'dependabot[bot]' }}
     needs:
     - eicrecon-gun
     steps:
     - uses: eic/trigger-gitlab-ci@v2
+      if: ${{ secret != '' }}
       id: trigger
       with:
         url: https://eicweb.phy.anl.gov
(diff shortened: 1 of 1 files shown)


veprbl:
@simonge Can you approve this? This is safe (tm).

simonge (dismissed):
Trusting...

veprbl:
I've changed timeout from 3h to 6h. Let's try again.

veprbl:
Not going to work https://github.com/eic/EICrecon/actions/runs/8819319775/job/24210808245
What we use in eic/epic does not require a synced branch

veprbl:
This needs
https://github.com/eic/detector_benchmarks/blob/16fd651e6f9ce0c670eefe047d337cbe7bbc8296/.github/workflows/mirror.yaml#L19-L29
plus, if I remember correctly, also merge group ref somehow didn't work, so a fix is needed.

I will disable the container requirement for now.

.github/workflows/linux-eic-shell.yml
```diff
@@ -884,7 +884,7 @@ jobs:
 
   trigger-container:
     runs-on: ubuntu-latest
-    if: ${{ github.event_name != 'merge_group' && github.actor != 'dependabot[bot]' }}
+    if: ${{ github.actor != 'dependabot[bot]' }}
```
wdconinc, line 887: While we're here:
```suggestion
    env: 
      secret: ${{ secrets.EICWEB_CONTAINER_TRIGGER }}
    if: ${{ github.actor != 'dependabot[bot]' && secret != '' }}
```

veprbl: Why can't we compare the value directly?

veprbl: Okay, I was confused, this is not a `step.*.if`, but a `jobs.*.if`, where `env` should not be accessible, allegedly.

veprbl:
@wdconinc Let's try again, since, I think, I resolved the 404 issue. Please, approve.

wdconinc (approved):
Another one for the history books!

veprbl:
We require eicweb benchmarks for eic/epic repository merges, yet somehow here it failed. In addition to container build failures and slowness, I don't think this is reliable enough to enable this now.

PR #1402, opened 2024-04-30
Add minbias test

simonge:
Adds tests for the pythia6 min bias events at all standard beam energies. This includes electrons within the Low-Q2 tagger acceptance which currently no other samples cover.


simonge:
Am I missing something to have the new tests included in the capybara summary or is this because it fails to find a previous artifact for the tests?

veprbl:
I believe, the "Compare to previous artifacts" step is skipped because main doesn't have those yet.

veprbl (approved):

Issue #1403, opened 2024-05-01
cluster merging for neutron reconstruction


sebouh137:
**Is your feature request related to a problem? Please describe.**
We propse to add a new feature wherein neutron candidates are reconstructed by merging all clusters in a particular detector, for instance the ZDC HCAL.

**Describe the solution you'd like**
Create new algorithm, and add it to the recon for the ZDC.  

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1404, opened 2024-05-01
added a 'neutron reconstruction' algorithm which creates a neutron candidate that merges all clusters in a given bank

sebouh137:
Adds a cluster-merging algorithm that creates a "neutron candidate" (of type ReconstructedParticle) that merges all clusters in a given bank together.  

- [ ] Bug fix (issue #__)
- [X] New feature (issue #1403 )
- [ ] Documentation update
- [ ] Other: __

no
Yes, changes ZDC reconstruction to use ImagingTopo clustering instead of island clustering.   Also tweaked the ImagingTopo algorithm parameters that are used in ZDC recon


wdconinc:
@sebouh137 Please fix merge conflicts.

sebouh137:
> @sebouh137 Please fix merge conflicts.

Fixed

src/algorithms/calorimetry/NeutronReconstructionConfig.h
```diff
@@ -0,0 +1,12 @@
+// Subject to the terms in the LICENSE file found in the top-level directory.
+
+#pragma once
+
+namespace eicrecon {
+
+struct NeutronReconstructionConfig {
+
+
+};
```
ruse-traveler, line 10: Should there be parameters here? If not, I'd suggest using `WithPodConfig<NoConfig>` and removing this file.

ruse-traveler (changes requested):
Thanks for the development, Sebouh! I have a minor fix suggested on the code itself, but I have some larger conceptual suggestions:

- I think this might make more sense alongside the `reco` algorithms since this is taking the final step from detector objects (topoclusters) to physics objects (neutron candidates)
- I would suggest renaming this to something like `FarForwardNeutronReconstrution` since neutrons will also be reconstructed elsewhere in ePIC with very different algorithms.

src/algorithms/calorimetry/NeutronReconstructionConfig.h
```diff
@@ -0,0 +1,12 @@
+// Subject to the terms in the LICENSE file found in the top-level directory.
```
veprbl, line 1:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024 Sebouh Pual
```

veprbl: Where did you copy this from?

src/algorithms/calorimetry/NeutronReconstructionConfig.h
sebouh137: Done as requested

sebouh137:
> Thanks for the development, Sebouh! I have a minor fix suggested on the code itself, but I have some larger conceptual suggestions:
> 
> * I think this might make more sense alongside the `reco` algorithms since this is taking the final step from detector objects (topoclusters) to physics objects (neutron candidates)
Ok, I have moved it there.  
> * I would suggest renaming this to something like `FarForwardNeutronReconstrution` since neutrons will also be reconstructed elsewhere in ePIC with very different algorithms.
Ok, I have changed the name.

src/detectors/ZDC/ZDC.cc
```diff
@@ -195,6 +195,13 @@ extern "C" {
           app   // TODO: Remove me once fixed
         ));
 
+        app->Add(new JOmniFactoryGeneratorT<FarForwardNeutronReconstruction_factory>(
+           "HcalFarForwardZDCNeutronCandidates",
+          {"HcalFarForwardZDCClusters"},  // edm4eic::ClusterCollection
+          {"HcalFarForwardZDCNeutronCandidates"}, // edm4eic::ReconstrutedParticleCollection,
+          app   // TODO: Remove me once fixed
+        ));
+
```
ruse-traveler, line 204: Whoops! Should've commented on this in my last review: alongside moving the algo/factory to reco, I would suggest moving this to the reco plugin (`global/reco/reco.cc`) and renaming the output to `ReconstructedFarForwardNeutrons` (or similar).

ruse-traveler (changes requested):
Nice! Thanks for incorporating the feedback! I see the conflicts and copyrights are fixed, so aside from the one suggestion I forgot to include in my last review, I'm happy with this!

src/algorithms/calorimetry/NeutronReconstructionConfig.h
sebouh137: I think this was in one of the other Config.h files, but I forget which one.  However, since I have removed this file (due to there being no parameters), it is no longer relevant.

src/detectors/ZDC/ZDC.cc
sebouh137: Done, thanks!

ruse-traveler: Have you pushed the changes? I'm still seeing the factory declaration in `zdc.cc`...

sebouh137: oops; made the change locally but forgot to git add; git commit ..., before doing git pull; git push.   Pushed to the github now.

ruse-traveler: Thanks! One final suggestion and then I'm happy to merge! I'd suggest renaming the output to `"ReconstructedFarForwardZDCNeutrons"` just to make it consistent with most of the other reconstructed particle collections

sebouh137: done

Issue #897, opened 2023-09-05
Reconstruction: **[Particle Flow] **Validation of existing cluster-splitting


ruse-traveler:

ruse-traveler:
Comment for posterity: I've added a flag indicating that this was identified a high-priority TDR need from the calorimetry side.

PR #1406, opened 2024-05-01
Add Track-Based Cluster Merging Algorithm

ruse-traveler:

This PR introduces the `TrackClusterMergeSplitter` algorithm, an algorithm for merging and splitting calorimeter clusters based on matched track projections. Inspired by the [Particle Flow algorithm from ATLAS](https://doi.org/10.1140/epjc/s10052-017-5031-2).

- [ ] Bug fix (issue #__)
- [X] New feature (issue #897 )
- [ ] Documentation update
- [ ] Other: __

No.

Yes. Now "SplitMerge" `edm4eic::Cluster` collections and corresponding associations will be created for 6 of the central calorimeters (`NHCal`, `EEEMCal`, `BHCal`, `BIC`, `LFHCal`, and `FEMC`). Track projections will need to be extended to other calorimeters before the algorithm introduced here can be used on them.

### To-Do

The split-merge parameters for each calorimeter need to be updated from the current dummy values used for testing. In the interest of getting the functionality of this algorithm merged, this could update could be done in a separate PR.


ruse-traveler:
All of the logic is here now, and it's functioning as I would expect for the (very) arbitrary parameters I've plugged in so far...

One thing I realized in the course of coding this up is that this would be better served as a step between the initial protocluster formation and ClusterRecoCoG (I ended up unnecessarily duplicating a lot of code from the former). This will be a relatively straightforward change since the logic is identical and a _lot_ of code can just be simply deleted...

ruse-traveler:
@veprbl I think the HCal parameters are now good enough for this initial implementation and to start really kicking the tires on this.

Due to some internet connectivity issues and some condor issues on SDCC, I have to rerun the jobs for the ECal parameters (which I'm tuning to single electron/positron simulations). It's a relatively quick turnaround and I could get things done by later tonight, but I'd also be okay dropping the ECal instantiations for this PR and getting them in as a bugfix... What are your thoughts?

veprbl:
It's great to see that this is ready for review! However, I don't think we are going to be able to review this in time for 24.07.0. This is a big chunk of new code, would be nice to have a proper review.

ruse-traveler:
That's fair! We can mark this for 24.08.0...

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Derek Anderson
+
+#define EICRECON_TRACKCLUSTERMERGESPLITTER_CC
+
```
veprbl, line 5:
```suggestion
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+#include <cmath>
+#include <regex>
+#include <utility>
+#include <algorithm>
+// dd4hep utilities
+#include <DD4hep/Readout.h>
+// jana utilities
+#include <JANA/JException.h>
```
veprbl, line 13:
```suggestion
```
Not supposed to use JANA things in algorithms.

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
```diff
@@ -0,0 +1,111 @@
...
+      const dd4hep::Detector* m_detector {NULL};
+      const dd4hep::rec::CellIDPositionConverter* m_converter {NULL};
+
+      // bookkeeping members
+      mutable MapToFlag m_mapIsConsumed;
+      mutable MapOneToOne m_mapClustProject;
+      mutable MapOneToMany m_mapClustToMerge;
+      mutable MapOneToMany m_mapProjToMerge;
+      mutable VecTrkPoint m_vecProject;
+      mutable VecCluster m_vecClust;
```
veprbl, line 105:
```suggestion
```
Please book the variables in process or within `process()` or downstream, as close as possible to where they are used. This is unreadable and not thread-safe.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+    const TrackClusterMergeSplitter::Output& output
+  ) const {
+
+    // grab inputs/outputs
+    const auto [in_protoclusters, in_projections] = input;
+    auto [out_protoclusters] = output;
+
+    // exit if no clusters in collection
+    if (in_protoclusters -> size() == 0) {
+      throw JException("No proto-clusters in input collection!");
```
veprbl, line 58:
```suggestion
      throw std::runtime_exception("No proto-clusters in input collection!");
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    // reset bookkeeping containers
+    reset_bookkeepers();
+
+    // flag all clusters as not being consumed
+    for (auto cluster : *in_protoclusters) {
+      m_mapIsConsumed.insert(
+        {cluster.getObjectID().index, false}
```
veprbl, line 67: Using just the index in the collection without the full id will break if input collection is a subset collection combining several other inputs. Is it possible to use full ObjectID's?

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+    m_mapClustToMerge.clear();
+    m_mapProjToMerge.clear();
+    m_vecProject.clear();
+    m_vecClust.clear();
+    trace("Reset bookkeeping containers");
+
+  }  // end 'reset_bookkeepers()'
+
+
+
```
veprbl, line 234:
```suggestion
```
This should not exist. Use local variables in the methods instead.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+    // get readout
+    auto readout = m_converter -> findReadout(
+      m_converter -> findDetElement(
+        m_converter -> position(hit.getCellID())
+      )
+    );
+
+    // grab detector id
+    const int id = m_detector -> constant<int>(
+      std::regex_replace(readout.name(), std::regex("Hits"), "_ID")
```
veprbl, line 258:
```suggestion
      readout.name().replace("Hits", "_ID")
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+      debug("No projections in input collection.");
+      return;
+    }
+
+    // get readout
+    auto readout = m_converter -> findReadout(
+      m_converter -> findDetElement(
+        m_converter -> position(hit.getCellID())
+      )
+    );
```
veprbl, line 254: This is a dirty hack. Just provide subsystem id constant name via configuration.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+
+    // loop over relevant projections
+    for (uint32_t iProject = 0; iProject < m_vecProject.size(); ++iProject) {
+
+      // grab projection
+      auto project = m_vecProject[iProject];
+
+      // get eta, phi of projection
+      const float etaProj = edm4hep::utils::eta(project.position);
+      const float phiProj = atan2(project.position.y, project.position.x);
```
veprbl, line 297: Is there a function for azimuthal angle?

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+      // find closest cluster
+      bool  foundMatch = false;
+      float dMatch = std::numeric_limits<float>::max();
+      int   iMatch = std::numeric_limits<int>::max();
+      for (auto cluster : *clusters) {
+
+        // get eta, phi of cluster
+        const auto  posClust = get_cluster_position(cluster);
+        const float etaClust = edm4hep::utils::eta(posClust);
+        const float phiClust = std::atan2(posClust.y, posClust.x);
```
veprbl, line 308: Is there a function for azimuthal angle?

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+
+        // calculate distance to centroid
+        const float dist = std::hypot(
+          etaProj - etaClust,
+          phiProj - phiClust
```
veprbl, line 313:
```suggestion
          std::remainder(phiProj - phiClust, 2 * M_PI);
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+      auto project = m_vecProject[iProject];
+
+      // get eta, phi of projection
+      const float etaProj = edm4hep::utils::eta(project.position);
+      const float phiProj = atan2(project.position.y, project.position.x);
+
+      // find closest cluster
+      bool  foundMatch = false;
+      float dMatch = std::numeric_limits<float>::max();
+      int   iMatch = std::numeric_limits<int>::max();
```
veprbl, line 302:
```suggestion
      float dMatch;
      int   iMatch;
```
foundMatch should be enough to prevent UB in the current code. If somebody ever breaks, UB is preferable to inserting nonsense, to my taste.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+    trace("Collected relevant projections: {} to process", m_vecProject.size());
+
+  }  // end 'get_projections(edm4eic::CalorimeterHit&, edm4eic::TrackSegmentCollection&)'
+
+
+
+  // --------------------------------------------------------------------------
+  //! Match clusters to track projections
+  // --------------------------------------------------------------------------
+  /*! FIXME this might be better handled in a separate algorithm
```
veprbl, line 282: Would be nice to have a single track-cluster matcher, yes.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+  // --------------------------------------------------------------------------
+  void TrackClusterMergeSplitter::merge_clusters(
+    const edm4eic::TrackPoint& matched_trk,
+    const VecCluster& to_merge,
+    edm4eic::MutableProtoCluster& merged_clust
+  ) const {
+
+    // get track eta, phi
+    const float etaTrk = edm4hep::utils::eta(matched_trk.position);
+    const float phiTrk = atan2(matched_trk.position.y, matched_trk.position.x);
```
veprbl, line 349: Same as above.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+      for (auto hit : old_clust.getHits()) {
+
+        // get hit eta, phi
+        const float etaHit = edm4hep::utils::eta(hit.getPosition());
+        const float phiHit = std::atan2(hit.getPosition().y, hit.getPosition().x);
+
+        // get distance to track
+        const float dist = std::hypot(
+          etaHit - etaTrk,
+          phiHit - phiTrk
```
veprbl, line 365:
```suggestion
          std::remainder(phiHit - phiTrk, 2 * M_PI);
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+  ) const {
+
+    // get track eta, phi
+    const float etaTrk = edm4hep::utils::eta(matched_trk.position);
+    const float phiTrk = atan2(matched_trk.position.y, matched_trk.position.x);
+
+    // grab hits from each cluster to merge
+    float eTotal = 0.;
+    float tClust = std::numeric_limits<float>::max();
+    float tErrClust = std::numeric_limits<float>::max();
```
veprbl, line 354:
```suggestion
```
Unused.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+      auto projSeed = m_vecProject[clustAndProject.second];
+
+      // add cluster to lists and flag as used
+      m_mapClustToMerge.insert(
+        {clustAndProject.first, {}}
+      );
+      m_mapProjToMerge.insert(
+        {clustAndProject.first, {clustAndProject.second}}
+      );
+      m_mapIsConsumed[clustAndProject.first] = true;
```
veprbl, line 101: Just use `std::set`/`std::unordered_set`.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+
+        // skip if seed cluster
+        if (iCluster == clustAndProject.first) {
+          continue;
+        }
+
+        // get eta, phi of cluster
+        const auto  posClust = get_cluster_position(in_cluster);
+        const float etaClust = edm4hep::utils::eta(posClust);
+        const float phiClust = std::atan2(posClust.y, posClust.x);
```
veprbl, line 141: Any function for azimuthal angle in utils?

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+          m_mapProjToMerge[clustAndProject.first].size()
+        );
+      }  // end cluster loop
+    }  // end matched cluster-projection loop
+
+    // do cluster merging
+    for (auto clustToMerge : m_mapClustToMerge) {
+
+      // add clusters to be merged
+      m_vecClust.clear();
```
veprbl, line 182: Use local variables, please.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,460 @@
...
+
+    // get readout
+    auto readout = m_converter -> findReadout(
+      m_converter -> findDetElement(
+        m_converter -> position(hit.getCellID())
+      )
+    );
+
+    // grab detector id
+    const int id = m_detector -> constant<int>(
```
veprbl, line 257: Please, move lookups to the `init`. Also, why not just make a decoder for the `"system"` field and extract it from cellID that way?

veprbl (commented):
Trying to wrap my head around this. Looks good, but I have some suggestions.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Whoops! Good catch!

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Ah! I didn't even think to look, but there is! I've swapped out all the instances of `atan2(...)` for that...

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
ruse-traveler: Ahhh I didn't even think about thread safety! I've turned these into local variables and updated the relevant methods accordingly...

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Oops! They are indeed...

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Actually, I could provide the name of the ID as a parameter, then this gets nicely reduced to looking up the constant in `init()`!

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Like I mentioned up above, an alternative would be to provide the ID name as a parameter. Then we wouldn't even need the decoder...

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Ah! That's something I hadn't considered! It should be, but I might need to rework how I'm doing my bookkeeping (which might be a good thing since I can probably simplify things)...

ruse-traveler: Thinking through this: would subsets be an issue if rather than using `ObjectID().index` I used the index of the cluster/projection in the input collections, e.g. `iCluster` in (*in_protoclusters)[iCluster]`?

veprbl: It's the same value, no?

veprbl: I guess, it's not for the subset collections. Not sure why not just use ObjectID, but this would be fine too.

ruse-traveler: Gotcha! I've been working with just the indices since there are a few times where I retrieve an object from the input using the index (e.g. at [line 181](https://github.com/eic/EICrecon/blob/a72118deb86a22368b9f445425b37b67881a74d0/src/algorithms/calorimetry/TrackClusterMergeSplitter.cc#L181C1-L186C69)).:
```
      // add clusters to be merged
      vecClust.clear();
      for (const int& iClustToMerge : clustToMerge.second) {
        vecClust.push_back( (*in_protoclusters)[iClustToMerge] );
      }
      vecClust.push_back( (*in_protoclusters)[clustToMerge.first] );
```

I definitely could key the maps on ObjectID's, and then use `ObjectID().index` when needed... But I would think that would still run into the subset issue, right?

veprbl: That would not, as object id also contains collection id.

ruse-traveler: Gotcha! I'll just do that, then!

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: I'm a little unclear on how I might use `std::set` here... Would this be for `m_mapIsConsumed` or for `m_mapProjToMerge`? (Or both?)

veprbl: Just for `m_mapIsConsumed`. There is no point to store "true", the mere presence in a set carries that information.

ruse-traveler: Ah! I see! Gotcha!

ruse-traveler:
Hi all, I think this is ready for a re-review. But I am seeing some odd failures on the checks, e.g. on eicrecon-dis (clang, 10x100, 0, craterlake):
```
ValueError: failed to validate Range1d(id='p166776', ...).bounds: expected either None or a value of type MinMaxBounds(Auto, Tuple(Float, Float), Tuple(Nullable(Float), Float), Tuple(Float, Nullable(Float)), Tuple(TimeDelta, TimeDelta), Tuple(Nullable(TimeDelta), TimeDelta), Tuple(TimeDelta, Nullable(TimeDelta)), Tuple(Datetime, Datetime), Tuple(Nullable(Datetime), Datetime), Tuple(Datetime, Nullable(Datetime))), got (1.8247462191085126e+19, 1.8247462191085126e+19)
```
Which I'm not quite sure what it's trying to tell me... 🤔

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+    const TrackClusterMergeSplitter::Output& output
+  ) const {
+
+    // grab inputs/outputs
+    const auto [in_protoclusters, in_projections] = input;
+    auto [out_protoclusters] = output;
+
+    // exit if no clusters in collection
+    if (in_protoclusters -> size() == 0) {
+      throw std::runtime_error("No proto-clusters in input collection!");
```
veprbl, line 75:
```suggestion
      debug("No proto-clusters in input collection.");
      return
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+    const TrackClusterMergeSplitter::Input& input,
+    const TrackClusterMergeSplitter::Output& output
+  ) const {
+
+    // grab inputs/outputs
+    const auto [in_protoclusters, in_projections] = input;
+    auto [out_protoclusters] = output;
+
+    // exit if no clusters in collection
+    if (in_protoclusters -> size() == 0) {
```
veprbl, line 74:
```suggestion
    if (in_protoclusters->size() == 0) {
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+      const float phiSeed = edm4hep::utils::angleAzimuthal(posSeed);
+
+      // loop over other clusters
+      float eClustSum = eClustSeed;
+      float sigSum = sigSeed;
+      for (auto in_cluster : *in_protoclusters) {
+
+        // grab id and ignore used clusters
+        const podio::ObjectID idCluster = in_cluster.getObjectID();
+        if (setUsedClust.find(idCluster) != setUsedClust.end()) {
```
veprbl, line 150:
```suggestion
        if (setUsedClust.count(idCluster)) {
```

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+      }
+    }  // end clusters to merge loop
+
+    // ------------------------------------------------------------------------
+    // copy unused clusters to output
+    // ------------------------------------------------------------------------
+    for (auto in_cluster : *in_protoclusters) {
+
+      // ignore used clusters
+      if (setUsedClust.find(in_cluster.getObjectID()) != setUsedClust.end()) {
```
veprbl, line 233:
```suggestion
      if (setUsedClust.count(in_cluster.getObjectID())) {
```

veprbl (commented):
This looks good. It's a large PR, I have more suggestions.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+        const float weight = std::exp(-1. * dist / m_cfg.distScale);
+        const float eContrib = hit.getEnergy() * weight;
+
+        // increment sum and print debugging message
+        eTotal += eContrib;
+        trace("Added hit: eHit = {}, weight = {}, eContrib = {}, running total = {}", hit.getEnergy(), weight, eContrib, eTotal);
+
+        // set hit one-to-many/vector relations
+        merged_clust.addToHits( hit );
+        merged_clust.addToWeights( weight );
```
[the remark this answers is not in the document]
ruse-traveler: Ah! Good catch!! 😓

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+    for (auto in_cluster : *in_protoclusters) {
+
+      // ignore used clusters
+      if (setUsedClust.find(in_cluster.getObjectID()) != setUsedClust.end()) {
+        continue;
+      }
+
+      // copy cluster and add to output collection
+      edm4eic::MutableProtoCluster out_cluster = out_protoclusters->create();
+      copy_cluster(in_cluster, out_cluster);
```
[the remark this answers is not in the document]
ruse-traveler: Ah! Didn't realize you could do that!

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
```diff
@@ -0,0 +1,121 @@
...
+
+
+
+  // --------------------------------------------------------------------------
+  //! Convenience types
+  // --------------------------------------------------------------------------
+  typedef std::set<podio::ObjectID, CompareObjectID> SetOfIDs;
+  typedef std::map<podio::ObjectID, int, CompareObjectID> MapOneToIndex;
+  typedef std::map<podio::ObjectID, std::vector<int>, CompareObjectID> MapProjToMerge;
+  typedef std::map<podio::ObjectID, std::vector<podio::ObjectID>, CompareObjectID> MapClustToMerge;
```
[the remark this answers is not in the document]
ruse-traveler: Yeah... I think it would also eliminate the issue with subsets...

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+
+    float norm = 1.;
+    if (wTotal == 0.) {
+      warning("Total weight of 0 in position calculation!");
+    } else {
+      norm = wTotal;
+    }
+    return position / norm;
+
+  }  // end 'get_cluster_position(edm4eic::ProtoCluster&)'
```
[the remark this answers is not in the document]
ruse-traveler: This is something I've been thinking about for a bit now... I was initially planning on using `edm4eic::Cluster` as the input, but I opted for `edm4eic::Protocluster` instead since I felt that I would be duplicating more of ClusterRecoCoG's functionality if I stuck with just clusters (namely rerunning the cluster shape calculations, the CoG calculation, and handling the associations).

The trade-off, though, is that I lose the position as a member of the object. Which is why it gets duplicated here...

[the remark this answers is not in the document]
ruse-traveler: I've been thinking that routines like this, the cluster shape calculations, and similar that could get used across multiple algorithms might be useful to store somewhere centralized like an `edm4eic::utils`. @wdconinc and I briefly chatted about at the UGM...

But, of course, this is more just food for thought since this would be outside the scope of this PR 🤷

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -0,0 +1,462 @@
...
+      match_clusters_to_tracks(in_protoclusters, vecProject, mapClustProject);
+    }
+
+    // ------------------------------------------------------------------------
+    // 3. Loop over projection-cluster pairs to check if merging is needed
+    // ------------------------------------------------------------------------
+    SetOfIDs setUsedClust;
+    MapClustToMerge mapClustToMerge;
+    MapProjToMerge mapProjToMerge;
+    for (auto clustAndProject : mapClustProject) {
```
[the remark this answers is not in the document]
ruse-traveler: I've updated the bookkeepers to be hold actual proxy objects, which (I think) has substantially improved the clarity of blocks of code like this one...

Let me know, though, if you still think we need to define aliases for like `.first` , `.second`, etc...

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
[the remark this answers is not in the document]
ruse-traveler: Just now pushed a commit to update this procedure: the idea was to split the merged cluster up for each track pointing to it in a way parallel to how the splitting is done for multiple maxima in the island clustering. But I was space cadet and didn't read the island clustering code carefully enough the first time 😓

ruse-traveler:
Hi all, just wanted to check in on this: do the recent round of commits resolve the open issues? Or is there more to discuss? Thanks!

veprbl:
@ruse-traveler I'd like to spend some more time checking the actual change. Have you done any benchmarking on your end?

ruse-traveler:
Mostly just spot-checks to make sure that it runs and isn't producing stuff that's completely out-to-lunch. I can run some more detailed checks like looking at the E_{clust}/E_{par} in some DIS events.

veprbl (approved):
Ship it

veprbl (approved):

PR #1410, opened 2024-05-03
CalorimeterHitDigi: use EvaluatorSvc for corrMeanScale

veprbl:
This is a sequel to #1349, now for digitization.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -13,13 +13,13 @@
 
 #include "CalorimeterHitDigi.h"
 
-#include <algorithms/service.h>
 #include <DD4hep/Detector.h>
 #include <DD4hep/IDDescriptor.h>
 #include <DD4hep/Readout.h>
 #include <DD4hep/config.h>
 #include <DDSegmentation/BitFieldCoder.h>
 #include <Evaluator/DD4hepUnits.h>
+#include <algorithms/service.h>
 #include <edm4hep/CaloHitContributionCollection.h>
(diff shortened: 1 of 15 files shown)


src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -96,11 +98,25 @@ void CalorimeterHitDigi::init() {
...
     if (!m_cfg.fields.empty()) {
         for (auto & field : m_cfg.fields) {
-            id_inverse_mask |= id_desc.field(field)->mask();
+            id_inverse_mask |= id_spec.field(field)->mask();
         }
         debug("ID mask in {:s}: {:#064b}", m_cfg.readout, id_mask);
     }
     id_mask = ~id_inverse_mask;
+
+    std::function hit_to_map = [this](const edm4hep::SimCalorimeterHit &h) {
```
wdconinc, line 107: Optional, deferrable. Can this be reused from CalorimeterHitReco, somehow? It's essentially the same code. Templated lambda with concept `has_cellID`?

wdconinc (approved):
Not sure I like the computational complexity this adds for every hit, when at this point (and probably for some detectors never) this isn't used. For a corrMeanScale of 1, this changes a noop to a looped set of string lookups in the index, right?

src/algorithms/calorimetry/CalorimeterHitDigi.cc
veprbl: That is understood. What goes into the function is still in flux at this point, will likely add more fields or change how this works anyway.

ruse-traveler (approved):
Nice! 🙌

PR #1413, opened 2024-05-04
FarDetectorTrackerCluster.cc: ulong -> unsigned long

veprbl:
ulong from sys/types is not a standard C type

This fixes a compilation error:
```
src/algorithms/fardetectors/FarDetectorTrackerCluster.cc:113:26: error: use of undeclared identifier 'ulong'; did you mean 'long'?
      ROOT::VecOps::RVec<ulong> clusterList = {maxIndex};
                         ^~~~~
                         long
```

### DIFF
--- a/src/algorithms/fardetectors/FarDetectorTrackerCluster.cc
+++ b/src/algorithms/fardetectors/FarDetectorTrackerCluster.cc
@@ -110,7 +110,7 @@ namespace eicrecon {
 
       available[maxIndex] = 0;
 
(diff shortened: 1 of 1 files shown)


simonge (approved):
Was this being silently flagged in any tests?

veprbl:
No, actually this is from a private build on macOS. IIRC the GNU libc headers allow ulong and such by default, and sys/types gets included somewhere along the way.

PR #1414, opened 2024-05-04
cmake: use CONFIG mode for onnxruntime

veprbl:
This prevents EICrecon from picking up Findonnxruntime.cmake from Acts. The actual onnxruntime only ever provided the config, to my best knowledge.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,7 +222,7 @@ endif()
 # ONNX Runtime
 option(USE_ONNX "Compile with ONNX support" ON)
 if(${USE_ONNX})
-  find_package(onnxruntime)
+  find_package(onnxruntime CONFIG)
 endif()
 
 # Add CMake additional functionality:
--- a/cmake/jana_plugin.cmake
+++ b/cmake/jana_plugin.cmake
@@ -381,7 +381,7 @@ endmacro()
 macro(plugin_add_onnxruntime _name)
 
   if(NOT onnxruntime_FOUND)
-    find_package(onnxruntime)
+    find_package(onnxruntime CONFIG)
   endif()
 
   # Add libraries


wdconinc (approved):
Ha, how did you figure out this one?

veprbl:
> Ha, how did you figure out this one?

The easy way, of course. It fails my builds like so:
```
CMake Warning at CMakeLists.txt:205 (message):
  Acts has not been built with the requested C++ standard 20.
-- Found OnnxRuntime library at /lib/libonnxruntime.dylib
-- Found OnnxRuntime includes at /include (API version: 16)
CMake Error at /lib/cmake/Acts/Modules/Findonnxruntime.cmake:36 (string):
  string sub-command REPLACE requires at least four arguments.
Call Stack (most recent call first):
  CMakeLists.txt:225 (find_package)
CMake Warning (dev) at /share/cmake-3.27/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args`
  (OnnxRuntime) does not match the name of the calling package (onnxruntime).
  This can lead to problems in calling code that expects `find_package`
  result variables (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  /lib/cmake/Acts/Modules/Findonnxruntime.cmake:43 (find_package_handle_standard_args)
  CMakeLists.txt:225 (find_package)
```

wdconinc:
Should this be considered a bug elsewhere? Acts or spack?

veprbl:
> Should this be considered a bug elsewhere? Acts or spack?

It depends on your viewpoint. Perhaps, it's not very wise of Acts to ship a module that requires a `_acts_onnxruntime_version` variable to be set. But also, I don't expect CMake modules to work flawlessly in the first place.

PR #1417, opened 2024-05-05
Separate tracking and PID factories, use algorithms interface

veprbl:
The original approach in #829 was to add more functionality into a single factory. This pushes back on that by separating the two responsibilities into two algorithms. This should improve modularity. This also converts algorithms to the algorithms interface. The factory code is rewritten from scratch and licensed to LGPL3.

### DIFF
--- a/src/global/pid/pid.cc
+++ b/src/global/pid/pid.cc
@@ -18,15 +18,6 @@ extern "C" {
 
     using namespace eicrecon;
 
-    // configuration parameters ///////////////////////////////////////////////
-
-    // linking of reconstructed particles to PID objects
-    ParticlesWithPIDConfig link_cfg;
-    link_cfg.momentumRelativeTolerance = 100.0; /// Matching momentum effectively disabled
-    link_cfg.phiTolerance              = 0.1; /// Matching phi tolerance [rad]
-    link_cfg.etaTolerance              = 0.2; /// Matching eta tolerance
-
-
     // wiring between factories and data ///////////////////////////////////////
     // clang-format off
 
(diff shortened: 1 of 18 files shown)


src/global/tracking/tracking.cc
```diff
@@ -171,5 +173,35 @@ void InitPlugin(JApplication *app) {
...
+       .phiTolerance              = 0.1, /// Matching phi tolerance [rad]
+       .etaTolerance              = 0.2, /// Matching eta tolerance
+     };
+
+     app->Add(new JOmniFactoryGeneratorT<ParticlesWithTruthPID_factory>(
+             "ChargedParticlesWithAssociations",
+             {"MCParticles",                                    // edm4hep::MCParticle
+             "CentralCKFTracks",                                // edm4eic::Track
+             },
+             {"ReconstructedChargedWithoutPIDParticles",                  //
```
wdconinc, line 188: Is it clearer to refer to this as TruthPID then?
```suggestion
             {"ReconstructedChargedWithTruthPIDParticles",                  //
```

src/global/pid/pid.cc
```diff
@@ -18,48 +18,33 @@ extern "C" {
...
           app
           ));
 
     app->Add(new JOmniFactoryGeneratorT<ParticlesWithPID_factory>(
           "SeededChargedParticlesWithAssociations",
           {
-            "MCParticles",                      // edm4hep::MCParticle
-            "CentralCKFSeededTracks",           // edm4eic::Track
-            "DRICHMergedIrtCherenkovParticleID" // edm4eic::CherenkovParticleID
+            "ReconstructedSeededChargedWithoutPIDParticles", // edm4eic::ReconstructedParticle
```
wdconinc, line 41: This would also have to change, and make it clear we are transitioning from TruthPID to "real" PID.

wdconinc (dismissed):
Assuming this started with a straightforward cp from src/algorithms/pid/ParticlesWithPID.cc into src/algorithms/tracking/ParticlesWithTruthPID.cc, I didn't carefully look at that removed lines in the former and added lines in the latter.

src/global/tracking/tracking.cc
veprbl: The truth PID is to be removed when LUTs come. Users can lookup truth MCParticle via associations.

veprbl:
> Assuming this started with a straightforward cp from src/algorithms/pid/ParticlesWithPID.cc into src/algorithms/tracking/ParticlesWithTruthPID.cc, I didn't carefully look at that removed lines in the former and added lines in the latter.

Yes, the splitting was quite clean, I'm not sure why this had to be originally combined.

veprbl:
collect-docs is broken, but capybara report is clean now: https://veprbl.github.io/capybara-reports/c75e9e7160558ec6b929117b464291f5/

wdconinc (approved):
Capybara hath spoken and approves.

PR #1422, opened 2024-05-06
linux-eic-shell.yml: specify clang++ explicitly

veprbl:
```
/usr/bin/c++ --version
c++ (Debian 12.2.0-14) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```


wdconinc:
The current behavior here is explained in https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations. Later definitions apply to all earlier definitions that match without overwrites, and create a new entry if there is no match. Since we define a clang/clang++ combo earlier, simply using clang refers to this. The proposed changes don't hurt, though, and may be better for clarity.

wdconinc (approved):

PR #1432, opened 2024-05-08
pid_lut: support reading compressed files for hpdirc.lut.gz

veprbl:
This requires https://github.com/eic/epic/pull/724


veprbl:
Probably, not critical for 24.05.0

wdconinc (approved):

PR #1450, opened 2024-05-11
feat: set track relation for calorimeter and DRICH propagations

wdconinc:
The tracks propagated to the calorimeter surfaces and the DRICH aerogel and gas surfaces did not contain the relation to the track.

This PR adds the track relations to CalorimeterTrackProjections and the DRICH{Aerogel,Gas}Tracks collections.

Note the methodology here. We don't use edm4eic::Tracks to do the propagation, so this here relies on keeping the order in the edm4eic::Tracks and ActsExamples::Trajectories.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: no track relations in projections)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackPropagation.cc
+++ b/src/algorithms/tracking/TrackPropagation.cc
@@ -107,32 +107,46 @@ void TrackPropagation::init(const dd4hep::Detector* detector,
 
 
 void TrackPropagation::propagateToSurfaceList(
-          const std::tuple<const std::vector<const ActsExamples::Trajectories*>, const std::vector<const ActsExamples::ConstTrackContainer*>> input,
+          const std::tuple<const edm4eic::TrackCollection&, const std::vector<const ActsExamples::Trajectories*>, const std::vector<const ActsExamples::ConstTrackContainer*>> input,
           const std::tuple<edm4eic::TrackSegmentCollection*> output) const
 {
-    const auto [acts_trajectories, acts_tracks] = input;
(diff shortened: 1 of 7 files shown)


wdconinc:
Looks sensible...

Screenshot_20240511-105302

wdconinc:
Both e and pi singles are missing 4 of 100 entries, https://eicrecon.epic-eic.org/pr/1450/capybara/rec_e_1GeV_20GeV_craterlake/index.html#_CalorimeterTrackProjections_track

I could imagine this is simply projections that don't end up anywhere. The same 4 of 100 don't have points to begin with, https://eicrecon.epic-eic.org/pr/1450/capybara/rec_e_1GeV_20GeV_craterlake/index.html#CalorimeterTrackProjections

tylerkutz:
This looks good and the I agree the resulting numbers are reasonable (at least based on my own studies of the track propagations).

Just to confirm, this solution assumes the same ordering of the `edm4eic::TrackCollection` and the vector of `ActsExamples::Trajectories`s?  (As you mentioned in the reconstruction meeting).

wdconinc:
> Just to confirm, this solution assumes the same ordering of the `edm4eic::TrackCollection` and the vector of `ActsExamples::Trajectories`s?  (As you mentioned in the reconstruction meeting).

Yes, it assumes the thing I've told people not to assume. For now, it's valid; it won't be when we go to more complex running strategies. By then we will have phased this out.

tylerkutz (approved):

Issue #1318, opened 2024-03-04
InclusiveKinematicsJB should include neutrals too, not just ReconstructedChargedParticles


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? No
- Any special versions of ROOT or Geant4? No

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Look at reco.cc:
```cpp
    app->Add(new JOmniFactoryGeneratorT<InclusiveKinematicsJB_factory>(
        "InclusiveKinematicsJB",
        {
          "MCParticles",
          "ReconstructedChargedParticles",
          "ReconstructedChargedParticleAssociations"
        },
        {
          "InclusiveKinematicsJB"
        },
        app
    ));
```

### Expected Result: (what do you expect when you execute the steps above)
JB should use ReconstructedParticles and ReconstructedParticleAssociations, but that requires that those actually include the neutrals too.

### Actual Result: (what do you get when you execute the steps above)
JB ignores neutrals.

### Additional Context
There may be other inclusive kinematics algorithms that are similarly affected.

Cc: @tylerkutz @bschmookler @veprbl

PR #1453, opened 2024-05-14
Reconstructed inclusive kinematics algorithms now use centralized algorithms for scattered electron ID and hadronic final state calculation

tylerkutz:

All reconstructed inclusive kinematics algorithms now obtain:
- Scattered electron from `ScatteredElectronsTruth` algorithm
- Hadronic final state from `HadronicFinalState` algorithm

Previously, these steps were repeated as-needed in each individual inclusive kinematics algorithm.  Now, improvements/fixes made to `ScatteredElectronTruth` or `HadronicFinalState` will automatically be included in the inclusive kinematics algorithms.

This also fixes issue #1318 , as the `HadronicFinalState` algorithm uses all reconstructed particles, not just reconstructed charged particles.

- [ x] Bug fix (issue #1318 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, although the specific values of the reconstructed inclusive kinematics might be different (see below) the output collections are in the same format.

The electron reconstruction method should be identical, as the only difference is where the scattered electron is found.

The reconstruction methods using any information from the hadronic final state (JB, DA, [e]Sigma) will be different, as the hadronic final state now includes all reconstructed particles, not just charged particles.


simonge:
Any chance you also fancy swapping out MCParticles and using MCBeamElectrons and MCBeamProtons instead? This will stop the algorithms from independently needing to carry out the search, it could also be restricted to call once as it doesn't change between events, although I'm sure none of this is very expensive.

Ideally the information needed from the MCParticles would be contained in the metadata and used in the initialisation.

tylerkutz:
@simonge Great suggestion, I was actually not aware of these.  I will switch to using these instead.

veprbl:
I agree with @simonge's suggestion. If this is not implemented yet, can we, please, hold this off for another PR? This would ease the review process a bit.

veprbl:
My proposal is to wrap the 4 .cc files in src/algorithms/reco entirely in

```c++
#include <edm4eic/EDM4eicVersion.h>
#if EDM4EIC_VERSION_MAJOR >= 6
// Copyright (C) 202....

// contents of the file
#endif
```
and do some more careful guarding of includes and factories in reco.cc. The header files would end up unincluded as the result of this. This will disable kinematics factories for EDM4eic 5, but would allow to still test everything else. It would be easy to revert that later, once that's deprecated. @tylerkutz, what do you think?

veprbl:
Saving capybara report for the current state (c6fa0447df64af2f1fb25f5f67c7df3f6f4a76aa) to
https://veprbl.github.io/capybara-reports/a664ff9eb9cd58755d862771ff2bab0f/

veprbl:
As discussed in the meeting today, let's merge this and limit changes to refactoring and switching from ReconstructedChargedParticles to ReconstructedParticles. Other changes/reverts can follow.

PR #1454, opened 2024-05-14
Neutron energy scale

sebouh137:
- [ ] Bug fix (issue #__)
- [x] New feature (issue #1455)
- [ ] Documentation update
- [ ] Other: __  

no
yes.  Changes default behavior of the FarForwardNeutronReconstruction algorithm to include energy corrections.


src/algorithms/reco/FarForwardNeutronReconstruction.cc
```diff
@@ -41,9 +46,11 @@ namespace eicrecon {
       }
       if (Etot>0){
           auto rec_part = out_neutrons->create();
+          double corr=calc_corr(Etot);
+          Etot=Etot/(1+corr);
```
veprbl, line 50: Instead of adding ad-hoc functionality to what is essentially ReconstructedNeutralParticles factory, can we add evaluator expression support to ClusterRecoCoG sampFrac parameter, like it's done in #1410.

veprbl: That would need to make sampFrac depend not only on cellID, but also on original cluster energy.

sebouh137: A single neutron may produce multiple topological clusters, so the correction for single-neutron events would be different depending on the total shower energy, rather than having separate sampling fractions for each cluster.  Further, the sampling fractions we are using are based on an EM response (since the ZDC Hcal will measure not only hadrons but also mid-to-high energy photons as well).  This correction compensates for the different between EM and hadronic responses.  So, we'd prefer to keep this as we have implemented it in this pull request.

src/algorithms/reco/FarForwardNeutronReconstruction.cc
```diff
@@ -7,20 +7,25 @@
...
+#include <vector>
 
 /**
- Creates a "neutron candidate" Reconstructed Particle consisting of all clusters in a given ClusterCollection.
- Its energy is the sum of the energies of the constituent clusters, and its direction is the direction from the
-origin to the position of the most energetic cluster
+ Creates a "neutron candidate" Reconstructed Particle consisting of all clusters in a
+ given ClusterCollection.  Its energy is the sum of the energies of the constituent clusters
+ times a correction factor, and its direction is the direction from the origin to the position
+ of the most energetic cluster.  The correction factor is given by 1/(1+c[0]+c[1]/sqrt(E)+c[2]/E),
```
veprbl, line 16: Is there a reference for this form of correction? If yes, can we add a link to this comment?

sebouh137: Hi @veprbl ,
I decided on this form of a correction as an "empirical fit" to the distributions of the discrepancy between the uncorrected reconstructed energies and the truth energies in the MC simulations in some studies I have done.  I did not derive this from first principles, nor have I seen this formula somewhere else that I can refer back to.  How I got this was I noticed that at lower energies, the discrepancies were larger and at higher energies they leveled off, and were approximately flat.  I first tried fitting to a+c/E, but this clearly didn't fit very well.  Then I tried a+b/sqrt(E)+c/E, and this worked very well.  
Best regards,
Sebouh

src/algorithms/reco/FarForwardNeutronReconstruction.cc
```diff
@@ -7,20 +7,27 @@
...
+ where c is the coefficients and E is the uncorrected energy in GeV.  This form was chosen
+ empirically based on the discrepancies in single-neutron MC simulations between the uncorrected
+ reconstructed energies and the truth energies of the neutrons.  
  */
 
 namespace eicrecon {
 
     void FarForwardNeutronReconstruction::init() {  }
```
veprbl, line 24: Would you mind adding a check that coeffs has at least 3 elements?

src/algorithms/reco/FarForwardNeutronReconstruction.cc
veprbl: Understood. Thanks for updating the comment.

src/algorithms/reco/FarForwardNeutronReconstruction.cc
sebouh137: I agree with this; should this be in the init function or when processing events?   Should it shut down the process or just give a warning?  Do you have an example of what other algorithms do when checking for valid/invalid configuration values?
Thanks,
Sebouh

sebouh137: ok, found that other algorithms throw a JException with a description of what went wrong with improper configuration.  I have edited the algorithm accordingly

sebouh137:
@veprbl ,
IWYU is trying to get me to change one of the include statements to use a header that does not exist.  
In src/global/reco/reco.cc, it wants me to change #include <JANA/JApplication.h> to #include <JANA/JApplicationFwd.h>, however, doing so causes the compiler to complain that this header does not exist.  What's going wrong here, and can we override this somehow?  
Best regards,
Sebouh

veprbl:
@sebouh137 Good observation. Let's not use the Fwd header just yet. It breaks compilation in older containers, it seems like.

src/algorithms/reco/FarForwardNeutronReconstruction.cc
```diff
@@ -1,26 +1,37 @@
...
+ reconstructed energies and the truth energies of the neutrons.
  */
 
 namespace eicrecon {
 
-    void FarForwardNeutronReconstruction::init() {  }
-
+    void FarForwardNeutronReconstruction::init() {
+      if (m_cfg.scale_corr_coeff.size() < 3) {
+        throw JException("Invalid configuration.  m_cfg.scale_corr_coeff should have at least 3 parameters");
```
veprbl, line 26: IIRC, for now we still need to print the message, also we can't use JANA anything in algorithms.
```suggestion
        error("Invalid configuration.  m_cfg.scale_corr_coeff should have at least 3 parameters");
        throw std::runtime_error("Invalid configuration.  m_cfg.scale_corr_coeff should have at least 3 parameters");
```

sebouh137:
@veprbl ,
the only checks that are failing are the get_docs_from_main and eicweb/eic_container.  Are there things that I need to fix on my end to make this branch pass these checks?  If not, let me know anything else I need to do in order to get this pull request approved.
Thanks,
Sebouh

veprbl:
It would be nice to have that fixed. I'm trying to fix it via #1459

src/global/reco/reco.cc
```diff
@@ -3,14 +3,15 @@
...
+
 #include <JANA/JApplication.h>
 #include <edm4eic/Cluster.h>
 #include <edm4eic/EDM4eicVersion.h>
 #include <edm4eic/MCRecoClusterParticleAssociation.h>
 #include <edm4eic/MCRecoParticleAssociation.h>
 #include <edm4eic/ReconstructedParticle.h>
 #include <edm4hep/MCParticle.h>
-#include <algorithm>
+//#include <algorithm>
```
ruse-traveler, line 14: Is there a reason to comment this out rather than just delete this include?

ruse-traveler (commented):
One tiny comment, but overall looks good!

src/global/reco/reco.cc
veprbl:
```suggestion
```

veprbl:
`HcalFarForwardZDCNeutronCandidates` doesn't show up in capybara. Is it produced at all?

ruse-traveler:
> `HcalFarForwardZDCNeutronCandidates` doesn't show up in capybara. Is it produced at all?

It should be in `ReconstructedFarForwardZDCNeutrons`... Which no longer seems to be listed in `JEventProcessorPODIO.cc`?

veprbl:
https://github.com/eic/EICrecon/pull/1404/files#diff-d750a62ff8386199706241b174c0b6466ec5714a43057a4fd591df3dd77de239R290
vs
https://github.com/eic/EICrecon/pull/1404/files#diff-08c369ce77f61e0738a1b6bea175ebb10a4df985a97ced435296883bcca0992cR279

sebouh137:
> > `HcalFarForwardZDCNeutronCandidates` doesn't show up in capybara. Is it produced at all?
> 
> It should be in `ReconstructedFarForwardZDCNeutrons`... Which no longer seems to be listed in `JEventProcessorPODIO.cc`?

I must have forgotten to change the name of this bank in JEventProcessorPODIO.cc during the review of the earlier pull request.  Just fixed this now.

ruse-traveler:
> I must have forgotten to change the name of this bank in JEventProcessorPODIO.cc during the review of the earlier pull request. Just fixed this now.

Thanks! Let's wait for the capybara, and then if that looks fine, I think this is good to go... Are there any concerns you have @veprbl?

veprbl (approved):
Looks fine to me

PR #1458, opened 2024-05-16
linux-eic-shell.yml: use github.head_ref

veprbl:
The documentation https://docs.github.com/en/actions/learn-github-actions/contexts#github-context says that it only works for pull_request*, but there is some hope that it can inherit from merge_group payload too https://docs.github.com/en/webhooks/webhook-events-and-payloads#merge_group . Then that would fix issues with missing capybara references on main: https://github.com/eic/EICrecon/issues/1264

Fixes: #1264

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -102,11 +102,11 @@ jobs:
       uses: actions/cache@v4
       with:
         path: .ccache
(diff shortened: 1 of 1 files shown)


veprbl:
@wdconinc Should we try merging this? I hope to get main to green again.

wdconinc (approved):
Let's try it.

PR #1459, opened 2024-05-17
linux-eic-shell.yml: use github.event.merge_group.base_ref

veprbl:
`github.base_ref` doesn't work for merge_group: https://github.com/eic/EICrecon/actions/runs/9120955781/job/25084018811#step:13:21

It's put after github.base_ref, so that we can later test if this can be reverted by re-running with enabled debug logging.


veprbl:
@wdconinc I have no way to debug this in a private fork (unless I make a private org, so that I can enable merge queues). So here is the second attempt.

wdconinc (approved):
Yeah, this is the kind of CI pipeline testing that is often hard to do...

Issue #1336, opened 2024-03-19
Introduce service which provides access to podio metadata tree


simonge:
Currently there isn't a way to access the contents of the "metadata" tree in EICrecon. I would like to see a service which provides access to the members so algorithms can use this information directly.

Currently work arounds are needed which require this information to hard coded/passed as a configuration or inferred from the data event by event. This can cause errors from the user where the configuration may not be carried out correctly.

Accessing the beamline energy settings is the focus of my request but there will other applications for this.

nathanwbrei:
How urgently is this needed?

simonge:
In the order of a few weeks to a month would be ideal.

DraTeots:
> Accessing the beamline energy settings is the focus of my request but there will other applications for this.

This is important for my current tasks also

PR #1467, opened 2024-05-21
Add low Q2 NN inference

simonge:
Adds the inference step, reconstructing the initial momentum of the low-Q2 electrons from their tracks. This relies on https://github.com/eic/epic-data/pull/17 to add the inference model to the expected default location (so far tested by manually copying the file there).

This will be followed by the final PR which finally injects the low-Q2 electron tracks into the main tracking data flow so they get treated equally by the following algorithms.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


simonge:
The build-docs is complaining that none of the capybara files are available. The reconstruction CI steps cannot find the previous versions for capybara to compare against. I can't see anything in this PR which could have influenced this?

veprbl:
I've restarted CI on main (looks like it was cancelled?), that should fix it.

src/factories/fardetectors/FarDetectorMLReconstruction_factory.h
```diff
@@ -0,0 +1,57 @@
+// Created by Simon Gardner
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
```
veprbl, line 3:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2024, Simon Gardner
```

simonge:
The whole workflow seems to be a bit unstable at the moment. I might be wrong in some of this but, seemingly random jobs have been skipped, sometimes taken much longer than usual or been unable to find artefacts from either main or which have just been produced.

.github/workflows/linux-eic-shell.yml
```diff
@@ -852,32 +859,32 @@ jobs:
           export DETECTOR_CONFIG=${DETECTOR}_${{ matrix.detector_config }}
           export LD_LIBRARY_PATH=$PWD/install/lib:$LD_LIBRARY_PATH
           export JANA_PLUGIN_PATH=$PWD/install/lib/EICrecon/plugins:/usr/local/plugins
-          $PWD/install/bin/eicrecon $JANA_OPTIONS -Ppodio:output_file=rec_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4eic.root sim_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root -Pacts:WriteObj=true -Pacts:WritePly=true -Pplugins=janadot,janatop $(<${{ github.workspace }}/.github/janadot.groups)
+          $PWD/install/bin/eicrecon $JANA_OPTIONS -Ppodio:output_file=rec_dis_${{matrix.beamelectron}}x${{matrix.beamproton}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4eic.root sim_dis_${{matrix.beamelectron}}x${{matrix.beamproton}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root -PLOWQ2:TaggerTrackerTrajectories:electron_beamE=${{matrix.beamelectron}} -Pacts:WriteObj=true -Pacts:WritePly=true -Pplugins=janadot,janatop $(<${{ github.workspace }}/.github/janadot.groups)
```
veprbl, line 862: What is the blocker for propagating beam energy from afterburner through DD4hep to EDM4hep metadata?

simonge: The afterburner PR is still open and has been for a while, my implementation broke a couple of ways it could be configured previously but I wanted to know if anyone was actually using them or if they could be swapped to a different approach. 

If the hepmc file has the information in its RunAttributes dd4hep was updated so that it passes this information through the simulation. I tested this but it might have needed some additional flag, I'll have to go back and check.

The main hold up will be a metadata service in EICrecon, requested in #1336 which I believe is on the todo list of @nathanwbrei

veprbl: I don't think we can hand this off like this to production. What about the constants defined in DD4hep https://github.com/eic/epic/blob/7c7f26f34371e70754ad582487b32f79acb0ac4c/compact/fields/beamline_10x100.xml#L6 ?

simonge: What would be the problem with having this in production?

We would then need to separately set the xml for each eicrecon configuration in the workflows which I believe isn't currently done as the reconstruction behaves the same, although users are probably more likely to keep the xml identical between simulation and reconstruction in their own scripts. Both solutions are stop gaps for having the metadata present in the file but having a command line config parameter with the ability to override the meta data makes more sense to me longer term.

In a final solution what would the precedence of the configuration be?

1. Command line config
2. Metadata
3. Detector geometry

veprbl: The issue is that this implies that production and users have to populate every such config option.

I understand your point about xml, my assumption was that this might be already incorporated in the production. Looks like I'm likely wrong.

What do you think about re-using logic from https://github.com/eic/EICrecon/blob/fa7377c9261724b849e9e6f54a581ebe1fe42522/src/algorithms/reco/InclusiveKinematicseSigma.cc#L42-L53 ?

wdconinc: This is also used like that in the far forward region: https://github.com/eic/EICrecon/blob/fa7377c9261724b849e9e6f54a581ebe1fe42522/src/algorithms/fardetectors/MatrixTransferStatic.cc#L57-L60

Adding a configuration option and expecting someone downstream to provide the right value is not workable in the short term (i.e. 24.06).

simonge: While it has some downsides I will concede the far-forward solution is currently best from the user point of view.

src/factories/fardetectors/FarDetectorMLReconstruction_factory.h
```diff
@@ -0,0 +1,57 @@
...
+class FarDetectorMLReconstruction_factory :
+        public JOmniFactory<FarDetectorMLReconstruction_factory,FarDetectorMLReconstructionConfig> {
+
+public:
+    using AlgoT = eicrecon::FarDetectorMLReconstruction;
+private:
+    std::unique_ptr<AlgoT> m_algo;
+
+    PodioInput<edm4eic::TrackParameters>  m_trackparam_input     {this};
+    PodioInput<edm4hep::MCParticle>       m_beamelectrons_intput {this};
```
veprbl, line 29:
```suggestion
    PodioInput<edm4hep::MCParticle>       m_beamelectrons_input {this};
```

src/factories/fardetectors/FarDetectorMLReconstruction_factory.h
```diff
@@ -0,0 +1,57 @@
...
+        m_algo->level((algorithms::LogLevel)logger()->level());
+        m_algo->applyConfig(config());
+        m_algo->init();
+    }
+
+    void ChangeRun(int64_t run_number) {
+    }
+
+    void Process(int64_t run_number, uint64_t event_number) {
+        m_algo->process({m_trackparam_input(),m_beamelectrons_intput()}, {m_trajectory_output().get(), m_trackparam_output().get(), m_track_output().get()});
```
veprbl, line 53:
```suggestion
        m_algo->process({m_trackparam_input(),m_beamelectrons_input()}, {m_trajectory_output().get(), m_trackparam_output().get(), m_track_output().get()});
```

src/factories/fardetectors/FarDetectorMLReconstruction_factory.h
```diff
@@ -0,0 +1,57 @@
...
+    PodioInput<edm4hep::MCParticle>       m_beamelectrons_intput {this};
+    PodioOutput<edm4eic::Trajectory>      m_trajectory_output    {this};
+    PodioOutput<edm4eic::TrackParameters> m_trackparam_output    {this};
+    PodioOutput<edm4eic::Track>           m_track_output         {this};
+
+
+    ParameterRef<std::string> modelPath       {this, "modelPath",       config().modelPath       };
+    ParameterRef<std::string> methodName      {this, "methodName",      config().methodName      };
+
+    ParameterRef<float> electron_beamE  {this, "electron_beamE",  config().electronBeamE };
```
veprbl, line 38:
```suggestion
    ParameterRef<std::string> m_modelPath       {this, "modelPath",       config().modelPath       };
    ParameterRef<std::string> m_methodName      {this, "methodName",      config().methodName      };
```

src/algorithms/fardetectors/FarDetectorMLReconstructionConfig.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <DD4hep/DD4hepUnits.h>
+
+namespace eicrecon {
+  struct FarDetectorMLReconstructionConfig {
+
+    std::string modelPath{"calibrations/tmva/LowQ2_DNN_CPU.weights.xml"};
+    std::string methodName{"DNN_CPU"};
+
+    float electronBeamE{10*dd4hep::GeV};
```
veprbl, line 14:
```suggestion
```

src/algorithms/fardetectors/FarDetectorMLReconstructionConfig.h
```diff
@@ -0,0 +1,17 @@
...
+
+#pragma once
+
+#include <DD4hep/DD4hepUnits.h>
+
+namespace eicrecon {
+  struct FarDetectorMLReconstructionConfig {
+
+    std::string modelPath{"calibrations/tmva/LowQ2_DNN_CPU.weights.xml"};
+    std::string methodName{"DNN_CPU"};
```
veprbl, line 12: Move these values to LOWQ2.cc, please

veprbl (approved):

Issue #1289, opened 2024-02-09
Excessive cluster splitting at high pseudorapidity in EcalBarrelScFi


veprbl:
![EcalBarrelScFi](https://github.com/eic/EICrecon/assets/245573/e5f59432-95ee-44dc-ad09-a47149a5e426)

Reproducer using single 1GeV electrons from the 23.12 campaign https://gist.github.com/veprbl/527906251dc8f86c45c99ca502bf5528

cc @Chao1009

veprbl:
@AkshayaVijay Could you, please, look into this? I think, this is quite important to get resolved.

cc @mariakzurek

AkshayaVijay:
> @AkshayaVijay Could you, please, look into this? I think, this is quite important to get resolved.
> 
> cc @mariakzurek

Sure.

wdconinc:
@AkshayaVijay Can you file a pull request with the changes to the clustering parameters that we think are now better than what is there in EICrecon? You can link your slides in that PR for documentation purposes. Even if we don't have the full cluster-to-'relevant'-MCParticle association working yet, it would be helpful to include this in the June campaign.

Issue #1475, opened 2024-05-30
Proposal for improved calorimeter cluster associations


wdconinc:
**Is your feature request related to a problem? Please describe.**
Right now, the cluster associations are determined using the following strategy:
https://github.com/eic/EICrecon/blob/fa7377c9261724b849e9e6f54a581ebe1fe42522/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc#L78-L81
This leaves something to be desired, in particular when studying undesired cluster splitting.
- The algorithm currently in use by definition always returns a single assocation (weight of 1) between the cluster and the MCParticle that caused its highest energy deposition. 
- Most of the time (e.g. https://eicrecon.epic-eic.org/pr/1422/capybara/rec_dis_10x100_minQ2=1000_craterlake/index.html#EcalBarrelScFiClusterAssociations), the MCParticles are not available because they are shower secondaries created inside the calorimeter, not primary or relevant secondaries.

In particular, these issues are hampering studies like: how many clusters are actually composed of two particles that were incorrectly lumped together in one cluster?

**Describe the solution you'd like**
We (well, @AkshayaVijay) is working on an analysis where we go from clusters to hits to hit contributions to MCParticles, and then recursively follow the MCParticle parent relations until an MCParticle has a vertex inside the tracking volume (i.e. up to a maximum radius of the inner radius of the calorimeter). When combined with the energy deposition in the cluster, this could result in a better association of the 'real' MCParticle to be associated with the cluster. 

**Describe alternatives you've considered**
Another approach could be DD4hep plugin that keeps tracks of the parents, or rewrites some of the particle decay graph for particles that are created outside the tracking volume so a shower would more easily be able to traced back to its parents.

wdconinc:
RFC @ruse-traveler @veprbl

veprbl:
This is implemented in #1396, where we didn't see a change.

wdconinc:
No, this is different. The changes in #1396 look at ALL contributors (good), and the MCParticle of the highest contribution. This feature request is to allow for potentially different MCParticles per cluster. In particular #1396 still has weight equal to 1:
https://github.com/eic/EICrecon/blob/080d9ae7586d32766c0b222165c9b83f5850ecd7/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc#L367

What I would like as an outcome here is:
```
MCParticles:
0. E = 10 GeV, PDG = 22
[...]

PR #1478, opened 2024-05-31
Update the localDistXZ paramter in BEMC.cc

AkshayaVijay:
The clustering parameter, localDistXZ  is changed from {40 mm, 40mm} to {80mm, 80mm}. 

Minimizes the cluster splitting at the pseudorapidities corresponding to forward and backward region. The results can be seen here:
https://docs.google.com/presentation/d/1DympCgYqOKZBqyOqESEXCtlv073pGnkf266O1KRoba8/edit?usp=sharing

- [x] Bug fix (issue #1289 __)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/detectors/BEMC/BEMC.cc
+++ b/src/detectors/BEMC/BEMC.cc
@@ -79,7 +79,7 @@ extern "C" {
           "EcalBarrelScFiProtoClusters", {"EcalBarrelScFiRecHits"}, {"EcalBarrelScFiProtoClusters"},
           {
             .sectorDist = 50. * dd4hep::mm,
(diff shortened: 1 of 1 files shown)


veprbl:
eicrecon_issue1289
Works for me.

veprbl (approved):

wdconinc (approved):
At some point in the future (after https://github.com/eic/EICrecon/issues/1475 is address to allow detection of clusters that should be split), we can re-evaluate at what value we get too much splitting. For now, this is good.

PR #1483, opened 2024-06-03
fix: convert to algorithms logger, no m_log in each algorithm class

wdconinc:
This PR moves a lot of algorithms and factories over to using the internal logger functionality that comes with the algorithms interface. It means that individual algorithms don't need to initialize their own `m_log`. It also means that these algorithms are now properly conformant to the algorithms interface, i.e. `init()` and `process(input, output) const`.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/digi/SiliconTrackerDigi.cc
+++ b/src/algorithms/digi/SiliconTrackerDigi.cc
@@ -20,10 +20,7 @@
 
 namespace eicrecon {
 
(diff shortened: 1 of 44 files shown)


veprbl (approved):
diff LGTM

veprbl (approved):

PR #1487, opened 2024-06-04
feat: particleSvc to distribute mass by PDG

wdconinc:
Instead of storing the mass of the proton etc in various classes, this PR adds an algorithms::ParticleSvc that distributes the correct masses and charge etc by PDG number.

Note: I was playing around with using the algorithms::ParticleSvc as a thin interface-only, and let a JService read a data file and provide the actual service. But I gave up since it seemed pointless.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/algorithms/reco/InclusiveKinematicsSigma.cc
```diff
@@ -88,6 +82,7 @@ namespace eicrecon {
     auto sigma_tot = sigma_e + sigma_h;
 
     // Calculate kinematic variables
+    const auto m_proton = m_particleSvc.particle(2122).mass;
```
wdconinc, line 85:
```suggestion
    const auto m_proton = m_particleSvc.particle(2212).mass;
```

veprbl (commented):
This could use some extra `static` variable declarations, but already LGTM.

src/algorithms/interfaces/ParticleSvc.h
```diff
@@ -0,0 +1,51 @@
...
+
+  virtual Particle& particle(int pdg) const {
+    if (m_particleMap->count(pdg) == 0) {
+      return m_particleMap->at(0);
+    }
+    return m_particleMap->at(pdg);
+  };
+
+protected:
+  std::shared_ptr<ParticleMap> m_particleMap{nullptr};
```
wdconinc, line 46: @veprbl You want this static?

```suggestion
  static std::shared_ptr<ParticleMap> m_particleMap{nullptr};
```

I think that would then call for not setting this on constructor, and probably makes it harder to set based on external input.

src/algorithms/pid/IrtCherenkovParticleID.cc
```diff
@@ -111,11 +111,9 @@ void IrtCherenkovParticleID::init(
   }
 
   // get PDG info for the particles we want to identify in PID
-  // FIXME: cannot use `TDatabasePDG` since it is not thread safe; until we
-  // have a proper PDG database service, we hard-code the masses in Tools.h
   m_log->debug("List of particles for PID:");
   for(auto pdg : m_cfg.pdgList) {
-    auto mass = Tools::GetPDGMass(pdg);
+    auto mass = m_particleSvc.particle(pdg).mass;
```
veprbl, line 116:
```suggestion
    static const auto mass = m_particleSvc.particle(pdg).mass;
```

src/algorithms/reco/InclusiveKinematicsJB.cc
```diff
@@ -75,6 +69,7 @@ namespace eicrecon {
     }
 
     // Calculate kinematic variables
+    const auto m_proton = m_particleSvc.particle(2212).mass;
```
veprbl, line 72:
```suggestion
    static const auto m_proton = m_particleSvc.particle(2212).mass;
```

src/algorithms/reco/InclusiveKinematicsSigma.cc
```diff
@@ -88,6 +82,7 @@ namespace eicrecon {
     auto sigma_tot = sigma_e + sigma_h;
 
     // Calculate kinematic variables
+    const auto m_proton = m_particleSvc.particle(2212).mass;
```
veprbl, line 85:
```suggestion
    static const auto m_proton = m_particleSvc.particle(2212).mass;
```

src/algorithms/reco/InclusiveKinematicsTruth.cc
```diff
@@ -50,7 +42,7 @@ namespace eicrecon {
     }
     const auto ei_p = ei_coll[0].getMomentum();
     const auto ei_p_mag = edm4hep::utils::magnitude(ei_p);
-    const auto ei_mass = m_electron;
+    const auto ei_mass = m_particleSvc.particle(11).mass;
```
veprbl, line 45:
```suggestion
    static const auto ei_mass = m_particleSvc.particle(11).mass;
```

src/algorithms/reco/InclusiveKinematicsTruth.cc
```diff
@@ -61,7 +53,7 @@ namespace eicrecon {
     }
     const auto pi_p = pi_coll[0].getMomentum();
     const auto pi_p_mag = edm4hep::utils::magnitude(pi_p);
-    const auto pi_mass = pi_coll[0].getPDG() == 2212 ? m_proton : m_neutron;
+    const auto pi_mass = m_particleSvc.particle(pi_coll[0].getPDG()).mass;
```
veprbl, line 56:
```suggestion
    static const auto pi_mass = m_particleSvc.particle(pi_coll[0].getPDG()).mass;
```

src/algorithms/reco/InclusiveKinematicsTruth.cc
```diff
@@ -76,7 +68,7 @@ namespace eicrecon {
     }
     const auto ef_p = ef_coll[0].getMomentum();
     const auto ef_p_mag = edm4hep::utils::magnitude(ef_p);
-    const auto ef_mass = m_electron;
+    const auto ef_mass = m_particleSvc.particle(11).mass;
```
veprbl, line 71:
```suggestion
    static const auto ef_mass = m_particleSvc.particle(11).mass;
```

veprbl (commented):
I meant like that.

src/algorithms/interfaces/ParticleSvc.h
veprbl: Not this.

src/algorithms/pid/IrtCherenkovParticleID.cc
wdconinc: `static` here would only get mass for the first entry.

src/algorithms/reco/InclusiveKinematicsTruth.cc
wdconinc: `static` would set `pi_mass` to the same value, regardless of the PDG value. The initial nucleon can be proton or neutron, with different mass, and this can change from event to event in e.g. ion beams.

Note: `pi` stands for `p` of the *initial* nucleon, so this is not the pion mass.

src/algorithms/reco/InclusiveKinematicseSigma.cc
```diff
@@ -96,6 +90,7 @@ namespace eicrecon {
     const auto Q2_sig = (pt_e*pt_e) / (1. - y_sig);
     const auto x_sig = Q2_sig / (4.*ei.energy()*pi.energy()*y_sig);
 
+    const auto m_proton = m_particleSvc.particle(2212).mass;
```
wdconinc, line 93:
```suggestion
    static const auto m_proton = m_particleSvc.particle(2212).mass;
```

src/algorithms/reco/InclusiveKinematicsSigma.cc
wdconinc: This is arguably too narrow, since it assumes scattering off the proton.

src/algorithms/reco/InclusiveKinematicsElectron.cc
```diff
@@ -128,6 +122,7 @@ namespace eicrecon {
     }
 
     // DIS kinematics calculations
+    const auto m_proton = m_particleSvc.particle(2212).mass;
```
wdconinc, line 125:
```suggestion
    static const auto m_proton = m_particleSvc.particle(2212).mass;
```

src/algorithms/reco/InclusiveKinematicsDA.cc
```diff
@@ -87,6 +81,7 @@ namespace eicrecon {
     }
 
     // Calculate kinematic variables
+    const auto m_proton = m_particleSvc.particle(2212).mass;
```
wdconinc, line 84:
```suggestion
    static const auto m_proton = m_particleSvc.particle(2212).mass;
```

src/algorithms/reco/InclusiveKinematicsTruth.cc
veprbl: Yeah, only for the edge case when input has several different beam configurations. But yeah, good thinking, I haven't even considered.

src/algorithms/pid/IrtCherenkovParticleID.cc
veprbl: Good catch

veprbl (approved):

PR #1491, opened 2024-06-04
Inject Low-Q2 electrons into tracking

simonge:
Injects the LowQ2 tracks into the tracking workflow so they can finally be used like any other particle.

This might not be the best place to add the information from the events in the longer term as there will may be some extra looping over calorimeter and pid hits which aren't possible. At the moment though there isn't another easy place to add them while there is a mix of ReconstructedChargedParticles and ReconstructedParticles used in different places.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Low-Q2 electrons should show up in ReconstructedParticle collections and easily identifiable in the InclusiveKinematics collections

### DIFF
--- a/src/global/tracking/tracking.cc
+++ b/src/global/tracking/tracking.cc
@@ -251,6 +251,14 @@ void InitPlugin(JApplication *app) {
             app
             ));
 
+     // Add central and other tracks
+    app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::Track>>(
+            "CombinedTracks",
+            {"CentralCKFTracks", "TaggerTrackerTracks"},
+            {"CombinedTracks"},
+            app
+            ));
+
      // linking of reconstructed particles to PID objects
      TracksToParticlesConfig link_cfg {
        .momentumRelativeTolerance = 100.0, /// Matching momentum effectively disabled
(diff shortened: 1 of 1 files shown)


simonge:
Is there any chance this could be backported so the electrons would be included usefully in this months campaign?

src/global/tracking/tracking.cc
```diff
@@ -3,6 +3,7 @@
 
 #include <DD4hep/Detector.h>
 #include <JANA/JApplication.h>
+#include <edm4eic/Track.h>
```
wdconinc, line 6: Why did this stop suggesting `*Collection.h` as include?

wdconinc: https://github.com/eic/EICrecon/blob/23e24712e949a752faeb1baecae95fe07541f7d3/.github/iwyu.imp#L26

simonge: Is this something that needs to be addressed here?

wdconinc:
```suggestion
#include <edm4eic/TrackCollection.h>
```

src/global/tracking/tracking.cc
```diff
@@ -3,6 +3,7 @@
...
 #include <edm4eic/TrackerHit.h>
```
wdconinc, line 7:
```suggestion
#include <edm4eic/TrackerHitCollection.h>
```

veprbl (approved):

epic-capybara:
Successfully created backport PR for `v1.14`:
- #1501

simonge:
Fantastic, thanks!

PR #1493, opened 2024-06-04
fixed LFHCAL adjacency matrix

steinber:

Fix to LFHCAL clustering, avoiding sign error on local coordinates

- [X ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ X] Documentation has been added / updated
- [ X] Changes have been communicated to collaborators

Nothing breaks and no changes needed
Yes - it no longer splits clusters

### DIFF
--- a/src/detectors/FHCAL/FHCAL.cc
+++ b/src/detectors/FHCAL/FHCAL.cc
@@ -186,7 +186,7 @@ extern "C" {
         // Magic constants:
         //  54 - number of modules in a row/column
         //  2  - number of towers in a module
-        // sign for towerx and towery now *negative*, needed to restore linearity with global X and Y (PAS)
+        // sign for towerx and towery are *negative* to maintain linearity with global X and Y
         std::string cellIdx_1  = "(54*2-moduleIDx_1*2-towerx_1)";
         std::string cellIdx_2  = "(54*2-moduleIDx_2*2-towerx_2)";
         std::string cellIdy_1  = "(54*2-moduleIDy_1*2-towery_1)";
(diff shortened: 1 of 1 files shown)


src/detectors/FHCAL/FHCAL.cc
```diff
@@ -186,10 +186,11 @@ extern "C" {
         // Magic constants:
         //  54 - number of modules in a row/column
         //  2  - number of towers in a module
-        std::string cellIdx_1  = "(54*2-moduleIDx_1*2+towerx_1)";
-        std::string cellIdx_2  = "(54*2-moduleIDx_2*2+towerx_2)";
-        std::string cellIdy_1  = "(54*2-moduleIDy_1*2+towery_1)";
-        std::string cellIdy_2  = "(54*2-moduleIDy_2*2+towery_2)";
+        // sign for towerx and towery now *negative*, needed to restore linearity with global X and Y (PAS)
```
veprbl, line 189:
```suggestion
```
No need to document whole history in the code. git does that for us.

steinber: people don't always read the git comments and this may be something that may show up elsewhere, so I don't see a downside in keeping it for now.  However, Wouter did object to me adding my initials, and I'm happy to remove that.

veprbl: Okay.

veprbl (approved):
LGTM

src/detectors/FHCAL/FHCAL.cc
wdconinc: `PAS` is also just an obscure way to refer to people. The only reason I would know who it refers to is here and now because of obvious context.

epic-capybara:
Successfully created backport PR for `v1.14`:
- #1495

PR #1503, opened 2024-06-10
Make TaggerTracker and/or central tracks optional

simonge:
This should restore the ability to write out tracks from the central detector while using an xml configuration which doesn't contain the far-backward detectors (and vise versa)

### DIFF
--- a/src/global/tracking/tracking.cc
+++ b/src/global/tracking/tracking.cc
@@ -253,12 +253,12 @@ void InitPlugin(JApplication *app) {
             ));
 
 
-    
+
     std::vector<std::string> input_track_collections;
     //Check size of input_collections to determine if CentralCKFTracks should be added to the input_track_collections
     if (input_collections.size() > 0) {
         input_track_collections.push_back("CentralCKFTracks");
-    }    
+    }
     //Check if the TaggerTracker readout is present in the current configuration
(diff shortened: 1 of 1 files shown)


veprbl (approved):
This recovers tracks for the epic_tracking_only configuration

simonge:
Where do we stand with having optional collections, at least for the very simplest scenario where all input collections are optional?

veprbl:
> Where do we stand with having optional collections, at least for the very simplest scenario where all input collections are optional?

All input collections are currently "optional", in that the factories just disable themselves and don't produce any output.

nathanwbrei:
@simonge It's present in the upstreamed version of OmniFactory that lives under `JANA/Omni/JOmniFactory`. Both `PodioInputs` and `VariationalPodioInputs` accept an `is_optional` flag. If you set this flag to true, the collection pointers will be allowed to be null, so you have to check for that. I haven't updated EICrecon to use the upstreamed version of OmniFactory anywhere yet, but you could experiment with it if you like, and report back any problems you find. 

Note that if you use the upstreamed JOmniFactory, you'll probably have to use the upstreamed JOmniFactoryGenerator that goes with it. Both of these live in the `jana::omni` namespace, so they hopefully shouldn't conflict with the EICrecon version.

PR #1511, opened 2024-06-20
CalorimeterClusterRecoCoG: set intrinsic{Phi,Theta}

veprbl:
This partially reverts effect of #1391. The clusterShape has become a way to extend EDM without going through extending it. We should sort those variable out into proper members, this is a first step.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -324,17 +324,14 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
   cl.addToShapeParameters( eigenValues_3D[1].real() ); // 3D x-y-z cluster width 2
   cl.addToShapeParameters( eigenValues_3D[2].real() ); // 3D x-y-z cluster width 3
 
-  if (axis_z != 0) {
+  if (m_cfg.longitudinalShowerInfoAvailable) {
     cl.setIntrinsicPhi(atan2(axis_y, axis_x));
     cl.setIntrinsicTheta(atan2(std::hypot(axis_x, axis_y), axis_z));
+    // TODO intrinsicDirectionError
   } else {
-    // best estimate on the cluster direction is the cluster position
-    // for simple 2D CoG clustering
-    cl.setIntrinsicTheta(edm4hep::utils::anglePolar(cl.getPosition()));
-    cl.setIntrinsicPhi(edm4hep::utils::angleAzimuthal(cl.getPosition()));
+    cl.setIntrinsicTheta(NAN);
+    cl.setIntrinsicPhi(NAN);
   }
-  // TODO intrinsicDirectionError
-
 
   return std::move(cl);
 }
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -326,7 +326,7 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
 
   if (axis_z != 0) {
     cl.setIntrinsicPhi(atan2(axis_y, axis_x));
-    cl.setIntrinsicTheta(atan2(std::hypot(axis_x, axis_y), axis_z);
+    cl.setIntrinsicTheta(atan2(std::hypot(axis_x, axis_y), axis_z));
   } else {
     // best estimate on the cluster direction is the cluster position
     // for simple 2D CoG clustering
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -242,12 +242,11 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
   Eigen::Vector3f sum1_3D = Eigen::Vector3f::Zero();
   Eigen::Vector2cf eigenValues_2D = Eigen::Vector2cf::Zero();
   Eigen::Vector3cf eigenValues_3D = Eigen::Vector3cf::Zero();
-  //the axis is the direction of the eigenvalue corresponding to the largest eigenvalue.
-  double axis_x=0, axis_y=0, axis_z=0;
+  // the axis is the direction of the eigenvalue corresponding to the largest eigenvalue.
+  edm4hep::Vector3f axis;
+
   if (cl.getNhits() > 1) {
-
     for (const auto& hit : pcl.getHits()) {
-
       float w = weightFunc(hit.getEnergy(), totalE, logWeightBase, 0);
 
       // theta, phi
@@ -300,19 +299,15 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
             return std::real(a) < std::real(b);
         }
       );
-      auto axis = eigenvectors.col(std::distance(
+      auto axis_eigen = eigenvectors.col(std::distance(
             eigenValues_3D.begin(),
             max_eigenvalue_it
         ));
-      axis_x=axis(0,0).real();
-      axis_y=axis(1,0).real();
-      axis_z=axis(2,0).real();
-      double norm=sqrt(axis_x*axis_x+axis_y*axis_y+axis_z*axis_z);
-      if (norm!=0){
-        axis_x/=norm;
-        axis_y/=norm;
-        axis_z/=norm;
-      }
+      axis = {
+        axis_eigen(0,0).real(),
+        axis_eigen(1,0).real(),
+        axis_eigen(2,0).real(),
+      };
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc
```diff
@@ -97,10 +89,8 @@ TEST_CASE( "the calorimeter CoG algorithm runs", "[CalorimeterClusterRecoCoG]" )
 
 
   for (auto clust : *clust_coll){
-    // require that this cluster's axis is 0,0,1
-    REQUIRE(clust.getShapeParameters()[7] == 0);
-    REQUIRE(clust.getShapeParameters()[8] == 0);
-    REQUIRE(abs(clust.getShapeParameters()[9]) == 1);
+    REQUIRE_THAT(clust.getIntrinsicTheta(), Catch::Matchers::WithinAbs(M_PI * 3 / 4, EPSILON));
+    REQUIRE_THAT(clust.getIntrinsicPhi(), Catch::Matchers::WithinAbs(0, EPSILON));
```
wdconinc, line 93: Shouldn't this be:
```suggestion
    REQUIRE_THAT(clust.getIntrinsicTheta(), Catch::Matchers::WithinAbs(M_PI / 4, EPSILON));
    REQUIRE_THAT(clust.getIntrinsicPhi(), Catch::Matchers::WithinAbs(M_PI, EPSILON));
```

My mental model is that for a gamma and a perfect detector the cluster intrinsic directions should agree with the direction of the gamma, and electrons it may disagree in phi due to track bending.

This means we need to flip the eigenvector to point away from the origin, or dot product of cluster cog position and principal eigenvector should be positive.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -329,10 +323,19 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
   cl.addToShapeParameters( eigenValues_3D[0].real() ); // 3D x-y-z cluster width 1
   cl.addToShapeParameters( eigenValues_3D[1].real() ); // 3D x-y-z cluster width 2
   cl.addToShapeParameters( eigenValues_3D[2].real() ); // 3D x-y-z cluster width 3
-  //last 3 shape parameters are the components of the axis direction
-  cl.addToShapeParameters( axis_x );
-  cl.addToShapeParameters( axis_y );
-  cl.addToShapeParameters( axis_z );
+
+  if (axis_z != 0) {
```
wdconinc, line 327: This is solely intended to catch singular cases where an eigenvector cannot be found, it seems (there is not div0 reason for it and only meaningfully test that the value is still at its initial value).

Are there more than the following two classes of these cases:
- two hits only,
- multiple hits where digitization and position quantization result in the perfect singularity of the cluster.

Presumably the second case is infrequent to the point of being negligible. The first may be numerous in the presence of low energy backgrounds.

I feel that adding the fallback gives clusters with two hits a different behavior which may lead to confusion. I also think that writing the exact zero in the intrinsic phi and theta may lead to confusion... I wish we could indicate validity of the eigenvalues somehow (without a special value, zero or otherwise, in the directions or their errors).

src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc
veprbl: You are correct. I was saving the fix for another PR, but can implement here.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
veprbl: The intention was to not handle trivial clusters, but handle detectors that don't provide 3d profile of the shower. I think, we should add a config flag to switch this instead.

wdconinc: Ok, I see. There are then:
- trivial clusters because of number of hits,
- trivial clusters because of detector planarity,
- accidentally singular clusters.

I think the data model should define what to do for trivial clusters because of detector planarity. It seems that having this depend on a config flag (which is still hard to access in our output files) could cause confusion too.

src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc
wdconinc: Descoping here is fine with me. Last week before July, so better to get this in before the rush.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
wdconinc: (Same comment as above about descoping applies here too.)

wdconinc:
(Outside of reviewable context)

This is a symmetric real (positive semi-definite), i.e. self-adjoint problem, so the `Eigen::EigenSolver<Eigen::Matrix2f>` and `Eigen::EigenSolver<Eigen::Matrix3f>` on 2D and 3D covariance matrices can be replaced by `Eigen::SelfAdjointEigenSolver<Eigen::Matrix2f>` and `Eigen::SelfAdjointEigenSolver<Eigen::Matrix3f>`. This then allows to change `Eigen::Vector2cf eigenValues_2D` and `Eigen::Vector3cf eigenValues_3D` to purely real `Eigen::Vector2f eigenValues_2D` and `Eigen::Vector3f eigenValues_3D`, and there is no need to explicit `real` calls or carrying zero imaginary parts.

Since the quadratic forms that add to the covariance matrix are symmetric, we only need to compute/store the lower triangular part, but it doesn't seem like Eigen3 provides a self-adjoint matrix class itself (only a view), so I don't think this can be simplified further, computationally.

src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc
veprbl: Implemented.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
veprbl: Implemented. No handling of trivial clusters yet.

src/tests/algorithms_test/calorimetry_CalorimeterClusterRecoCoG.cc
```diff
@@ -28,16 +29,19 @@ using eicrecon::CalorimeterClusterRecoCoGConfig;
...
   std::shared_ptr<spdlog::logger> logger = spdlog::default_logger()->clone("CalorimeterClusterRecoCoG");
   logger->set_level(spdlog::level::trace);
 
   CalorimeterClusterRecoCoGConfig cfg;
   cfg.energyWeight = "log";
-  cfg.sampFrac = 0.0203,
-  cfg.logWeightBaseCoeffs={5.0,0.65,0.31},
-  cfg.logWeightBase_Eref=50*dd4hep::GeV,
+  cfg.sampFrac = 0.0203;
+  cfg.logWeightBaseCoeffs = {5.0,0.65,0.31},
```
wdconinc, line 42:
```suggestion
  cfg.logWeightBaseCoeffs = {5.0,0.65,0.31};
```

wdconinc: OT why doesn't clang-tidy catch this? I know it's valid but it is also clearly unintended.

wdconinc:
Can you rebase to allow for review without unrelated changes?

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -329,10 +318,22 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
-  cl.addToShapeParameters( axis_y );
-  cl.addToShapeParameters( axis_z );
+
+
+  double dot_product = cl.getPosition() * axis;
+  if (dot_product < 0) {
+    axis = -1 * axis;
+  }
+
+  if (m_cfg.longitudinalShowerInfoAvailable) {
```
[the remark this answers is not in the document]
veprbl: An easy check was to see if `axis_z != 0`, but that is questionable.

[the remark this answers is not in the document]
wdconinc: This situation is the singular matrix exception, right? Two eigenvalues non-zero and last eigenvalue zero.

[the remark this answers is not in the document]
veprbl: No it's a possibility for "flat" arrangement, there is no inversion going on.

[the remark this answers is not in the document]
wdconinc: But still, flat arrangement results in a 2D cluster with two principal directions in the detector plane, and the third eigenvector normal to the plane with an eigenvalue of zero (size of cluster in that direction). Having one zero eigenvalue makes this a singular matrix.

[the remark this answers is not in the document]
veprbl: Yes, singular matrices have eigenvalues. If lowest eigenvalue is zero, that means that cluster centers are on a single plane. This is practically only possible in a "flat" detector that doesn't provide 3d shower profile.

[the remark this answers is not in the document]
wdconinc: As discussed and agreed with you, the preference is for making detectors explicitly 'flat' to avoid having to introduce arbitrary tolerance parameters to allow for 3d clusters solely introduced by minor misalignment in the block perpendicular to their arrangement plane.

wdconinc (approved):

PR #1512, opened 2024-06-22
added functionality to use ZDC Ecal clusters in neutron recon

sebouh137:
Includes clusters in the ZDC Ecal in the total energy when reconstructing neutrons

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1516 )
- [ ] Documentation update
- [ ] Other: __

no
Yes.  includes ZDC Ecal clusters in neutron recon.


src/algorithms/reco/FarForwardNeutronReconstruction.cc
```diff
@@ -6,14 +6,14 @@
...
 #include <stdexcept>
+#include <vector>
 
 #include "FarForwardNeutronReconstruction.h"
 
 /**
- Creates a "neutron candidate" Reconstructed Particle consisting of all clusters in a
- given ClusterCollection.  Its energy is the sum of the energies of the constituent clusters
+ Creates a "neutron candidate" Reconstructed Particle consisting of all clusters in the
+ ZDC (both Hcal and Ecal collections).  Its energy is the sum of the energies of the constituent clusters
```
ruse-traveler, line 16: Remember, we want algorithms to be as generic as possible: while the ZDC will likely be the primary user of this algorithm for the time being, this could be used by other combinations of subsystems. So please make sure code and comments reflect this.

Not to mention, the ECal clusters are an optional input, so how about the following?
```
Creates a candidate neutron by summing the energy of all clusters from a specified hadronic calorimeter and (optionally) from a specified Electromagnetic calorimeter.
```

src/algorithms/reco/FarForwardNeutronReconstruction.h
```diff
@@ -29,9 +31,9 @@ using FarForwardNeutronReconstructionAlgorithm = algorithms::Algorithm<
        public:
          FarForwardNeutronReconstruction(std::string_view name)
                   : FarForwardNeutronReconstructionAlgorithm{name,
-                                        {"inputClusters"},
+                                                             {"inputClustersHcal", "inputClustersEcal"},
                                         {"outputNeutrons"},
```
ruse-traveler, line 35: Minor nitpick: please keep ctor arguments aligned

src/algorithms/reco/FarForwardNeutronReconstruction.h
```diff
@@ -17,7 +18,8 @@ namespace eicrecon {
 
 using FarForwardNeutronReconstructionAlgorithm = algorithms::Algorithm<
    algorithms::Input<
-       const edm4eic::ClusterCollection
+     const edm4eic::ClusterCollection,
+     std::optional<edm4eic::ClusterCollection>
     >,
     algorithms::Output<
        edm4eic::ReconstructedParticleCollection
```
ruse-traveler, line 25: Minor nitpick: please be consistent with alignment
```suggestion
       const edm4eic::ClusterCollection,
       std::optional<edm4eic::ClusterCollection>
```

src/detectors/ZDC/ZDC.cc
```diff
@@ -82,7 +82,7 @@ extern "C" {
              "EcalFarForwardZDCTruthClusterAssociations"}, // edm4eic::MCRecoClusterParticleAssociation
             {
               .energyWeight = "log",
-              .sampFrac = 1.0,
+              .sampFrac = 0.701,
```
ruse-traveler, line 85: For my own understanding: is this a correction to the EM scale? Or is this a scale factor to adjust the relative weighting between the ECal and HCal components?

If it's the latter, this should be a parameter for the Neutron Reconstruction algorithm rather than overloading the sampling fraction parameter for the cluster reconstruction.

src/detectors/ZDC/ZDC.cc
```diff
@@ -99,7 +99,7 @@ extern "C" {
              "EcalFarForwardZDCClusterAssociations"}, // edm4eic::MCRecoClusterParticleAssociation
             {
               .energyWeight = "log",
-              .sampFrac = 1.0,
+              .sampFrac = 0.701,
```
ruse-traveler, line 102: Same here...

ruse-traveler (changes requested):
Hi Sebouh, thanks for the development! I alluded to this in one of my comments, but one thing I think this could use is to add a parameter to adjust the relative weighting of the ECal and HCal clusters (even if they're both 1.0 for the time being).

src/detectors/ZDC/ZDC.cc
sebouh137: I determined the value of the sampling fraction by running gamma simulations in the Ecal, and taking the ratio between the generated photon energy and the sum of simulated hit energies.  I think this was set to 1.0 as a placeholder earlier.

src/detectors/ZDC/ZDC.cc
sebouh137: same as above

src/algorithms/reco/FarForwardNeutronReconstruction.cc
sebouh137: Thanks, done

src/algorithms/reco/FarForwardNeutronReconstruction.h
sebouh137: thanks, done

src/algorithms/reco/FarForwardNeutronReconstruction.h
sebouh137: Fixed, thanks!

src/detectors/ZDC/ZDC.cc
ruse-traveler: Gotcha! Makes sense! That being said, I do think including a parameter to adjust the relative weighting of the HCal vs. ECal would be a useful. Something like so:
```
float ecalWeight = 1.0;
float hcalWeight = 1.0;
```

Where, of course, the parameters could be updated when needed. I suggest this since (and correct if I'm wrong), the HCal part of the ZDC is non-compensating, right? So e/h will differ between the ECal and HCal parts...

(Or if it is, this algorithm may get used by a different combination of subsystems where the HCal part is non-compensating...)

sebouh137:
> Hi Sebouh, thanks for the development! I alluded to this in one of my comments, but one thing I think this could use is to add a parameter to adjust the relative weighting of the ECal and HCal clusters (even if they're both 1.0 for the time being).

I think that's a good idea to have different corrections for the HCal and ECal.  Currently they are set to be exactly the same as one another.  One variant of what you propose is we could do is have two sets of parameters, one for the Hcal, and a second set of parameters in the same form for the Ecal.  For now, I'll have both parameters be the same, and if this parameterization needs to be tweaked at a later point, someone can do that.

ruse-traveler:
> I think that's a good idea to have different corrections for the HCal and ECal. Currently they are set to be exactly the same as one another. One variant of what you propose is we could do is have two sets of parameters, one for the Hcal, and a second set of parameters in the same form for the Ecal. For now, I'll have both parameters be the same, and if this parameterization needs to be tweaked at a later point, someone can do that.

🙌 Awesome! Sounds good!

sebouh137:
> > I think that's a good idea to have different corrections for the HCal and ECal. Currently they are set to be exactly the same as one another. One variant of what you propose is we could do is have two sets of parameters, one for the Hcal, and a second set of parameters in the same form for the Ecal. For now, I'll have both parameters be the same, and if this parameterization needs to be tweaked at a later point, someone can do that.
> 
> 🙌 Awesome! Sounds good!

Added this feature.  it should be ready now.  Let me know if you have anything else you'd like to add, otherwise please approve the pull request.  Thanks.

sebouh137:
The pre-commit has been acting up today; I'm not sure why it's failing.  @ruse-traveler do you have any insights as to why this is happening?

wdconinc:
githubstatus.com

sebouh137:
@ruse-traveler I have added the changes you have requested, and all of the relevant checks have passed.  Please approve my pull request, if you don't have any other questions or comments.  
Thanks,
Sebouh

veprbl:
I have a question: should this, in principle, do some sort of geometric matching between the two types of clusters? Is that the plan for follow-up changes?

sebouh137:
> I have a question: should this, in principle, do some sort of geometric matching between the two types of clusters? Is that the plan for follow-up changes?

This is a fair suggestion to include in a future pull request.  I think that would add a cut that only includes Ecal clusters within some distance of the most energetic Hcal cluster.  But for now let's just get this pull request merged.

ruse-traveler:
IIRC I believe you mentioned looking at individual clusters/showers for neutron reconstruction in the ZDC in the LFHCAL meeting at some point. That definitely would need some geometric matching between the ECal and HCal clusters...

But agreed! That's for another time; I'm happy with this PR as it is now!

ruse-traveler (approved):

PR #1513, opened 2024-06-22
fix: ensure USE_ONNX=OFF can compile

wdconinc:
This PR passes USE_ONNX as a compile definition and guards use of ONNX functionality with an ifdef.

- [x] Bug fix (issue: compilation failure when USE_ONNX=OFF)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,6 +233,7 @@ endif()
 option(USE_ONNX "Compile with ONNX support" ON)
 if(${USE_ONNX})
   find_package(onnxruntime CONFIG)
+  add_compile_definitions(USE_ONNX)
 endif()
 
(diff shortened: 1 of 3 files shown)


wdconinc:
This approach continues to break downstream compilation due to the use of the ONNX functionality in headers. We should promote USE_ONNX as a target compile definition attached to one of our installed targets or somehow propagate USE_ONNX through EICreconConfig.cmake. This is similar to the issue in JANA2 with HAVE_PODIO.

I'm considering this issue out of scope for this PR which just aims to fix compilation within EICrecon.

veprbl (approved):
Diff LGTM

PR #1518, opened 2024-06-28
PIDLookup: define momentum cuts to set an explicit boundary between TOF and hpDIRC domains

veprbl:

Currently DIRC is processed after TOF, and will replace previously looked up TOF values whenever DIRC has a table entry.
The new approach is to use DIRC values only for $|\vec{p}| > 2.5$ GeV.

- [x] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators (the idea originally was voiced by TC, this was discussed a month ago in Software and Analysis join meeting)

No

Yes


wdconinc:
@ohartbri Will you have a chance to review this in the next week?

wdconinc:
Code looks fine. But does it make sense that the number of DIRC particle ID entries drops so precipitously, from nearly 250 to 4? (Note to self: failure mode detected for Kolmogorov-Smirnov comparisons...)
https://eicrecon.epic-eic.org/pr/1518/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#DIRCParticleIDs
image

veprbl:
I think the effect is drastic only for Q2min = 0, because at 1000 GeV its not:
image
It must be that not enough particles could reach the 2.5 GeV threshold.

A related, but separate point, is that we should not remove lookup results based on cuts. Only naively sampled PID field itself should be affected, the full set of succesfull table lookups should be available regardless of any cuts.

wdconinc:
@veprbl Is this still a relevant PR?

PR #1531, opened 2024-07-22
WIP: Modified JEventProcessorPODIO::Process to acquire lock only when need…

RaiqaRasool:
This PR introduces changes to the `JEventProcessorPODIO::Process` to limit the scope of locks, enabling the use of JANA's parallelism and improving the performance by avoiding unnecessary serialization of events.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, this PR does not introduce breaking changes. Users do not need to make any changes to their existing code.

Yes, the PR changes the default behavior by enabling JANA's parallelism through the use of `std::unique_lock` with `std::defer_lock`, allowing sections of the code to run in parallel rather than in sequence. The main change is the introduction of `std::unique_lock` with `std::defer_lock` to manage the mutex locking more precisely, allowing for better parallelism by unlocking the mutex during non-critical sections of the code.

### DIFF
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -372,7 +372,7 @@ void JEventProcessorPODIO::FindCollectionsToWrite(const std::shared_ptr<const JE
 void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent> &event) {
 
     std::unique_lock<std::mutex> lock(m_mutex, std::defer_lock);
-    lock.lock(); 
+    lock.lock();
     if (m_is_first_event) {
         FindCollectionsToWrite(event);
         m_is_first_event = false;
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -371,10 +371,13 @@ void JEventProcessorPODIO::FindCollectionsToWrite(const std::shared_ptr<const JE
 
 void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent> &event) {
 
-    std::lock_guard<std::mutex> lock(m_mutex);
+    std::unique_lock<std::mutex> lock(m_mutex, std::defer_lock);
+    lock.lock(); 
     if (m_is_first_event) {
         FindCollectionsToWrite(event);
+        m_is_first_event = false;
     }
+    lock.unlock();
 
     for (const std::string& coll : m_collections_to_write) {
         try {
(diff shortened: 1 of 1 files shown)


src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -371,10 +371,13 @@ void JEventProcessorPODIO::FindCollectionsToWrite(const std::shared_ptr<const JE
...
 void JEventProcessorPODIO::Process(const std::shared_ptr<const JEvent> &event) {
 
-    std::lock_guard<std::mutex> lock(m_mutex);
+    std::unique_lock<std::mutex> lock(m_mutex, std::defer_lock);
+    lock.lock();
     if (m_is_first_event) {
         FindCollectionsToWrite(event);
+        m_is_first_event = false;
     }
+    lock.unlock();
```
wdconinc, line 380: If I understand correctly (`m_is_first_event` monotonically changing from true to false), the lock can be in `if`, avoiding a lock/unlock after the first event. I'd want to avoid an unnecessary lock since it potentially introduces a wait for the write operation.

The (new) race condition to avoid then is `m_is_first_event` changing from true to false while the condition is evaluated, so we'd have to do a second locked `if` inside the unlocked `if`, but it would virtually never be used.

wdconinc (commented):
Was that all that was needed? In transit but I thought it wasn't that easy...

src/services/io/podio/JEventProcessorPODIO.cc
nathanwbrei: You're describing the ABA problem. I think `std::call_once` is the tool for this purpose. (Just make sure that any code inside call_once is exception free because we've been bitten by bad standard library implementations before)

nathanwbrei:
> Was that all that was needed? In transit but I thought it wasn't that easy...

The goal is to find out. I'm also resuscitating #1069 as part of this effort.

src/services/io/podio/JEventProcessorPODIO.cc
wdconinc: > any code inside call_once is exception free

The code inside is an entire event processing in EICrecon :=)

nathanwbrei: Explanation of double-checked locking and ABA problem for anyone who is interested: https://en.wikipedia.org/wiki/Double-checked_locking#Motivation_and_original_pattern

However, this is an optimization detail that I would avoid for now in favor of the simple, obvious solution. We'd like to verify:
- Consistent results in the output file
- A sensible cores vs throughput curve
- No data races uncovered by valgrind/tsan/whatever

nathanwbrei: > > any code inside call_once is exception free
> 
> The code inside is an entire event processing in EICrecon :=)

Nope, just `FindCollectionsToWrite()` thankfully :)

wdconinc: Ok, I see. At some point this ran through the whole first event to figure out which output names were activated, right? Or do I remember that wrong?

wdconinc:
Closing since #975 is merged and actively used.

PR #1532, opened 2024-07-22
lowered the amount of distance neighbor hits in topoclustering in the ZDC

sebouh137:
Redefines neighboring subcell hits in the ZDC (in the context of topo clustering) to be the nearest neighbors only for hits in the same layer, and the same position for hits in adjacent layers.   This change reduces the cases where showers from different particles are merged into a single cluster

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: parameter tweak

no.  
yes.  Changes the default behavior of the topo clustering in the ZDC


src/detectors/ZDC/ZDC.cc
```diff
@@ -158,12 +158,12 @@ extern "C" {
...
                 .neighbourLayersRange = 1,
-                .localDistXY = {0.76*side_length, 0.76*side_length*sin(M_PI/3)},
-                .layerDistXY = {0.76*side_length, 0.76*side_length*sin(M_PI/3)},
+                .localDistXY = {0.5*side_length, 0.5*side_length*sin(M_PI/3)},
+                .layerDistXY = {0.25*side_length, 0.25*side_length*sin(M_PI/3)},
                 .layerMode=eicrecon::ImagingTopoClusterConfig::ELayerMode::xy,
                 .sectorDist = 10.0 * dd4hep::cm,
                 .minClusterHitEdep = 100.0 * dd4hep::keV,
-                .minClusterCenterEdep = 3.0 * dd4hep::MeV,
+                .minClusterCenterEdep =3.0 * dd4hep::MeV,
```
ruse-traveler, line 166: Minor nitpick: please keep spacing consistent
```suggestion
                .minClusterCenterEdep = 3.0 * dd4hep::MeV,
```

ruse-traveler (commented):
Looks reasonable to me! I have an extremely minor nitpick, but otherwise seems good.

Just so I could educate myself: are there some studies that shows the impact of this tweak? I was having a hard time getting a feel for how this changes things just from the capybara...

veprbl:
@sebouh137 Is this ready? Can you address the review question?

sebouh137:
> @sebouh137 Is this ready? Can you address the review question?

Hi @veprbl I determined that this change was necessary in the context of the Lambda benchmark in the ZDC, 
 https://github.com/eic/physics_benchmarks/pull/19 , since I was seeing a lot of events where clusters were being merged together when they shouldn't be merging (causing a lot of events to have only one cluster, when they should have two or ideally three if the two photons's and the neutron's showers can be distinguished from one another).  I can make some plots with and without this change to the EICrecon

veprbl (approved):

veprbl:
Feel free to post figures as they become available.

PR #1533, opened 2024-07-22
ImagingTopoCluster: fix Init printouts by adding missing break statements

sebouh137:
Fixed an issue where "break" is not included in a case/break statement.

When initializing the ImagingTopoCluster algorithm, there is a switch statement wherein different things are printed to inform the user what the parameters are (depending on the layer mode).  This lacked breaks for the cases, causing the other print statements to be executed.  This patch fixes it.  

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no.  no
yes.  only prints out the information about the cuts for defining neighbors that is relevent based on the layer mode.


veprbl (approved):
LGTM

PR #1534, opened 2024-07-25
TracksToParticles: avoid initializing mass from truth

veprbl:
People were confused by us removing truth PDG in ReconstructedParticles, so we better remove this too, before it is relied upon for too much.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes


wdconinc (approved):

wdconinc (approved):
You should be able to get to the truth through associations only.

PR #1536, opened 2024-07-26
BTOF,DRICH,DIRC,PFRICH: unroll PID LUT factory initialization loops

veprbl:
This is to help developers to find the code using simple search.

### DIFF
--- a/src/detectors/BTOF/BTOF.cc
+++ b/src/detectors/BTOF/BTOF.cc
@@ -8,15 +8,13 @@
 #include <DD4hep/Detector.h>
 #include <Evaluator/DD4hepUnits.h>
 #include <JANA/JApplication.h>
-#include <fmt/core.h>
 #include <algorithm>
 #include <gsl/pointers>
 #include <memory>
 #include <stdexcept>
-#include <string>
-#include <vector>
 
 #include "algorithms/interfaces/WithPodConfig.h"
+#include "algorithms/pid_lut/PIDLookupConfig.h"
 #include "extensions/jana/JOmniFactoryGeneratorT.h"
 #include "factories/digi/SiliconTrackerDigi_factory.h"
 #include "factories/tracking/TrackerHitReconstruction_factory.h"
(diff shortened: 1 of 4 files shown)


wdconinc:
I think this programmatic creating of collection names to save a repeat call is obscuring quite a lot, and makes grepping ineffective. Adding the comment is a bandaid, and of course I like bandaids to stop the bleeding but I also want the patient to stop playing with sharp knives.

wdconinc (dismissed):

veprbl:
I agree, this is now addressed.

src/detectors/BTOF/BTOF.cc
```diff
@@ -64,32 +62,46 @@ void InitPlugin(JApplication *app) {
...
-              app
-              ));
-    }
+    PIDLookupConfig pid_cfg {
+      .filename="calibrations/tof.lut",
+      .system=BarrelTOF_ID,
+      .pdg_values={11, 211, 321, 2212},
+      .charge_values={1},
+      .momentum_edges={0.0, 0.3, 0.6, 0.9, 1.2, 1.5, 1.8, 2.1, 2.4, 2.7, 3.0, 3.3, 3.6, 3.9, 4.2, 4.5, 4.8, 5.1, 5.4, 5.7, 6.0},
+      .polar_edges={2.50, 10.95, 19.40, 27.85, 36.30, 44.75, 53.20, 61.65, 70.10, 78.55, 87.00, 95.45, 103.90, 112.35, 120.80, 129.25, 137.70, 146.15, 154.60},
```
wdconinc, line 71: (Optional) Lots of magic and opportunities for typos in these numbers. Maybe we need a `linspace` helper (https://godbolt.org/z/jd657W4qK):
```cpp
    auto linspace = [](std::floating_point auto start, std::floating_point auto stop, size_t num){
        std::vector<decltype(start)> v(num);
        std::ranges::generate(v, [&start, n = 0, diff = (stop - start)/(num-1)]() mutable {
            return start + n++ * diff;
        });
        return std::move(v);
    };
```

wdconinc (approved):

src/detectors/BTOF/BTOF.cc
veprbl: TBH, I'd rather not. Most numbers came directly from DSC's, they should do the updates, if they want to.

PR #1538, opened 2024-07-27
fix: move static ParticleMap into header as inline for downstream use

wdconinc:
Downstream plugins that use the ParticleSvc as an interface just include the header and are supposed to link against the library to get the static member. Having to load a shared library for just a static list is a bit annoying for a servicde interface definition.

This PR moves the static default list into an inline static definition (not a shared_ptr), and everything now lives in the header.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


rahmans1 (commented):
Compiles and runs without issue. Slightly different output compared to what I get from standard eicrecon. Does this matter? @wdconinc  

Screenshot 2024-07-29 124711b
Screenshot 2024-07-29 124649a

wdconinc:
Capybara claims no differences. This shouldn't introduce any differences at all, so I'm surprised it's showing differences for you.

veprbl (approved):

PR #1541, opened 2024-07-31
Update ZDC LYSO Pedestal Sigma Parameter

JiaJunHuang120000:
Correct ZDC LYSO pedestal sigma parameter to for accurate reconstruction of low energy photons. Currently each ADC channel correspond to 0.06 MeV, and with previous value of pedestal sigma towers with energy less than 0.08 MeV were removed. Now, towers with lower energy deposites are kept.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: _Parameter Update_

- [x] Changes have been communicated to collaborators @bschmookler 

No.

Only in ZDC ECal.


src/detectors/ZDC/ZDC.cc
```diff
@@ -33,7 +33,7 @@ extern "C" {
             .capADC = 32768,
             .dyRangeADC = 2000 * dd4hep::MeV,
             .pedMeanADC = 400,
-            .pedSigmaADC = 3.2,
+            .pedSigmaADC = 0.2,
```
veprbl, line 36: Basically, here you claim that 15-bit ADC can not resolve your pedestal. That is unrealistic.

veprbl:
This appears to break your benchmark: https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/3574966#L3615

src/detectors/ZDC/ZDC.cc
bschmookler: The current parameters cause issues for O(5-10 MeV) photons. 5-10 MeV photons may be a reasonable lower limit target for the LYSO.

We do not have expected noise estimates yet. But the current version has a threshold cut at 4*3.2 = 12.8 ADC channels, which is a deposited energy per tower of about 0.8 MeV. For LYSO, a 0.8 MeV deposit corresponds to a light yield of about 25,000 photons.

If we don't want to decrease to pedestal width from 3.2 ADC channels, we maybe can decrease the dynamic range of the LYSO from the current 2 GeV value.

bschmookler: Sounds good. Thanks. We'll chat with the DSC about the parameter values.

src/detectors/ZDC/ZDC.cc
```diff
@@ -64,7 +64,7 @@ extern "C" {
           {
             .sectorDist = 5.0 * dd4hep::cm,
             .localDistXY = {50 * dd4hep::cm, 50 * dd4hep::cm},
-            .dimScaledLocalDistXY = {50.0*dd4hep::mm, 50.0*dd4hep::mm},
+            .dimScaledLocalDistXY = {5, 5},
```
wdconinc, line 67: Likely not going to work since you rely on buggy behavior fixed in #1598.

PR #1543, opened 2024-08-01
InclusiveKinematicsML: gracefully handle non-tensor types

veprbl:
This should save developers from some confusing crashes.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


ruse-traveler (approved):
Looks good! For my own understanding: are there ever cases we would be interested in a type other than `ONNX_TYPE_TENSOR`?

veprbl:
That is a good question. It seems like different exporters may choose to use non-tensor output nodes. For example, I've encountered a `ONNX_TYPE_SEQUENCE` type for one of output nodes of a classifier model exported from CatBoost. Not clear to me if we are to deal with those, or somehow adjust the graph to only produce tensors, but we definitely don't want to crash EICrecon over this, hence the PR.

ruse-traveler:
Gotcha! Thanks for the clarification! (Also fully agree!)

PR #1544, opened 2024-08-01
Added missing seed finder parameter

bschmookler:
The _rMin_ parameter defined in the seeder configuration file was not being passed to the seed finder. This PR adds this. The parameter is currently set to the default value defined here:

https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/SeedFinderOrthogonalConfig.hpp#L39
https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/SeedFinderConfig.hpp#L43

So, there should be no change to the output. But it will be needed in the future if we ever want to search for seeds in an outer part of the tracking detector volume.

In addition, the _bFieldMin_ parameter defined in the seeder configuration file but not passed to the seed finder has been removed. This parameter does not seem to be a Acts seed finder parameter in any event. It seems to only be used if the _EstimateTrackParamsFromSeed_ class is called to fit the seed parameters (which we do not use):

https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp#L148-L149

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
+++ b/src/algorithms/tracking/OrthogonalTrackSeedingConfig.h
@@ -35,7 +35,6 @@ namespace eicrecon {
     float beamPosX         = 0; // x offset for beam position
     float beamPosY         = 0; // y offset for beam position
     float impactMax        = 3. * Acts::UnitConstants::mm; // Maximum transverse PCA allowed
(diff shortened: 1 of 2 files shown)


veprbl (approved):
Looks good

PR #1547, opened 2024-08-04
fix(ci): suppress more UB false positives out of our scope

wdconinc:
We're at 71 warnings (https://github.com/eic/EICrecon/actions/runs/10238130891), most outside of our control in boost iostream, eigen3, and acts. This PR causes UBSan to print the actual error name to use in the suppression file, and then uses that to suppress the false positives (well, out of our scope errors). This bring us down to 34 errors.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
LGTM

PR #1548, opened 2024-08-04
fix: avoid division by zero in neutron identification when no Hcal cluster

wdconinc:
When there is no Hcal cluster found, then the position is still initialized to zero. Setting the momentum components based on a normalized position fails with a division by zero, see e.g. https://github.com/eic/EICrecon/actions/runs/10237639088/job/28321546895#step:8:6883.

This PR requires a cluster in the Hcal for there to be a neutron identification with a momentum direction given by that cluster.

This PR also modifies some of the calls to use a vector type, which is semantically cleaner as it reduces the number of independent variables.

- [x] Bug fix (issue: division by zero)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @sebouh137 @ruse-traveler 

No.

No.

### DIFF
--- a/src/algorithms/reco/FarForwardNeutronReconstruction.cc
+++ b/src/algorithms/reco/FarForwardNeutronReconstruction.cc
@@ -4,6 +4,7 @@
 #include <edm4eic/ClusterCollection.h>
 #include <edm4eic/ReconstructedParticleCollection.h>
 #include <edm4hep/Vector3f.h>
+#include <edm4hep/utils/vector_utils.h>
 #include <math.h>
 #include <gsl/pointers>
 #include <stdexcept>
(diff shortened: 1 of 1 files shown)


veprbl:
@rahmans1 Are you able to review this today? If not, let's reassign.

sebouh137 (approved):
Thanks Wouter for this pull request.  I approve.

wdconinc:
The relevant diffs in capybara are as follows (subset):
image
We lost 19 'zero energy' neutrons (where no clusters were in the Hcal), but other non-zero energy neutrons (with clusters in the Hcal) are unchanged. These 19 neutrons are also removed from other entries, such as the covariance matrices etc.

PR #1540, opened 2024-07-31
BHCAL: drop support for legacy sectors

veprbl:
This readout field was removed in 24.03.0. We don't need to support it.

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No

### Does this PR change default behavior?
No

PR #1551, opened 2024-08-04
fix: avoid DD4hep printout for missing Barrel Hcal sector field

wdconinc:
For every run, the BHcal tries to find whether a `sector` field is in the id specification, and every time DD4hep prints out an error because it doesn't exist (e.g. https://github.com/eic/EICrecon/actions/runs/10237639088/job/28321546930#step:8:178).

This PR changes our strategy to looking in the field list, and avoiding the error message from DD4hep.

- [x] Bug fix (issue: error message for correct operation)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @ruse-traveler 

No.

No.

### DIFF
--- a/src/detectors/BHCAL/BHCAL.cc
+++ b/src/detectors/BHCAL/BHCAL.cc
@@ -39,13 +39,9 @@ extern "C" {
         }
 
         // check if sector field is present
-        bool useSectorIndex = false;
-        try {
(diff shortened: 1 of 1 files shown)


wdconinc:
Or #1540 yeah. Dropping code is better :-D

PR #1554, opened 2024-08-05
CalorimeterIslandCluster: peakNeighbourhoodMatrix to control how far peaks need to be to split

veprbl:

This reduces excessive splitting observed in EEEMCal. The issue is observed by looking at E/p in log scale. The splitting causes an extended tail towards lower E/p, which reduces electron efficiency.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
+++ b/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
@@ -80,6 +80,7 @@ static edm4hep::Vector2f globalDistEtaPhi(const CaloHit &h1, const CaloHit &h2) 
 // AlgorithmInit
 //------------------------
 void CalorimeterIslandCluster::init() {
+
     static std::map<std::string,
                 std::tuple<std::function<edm4hep::Vector2f(const CaloHit&, const CaloHit&)>, std::vector<double>>>
     distMethods{
--- a/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
+++ b/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
@@ -80,7 +80,6 @@ static edm4hep::Vector2f globalDistEtaPhi(const CaloHit &h1, const CaloHit &h2) 
 // AlgorithmInit
 //------------------------
 void CalorimeterIslandCluster::init() {
-
     static std::map<std::string,
                 std::tuple<std::function<edm4hep::Vector2f(const CaloHit&, const CaloHit&)>, std::vector<double>>>
     distMethods{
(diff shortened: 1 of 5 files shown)


veprbl:
pred_log_epic_craterlake_24 07 0
pred_log_reduced_splitting
pion_rej_epic_craterlake_24 07 0
pion_rej_reduced_splitting

This removes the tail and improves cluster-based pion rejection in single-particle simulations.

src/detectors/EEMC/EEMC.cc
```diff
@@ -69,6 +69,7 @@ extern "C" {
           "EcalEndcapNIslandProtoClusters", {"EcalEndcapNRecHits"}, {"EcalEndcapNIslandProtoClusters"},
           {
             .adjacencyMatrix = "(abs(row_1 - row_2) + abs(column_1 - column_2)) == 1",
+            .peakNeighbourhoodMatrix = "max(abs(row_1 - row_2), abs(column_1 - column_2)) == 1",
```
ruse-traveler, line 72: I'm having a little trouble following what this does: does this matrix define how far apart two maxima need to be in order to be considered two distinct maxima? So here, this would mean that two cells need to be at least more than 1 cell (either in a row or column) away to be considered two maxima?

ruse-traveler (commented):
Nice! There's a clear improvement in the electron-pion discrimination! I think this looks pretty good, but I'm having a little trouble visualizing how this works...

src/algorithms/calorimetry/CalorimeterIslandCluster.cc
```diff
@@ -80,7 +80,6 @@ static edm4hep::Vector2f globalDistEtaPhi(const CaloHit &h1, const CaloHit &h2)
 // AlgorithmInit
 //------------------------
 void CalorimeterIslandCluster::init() {
-
```
ruse-traveler, line 83: Also extremely minor nitpick: I'd prefer leaving a space here (keeps the function name visually separated from its body and keeps it consistent with the other functions' implementation).

src/detectors/EEMC/EEMC.cc
veprbl: In order for a cell to be considered a peak, it needs to dominate its neighbours. This parameter allows to define what is considered a neighbour.

`(abs(row_1 - row_2) + abs(column_1 - column_2)) == 1` corresponds to
```
 x
xox
 x
```
whereas the `max(abs(row_1 - row_2), abs(column_1 - column_2)) == 1`
```
xxx
xox
xxx
```
Basically, with the new condition, the tower has to dominate 8 neighbouring towers instead of 4, like before.

ruse-traveler: Ahh I see! Makes sense!

ruse-traveler (approved):
🚀

PR #1557, opened 2024-08-07
fix: avoid integer wraparound in CalorimeterHitDigi

wdconinc:
In some configurations (EcalBarrelImagingRawHits), the pedestal mean (14 a.u.) and sigma (5 a.u.) are such that negative hit pedestal channel numbers are not uncommon in a simple Gaussian model. For low energy hits, this pushes the ADC value into negative double range, and conversion to unsigned long long results in integer wraparound. E.g. https://github.com/eic/EICrecon/actions/runs/10239218932/job/28324710737#step:9:6119.
```
Error: /home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterHitDigi.cc:186:38: runtime error: implicit conversion from type 'long long' of value -2 (64-bit, signed) to type 'unsigned long long' changed the value to 18446744073709551614 (64-bit, unsigned)
```

This PR prevents the hit pedestal value from being negative. This PR does not address other reasons why we can get negative values, through intentionally incorrect user input (negative dynamic range).

- [x] Bug fix (issue: UBSan implicit-integer-sign-change in CalorimeterHitDigi)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -184,7 +184,7 @@ void CalorimeterHitDigi::process(
         double corrMeanScale_value = corrMeanScale(leading_hit);
 
         // Pedestal in ADC channels must be positive
-        double ped = std::max(m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC, 0);
+        double ped = std::max(m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC, 0.0);
 
         // Note: both adc and tdc values must be positive numbers to avoid integer wraparound
         unsigned long long adc = std::llround(ped + edep * corrMeanScale_value * (1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC);
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -183,10 +183,10 @@ void CalorimeterHitDigi::process(
                 : 0;
         double corrMeanScale_value = corrMeanScale(leading_hit);
 
-        # Pedestal in ADC channels must be positive
+        // Pedestal in ADC channels must be positive
         double ped = std::max(m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC, 0);
 
-        # Note: both adc and tdc values must be positive numbers to avoid integer wraparound
+        // Note: both adc and tdc values must be positive numbers to avoid integer wraparound
         unsigned long long adc = std::llround(ped + edep * corrMeanScale_value * (1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC);
         unsigned long long tdc = std::llround((time + m_gaussian(m_generator) * tRes) * stepTDC);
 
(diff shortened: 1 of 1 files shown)


src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -181,10 +182,14 @@ void CalorimeterHitDigi::process(
...
-        double    ped     = m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC;
-        unsigned long long adc     = std::llround(ped + edep * corrMeanScale_value * ( 1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC);
-        unsigned long long tdc     = std::llround((time + m_gaussian(m_generator) * tRes) * stepTDC);
+        double corrMeanScale_value = corrMeanScale(leading_hit);
+
+        // Pedestal in ADC channels must be positive
+        double ped = std::max(m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC, 0.0);
+
+        // Note: both adc and tdc values must be positive numbers to avoid integer wraparound
+        unsigned long long adc = std::llround(ped + edep * corrMeanScale_value * (1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC);
```
veprbl, line 191: I'm not aware why noise contribution has to be positive in our modeling.
```suggestion
        double ped = m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC;

        // Note: both adc and tdc values must be positive numbers to avoid integer wraparound
        unsigned long long adc = std::max(std::llround(ped + edep * corrMeanScale_value * (1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC), 0);
```

src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -181,10 +182,13 @@ void CalorimeterHitDigi::process(
...
-        double    corrMeanScale_value = corrMeanScale(leading_hit);
-        double    ped     = m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC;
-        unsigned long long adc     = std::llround(ped + edep * corrMeanScale_value * ( 1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC);
-        unsigned long long tdc     = std::llround((time + m_gaussian(m_generator) * tRes) * stepTDC);
+        double corrMeanScale_value = corrMeanScale(leading_hit);
+
+        double ped = m_cfg.pedMeanADC + m_gaussian(m_generator) * m_cfg.pedSigmaADC;
+
+        // Note: both adc and tdc values must be positive numbers to avoid integer wraparound
+        unsigned long long adc = std::max(std::llround(ped + edep * corrMeanScale_value * (1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC), 0);
```
veprbl, line 190:
```suggestion
        unsigned long long adc = std::max(std::llround(ped + edep * corrMeanScale_value * (1.0 + eResRel) / m_cfg.dyRangeADC * m_cfg.capADC), 0LL);
```

veprbl (approved):
Diff LGTM. Although, would be interesting to know if those occurrences ever showed up in our simulations.

wdconinc:
There is a `(adc > m_cfg.capADC ? m_cfg.capADC : adc)` later in the digitization class. This would have capped these overflows at 8192 for the 13-bit ADC in the imaging calorimeter. With the new treatment these are going into the zero bin.

Therefore, the capybara diffs show 2 hits going from overflow to underflow:
image
image

PR #1558, opened 2024-08-07
feat: print error when raw hit has amplitude > ADC capacity

wdconinc:
When a raw hit is encountered with an energy amplitude greater than the ADC capacity we think we are working with, then this almost certainly is an indication of a digitization configuration error. E.g. https://github.com/eic/EICrecon/actions/runs/10239218932/job/28324710737#step:9:6119.

This PR adds an error message for these raw hits in calorimetry.

- [x] Bug fix (issue: https://github.com/eic/EICrecon/actions/runs/10239218932/job/28324710737#step:9:6119)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitReco.cc
@@ -171,7 +171,7 @@ void CalorimeterHitReco::process(
         if (rh.getAmplitude() < m_cfg.pedMeanADC + thresholdADC) {
             continue;
         }
- 
+
         if (rh.getAmplitude() > m_cfg.capADC) {
             error("Encountered hit with amplitude {} outside of ADC capacity {}", rh.getAmplitude(), m_cfg.capADC);
             continue;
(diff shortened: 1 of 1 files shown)


wdconinc:
Not seeing the warning when I expect it...

veprbl (approved):

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -172,6 +172,11 @@ void CalorimeterHitReco::process(
             continue;
         }
 
+        if (rh.getAmplitude() > m_cfg.capADC) {
+            error("Encountered hit with amplitude {} outside of ADC capacity {}", rh.getAmplitude(), m_cfg.capADC);
```
veprbl, line 176: (I wonder if this makes more sense to put in Digi.)

wdconinc: I think it should be in both. This is intended as the check on the digitization.

wdconinc:
> Not seeing the warning when I expect it...

Now realizing why I'm not seeing it: we are capping high values in the digitization (good), and we don't have any bug :-)

PR #1559, opened 2024-08-07
CalorimeterIslandCluster: avoid complaining about missing readouts

veprbl:
This fixes a small regression from https://github.com/eic/EICrecon/pull/1554 pointed out by @wdconinc. Some error messages would appear for detectors that are not using matrices and don't specify "readout" field. This allows the current status quo to continue, but removes the messages.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
+++ b/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
@@ -135,7 +135,9 @@ void CalorimeterIslandCluster::init() {
     };
 
     if (m_cfg.readout.empty()) {
-      error("readoutClass is not provided, it is needed to know the fields in readout ids");
+      if ((!m_cfg.adjacencyMatrix.empty()) || (!m_cfg.peakNeighbourhoodMatrix.empty())) {
+        throw std::runtime_error("readoutClass is not provided, it is needed to know the fields in readout ids");
+      }
     } else {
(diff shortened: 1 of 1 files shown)


wdconinc (dismissed):
Looks good to me.

Ultimately the reason why we can't guard `m_cfg.readout` usage behind existing `m_cfg.adjacencyMatrix` and `m_cfg.peakNeighbourhoodMatrix` checks is that it would duplicate error handling for empty `readout`.

src/algorithms/calorimetry/CalorimeterIslandCluster.cc
```diff
@@ -135,7 +135,9 @@ void CalorimeterIslandCluster::init() {
     };
 
     if (m_cfg.readout.empty()) {
-      error("readoutClass is not provided, it is needed to know the fields in readout ids");
+      if ((!m_cfg.adjacencyMatrix.empty()) || (!m_cfg.peakNeighbourhoodMatrix.empty())) {
+        throw std::runtime_error("readoutClass is not provided, it is needed to know the fields in readout ids");
```
wdconinc, line 139: The config variable is not named `readoutClass`, so the error is confusing.
```suggestion
        throw std::runtime_error("'readout' is not provided, it is needed to know the fields in readout ids");
```

wdconinc (approved):

PR #1560, opened 2024-08-07
PIDLookup: set mass consistent with the assigned PID

veprbl:

The #1534 makes EICrecon no longer leak mass value known from truth. This changes makes it so that mass is set by PID lookup table facility according to the mass of the particle for the PDG code assigned.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes


src/algorithms/pid_lut/PIDLookup.cc
```diff
@@ -130,6 +130,7 @@ void PIDLookup::process(const Input& input, const Output& output) const {
 
     if (identified_pdg != 0) {
       recopart.setPDG(std::copysign(identified_pdg, (identified_pdg == 11) ? -charge : charge));
+      recopart.setMass(m_particleSvc.particle(identified_pdg).mass);
```
bspage912, line 133: How is the mass value initiated? What is the reconstructed particle mass if no PID is assigned?

bspage912 (commented):
See comment about default mass value in PIDLookup.cc

src/algorithms/pid_lut/PIDLookup.cc
veprbl: The mass values are looked up according to masses from PDG https://github.com/eic/EICrecon/blob/97b17d7c4480a29011d4c75a4388c6f66eff9e8f/src/algorithms/interfaces/ParticleSvc.cc#L18
If no PID is assigned the mass is not touched by this algorithm. In the current setup it should remain to be zero.

bspage912: Ok, great. It the mass is set to (kept at) zero in the absence of a PID assignment, I'm happy

bspage912 (approved):

PR #1535, opened 2024-07-26
Add one-to-one relation with raw tracker/calorimeter hit for reconstructed tracker/calorimeter hit

rahmans1:
### Briefly, what does this PR introduce?
Add one-to-one relation with raw tracker/calorimeter hit for reconstructed tracker/calorimeter hit. Contingent on https://github.com/eic/EDM4eic/pull/86. 

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

PR #1564, opened 2024-08-09
feat: add track associations based on raw hit association, weighted by number of hits

wdconinc:
This PR builds on #1535 to create `edm4eic::Track` associations to `edm4hep::MCParticles`. Results in weights like this:
```
events->Draw("CentralCKFTrackAssociations.weight")
```
image

```
events->Draw("sqrt(MCParticles[_CentralCKFSeededTrackAssociations_sim.index].momentum.x**2+MCParticles[_CentralCKFSeededTrackAssociations_sim.index].momentum.y**2+MCParticles[_CentralCKFSeededTrackAssociations_sim.index].momentum.z**2)")
```
image

- [ ] Bug fix (issue #__)
- [x] New feature (issue: track associations)
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/algorithms/tracking/ActsToTracks.h
```diff
@@ -17,13 +19,36 @@ namespace eicrecon {
 
 using ActsToTracksAlgorithm =
     algorithms::Algorithm<
-      algorithms::Input<edm4eic::Measurement2DCollection, std::vector<ActsExamples::Trajectories>>,
-      algorithms::Output<edm4eic::TrajectoryCollection, edm4eic::TrackParametersCollection, edm4eic::TrackCollection>
+      algorithms::Input<
+        edm4eic::Measurement2DCollection,
+        std::vector<ActsExamples::Trajectories>,
+        std::optional<edm4eic::MCRecoTrackerHitAssociationCollection>
```
bschmookler, line 25: Why do you use std::optional here, since you always pass in the "CentralTrackingRawHitAssociations" and the association between the raw and sim hits should exist for each raw hit?

wdconinc: The associations only exist for simulation input. We want algorithms to work for non-simulation input too.

bschmookler: Got it. Thanks.

src/algorithms/tracking/ActsToTracks.cc
```diff
@@ -185,6 +207,22 @@ void ActsToTracks::process(const Input& input, const Output& output) const {
 
       });
 
+      // Store track associations if hit associations provided
+      // FIXME: not able to check whether optional inputs were provided
+      //if (raw_hit_assocs->has_value()) {
+        double total_weight = std::accumulate(
+          mcparticle_weight_by_hit_count.begin(), mcparticle_weight_by_hit_count.end(),
+          0, [](const double sum, const auto& i) { return sum + i.second; });
+        for (const auto& [mcparticle, weight] : mcparticle_weight_by_hit_count) {
```
bschmookler, line 216: So, if a given track has measurement hits coming from multiple MCParticles, you create an association with each 
of these MCParticle; not just the MCParticle with the largest weight. This saves more information, which is good, but we'll need to  inform users that this is what the association contains.

wdconinc: This is how all associations are done.

bschmookler (approved):
Looks good to me

veprbl (approved):
Everything looks good to me.

wdconinc:
I propose to ignore sonarcloud since I don't see why there is any duplicated code here...

veprbl (approved):

PR #1565, opened 2024-08-13
Moved exceptions from init and added debug statements to process

simonge:
Highlighted by @ShujieL and @wdconinc [https://chat.epic-eic.org/main/pl/uiei8z865tbdjn4mpou19o4tuo](https://chat.epic-eic.org/main/pl/uiei8z865tbdjn4mpou19o4tuo). Failing to initialise the FarDetectorTrackerCluster algorithm, due to missing detectors in the geometry, calls an exception meaning JANA tries to init it again on the next event resulting in errors printed for each event.

This PR tries to tidy up the output by removing the exception so that the initialisation error is only printed once. A check is instead moved to the start of process() where repeated warning messages can still be accessed by setting the logger to debug.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

NA
Fewer repeated error statements should be printed. Otherwise no.

### DIFF
--- a/src/algorithms/fardetectors/FarDetectorMLReconstruction.cc
+++ b/src/algorithms/fardetectors/FarDetectorMLReconstruction.cc
@@ -59,7 +59,7 @@ namespace eicrecon {
     if(!m_initialized){
       debug("Initialization did not complete");
       return;
-    }    
+    }
 
     //Set beam energy from first MCBeamElectron, using std::call_once
     std::call_once(m_initBeamE,[&](){
--- a/src/algorithms/fardetectors/FarDetectorMLReconstruction.cc
+++ b/src/algorithms/fardetectors/FarDetectorMLReconstruction.cc
@@ -32,10 +32,10 @@ namespace eicrecon {
 
     // Locate and load the weight file
     // TODO - Add functionality to select passed by configuration
-    bool methodFound = false;
     if(!m_cfg.modelPath.empty()){
       try{
         m_method = dynamic_cast<TMVA::MethodBase*>(m_reader->BookMVA( m_cfg.methodName, m_cfg.modelPath ));
+        m_initialized = true;
       }
       catch(std::exception &e){
         error(fmt::format("Failed to load method {} from file {}: {}", m_cfg.methodName, m_cfg.modelPath, e.what()));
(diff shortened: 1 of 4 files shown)


wdconinc (commented):
You could use `m_method == nullptr` in `FarDetectorMLReconstruction` and `m_id_dec == nullptr` in `FarDetectorTrackerCluster` instead of adding a new initialization variable. I'll defer to @veprbl if that's better style or not :-)

src/algorithms/fardetectors/FarDetectorMLReconstruction.cc
```diff
@@ -55,6 +55,12 @@ namespace eicrecon {
     const auto [inputTracks,beamElectrons] = input;
     auto [outputFarDetectorMLTrajectories, outputFarDetectorMLTrackParameters, outputFarDetectorMLTracks] = output;
 
+    // Check if the algorithm has been initialized properly
+    if(!m_initialized){
+      debug("Initialization did not complete");
+      return;
```
veprbl, line 61: Unless you throw here, the Process will be called over and over.
```suggestion
      throw std::runtime_exception("Initialization did not complete");
```

veprbl:
This solves a global issue locally. We should figure out why exceptions in `Init` are swallowed silently. I like that this removes use of JException in algorithms, we should take this in, IMO.

src/algorithms/fardetectors/FarDetectorMLReconstruction.cc
simonge: Will it not call process over and over anyway? I thought the behaviour we wanted was to flag the error but still return an empty collection so that the reconstruction can be carried out with the limited geometry provided. It would be nicer if upon throwing the exception in init, JANA knows not to try again.

veprbl: It will not. The exceptions in Process trickle into our JEventProcessorPODIO, where we try to print error messages only once, which is the desired effect. We could do some accounting to disable collections, but, you are right, they don't get disabled from being called currently.

ruse-traveler:
Hey @simonge, just wanted to check-in on this: is this something you'd like to continue pursuing, or has it been superseded another PR?

veprbl:
If we do, I believe, it might be worth investigating if exception handling for Init may have changed in recent JANA2.

ruse-traveler:
Discussed this in the [08.17.2026 Reco WG meeting](https://indico.bnl.gov/event/33693/). We'll close this for now, but can reopen if needed!

PR #1566, opened 2024-08-13
MatrixTransferStatic only initalizes matrix once

simonge:
From comments https://chat.epic-eic.org/main/pl/qo3sdqkdgibijpetrhxocoio5e this should result in the initialization of the transfer matrix only being carried out once so the logger won't flag up errors every event.

A couple of slight functional changes which will need checking and possibly reverting:

1. The nominal momentum for selecting the correct matrix is taken from only the first beam/scattered proton in the collection rather than the average. 
2. The nominal momentum is set to the configuration momentum chosen rather than remaining as the average of the MC collection values.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

### DIFF
--- a/src/algorithms/fardetectors/MatrixTransferStatic.cc
+++ b/src/algorithms/fardetectors/MatrixTransferStatic.cc
@@ -22,6 +22,12 @@
 
 void eicrecon::MatrixTransferStatic::init() {
 
+  m_nomMomentum           = m_cfg.nomMomentum; //extract the nominal value first -- will be overwritten by MCParticle
+  m_local_x_offset        = m_cfg.local_x_offset;
+  m_local_y_offset        = m_cfg.local_y_offset;
+  m_local_x_slope_offset  = m_cfg.local_x_slope_offset;
+  m_local_y_slope_offset  = m_cfg.local_y_slope_offset;
+  
 }
 
 void eicrecon::MatrixTransferStatic::process(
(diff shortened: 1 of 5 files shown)


ajentsch (commented):
Okay, so I wasn't expecting a major change right now - this is not what was discussed in the Mattermost. 

Since this code is used for 50% of the FF detectors, I really do not want major changes handled this way. We need to have a discussion about what changes will be made, and why before this is coded up and merged. It makes it very difficult for me if I have to read through new code blocks without knowing the purpose of the changes, and how they might affect future updates.

I wrote this code the way I did to handle the various MC files we deal with (not all MC files handle beam particles the same way, or even properly), and to handle particle gun input (hence the mcparts->size() == 1 conditional). We have to have flexibility for these different options - this flexibility exists for the tested items, and has been verified to work.

I am unfamiliar with the BeamProtons and ScatteredProtons - this seems to only be something which will work if you have a proton beam - what happens when you have a spectator from e + d or e + He3? What about a coherent final state like He4?

Using MCParticles allows us to easily adjust the algorithm to add extra conditions, and simply add the PDG code for the check - using custom classes for a beam particle seems less flexible, and adds more complication to tracking down a simple problem.

ajentsch:
Additionally, the reason I did not simply use the first particle in the list is because of the beam effects in MCParticles. If I just took the first particle, there's a chance that single particle is outside of the spec for the nominal beam we expect - taking the average over all of the beam protons smoothed out any outliers and ensured that we never threw away an entire set of events. That for loop over the particles does not take any time at all - the RP reconstructions is only ever eventually processing and storing 1, or at most 2 tracks.

simonge:
I wasn't proposing to implement this without a discussion and would be happy to remove aspects related to separating the proton from the MCParticles. Otherwise only very little has been changed and it was just meant to be a refactoring rather than a major update.

Longer term, more algorithms might depend on the identification of specific MCParticles so having them defined in a central location makes sense so it will actually be easier to limit any future discrepancies and can be expanded to handle other cases.

ajentsch:
That's fine - I think for now just fixing the issue that was mentioned on the Mattermost (making the algorithm quieter) is totally fine. I think changing the extraction of the beam particle should be discussed at a meeting - perhaps at a combined FF/FB meeting, or maybe even in a reconstruction working group meeting, since it will affect many people.

simonge:
Should the nominal momentum be set to the value that has been identified by the filter rather than remain as the value calculated from the MCParticles? Maybe that should be a separate bugfix.

ajentsch:
Yes, that's a good point. The presently-stored value is that average number used to determine which matrix is needed, which is not the correct number to reconstruct the full momentum with that particular matrix.

ruse-traveler:
Hey @simonge and @ajentsch, just wanted to check-in on this: is this an effort that would be good to revisit, or has this been addressed elsewhere?

ruse-traveler:
Discussed in the [08.17.2026 Reco WG meeting](https://indico.bnl.gov/event/33693/). We'll close for now, but can reopen if needed!

PR #1568, opened 2024-08-14
CalorimeterIslandCluster: fix incorrect distance parameter priorities

veprbl:
A bug that was introduced when EICrecon code was produced from copy-pasting code from Juggler. A `uprops` variable that set option priority used to be an `std::vector`
https://eicweb.phy.anl.gov/EIC/juggler/-/blob/main/JugReco/src/components/CalorimeterIslandCluster.cpp#L213
but later became an `std::map`
https://github.com/eic/EICrecon/commit/de2f4896ff04af17dff7e2a96e0aee7485005e34#diff-57b413772a7a45d51cd383a3e763c8f739f1783ff05c234a7fdd130dd23d3045R88
hence the priority ordering became platform-dependent.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Need to check

Yes

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
+++ b/src/algorithms/calorimetry/CalorimeterIslandCluster.cc
@@ -109,7 +109,7 @@ void CalorimeterIslandCluster::init() {
       return true;
     };
 
-    std::map<std::string, std::vector<double>> uprops{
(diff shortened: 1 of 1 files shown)


ruse-traveler (approved):
Ah! That's a subtle one! Good catch!

PR #1572, opened 2024-08-16
InclusiveKinematicseSigma -> InclusiveKinematicsESigma

veprbl:
This helps to highlight a difference to electron $\Sigma$ method vs $\Sigma$ method to avoid confusion (there were questions on mattermost).

### DIFF
--- a/src/algorithms/reco/InclusiveKinematicsESigma.cc
+++ b/src/algorithms/reco/InclusiveKinematicsESigma.cc
@@ -17,17 +17,17 @@
 
 #include "Beam.h"
 #include "Boost.h"
-#include "InclusiveKinematicseSigma.h"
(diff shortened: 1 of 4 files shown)


wdconinc (approved):
Fine with me :-)

Related, if we ever have `ePIC` in a CamelCase context, will it become `EPIC`? ;-)

wdconinc:
SonarCloud doesn't make it easy to find where the duplication is, https://sonarcloud.io/code?id=eic_EICrecon&pullRequest=1572&selected=eic_EICrecon%3Asrc%2Falgorithms%2Freco%2FInclusiveKinematicsESigma.cc ...

It's not wrong but it is also out of scope here. At some point we should create an InclusiveKinematics base class that does the common first part.

Issue #1575, opened 2024-08-16
Vertex output missing associated ReconstructedParticles


starsdong:
**Is your feature request related to a problem? Please describe.**
As stated in the title

**Describe the solution you'd like**
Add ReconstructedParticle as input to IVF. The links between ReconstructedParticles to Acts vertices are not explicit. Will compare trajectory parameters to create the link

**Describe alternatives you've considered**

**Additional context**

PR #1576, opened 2024-08-16
IterativeVertexFinder: fill vertex-to-particle relation

starsdong:
- added association for ReconstructionParticles to output vertices

- [ ] Bug fix (issue #__)
- [X] New feature (issue #1575)
- [ ] Documentation update
- [ ] Other: __

- ReconstructedParticle association added to the default output CKFCentralVertices


src/global/tracking/IterativeVertexFinder_factory.h
```diff
@@ -26,6 +27,7 @@ class IterativeVertexFinder_factory :
     std::unique_ptr<AlgoT> m_algo;
 
     Input<ActsExamples::Trajectories> m_acts_trajectories_input {this};
+    Input<edm4eic::ReconstructedParticle> m_edm4eic_reconParticles_input {this, "ReconstructedParticles"};
```
veprbl, line 30:
```suggestion
    PodioInput<edm4eic::ReconstructedParticle> m_edm4eic_reconParticles_input {this};
```
Not sure how a `const char*` converts to `InputOptions`, but it seems like this second argument is not even needed.

src/algorithms/tracking/IterativeVertexFinder.cc
```diff
@@ -26,16 +30,20 @@
...
+#include <Acts/Vertexing/TrackAtVertex.hpp>
 #include <Acts/Vertexing/Vertex.hpp>
 #include <Acts/Vertexing/VertexingOptions.hpp>
 #include <Acts/Vertexing/ZScanVertexFinder.hpp>
 #include <ActsExamples/EventData/Trajectories.hpp>
 #include <boost/container/vector.hpp>
 #include <edm4eic/Cov4f.h>
-#include <math.h>
+#include <edm4eic/TrackParameters.h>
+#include <edm4eic/Trajectory.h>
```
veprbl, line 41:
```suggestion
#include <edm4eic/Trajectory.h>
#include <edm4eic/unit_system.h>
```

src/algorithms/tracking/IterativeVertexFinder.cc
```diff
@@ -207,7 +220,36 @@ std::unique_ptr<edm4eic::VertexCollection> eicrecon::IterativeVertexFinder::prod
...
+      float loc_a = par.localPosition().x();
+      float loc_b = par.localPosition().y();
+
+      for (const auto part : reconParticles) {
+        const auto& tracks = part->getTracks();
+        for (const auto trk : tracks) {
+          const auto& traj = trk.getTrajectory();
+          const auto& trkPars = traj.getTrackParameters();
+          for (const auto par : trkPars) {
+            if(fabs(par.getLoc().a - loc_a) < 1.e-4 && fabs(par.getLoc().b - loc_b) < 1.e-4) {
```
veprbl, line 241:
```suggestion
            constexpr acts_to_edm4eic = edm4eic::unit::mm / Acts::UnitConstants::mm;
            if(fabs(par.getLoc().a - loc_a * acts_to_edm4eic) < 1.e-4 && fabs(par.getLoc().b - loc_b * acts_to_edm4eic) < 1.e-4) {
```

src/global/tracking/tracking.cc
```diff
@@ -226,7 +226,7 @@ void InitPlugin(JApplication *app) {
 
     app->Add(new JOmniFactoryGeneratorT<IterativeVertexFinder_factory>(
             "CentralTrackVertices",
-            {"CentralCKFActsTrajectories"},
+            {"CentralCKFSeededActsTrajectories","ReconstructedSeededChargedParticles"},
```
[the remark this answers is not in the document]
starsdong: Dmitri, we have shown the realistic seeded trajectories deliver good performance. And I believe we agreed in the collaboration to move to the Seeded objects for these downstream analysis.

src/global/tracking/IterativeVertexFinder_factory.h
starsdong, line 1: Dmitry, this changes break the compilation. I guess your proposal may work, but that would require also changes in the IterativeVertexFinder.cc to make the corresponding changes. For simplicity, shall we revert the commit and keep the original version in the PR?

src/algorithms/tracking/IterativeVertexFinder.cc
```diff
@@ -56,7 +65,8 @@ void eicrecon::IterativeVertexFinder::init(std::shared_ptr<const ActsGeometryPro
 }
 
 std::unique_ptr<edm4eic::VertexCollection> eicrecon::IterativeVertexFinder::produce(
-    std::vector<const ActsExamples::Trajectories*> trajectories) {
+    std::vector<const ActsExamples::Trajectories*> trajectories,
+    std::vector<const edm4eic::ReconstructedParticle*> reconParticles) {
```
wdconinc, line 69: This is what needs changing to fix your compilation. We should not introduce new interfaces that use `std::vecotr<podio_type>`.

wdconinc (changes requested):

src/global/tracking/tracking.cc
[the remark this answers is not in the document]
veprbl: Yes, but we should move centrally. The plan is to rename `ReconstructedSeededChargedParticles` to `ReconstructedChargedParticles` and rename the original `ReconstructedChargedParticles` to something else, and leave that for dedicated tracking studies.

src/algorithms/tracking/IterativeVertexFinder.cc
starsdong: addressed in the following commit

veprbl:
image
Associations are filled.

veprbl (dismissed):
LGTM

starsdong:
Hi Woulter and Dmitry,

Are these anything you would like to comment before merging this PR? The requested change Woulter brought up, I believe, is addressed in the later commits. Would appreciate if you can take a review to see whether we can proceed. Thanks

/xin

veprbl:
I don't understand why this hasn't merged over the weekend. Maybe we needed to have and approval form @wdconinc or dismiss his "1 change requested" review.

src/algorithms/tracking/IterativeVertexFinder.cc
```diff
@@ -207,7 +222,37 @@ std::unique_ptr<edm4eic::VertexCollection> eicrecon::IterativeVertexFinder::prod
...
+      float loc_b = par.localPosition().y();
+
+      for (const auto& part : *reconParticles) {
+        const auto& tracks = part.getTracks();
+        for (const auto trk : tracks) {
+          const auto& traj = trk.getTrajectory();
+          const auto& trkPars = traj.getTrackParameters();
+          for (const auto par : trkPars) {
+            constexpr double acts_to_edm4eic = edm4eic::unit::mm / Acts::UnitConstants::mm;
+            if(fabs(par.getLoc().a - loc_a * acts_to_edm4eic) < 1.e-4 && fabs(par.getLoc().b - loc_b * acts_to_edm4eic) < 1.e-4) {
```
wdconinc, line 244: Avoid dimensionful magic numbers.

wdconinc (approved):
The 'request changes' is always sticky and requires the same person who requested changes to give explicit approval. In this case, I requested changes since we needed to use the podio collection instead of std::vector<podio type>. That's been done now, so I approve.

One more comment though on a magic number that should be a configuration variable (best) or defined as a ratio of another dimensionful quantity. The reconstruction should not care whether we work in dd4hep units or, for some reason, another unit system.

src/algorithms/tracking/IterativeVertexFinder.cc
veprbl: Addressed. Let's assume that we don't need to change this value. The algorithm depends on numerical tolerance, and it will be replaced with another approach once that's supported by the EDM.

src/algorithms/tracking/IterativeVertexFinder.cc
```diff
@@ -207,7 +222,38 @@ std::unique_ptr<edm4eic::VertexCollection> eicrecon::IterativeVertexFinder::prod
          (float)vtx.time(),
     }); // vtxposition
     eicvertex.setPositionError(cov);                          // covariance
-  }
+
+    for (const auto& t : vtx.tracks()) {
+#if Acts_VERSION_MAJOR >= 33
+      const auto& trk = &t.originalParams;
+      const auto& par = finderCfg.extractParameters(trk);
```
wdconinc, line 229: Fails to compile for acts v33, with
```
#32 1016.3 /tmp/root/spack-stage/spack-stage-eicrecon-git.7f8feba7572a7b05cab02027929cd08659b1cdbd_main-xib6spjd5y4zv24smqpnu3i6uxgmhggo/spack-src/src/algorithms/tracking/IterativeVertexFinder.cc:229:25: error: no matching function for call to object of type 'InputTrack::Extractor' (aka 'Delegate<GenericBoundTrackParameters<ParticleHypothesis> (const Acts::InputTrack &)>')
#32 1016.3       const auto& par = finderCfg.extractParameters(trk);
#32 1016.3                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
#32 1016.3 /opt/software/linux-debian12-x86_64_v2/clang-16.0.6/acts-33.1.0-eihfjd5ywurf5qidbrexfxoreo7yj4nx/include/Acts/Utilities/Delegate.hpp:249:15: note: candidate function not viable: no known conversion from 'const InputTrack *const' to 'const Acts::InputTrack' for 1st argument; dereference the argument with *
#32 1016.3   return_type operator()(Args... args) const {
#32 1016.3               ^
#32 1016.3 1 error generated.
#32 1016.3 make[3]: *** [src/algorithms/tracking/CMakeFiles/algorithms_tracking_library.dir/build.make:163: src/algorithms/tracking/CMakeFiles/algorithms_tracking_library.dir/IterativeVertexFinder.cc.o] Error 1
```

wdconinc: Ref https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/3746414

wdconinc: #1610

PR #1577, opened 2024-08-16
reco.cc: add InclusiveKinematicseSigma backwards compatibility for now

veprbl:
Provide a graceful deprecation for https://github.com/eic/EICrecon/pull/1572


src/global/reco/reco.cc
```diff
@@ -171,6 +171,18 @@ void InitPlugin(JApplication *app) {
         app
     ));
 
+    // InclusiveKinematicseSigma is deprecated and will be removed, use InclusiveKinematicsESigma instead
+    app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::HadronicFinalState>>(
```
wdconinc, line 175:
```suggestion
    app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::InclusiveKinematics>>(
```

wdconinc:
What is this? Windows?

wdconinc (dismissed):

veprbl:
> What is this? Windows?

Dot files are not converting, there might be a regex issue.

wdconinc:
Offending dot file has a subgraph cluster_18 that's just empty... But I think what really fixes this for me is upgrading graphviz. I'll see to push that out to nightly.

wdconinc:
> Offending dot file has a subgraph cluster_18 that's just empty....

Must have always been empty due to LFHCAL duplicate naming:
https://github.com/eic/EICrecon/blob/4a2e46e3b688c9818cea6a1c4eb247312d176cfc/.github/janadot.groups#L14

veprbl:
No cigar

wdconinc:
> No cigar

Yeah, I'd restarted some jobs which should have picked up a new graphviz, and no luck.

Maybe we need to make the graphviz job to allow failure. And figure out the issue on the side.

wdconinc (approved):

Issue #1573, opened 2024-08-16
`HcalFarForwardZDCImagingProtoClusters` is slow


veprbl:
Per 1000 events:
```
39.81 s (  1.1%) edm4eic::CalorimeterHit:HcalFarForwardZDCSubcellHits
3538.76 s ( 98.7%) edm4eic::ProtoCluster:HcalFarForwardZDCImagingProtoClusters
```
https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/3651688#L4721

cc @sebouh137

wdconinc:
I guess I was a day late in realizing this. I'll remove `HcalFarForwardZDCImagingProtoClusters` for now.

PR #1581, opened 2024-08-17
fix: introduce sorted index proxy to allow early break on neighbor search

wdconinc:
This PR originally intended to add hit sorting based on layer to achieve early breaks in the neighbor searching (i.e. if neighbors have to be at most two layers different in a ZDC-like detector, then in a sorted list we can stop evaluating for neighbors as soon as we encounter the first hit with more than two layers difference).

This PR also makes several other changes to the topological clustering.
- Hits are now referred to by an index proxy ordered set. This set (with ordering by layer and then index in collection) is modified as hits are assigned to clusters, such that only eligible hits remain in the set (once ordered, erasure does of course not affect ordering). The need for a separate vector which keeps track of already assigned hits is avoided. As more hits are assigned, the remaining search space is reduced, allowing for more efficient searching.
- Clusters (groups) are now stored in a vector ordered by insertion sequence instead of a set (which is ordered but not in a useful way). This allows us to avoid rechecking already added hits for neighbors, and only check newly added hits for new neighbors to add.

When modifying containers that we are iterating over, we have to make sure we do not invalidate iterators. Note https://en.cppreference.com/w/cpp/container#Iterator_invalidation. A vector container may change capacity, which invalidates all iterators. Iterators on a set are only invalidated when the element pointed to is erased. Erasure returns an iterator to the next element. Therefore, we postpone erasure of the starting hit of each group until after the entire group has been collected (and we have to make sure to avoid picking it again as a hit). If we erased the starting hit right away, the next iterator returned by the erase operation would likely become invalid when the hit is (likely, due to ordering) used in the cluster.

- [x] Bug fix (issue #1573)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -147,7 +147,7 @@ namespace eicrecon {
         }
 
         // Group neighbouring hits
-        std::vector<std::vector<std::size_t>> groups;
+        std::vector<std::list<std::size_t>> groups;
         // because indices changes, the loop over indices requires some care:
         // - we must use iterators instead of range-for
         // - erase returns an incremented iterator and therefore acts as idx++
@@ -175,7 +175,7 @@ namespace eicrecon {
             }
 
             // create a new group, and group all the neighbouring hits
-            groups.emplace_back(std::vector{*idx});
+            groups.emplace_back(std::list{*idx});
             bfs_group(*hits, indices, groups.back(), *idx);
 
             // wait with erasing until after bfs_group to ensure iterator is not invalidated in bfs_group
@@ -241,7 +241,7 @@ namespace eicrecon {
     // grouping function with Breadth-First Search
     // note: template to allow Compare only known in local scope of caller
     template<typename Compare>
-    void bfs_group(const edm4eic::CalorimeterHitCollection &hits, std::set<std::size_t,Compare>& indices, std::vector<std::size_t> &group, const std::size_t idx) const {
+    void bfs_group(const edm4eic::CalorimeterHitCollection &hits, std::set<std::size_t,Compare>& indices, std::list<std::size_t> &group, const std::size_t idx) const {
 
       // not a qualified hit to participate clustering, stop here
       if (hits[idx].getEnergy() < m_cfg.minClusterHitEdep) {
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -127,6 +127,9 @@ namespace eicrecon {
 
         // Sort hit indices (podio collections do not support std::sort)
         auto compare = [&hits](const auto& a, const auto& b) {
+            if ((*hits)[a].getLayer() == (*hits)[b].getLayer()) {
+              return (*hits)[a].getCellID() < (*hits)[b].getCellID();
+            }
             return (*hits)[a].getLayer() < (*hits)[b].getLayer();
         };
         std::set<std::size_t, decltype(compare)> indices(compare);
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -150,8 +150,8 @@ namespace eicrecon {
         std::vector<std::vector<std::size_t>> groups;
         // because indices changes, the loop over indices requires some care:
         // - we must use iterators instead of range-for
-        // - erase returns an increment iterator and therefore acts as idx++
-        // - when the set becomes empty on erase, idx is invalid and idx++ will fail
+        // - erase returns an incremented iterator and therefore acts as idx++
+        // - when the set becomes empty on erase, idx is invalid and idx++ will be too
         // (also applies to loop in bfs_group below)
         for (auto idx = indices.begin(); idx != indices.end();
                  indices.empty() ? idx = indices.end() : idx) {
@@ -162,13 +162,13 @@ namespace eicrecon {
                          (*hits)[*idx].getEnergy()
             );
 
-            // not energetic enough for cluster center or other cluster hit
+            // not energetic enough for cluster center or other cluster hits, so erase
             if ((*hits)[*idx].getEnergy() < std::min(m_cfg.minClusterHitEdep,m_cfg.minClusterCenterEdep)) {
               idx = indices.erase(idx);
               continue;
             }
 
-            // not energetic enough for cluster center (could still be cluster hit)
+            // not energetic enough for cluster center, but could still be cluster hit
             if ((*hits)[*idx].getEnergy() < minClusterCenterEdep) {
               idx++;
               continue;
@@ -178,7 +178,7 @@ namespace eicrecon {
             groups.emplace_back(std::vector{*idx});
             bfs_group(*hits, indices, groups.back(), *idx);
 
-            // wait with erasing until after bfs_group to ensure valid iterator is returned
+            // wait with erasing until after bfs_group to ensure iterator is not invalidated in bfs_group
             idx = indices.erase(idx); // takes role of idx++
         }
         debug("found {} potential clusters (groups of hits)", groups.size());
@@ -239,6 +239,7 @@ namespace eicrecon {
     }
 
     // grouping function with Breadth-First Search
+    // note: template to allow Compare only known in local scope of caller
     template<typename Compare>
     void bfs_group(const edm4eic::CalorimeterHitCollection &hits, std::set<std::size_t,Compare>& indices, std::vector<std::size_t> &group, const std::size_t idx) const {
 
@@ -247,6 +248,7 @@ namespace eicrecon {
         return;
       }
 
+      // keep track of size as group grows
       std::size_t prev_size = 0;
       while (prev_size != group.size()) {
         auto prev_group = group;
@@ -256,6 +258,7 @@ namespace eicrecon {
               indices.empty() ? idx2 = indices.end() : idx2) {
 
             // skip idx1 and original idx
+            // (we cannot erase idx since it would invalidate iterator in calling scope)
             if (*idx2 == *idx1 || *idx2 == idx) {
               idx2++;
               continue;
@@ -267,6 +270,7 @@ namespace eicrecon {
             //}
 
             // not energetic enough for cluster center or other cluster hit
+            // whereas caller has removed earlier low energy hits, this removes ones that caller hasn't gotten to yet
             if (hits[*idx2].getEnergy() < std::min(m_cfg.minClusterHitEdep,m_cfg.minClusterCenterEdep)) {
               idx2 = indices.erase(idx2);
               continue;
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -162,12 +162,6 @@ namespace eicrecon {
                          (*hits)[*idx].getEnergy()
             );
 
-            // not energetic enough for cluster center or other cluster hits, so erase
-            if ((*hits)[*idx].getEnergy() < std::min(m_cfg.minClusterHitEdep,m_cfg.minClusterCenterEdep)) {
-              idx = indices.erase(idx);
-              continue;
-            }
-
             // not energetic enough for cluster center, but could still be cluster hit
             if ((*hits)[*idx].getEnergy() < minClusterCenterEdep) {
               idx++;
@@ -268,7 +262,7 @@ namespace eicrecon {
 
           // not energetic enough for cluster center or other cluster hit
           // whereas caller has removed earlier low energy hits, this removes ones that caller hasn't gotten to yet
-          if (hits[*idx2].getEnergy() < std::min(m_cfg.minClusterHitEdep,m_cfg.minClusterCenterEdep)) {
+          if (hits[*idx2].getEnergy() < std::min(m_cfg.minClusterHitEdep, m_cfg.minClusterCenterEdep)) {
             idx2 = indices.erase(idx2);
             continue;
           }
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -128,8 +128,8 @@ namespace eicrecon {
         // Sort hit indices (podio collections do not support std::sort)
         std::set<std::size_t> indices(hits->size());
         std::iota(indices.begin(), indices.end(), 0);
-        std::sort(
-            indices.begin(), indices.end(),
+        std::ranges::sort(
+            indices,
             [&hits](const auto& a, const auto& b) {
               return (*hits)[a].getLayer() < (*hits)[b].getLayer();
             });
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -66,7 +66,7 @@ namespace eicrecon {
 
     // unitless counterparts of the input parameters
     double localDistXY[2]{0, 0}, layerDistEtaPhi[2]{0, 0}, layerDistXY[2]{0, 0}, sectorDist{0};
-    double minClusterHitEdep{0}, minClusterCenterEdep{0}, minClusterEdep{0}, minClusterNhits{0};
+    double minClusterHitEdep{0}, minClusterCenterEdep{0}, minClusterEdep{0};
 
   public:
     void init() {
@@ -166,7 +166,7 @@ namespace eicrecon {
 
         // form clusters
         for (const auto &group : groups) {
-            if (static_cast<int>(group.size()) < m_cfg.minClusterNhits) {
+            if (group.size() < m_cfg.minClusterNhits) {
                 continue;
             }
             double energy = 0.;
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -127,6 +127,8 @@ namespace eicrecon {
 
         // Sort hit indices (podio collections do not support std::sort)
         auto compare = [&hits](const auto& a, const auto& b) {
+            // if !(a < b) and !(b < a), then a and b are equivalent
+            // and only one of them will be allowed in a set
             if ((*hits)[a].getLayer() == (*hits)[b].getLayer()) {
               return (*hits)[a].getCellID() < (*hits)[b].getCellID();
             }
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -84,6 +84,10 @@ namespace eicrecon {
             error( "Expected 2 values (eta_dist, phi_dist) for layerDistEtaPhi" );
             return;
         }
+        if (m_cfg.minClusterCenterEdep < m_cfg.minClusterHitEdep) {
+            error( "minClusterCenterEdep must be greater than or equal to minClusterHitEdep" );
+            return;
+        }
 
         // using juggler internal units (GeV, dd4hep::mm, dd4hep::ns, dd4hep::rad)
         localDistXY[0] = m_cfg.localDistXY[0] / dd4hep::mm;
@@ -241,11 +245,6 @@ namespace eicrecon {
     template<typename Compare>
     void bfs_group(const edm4eic::CalorimeterHitCollection &hits, std::set<std::size_t,Compare>& indices, std::list<std::size_t> &group, const std::size_t idx) const {
 
-      // not a qualified hit to participate clustering, stop here
-      if (hits[idx].getEnergy() < m_cfg.minClusterHitEdep) {
-        return;
-      }
-
       // loop over group as it grows, until the end is stable and we reach it
       for (auto idx1 = group.begin(); idx1 != group.end(); ++idx1) {
         // check neighbours (note comments on loop over set above)
@@ -264,16 +263,9 @@ namespace eicrecon {
           //  break;
           //}
 
-          // not energetic enough for cluster center or other cluster hit
-          // whereas caller has removed earlier low energy hits, this removes ones that caller hasn't gotten to yet
-          if (hits[*idx2].getEnergy() < std::min(m_cfg.minClusterHitEdep, m_cfg.minClusterCenterEdep)) {
-            idx2 = indices.erase(idx2);
-            continue;
-          }
-
-          // not energetic enough for cluster hit in this group
+          // not energetic enough for cluster hit
           if (hits[*idx2].getEnergy() < m_cfg.minClusterHitEdep) {
-            idx2++;
+            idx2 = indices.erase(idx2);
             continue;
           }
 
--- a/src/algorithms/calorimetry/ImagingTopoCluster.h
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.h
@@ -137,8 +137,7 @@ namespace eicrecon {
         // group neighbouring hits
         std::vector<bool> visits(hits->size(), false);
         std::vector<std::set<std::size_t>> groups;
-        for (std::size_t i = 0; i < indices.size(); ++i) {
-            std::size_t idx = indices[i];
+        for (const auto& idx : indices) {
 
             debug("hit {:d}: local position = ({}, {}, {}), global position = ({}, {}, {})", idx + 1,
                          (*hits)[idx].getLocal().x, (*hits)[idx].getLocal().y, (*hits)[idx].getPosition().z,
@@ -225,8 +224,7 @@ namespace eicrecon {
         prev_size = group.size();
         for (std::size_t idx1 : group) {
           // check neighbours
-          for (std::size_t i2 = 0; i2 < indices.size(); ++i2) {
-            std::size_t idx2 = indices[i2];
+          for (const auto& idx2 : indices) {
 
             // skip rest of list of hits when we're past relevant layers
             //if (hits[idx2].getLayer() - hits[idx1].getLayer() > m_cfg.neighbourLayersRange) {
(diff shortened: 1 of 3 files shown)


wdconinc:
Diff looks sensible. Clusters are found in different order so hits_begin and hits_end are different. Clusters parameters are the same.

Curiously no change at all for BIC.

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,32 @@ namespace eicrecon {
...
+            });
+
         // group neighbouring hits
         std::vector<bool> visits(hits->size(), false);
         std::vector<std::set<std::size_t>> groups;
-        for (size_t i = 0; i < hits->size(); ++i) {
+        for (std::size_t i = 0; i < indices.size(); ++i) {
+            std::size_t idx = indices[i];
+
             debug("hit {:d}: local position = ({}, {}, {}), global position = ({}, {}, {})", i + 1,
```
wdconinc, line 143:
```suggestion
             debug("hit {:d}: local position = ({}, {}, {}), global position = ({}, {}, {})", idx + 1,
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,32 @@ namespace eicrecon {
         const auto [hits] = input;
         auto [proto] = output;
 
+        // Sort hit indices (podio collections do not support std::sort)
+        std::vector<std::size_t> indices(hits->size());
```
wdconinc, line 129:
```suggestion
        std::set<std::size_t> indices(hits->size());
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,32 @@ namespace eicrecon {
...
+            [&hits](const auto& a, const auto& b) {
+              return (*hits)[a].getLayer() < (*hits)[b].getLayer();
+            });
+
         // group neighbouring hits
         std::vector<bool> visits(hits->size(), false);
         std::vector<std::set<std::size_t>> groups;
-        for (size_t i = 0; i < hits->size(); ++i) {
+        for (std::size_t i = 0; i < indices.size(); ++i) {
+            std::size_t idx = indices[i];
```
wdconinc, line 141:
```suggestion
        for (const auto& idx : indices) {
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,7 +210,7 @@ namespace eicrecon {
     }
 
     // grouping function with Breadth-First Search
-    void bfs_group(const edm4eic::CalorimeterHitCollection &hits, std::set<std::size_t> &group, std::size_t idx, std::vector<bool> &visits) const {
+    void bfs_group(const edm4eic::CalorimeterHitCollection &hits, const std::vector<std::size_t>& indices, std::set<std::size_t> &group, std::size_t idx, std::vector<bool> &visits) const {
```
wdconinc, line 213: `indices` as a set can serve as `visits`.
```suggestion
    void bfs_group(const edm4eic::CalorimeterHitCollection &hits, std::vector<std::size_t>& indices, std::set<std::size_t> &group, std::size_t idx) const {
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -213,7 +225,14 @@ namespace eicrecon {
         prev_size = group.size();
         for (std::size_t idx1 : group) {
           // check neighbours
-          for (std::size_t idx2 = 0; idx2 < hits.size(); ++idx2) {
+          for (std::size_t i2 = 0; i2 < indices.size(); ++i2) {
+            std::size_t idx2 = indices[i2];
```
wdconinc, line 229:
```suggestion
          for (const auto& idx2 : indices) {
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -213,7 +225,14 @@ namespace eicrecon {
...
+
+            // skip rest of list of hits when we're past relevant layers
+            //if (hits[idx2].getLayer() - hits[idx1].getLayer() > m_cfg.neighbourLayersRange) {
+            //  break;
+            //}
+
             // not a qualified hit to participate clustering, skip
             if (hits[idx2].getEnergy() < m_cfg.minClusterHitEdep) {
```
wdconinc, line 237:
```suggestion
             if (hits[idx2].getEnergy() < m_cfg.minClusterHitEdep) {
                indices.erase(idx2);
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,7 +210,7 @@ namespace eicrecon {
     }
 
     // grouping function with Breadth-First Search
-    void bfs_group(const edm4eic::CalorimeterHitCollection &hits, std::set<std::size_t> &group, std::size_t idx, std::vector<bool> &visits) const {
+    void bfs_group(const edm4eic::CalorimeterHitCollection &hits, const std::vector<std::size_t>& indices, std::set<std::size_t> &group, std::size_t idx, std::vector<bool> &visits) const {
       visits[idx] = true;
```
wdconinc, line 214:
```suggestion
       indices.erase(idx);
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,31 @@ namespace eicrecon {
         const auto [hits] = input;
         auto [proto] = output;
 
+        // Sort hit indices (podio collections do not support std::sort)
+        std::set<std::size_t> indices(hits->size());
+        std::iota(indices.begin(), indices.end(), 0);
+        std::sort(
+            indices.begin(), indices.end(),
```
wdconinc, line 132:
```suggestion
        std::ranges::sort(
            indices,
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,31 @@ namespace eicrecon {
         const auto [hits] = input;
         auto [proto] = output;
 
+        // Sort hit indices (podio collections do not support std::sort)
+        std::set<std::size_t> indices(hits->size());
```
wdconinc, line 129:
```suggestion
        std::set<std::size_t> indices(
            [&hits](const auto& a, const auto& b) {
              return (*hits)[a].getLayer() < (*hits)[b].getLayer();
            });
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,31 @@ namespace eicrecon {
...
+        std::iota(indices.begin(), indices.end(), 0);
+        std::ranges::sort(
+            indices,
+            [&hits](const auto& a, const auto& b) {
+              return (*hits)[a].getLayer() < (*hits)[b].getLayer();
+            });
```
wdconinc, line 135:
```suggestion
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,34 @@ namespace eicrecon {
         const auto [hits] = input;
         auto [proto] = output;
 
+        // Sort hit indices (podio collections do not support std::sort)
+        auto compare = [&hits](const auto& a, const auto& b) {
```
wdconinc, line 129:
```suggestion
        auto compare = [&hits](const auto& a, const auto& b) {
            // if !(a < b) and !(b < a), a and b are equivalent
            // and only one of them will be allowed in a set
```

wdconinc:
No diffs with `set`.

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: Implemented.

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: Implemented.

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,61 @@ namespace eicrecon {
...
+        // ensure no hits were dropped due to equivalency in set
+        if (hits->size() != indices.size()) {
+            error("equivalent hits were dropped: #hits {:d}, #indices {:d}", hits->size(), indices.size());
+        }
+
+        // Group neighbouring hits
+        std::vector<std::vector<std::size_t>> groups;
+        // because indices changes, the loop over indices requires some care:
+        // - we must use iterators instead of range-for
+        // - erase returns an increment iterator and therefore acts as idx++
```
wdconinc, line 153:
```suggestion
        // - erase returns an incremented iterator and therefore acts as idx++
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,61 @@ namespace eicrecon {
...
+        // - when the set becomes empty on erase, idx is invalid and idx++ will fail
```
wdconinc, line 154:
```suggestion
        // - when the set becomes empty on erase, idx is invalid and idx++ will be too
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,61 @@ namespace eicrecon {
...
+            debug("hit {:d}: local position = ({}, {}, {}), global position = ({}, {}, {}), energy = {}", *idx,
+                         (*hits)[*idx].getLocal().x, (*hits)[*idx].getLocal().y, (*hits)[*idx].getPosition().z,
+                         (*hits)[*idx].getPosition().x, (*hits)[*idx].getPosition().y, (*hits)[*idx].getPosition().z,
+                         (*hits)[*idx].getEnergy()
             );
-            // already in a group, or not energetic enough to form a cluster
-            if (visits[i] || (*hits)[i].getEnergy() < minClusterCenterEdep) {
-                continue;
+
+            // not energetic enough for cluster center or other cluster hit
```
wdconinc, line 165:
```suggestion
            // not energetic enough for cluster center or other cluster hits, so erase
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,61 @@ namespace eicrecon {
...
+            if ((*hits)[*idx].getEnergy() < std::min(m_cfg.minClusterHitEdep,m_cfg.minClusterCenterEdep)) {
+              idx = indices.erase(idx);
+              continue;
+            }
+
+            // not energetic enough for cluster center (could still be cluster hit)
```
wdconinc, line 171:
```suggestion
            // not energetic enough for cluster center, but could still be cluster hit
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,61 @@ namespace eicrecon {
...
+              continue;
             }
+
             // create a new group, and group all the neighbouring hits
-            groups.emplace_back();
-            bfs_group(*hits, groups.back(), i, visits);
+            groups.emplace_back(std::vector{*idx});
+            bfs_group(*hits, indices, groups.back(), *idx);
+
+            // wait with erasing until after bfs_group to ensure valid iterator is returned
```
wdconinc, line 181:
```suggestion
            // wait with erasing until after bfs_group to ensure iterator is not invalidated in bfs_group
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,35 +239,57 @@ namespace eicrecon {
     }
 
     // grouping function with Breadth-First Search
```
wdconinc, line 241:
```suggestion
    // grouping function with Breadth-First Search
    // note: template to allow Compare only known in local scope of caller
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,35 +239,57 @@ namespace eicrecon {
...
 
       // not a qualified hit to participate clustering, stop here
       if (hits[idx].getEnergy() < m_cfg.minClusterHitEdep) {
         return;
       }
 
-      group.insert(idx);
-      size_t prev_size = 0;
-
+      std::size_t prev_size = 0;
```
wdconinc, line 250:
```suggestion
      // keep track of size as group grows
      std::size_t prev_size = 0;
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,35 +239,57 @@ namespace eicrecon {
...
       while (prev_size != group.size()) {
-        prev_size = group.size();
-        for (std::size_t idx1 : group) {
-          // check neighbours
-          for (std::size_t idx2 = 0; idx2 < hits.size(); ++idx2) {
-            // not a qualified hit to participate clustering, skip
-            if (hits[idx2].getEnergy() < m_cfg.minClusterHitEdep) {
+        auto prev_group = group;
```
wdconinc, line 252: I wonder now if we need a copy of the group. As we add hits, group will expand but idx1 is always before the insertion point and remains valid. End is always invalidated but we don't store it. At the end of each loop iteration the != end() condition can be safely evaluated. Without prev_group we'd have fewer copy operations and fewer while loop iterations, maybe only one though I can prove it right now.

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,35 +239,57 @@ namespace eicrecon {
...
+        for (auto idx1 = std::next(prev_group.begin(), prev_size); idx1 != prev_group.end(); ++idx1) {
+          // check neighbours (note comments on loop over set above)
+          for (auto idx2 = indices.begin(); idx2 != indices.end();
+              indices.empty() ? idx2 = indices.end() : idx2) {
+
+            // skip idx1 and original idx
```
wdconinc, line 258:
```suggestion
            // skip idx1 and original idx
            // (we cannot erase idx since it would invalidate iterator in calling scope)
```

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,35 +239,57 @@ namespace eicrecon {
...
+              idx2++;
+              continue;
+            }
+
+            // skip rest of list of hits when we're past relevant layers
+            //if (hits[*idx2].getLayer() - hits[*idx1].getLayer() > m_cfg.neighbourLayersRange) {
+            //  break;
+            //}
+
+            // not energetic enough for cluster center or other cluster hit
```
wdconinc, line 269:
```suggestion
            // not energetic enough for cluster center or other cluster hit
            // whereas caller has removed earlier low energy hits, this removes one that caller hasn't gotten to yet
```

wdconinc:
18x275, Q2>0
main (https://github.com/eic/EICrecon/actions/runs/10446179053/job/28923287532#step:8:6553)
```
Factories:
2.06 s (  0.4%) edm4eic::CalorimeterHit:HcalFarForwardZDCRecHits
2.71 s (  0.5%) edm4eic::ProtoCluster:LFHCALIslandProtoClusters
2.84 s (  0.5%) Acts::TrackContainer<Acts::ConstVectorTrackContainer, Acts::ConstVectorMultiTrajectory, std::shared_ptr>:CentralCKFActsTracksUnfiltered
3.38 s (  0.6%) edm4eic::Measurement2D:CentralTrackerMeasurements
5.79 s (  1.1%) edm4eic::TrackSegment:CalorimeterTrackProjections
5.86 s (  1.1%) edm4eic::CalorimeterHit:HcalFarForwardZDCSubcellHits
9.97 s (  1.9%) edm4eic::ProtoCluster:HcalEndcapPInsertImagingProtoClusters
11.71 s (  2.2%) edm4eic::ProtoCluster:HcalFarForwardZDCIslandProtoClustersBaseline
28.77 s (  5.4%) edm4eic::CherenkovParticleID:DRICHAerogelIrtCherenkovParticleID
442.57 s ( 82.9%) edm4eic::ProtoCluster:HcalFarForwardZDCImagingProtoClusters
```
b8667debcdf58a952311b06e4565af0ce932bc42 (https://github.com/eic/EICrecon/actions/runs/10446415150/job/28923931855#step:8:6570)
```
Factories:
2.03 s (  1.8%) edm4eic::ProtoCluster:HcalEndcapPInsertImagingProtoClusters
2.04 s (  1.8%) edm4eic::CalorimeterHit:HcalFarForwardZDCRecHits
2.78 s (  2.5%) edm4eic::ProtoCluster:LFHCALIslandProtoClusters
2.81 s (  2.5%) Acts::TrackContainer<Acts::ConstVectorTrackContainer, Acts::ConstVectorMultiTrajectory, std::shared_ptr>:CentralCKFActsTracksUnfiltered
3.30 s (  3.0%) edm4eic::Measurement2D:CentralTrackerMeasurements
5.70 s (  5.2%) edm4eic::CalorimeterHit:HcalFarForwardZDCSubcellHits
5.86 s (  5.3%) edm4eic::TrackSegment:CalorimeterTrackProjections
11.96 s ( 10.8%) edm4eic::ProtoCluster:HcalFarForwardZDCIslandProtoClustersBaseline
27.69 s ( 25.1%) edm4eic::ProtoCluster:HcalFarForwardZDCImagingProtoClusters
28.32 s ( 25.6%) edm4eic::CherenkovParticleID:DRICHAerogelIrtCherenkovParticleID
```

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: Implemented.

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: Implemented.

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -198,35 +239,56 @@ namespace eicrecon {
...
+          if (hits[*idx2].getEnergy() < std::min(m_cfg.minClusterHitEdep,m_cfg.minClusterCenterEdep)) {
+            idx2 = indices.erase(idx2);
+            continue;
+          }
+
+          // not energetic enough for cluster hit in this group
+          if (hits[*idx2].getEnergy() < m_cfg.minClusterHitEdep) {
+            idx2++;
+            continue;
+          }
```
ruse-traveler, line 280: Out of curiosity: would the equivalent conditions between lines 164 - 167 guarantee that the hits being processed at this point already satisfy these conditions?

ruse-traveler (commented):
Nice! This is a significant improvement! I'm still getting my head around the changes, but so far looks good!

I did see some diffs in some things like the cluster energy in the ZDC HCal clusters for 18x275 Q^2 > 0.... But I also some diffs in the Barrel HCal clusters which should be unaffected, so I'm guessing these are unrelated to the PR?

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: There's a list of all hits. In lines 164 - 167 the loop hasn't gotten to the hits that are being considered here. I clearly need to make my comment `whereas caller has removed earlier low energy hits, this removes ones that caller hasn't gotten to yet` clearer... Any suggestions for how to improve it?

wdconinc:
> some diffs in some things like the cluster energy

There are diffs in the raw and rec hits which propagate. I think TRandomMinMax is producing different random numbers depending on which architecture it ends up on... or something like that.

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -12,6 +12,7 @@
 #pragma once
```
sly2j, line 12: Regarding the attribution comment at the start of the file: I think if we have top-level comments with attribution like this we should keep them up-to-date. In this case you are modifying the algorithm so that should be mentioned and attributed. Or we should at least mention that the "first version" was based on the paper IMO.

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: I agree with you that this is duplication and could be addressed.

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -65,7 +66,7 @@ namespace eicrecon {
 
     // unitless counterparts of the input parameters
     double localDistXY[2]{0, 0}, layerDistEtaPhi[2]{0, 0}, layerDistXY[2]{0, 0}, sectorDist{0};
```
sly2j, line 68: Two comments (not pertaining to this PR): these should all be defined on their own line. And if we really want to use arrays here we should at least use `std::array<double, 2>` instead of raw C-arrays.

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: The algorithm isn't actually getting modified. The only thing that's modified is the choice of containers (vector -> set), the way of encoding that hits have been picked already (set membership), and the way groups are grown. I don't see this as a different algorithm. This should return the 100% identical proto-cluster hits (possibly in a different order).

Even sorting hits at first, and then modifying the selection loop by not considering hits that will clearly not match the requirements is an optimization, not a change to the algorithm. Maybe that's just semantics.

src/algorithms/calorimetry/ImagingTopoCluster.h
sly2j: I think it would be good to have a clear separation of responsibilities between bfs_group and the caller without too many pre-conditions that need to be assumed. I would suggest removing the first instance of the check in the caller, as it won't really lead to extraneous execution time (bfs_group will immediately clear them right here).

src/algorithms/calorimetry/ImagingTopoCluster.h
```diff
@@ -124,21 +125,61 @@ namespace eicrecon {
...
+                         (*hits)[*idx].getLocal().x, (*hits)[*idx].getLocal().y, (*hits)[*idx].getPosition().z,
+                         (*hits)[*idx].getPosition().x, (*hits)[*idx].getPosition().y, (*hits)[*idx].getPosition().z,
+                         (*hits)[*idx].getEnergy()
             );
-            // already in a group, or not energetic enough to form a cluster
-            if (visits[i] || (*hits)[i].getEnergy() < minClusterCenterEdep) {
-                continue;
+
+            // not energetic enough for cluster center or other cluster hits, so erase
+            if ((*hits)[*idx].getEnergy() < std::min(m_cfg.minClusterHitEdep,m_cfg.minClusterCenterEdep)) {
```
sly2j, line 166: I propose dropping this check right here (as I explain below). The first thing bfs_group will do is repeat this check+erasure. That way there are less preconditions to keep in mind in bfs_group, and less code duplication.

src/algorithms/calorimetry/ImagingTopoCluster.h
sly2j: Second comment: The way it is written here the second if statement will never evaluate to true so can be removed. In the strange case that `minClusterHitEdep > minCLusterCenterEdep` all hits will just be eaten (do we have a check for this, as that should at least print a warning on init).

sly2j (changes requested):
I left some minor comments that need addressing, but overall I think this looks very nice.

sly2j:
One more comment (not pertaining to this PR): 
Screenshot 2024-08-19 at 2 25 19 PM

This code does an absolute x-y difference if layers are within a certain distance. I'm assuming this is what is used for the ZDC? It would be good to add more documentation to the meaning of these options. Also, it may be worthwhile for completeness and options to add an xyz option that does the same thing but also considering the z-coordinate.

src/algorithms/calorimetry/ImagingTopoCluster.h
ruse-traveler: > There's a set of all hits. In lines 164 - 167 the loop hasn't gotten to the hits that are being considered here. I clearly need to make my comment `whereas caller has removed earlier low energy hits, this removes ones that caller hasn't gotten to yet` clearer... Any suggestions for how to improve it?

I see! That makes sense! I think noting that we need this 2nd check since we're processing the set of all hits, so maybe some thing like `since looping over all hits, this removes low energy hits not removed earlier by caller`

ruse-traveler:
> > some diffs in some things like the cluster energy
> 
> There are diffs in the raw and rec hits which propagate. I think TRandomMinMax is producing different random numbers depending on which architecture it ends up on... or something like that.

Ahhhhh that could be it... They're all on the level of a couple counts, so I'm not too worried about it

ruse-traveler (dismissed):
LGTM! I'm happy with this once @sly2j's comments are addressed!

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: 3d151e22a

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: 714aec408

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: 26790cf0c

wdconinc: 3941b5765

src/algorithms/calorimetry/ImagingTopoCluster.h
wdconinc: 26790cf0ccac6b0941ff80580daccdd32c09df14

sly2j (approved):

PR #1584, opened 2024-08-19
fix: deterministic cluster time

wdconinc:
This PR makes the cluster time deterministic, and not based on the whim of which hit happens to be first. It defines the cluster time as the energy-weighted average of all hit times.

This is not the only deterministic choice here (earliest hit is another).

- [x] Bug fix (issue: cluster time depends on hit ordering)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -169,7 +169,7 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
     debug("hit energy = {} hit weight: {}", hit.getEnergy(), weight);
     auto energy = hit.getEnergy() * weight;
     totalE += energy;
-    time += (hit.getTime() * energy - time) / totalE;
+    time += (hit.getTime() - time) * energy / totalE;
     cl.addToHits(hit);
     cl.addToHitContributions(energy);
     const float eta = edm4hep::utils::eta(hit.getPosition());
--- a/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
@@ -161,14 +161,15 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
   // Used to optionally constrain the cluster eta to those of the contributing hits
   float minHitEta = std::numeric_limits<float>::max();
   float maxHitEta = std::numeric_limits<float>::min();
-  auto time       = pcl.getHits()[0].getTime();
-  auto timeError  = pcl.getHits()[0].getTimeError();
+  auto time      = 0;
+  auto timeError = 0;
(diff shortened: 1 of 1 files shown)


src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
```diff
@@ -161,14 +161,15 @@ std::optional<edm4eic::MutableCluster> CalorimeterClusterRecoCoG::reconstruct(co
...
-  auto timeError  = pcl.getHits()[0].getTimeError();
+  auto time      = 0;
+  auto timeError = 0;
   for (unsigned i = 0; i < pcl.getHits().size(); ++i) {
     const auto& hit   = pcl.getHits()[i];
     const auto weight = pcl.getWeights()[i];
     debug("hit energy = {} hit weight: {}", hit.getEnergy(), weight);
     auto energy = hit.getEnergy() * weight;
     totalE += energy;
+    time += (hit.getTime() * energy - time) / totalE;
```
ruse-traveler, line 172: Out of curiosity, is there a particular reason to prefer the energy-weighted time over the earliest?

ruse-traveler (commented):
🙏 Very nice!

This looks good to me, though I'm curious about the pros-and-cons of using the time-weighted time vs. earliest.

src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc
wdconinc: I wanted something that is relatively robust against the inclusion of noise hits. If a noise hit gets included in a cluster, I didn't want it to throw off the entire timing determination.

veprbl, line 172: `(T*E - T) / E`? Something is wrong with dimensions in this expression?

wdconinc: Right! I was adapting a running update formula without weighting.
```suggestion
    time += (hit.getTime() * energy - time * (totalE - energy)) / totalE;
```

I think it would be useful to have a helper class for running updates of mean and stdev when using weighting.

wdconinc: Rederived it.
```suggestion
    time += (hit.getTime() - time) * energy / totalE;
```
Meets with limit criteria that the earlier comment violates:
- `energy = 0` should result in no change

wdconinc: If you want something robust that is closer to earliest hit, then I guess we could do a CFD-style timing determination: the time at which we reach 50% of the maximum in a suitably binned histogram. I don't know how you'd code that up...

ruse-traveler: Ahhh gotcha! That makes sense! I don't know how to code up the CFD-style determination either 😅 

I'm happy with the energy-weighted time for now! I was just curious about the rationale for the choice

PR #1585, opened 2024-08-19
linux-eic-shell.yml: don't tweak dot files

veprbl:


veprbl:
Seems to work, but results look bad.

Issue #1583, opened 2024-08-19
Division by zero in CalorimeterClusterRecoCoG


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released): topocluster-sort (might be specific to branch)
- Which version (or `HEAD` for the most recent on git):
- Any specific OS or system where the issue occurs?
- Any special versions of ROOT or Geant4?

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. https://github.com/eic/EICrecon/actions/runs/10446101104/job/28923156561#step:8:6332

### Expected Result: (what do you expect when you execute the steps above)
No division by zero.

### Actual Result: (what do you get when you execute the steps above)
```
Error: /home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h:34:43: runtime error: division by zero
/usr/bin/llvm-symbolizer-16: /opt/local/lib/libcurl.so.4: no version information available (required by /usr/bin/llvm-symbolizer-16)
    #0 0x7f3b771794dc in logWeight(double, double, double, int) /home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h:34:43
    #1 0x7f3b7717d8ad in std::function<double (double, double, double, int)>::operator()(double, double, double, int) const /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/std_function.h:591:9
    #2 0x7f3b7717d8ad in eicrecon::CalorimeterClusterRecoCoG::reconstruct(edm4eic::ProtoCluster const&) const /home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc:204:25
    #3 0x7f3b7717a466 in eicrecon::CalorimeterClusterRecoCoG::process(std::tuple<gsl::not_null<edm4eic::ProtoClusterCollection const*>, edm4hep::SimCalorimeterHitCollection const*> const&, std::tuple<gsl::not_null<edm4eic::ClusterCollection*>, edm4eic::MCRecoClusterParticleAssociationCollection*> const&) const /home/runner/work/EICrecon/EICrecon/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc:71:21
[...]

PR #1593, opened 2024-08-21
fix: avoid division by zero by using realistic ZDC Hcal digitization

wdconinc:
Issue #1583 points out division by zero in cluster reconstruction. This is caused by the [term E / tE in the log-based weighting of hits](https://github.com/eic/EICrecon/blob/9d1b01ed457aa5dceb413d3dace910b2c799278b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h#L33-L35), when [totalE is zero](https://github.com/eic/EICrecon/blob/9d1b01ed457aa5dceb413d3dace910b2c799278b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc#L204). That happens when the energy in a deposited hit is zero, which happens when the [raw hit amplitude is exactly equal to the pedestal](https://github.com/eic/EICrecon/blob/9d1b01ed457aa5dceb413d3dace910b2c799278b/src/algorithms/calorimetry/CalorimeterHitReco.cc#L189-L190). That can only happen [when the threshold is set below the pedestal](https://github.com/eic/EICrecon/blob/9d1b01ed457aa5dceb413d3dace910b2c799278b/src/algorithms/calorimetry/CalorimeterHitReco.cc#L172-L174), which is unphysical. 

This PR sets the threshold for the ZDC Hcal to a physical value, based on what it is set to for the ZDC Ecal. Happy to change it to something else, but we need a realistic value here.

- [x] Bug fix (issue #1583)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
[HcalFarForwardZDCRecHits](https://eicrecon.epic-eic.org/pr/1593/capybara/rec_dis_18x275_minQ2=0_craterlake/index.html#HcalFarForwardZDCRecHits) are much reduced in the low energy bins, and uniformly suppressed in other variables.
image

[HcalFarForwardZDCClusters](https://eicrecon.epic-eic.org/pr/1593/capybara/rec_dis_18x275_minQ2=0_craterlake/index.html#HcalFarForwardZDCClusters) show hardly any change in the overall distributions or number of clusters.
image

wdconinc:
Note that the decrease of the number of hits (with low energy) will in itself cause the ZDC clustering to improve in speed...

wdconinc:
@sebouh137 Can you provide more realistic ZDC Hcal digitization parameters?

sebouh137:
> @sebouh137 Can you provide more realistic ZDC Hcal digitization parameters?

I am going to defer this question to @rymilton , as i think he'd know the answer to that question

sebouh137:
> > @sebouh137 Can you provide more realistic ZDC Hcal digitization parameters?
> 
> I am going to defer this question to @rymilton , as i think he'd know the answer to that question
I talked to Ryan, and he doesn't know either, and I think these values were copied from somewhere else (probably from the insert, which was copied from the FHCAL)

wdconinc:
So you're ok with the values I changed it into?

wdconinc:
I used your methodology and copied it from the EcalFarForwardZDCRecHits :-P

rymilton:
As far as I know, the ZDC Hcal digitization parameters were originally taken from the insert/forward Hcal parameters at the time and then adjusted based on the performance we were getting. But I'm not sure where the -100 threshold value came from since I wasn't involved with this optimization. If the numbers you put @wdconinc fix the issue, I say go for it. We can adjust them later if need be.

rymilton:
Do you agree @sebouh137?

veprbl (approved):

PR #1594, opened 2024-08-21
allow MultiSegmentation in calorimeter hit recon

sebouh137:
Allow multisegmentation without giving a warning about it.

There is an if statement in CalorimeterHitReco that checks if the segmentation type is one of several types of segmentations (cartesian or hexagonal) that prints a warning if the segmentation is not one of those.  Adds MultiSegmentation to this list.  

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: removal of unnecessary warning

no
no


src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -243,7 +243,7 @@ void CalorimeterHitReco::process(
         std::vector<double> cdim;
         // get segmentation dimensions
         auto segmentation_type = m_converter->findReadout(local).segmentation().type();
-        if (segmentation_type == "CartesianGridXY" || segmentation_type == "HexGridXY") {
+        if (segmentation_type == "CartesianGridXY" || segmentation_type == "HexGridXY" || segmentation_type=="MultiSegmentation") {
```
wdconinc, line 246: There's a reason for this `if`: It's here to make sure that we actually have a segmentation of a planar geometry with 2-dimensional cells where the X and Y coordinates give the size of the cell. That's true for CartesianGridXY and for HexGridXY (maybe). It's why we don't have CartesianGridXZ here, or any of the other non-XY segmentations.

This is simply not the right place to handle MultiSegmentation (though it may be for a MultiSegmentation iff the actual segmentation of the specific cellID is CartesianGridXY or HexGridXY).

wdconinc (changes requested):

src/algorithms/calorimetry/CalorimeterHitReco.cc
sebouh137: hmm... maybe we could check recursively if the underlying subsegmentation is either CartesianGridXY or HexGridXY?  Otherwise algorithms that require the cell size (like Hexplit) won't work for multisegmentation.

sebouh137: I just checked in this recursive version... @wdconinc what do you think?

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -242,7 +243,12 @@ void CalorimeterHitReco::process(
         const auto pos = local.nominal().worldToLocal(gpos);
         std::vector<double> cdim;
         // get segmentation dimensions
-        auto segmentation_type = m_converter->findReadout(local).segmentation().type();
+        auto segmentation = m_converter->findReadout(local).segmentation();
+        auto segmentation_type = segmentation.type();
+        while (segmentation_type == "MultiSegmentation"){
+            segmentation = ((MutltiSegmentation&)segmentation).subsegmentation(cellID);
```
wdconinc, line 249: Avoid c-style casts.

wdconinc (commented):
The MultiSegmentation resolution before dealing with specific segmentations seems sensible.

src/algorithms/calorimetry/CalorimeterHitReco.cc
sebouh137: I am not sure how to do this.  Please enlighten me with a suggested change.

wdconinc: I think you should be able to use a dynamic_cast on reference types here.

wdconinc: Something like
```
        Segmentation& segmentation = m_converter->findReadout(local).segmentation();
        auto segmentation_type = segmentation.type();
        while (segmentation_type == "MultiSegmentation"){
            auto& multi_segmentation = dynamic_cast<Segmentation&>(segmentation);
            segmentation= multi_segmentation.subsegmentation(cellID);
```
but I'm entirely unable to test this here right now.

wdconinc: Probably works with first auto instead of explicit Segmentation&.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -242,7 +243,12 @@ void CalorimeterHitReco::process(
         const auto pos = local.nominal().worldToLocal(gpos);
         std::vector<double> cdim;
         // get segmentation dimensions
-        auto segmentation_type = m_converter->findReadout(local).segmentation().type();
+        const Segmentation& segmentation = m_converter->findReadout(local).segmentation();
+        auto segmentation_type = segmentation.type();
+        while (segmentation_type == "MultiSegmentation"){
+            auto& multi_segmentation = dynamic_cast<DDSegmentation::MultiSegmentation&>(segmentation);
+            segmentation= multi_segmentation.subsegmentation(cellID);
+        }
```
wdconinc, line 251: Yeah, this is running into the DD4hep handle structures. Try something like this:
```suggestion
        auto segmentation = m_converter->findReadout(local).segmentation();
        auto segmentation_type = segmentation.type();
        while (segmentation_type == "MultiSegmentation"){
            auto* multi_segmentation = dynamic_cast<dd4hep::DDSegmentation::MultiSegmentation*>(segmentation.segmentation());
            auto& sub_segmentation = multi_segmentation->subsegmentation(cellID);
            segmentation = dd4hep::Handle<dd4hep::DDSegmentation::Segmentation>(&sub_segmentation);
        }
```

wdconinc: This compiles, but you'll want to test that it actually works.

wdconinc: @veprbl may have thoughts too.

sebouh137: I'm getting linker errors with this.  I'll try this locally after you get it to compile.

sebouh137: I was able to get this to work, thanks to @nathanwbrei

wdconinc: I see what you did there. Still curious how DD4hep would resolve this using the handles they impose on us :-)

Sorry, I shouldn't complain about handles...

https://github.com/AIDASoft/DD4hep/blob/9c6040cfa7918691b02024a7d161d9d340a5da30/DDCore/include/DD4hep/Handle.h#L76-L78

sebouh137:
@wdconinc @veprbl we got this to compile, and I was able to check that it is working on my laptop.  After this passes all checks (except the pre-commit that is failing for reasons outside of the scope of this pull request), please approve this pull request unless you have other issues you want me to address.  Thanks

wdconinc (approved):
Thanks for figuring this out :-)

PR #1595, opened 2024-08-22
HcalEndcapPInsertImagingProtoClusters: use a larger hit spacing threshold

sebouh137:
Uses a larger threshold for the distance between the neighboring hits in the insert, due to the new cell size.  Also include "side" field (ie, left vs right) in the hit merger algorithm

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: change of parameters

Probably.  Including the detector "side" as a field in the hit merger algorithm probably breaks the code if using a version of the epic repository from before pull request https://github.com/eic/epic/pull/771 is merged.  Therefore this pull request and https://github.com/eic/epic/pull/771 should be merged at the same time as one another
yes.  Changes the parameters of the topo clustering and also uses the "side" field in the hit merger.


src/detectors/FHCAL/FHCAL.cc
```diff
@@ -89,13 +89,14 @@ extern "C" {
         app   // TODO: Remove me once fixed
       ));
 
-      double side_length=18.89 * dd4hep::mm;
+      // define the distance between neighbors in terms of the largest possible distance between subcell hits
+      double side_length=31 * dd4hep::mm;
```
veprbl, line 93: Would it make sense to access the `HcalEndcapPInsertCellSizeLGRight` from the geometry? We'd need to provide a fallback value when it's not defined.

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -69,7 +69,7 @@ extern "C" {
           "HcalEndcapPInsertMergedHits", {"HcalEndcapPInsertRecHits"}, {"HcalEndcapPInsertMergedHits"},
           {
             .readout = "HcalEndcapPInsertHits",
-            .fields = {"layer", "slice"},
+            .fields = {"layer", "slice", "side"},
             .refs = {1, 0},
```
veprbl, line 73: Should `refs` have the same length as `fields`?

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -69,7 +69,7 @@ extern "C" {
           "HcalEndcapPInsertMergedHits", {"HcalEndcapPInsertRecHits"}, {"HcalEndcapPInsertMergedHits"},
           {
             .readout = "HcalEndcapPInsertHits",
-            .fields = {"layer", "slice"},
+            .fields = {"layer", "slice", "side"},
```
veprbl, line 72: We could check the readout for presence of the side field to support the older geometry.

veprbl: Right now doing it like it's done here is easy because EICrecon is lax with exceptions (they are not fatal, only insert would be affected), but we'd want to move to a more strict behavior. Whatever preparation can be done upfront, we should try to do.

src/detectors/FHCAL/FHCAL.cc
sebouh137: i don't know what refs does.  What does it do?

src/detectors/FHCAL/FHCAL.cc
sebouh137: that makes sense, I'll look into how to access that information

src/detectors/FHCAL/FHCAL.cc
sebouh137: now that I look at this more carefully, I am not sure if this is necessary for what I am doing, so I will remove this particular change and keep the other ones in this pull request

src/detectors/FHCAL/FHCAL.cc
sebouh137: thanks!  I'll try this.

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -89,13 +89,15 @@ extern "C" {
         app   // TODO: Remove me once fixed
       ));
 
-      double side_length=18.89 * dd4hep::mm;
+      // define the distance between neighbors in terms of the largest possible distance between subcell hits
+      auto detector = app->GetService<DD4hep_service>()->detector();
+      double side_length=std::max({detector->constant<double>("HcalEndcapPInsertCellSizeLGRight"),detector->constant<double>("HcalEndcapPInsertCellSizeLGLeft")});
```
veprbl, line 94:
```suggestion
      double side_length;
      try {
        side_length = std::max({detector->constant<double>("HcalEndcapPInsertCellSizeLGRight"), detector->constant<double>("HcalEndcapPInsertCellSizeLGLeft")});
      } except (std::runtime_exception&) {
        side_length = 0. * dd4hep::mm;
      }
```

src/detectors/FHCAL/FHCAL.cc
sebouh137: done

sebouh137:
@veprbl, all of the checks now pass, and I have tested to make sure that it runs locally on my laptop.  If there is no further change you need from me, please approve this pull requeset.

veprbl (approved):
Looks great. Don't see any major changes in the capybara report for the `HcalEndcapPInsertClusters`, but that's probably fine.

PR #1596, opened 2024-08-23
linux-eic-shell.yml: continue-on-error for graphviz rendering

veprbl:
The issues with rendering the factory graph are cropping in. Example PRs: #1577 #1576


wdconinc (approved):
Clever trick on the upload-artifact if-no-files-found: error.

PR #1598, opened 2024-08-23
fix: write local position in mm in CalorimeterHitsMerger

wdconinc:
This PR fixes an issue where the `local` position of merged hits is written in DD4hep units instead of EDM4eic units. The subsequent clustering expects the `local` position to be in EDM4eic units of course.

- [x] Bug fix (issue: no clustering in backward HCal, https://chat.epic-eic.org/main/pl/pg6fwpimetgf8esmz4zfotwc3c)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @lkosarz 

No.

Fixes backward HCal clusters (should have changing number of clusters, distributed over 10x larger area).

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
@@ -121,7 +121,7 @@ void CalorimeterHitsMerger::process(
                 gpos.x() / dd4hep::mm, gpos.y() / dd4hep::mm, gpos.z() / dd4hep::mm
         );
         const decltype(edm4eic::CalorimeterHitData::local) local(
-                pos.x(), pos.y(), pos.z()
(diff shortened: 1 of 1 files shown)


lkosarz:
Thanks! That fixed Reco clusters, but MC clusters are still wrong. I was suspecting the issue was with the algorithms.

h_nHCal_cluster_MC_pos_xy_log
h_nHCal_cluster_Rec_pos_xy_log

wdconinc:
Truth clustering was never intended to work with merged hits. Every other detector has moved on to real clusters so there's little incentive to fix truth clustering, I'm afraid.

lkosarz:
I checked the hits associated to truth clusters and they look OK, well separated. I also had a quick look at truth clustering codes and don't see an obvious weak point. It would be good to have correct truth clusters later, but if others abandoned them we may do the same in the end. For now, they are not really an useful reference.

veprbl (approved):

PR #1600, opened 2024-08-24
fix: use podio::ObjectID::index in CalorimeterHitDigi

wdconinc:
Instead of using our own counting index in the loop, we should use the `podio::ObjectID` index, which is semantically defined as the order in the collection. Reduces locals, but otherwise pedantic.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: pedanticism, is that even a word?

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -22,6 +22,7 @@
 #include <algorithms/service.h>
 #include <edm4hep/CaloHitContributionCollection.h>
 #include <fmt/core.h>
+#include <podio/ObjectID.h>
 #include <podio/RelationRange.h>
 #include <algorithm>
 #include <cmath>
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -130,16 +130,13 @@ void CalorimeterHitDigi::process(
 
     // find the hits that belong to the same group (for merging)
     std::unordered_map<uint64_t, std::vector<std::size_t>> merge_map;
-    std::size_t ix = 0;
     for (const auto &ahit : *simhits) {
         uint64_t hid = ahit.getCellID() & id_mask;
 
         trace("org cell ID in {:s}: {:#064b}", m_cfg.readout, ahit.getCellID());
         trace("new cell ID in {:s}: {:#064b}", m_cfg.readout, hid);
 
-        merge_map[hid].push_back(ix);
-
-        ix++;
+        merge_map[hid].push_back(ahit.getObjectID().index);
     }
 
     // signal sum


src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -130,16 +131,13 @@ void CalorimeterHitDigi::process(
...
     for (const auto &ahit : *simhits) {
         uint64_t hid = ahit.getCellID() & id_mask;
 
         trace("org cell ID in {:s}: {:#064b}", m_cfg.readout, ahit.getCellID());
         trace("new cell ID in {:s}: {:#064b}", m_cfg.readout, hid);
 
-        merge_map[hid].push_back(ix);
-
-        ix++;
+        merge_map[hid].push_back(ahit.getObjectID().index);
```
simonge, line 140: If the input hit collection is ever becomes a subset collection, e.g. through filtering the hits or merging other collection of hits this will stop working correctly. What could be neater, and safer to cover for these instances, is having a map to a vector of the hits themselves rather than indices?

ruse-traveler: We were actually discussing a similar situation in the context of the track-cluster merge/splitter:
https://github.com/eic/EICrecon/pull/1406#discussion_r1681998184

So for my own understanding: if an input collection is somehow a subset (particularly if it's from merging other collections), is there a way to access specific members of that subset safely _other_ than just insertion index (e.g. `(*input_collect)[index]`)?

simonge: Using the full objectID should be relatively safe. I am concerned it might stop being unique when we start moving between events/timeframes, particularly when collecting from several JEvents. My preferred alternative approach, which someone might tell me is bad practice, would be to have a map which uses the pointer to the hit itself as the key/value.

In this case e.g.
` std::unordered_map<uint64_t, std::vector<edm4hep::SimCalorimeterHit*>> merge_map;`

And in #1406 
`typedef std::map<edm4eic::ProtoCluster*, bool, CompareObjectID> MapToFlag;`

I guess some downsides would be that the variable is linked to a type which using the ObjectID isn't. If there are other issues with this approach hopefully someone else can pitch in.

wdconinc: Podio separates the user layer from the POD layer. Multiple user layer objects can refer to the same underlying POD data, in particular when they are a mutable and immutable user layer object. Storing a pointer to the user layer is therefore maybe not recommended.

wdconinc:
Closing since pedantic and good arguments against this change brought forward in https://github.com/eic/EICrecon/pull/1600#discussion_r1731614691.

src/algorithms/calorimetry/CalorimeterHitDigi.cc
ruse-traveler: I see! Thanks for the clarifications @simonge and @wdconinc!

PR #1383, opened 2024-04-16
implementation of ambiguity resolution solver from ACTS

minjungkim12:
### Briefly, what does this PR introduce?
* implementation of ambiguity resolution solver from ACTS; removing duplicate tracks from realistic seeding
* new factory called after CKFtracking; taking  CKFtracking outputs as inputs; provides a set of output collections

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
* No change required; 

### Does this PR change default behavior?
* adding 3 PodIO output collection

PR #1605, opened 2024-08-27
tracking: use Seeded tracks by default, rename previous default to TruthSeeded

veprbl:

As we agreed during the collaboration meeting, the real seeding workflow is mature and should become the default. This changes Seeded collections to not include that prefix, and prefixless collections are moved to TruthSeeding (name is long but not supposed to be used by general users). As a side effect, Jet finding and ReconstructedParticles and DRICH IRT-based PID will use real seeding.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes

### DIFF
--- a/src/detectors/BTOF/BTOF.cc
+++ b/src/detectors/BTOF/BTOF.cc
@@ -75,6 +75,21 @@ void InitPlugin(JApplication *app) {
     };
 
     app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
+          "CombinedTOFTruthSeededLUTPID",
+          {
+          "ReconstructedTruthSeededChargedWithPFRICHPIDParticles",
+          "ReconstructedTruthSeededChargedWithPFRICHPIDParticleAssociations",
+          },
+          {
+          "ReconstructedTruthSeededChargedWithPFRICHTOFPIDParticles",
+          "ReconstructedTruthSeededChargedWithPFRICHTOFPIDParticleAssociations",
+          "CombinedTOFTruthSeededParticleIDs",
+          },
+          pid_cfg,
+          app
+          ));
+
+    app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
           "CombinedTOFLUTPID",
           {
           "ReconstructedChargedWithPFRICHPIDParticles",
@@ -84,21 +99,6 @@ void InitPlugin(JApplication *app) {
           "ReconstructedChargedWithPFRICHTOFPIDParticles",
           "ReconstructedChargedWithPFRICHTOFPIDParticleAssociations",
           "CombinedTOFParticleIDs",
-          },
-          pid_cfg,
-          app
-          ));
-
-    app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
-          "CombinedTOFSeededLUTPID",
-          {
-          "ReconstructedSeededChargedWithPFRICHPIDParticles",
-          "ReconstructedSeededChargedWithPFRICHPIDParticleAssociations",
-          },
-          {
-          "ReconstructedSeededChargedWithPFRICHTOFPIDParticles",
-          "ReconstructedSeededChargedWithPFRICHTOFPIDParticleAssociations",
-          "CombinedTOFSeededParticleIDs",
           },
           pid_cfg,
           app
--- a/src/detectors/DIRC/DIRC.cc
+++ b/src/detectors/DIRC/DIRC.cc
@@ -101,6 +101,21 @@ extern "C" {
     };
 
     app->Add(new JOmniFactoryGeneratorT<PIDLookup_factory>(
+          "DIRCTruthSeededLUTPID",
+          {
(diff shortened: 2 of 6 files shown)


wdconinc (approved):
Changes make sense. I did not verify that this is the complete set of changes required. I assume you've grepped for other instances.

bspage912 (approved):

michael-pitt:
Hi @veprbl ,

I'm not sure if I need to open an issue or if I can report it here. It seems that this RP also impacted the reconstruction of one of the far-forward sub-systems - the B0 Tracker. In the past we were using the `ReconstructedChargedParticles` container to get the B0 tracks. Since this RP, the B0 tracks moved to `ReconstructedTruthSeededChargedParticles` container. This might be OK since the b0 tracks are also using truth seeding, but I see that this container is called from [src/detectors/DRICH/DRICH.cc#L211](https://github.com/eic/EICrecon/blob/main/src/detectors/DRICH/DRICH.cc#L211), which is not related to Far-Forward detectors. It wasn't easy to find out what had happened in this case. 

I'm also motived some naming mismatch in [src/global/tracking/tracking.cc](https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc):
[L162 ](https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L162)`CentralCKFTrajectories` -> `CentralCKFActsTrajectoriesUnfiltered`
[L95](https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L95) `CentralCKFTruthSeededTrajectories`->`CentralCKFTruthSeededActsTrajectoriesUnfiltered`

Will you confirm that the B0 tracks should be retrieved from `ReconstructedTruthSeededChargedParticles` container? 

Thanks,
Michael

veprbl:
> I'm not sure if I need to open an issue or if I can report it here. It seems that this RP also impacted the reconstruction of one of the far-forward sub-systems - the B0 Tracker. In the past we were using the `ReconstructedChargedParticles` container to get the B0 tracks. Since this RP, the B0 tracks moved to `ReconstructedTruthSeededChargedParticles` container. This might be OK since the b0 tracks are also using truth seeding, but I see that this container is called from [src/detectors/DRICH/DRICH.cc#L211](https://github.com/eic/EICrecon/blob/main/src/detectors/DRICH/DRICH.cc#L211), which is not related to Far-Forward detectors. It wasn't easy to find out what had happened in this case.
> 
> Will you confirm that the B0 tracks should be retrieved from `ReconstructedTruthSeededChargedParticles` container?

Two things here: One is that realistic seeding does not work for B0 tracking. We need to have it implemented. Until we do that, I think it's for the best that we are reminded that truth seeding is being used by having to access the `ReconstructedTruthSeededChargedParticles`. Truth seeding can produce quite misleading performances. Another thing is about data processing for `ReconstructedParticles` passing though PID factories, that is needed to provide PID LUTs. The reconstructed particles are not just about tracking, this is a holistic object, even more detectors will be used in the future.

> 
> I'm also motived some naming mismatch in [src/global/tracking/tracking.cc](https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc): [L162 ](https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L162)`CentralCKFTrajectories` -> `CentralCKFActsTrajectoriesUnfiltered` [L95](https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L95) `CentralCKFTruthSeededTrajectories`->`CentralCKFTruthSeededActsTrajectoriesUnfiltered`

This refers to the two previous PRs: https://github.com/eic/EICrecon/pull/1470 https://github.com/eic/EICrecon/pull/1383
The `CentralCKF*Acts*` tags are Acts structures that are not PODIO and can't be e.g. directly written to an edm4eic.root file. We employ conversion from Acts to PODIO for now, but in the future the goal is to have algorithms to exchange PODIO structures only.

michael-pitt:
OK, Thanks! (adding @ajentsch in case someone in FF group will have a similar question), 

Indeed, I see that the central tracks are included in both `ReconstructedTruthSeededChargedParticles` and `ReconstructedChargedParticles` while B0 tracks are only in the `ReconstructedTruthSeededChargedParticles` container. 

So, I will do the following:
Central tracks retrieved from `ReconstructedChargedParticles`
B0 tracks are retrieved from `ReconstructedTruthSeededChargedParticles` after applying an eta cut of 4.2 (`trk_eta>4.2` in the hadron beam coordinate system).

simonge:
Other than for simplicity is there a reason the B0 tracks are mixed into the same raw hit collection as the central tracker? Would it not be more useful to keep them separate until they are needed(?) by the vertex finder for timing. The same acts algorithms would be used with the same configuration, or something specifically tuned to the B0, without attempting to use B0 hits in central detector tracks and vice versa. This would allow a different approach to setting the PID or at least making a PID assumption.

PR #1606, opened 2024-09-02
JEventProcessorPODIO: make `podio:output_include_collection` deprecation warning more compelling

veprbl:

This improves the wording to actually help general users get rid of the delay, that they may not know about.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc (approved):
We should parametrize the wait time, randomized based on the number of days since mid-August, exponential increase, asymptote at Thanksgiving, and use that in the message.

Or what you have is fine too.

PR #1607, opened 2024-09-03
GeometrySplit functor isn't threadsafe

simonge:
Fixes issues with the initialization of the GeometrySplit functor across threads.

Currently only the std::once_flag is a shared pointer and none of the other variables which are being initialized. This meant that when copied to other threads the is_init flag was set but the decoder and geometry divisions had not been so the loop over the detector ids was always empty. This resulted in an apparent variable decrease in the efficiency of the Low-Q2 tagger depending on how many threads were asked for.

Unit tests for all of the meta factories and functors should be added as an issue to try head this off in the future but outwith this fix.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
All of the Low-Q2 Tagger hits will be converted into clusters rather than just those run on the first thread to call the functor.


src/factories/meta/CollectionCollector_factory.h
```diff
@@ -35,6 +35,9 @@ namespace eicrecon {
         }
         typename T::collection_type* merged_collection = m_output().get();
         m_algo->process(in_collections, merged_collection);
+        //Log how many hits were collected from N input collections
+        this->logger()->info("Collected {} hits from {} input collections", merged_collection->size(), in_collections.size());
```
veprbl, line 39:
```suggestion
        this->logger()->debug("Collected {} hits from {} input collections", merged_collection->size(), in_collections.size());
```

src/factories/meta/SubDivideCollection_factory.h
```diff
@@ -43,6 +43,12 @@ class SubDivideCollection_factory : public JOmniFactory<SubDivideCollection_fact
       split_collections.push_back(gsl::not_null<typename T::collection_type*>(split.get()));
     }
     m_algo->process(m_input(),split_collections);
+    //Log how the hits were divided between the collection names
+    this->logger()->info("Divided {} hits between {} collections", m_input()->size(), split_collections.size());
+    //How many hits in each output collection
+    for (size_t i = 0; i < split_collections.size(); i++) {
+      this->logger()->info("Collection {} takes {} hits", i, split_collections[i]->size());
```
veprbl, line 50:
```suggestion
    this->logger()->debug("Divided {} hits between {} collections", m_input()->size(), split_collections.size());
    //How many hits in each output collection
    for (size_t i = 0; i < split_collections.size(); i++) {
      this->logger()->debug("Collection {} takes {} hits", i, split_collections[i]->size());
```

veprbl (commented):
`info` might be a bit too loud

simonge:
Are any of the CI tests run on multiple threads which would be possible to compare the output between a single and multi-threaded run?

simonge:
> Are any of the CI tests run on multiple threads which would be possible to compare the output between a single and multi-threaded run?

Looks like this was being partially approached in #975. For cases like this having at least one job still run on a single core to compare the output would be useful.

Checking the default flags it appears nthreads should be 1 if not specified, however the behaviour change seen here was only identified by explicitly adding the -Pnthreads=1 flag. Doesn't sound like that should have made a difference, any idea why it did?

wdconinc:
> Checking the default flags it appears nthreads should be 1 if not specified, however the behaviour change seen here was only identified by explicitly adding the -Pnthreads=1 flag. Doesn't sound like that should have made a difference, any idea why it did?

@nathanwbrei may be the best person to address that.

simonge:
Please could this be reviewed independently of the issues around running/testing with different numbers of threads, I can flag that as separate issue.

src/factories/meta/CollectionCollector_factory.h
```diff
@@ -35,6 +35,9 @@ namespace eicrecon {
         }
         typename T::collection_type* merged_collection = m_output().get();
         m_algo->process(in_collections, merged_collection);
+        //Log how many hits were collected from N input collections
+        this->logger()->debug("Collected {} hits from {} input collections", merged_collection->size(), in_collections.size());
+
```
veprbl, line 40: These need to be in the algorithm. Factories are supposed to not contain custom logic.

src/factories/meta/SubDivideCollection_factory.h
```diff
@@ -43,6 +43,12 @@ class SubDivideCollection_factory : public JOmniFactory<SubDivideCollection_fact
       split_collections.push_back(gsl::not_null<typename T::collection_type*>(split.get()));
     }
     m_algo->process(m_input(),split_collections);
+    //Log how the hits were divided between the collection names
+    this->logger()->debug("Divided {} hits between {} collections", m_input()->size(), split_collections.size());
+    //How many hits in each output collection
+    for (size_t i = 0; i < split_collections.size(); i++) {
+      this->logger()->debug("Collection {} takes {} hits", i, split_collections[i]->size());
+    }
```
veprbl, line 51: Same for this.

src/algorithms/meta/SubDivideFunctors.h
```diff
@@ -66,19 +70,19 @@ class GeometrySplit {
...
 
     std::vector<std::vector<long int>> m_ids;
     std::vector<std::string> m_divisions;
     std::string m_readout;
 
-    mutable std::shared_ptr<std::once_flag> is_init = std::make_shared<std::once_flag>();
-    mutable dd4hep::DDSegmentation::BitFieldCoder* m_id_dec;
-    mutable std::vector<size_t> m_div_ids;
+    std::shared_ptr<std::once_flag> is_init;
+    std::shared_ptr<dd4hep::DDSegmentation::BitFieldCoder*> m_id_dec;
```
veprbl, line 84: We don't need to store this on the heap
```suggestion
    dd4hep::DDSegmentation::BitFieldCoder m_id_dec;
```

veprbl: Okay, maybe, I get why.

veprbl:
Do we understand why these are copied in each thread in the first place?

simonge:
> Do we understand why these are copied in each thread in the first place?

I'm not familiar enough with JANA2 resource management to comment but would be keen to hear an explanation.

src/factories/meta/CollectionCollector_factory.h
simonge: Done

veprbl (approved):
I get the idea, but not the details. Fine to merge IMO.

PR #1609, opened 2024-09-06
added PrimaryVertices factory (subCollection of CentralTrackVertices)

starsdong:
- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


wdconinc:
@simonge can you make this a meta factory subset collection?

src/algorithms/reco/PrimaryVertices.h
```diff
@@ -0,0 +1,29 @@
...
+
+    public:
+
+        void init(std::shared_ptr<spdlog::logger>& logger);
+        std::unique_ptr<edm4eic::VertexCollection> execute(
+                const edm4eic::VertexCollection *rcvtx
+        );
+
+    private:
+        std::shared_ptr<spdlog::logger> m_log;
```
veprbl, line 26: This is not how we want algorithms to be defined.

1. Have to `#include <algorithms/algorithm.h>`
2. Define your algorithm type using algorithms::Algorithm
3. `void init()` has to be `final` and not take any arguments (use services for logger)
4. `void process(const Input&, const Output&) const final;` is the signature and name of the main entry point, it does not return a value other than by reference `Output`.

See any other algorithm in EICrecon for a good example.

src/algorithms/reco/PrimaryVertices.cc
```diff
@@ -0,0 +1,102 @@
...
+   * @return std::unique_ptr<edm4eic::VertexCollection>
+   */
+  std::unique_ptr<edm4eic::VertexCollection> PrimaryVertices::execute(
+                const edm4eic::VertexCollection *rcvtx
+        ){
+
+    // this map will store intermediate results
+    // so that we can sort them before filling output
+    // collection
+    std::map<int, edm4eic::Vertex> primaryVertexMap;
```
veprbl, line 43:
```suggestion
    std::map<int, edm4eic::Vertex, std::greater<int>> primaryVertexMap;
```

src/algorithms/reco/PrimaryVertices.cc
```diff
@@ -0,0 +1,102 @@
...
+      int N_trk = vtx.getAssociatedParticles().size();
+      m_log->trace( "\t N_trk = {}", N_trk );
+      primaryVertexMap[ N_trk ] = vtx;
+    } // vertex loop
+
+    // map sorts in descending order by default
+    // sort by descending
+    bool first = true;
+    // for (auto kv : primaryVertexMap) {
+    for (auto kv = primaryVertexMap.rbegin(); kv != primaryVertexMap.rend(); ++kv) {
```
veprbl, line 75:
```suggestion
    bool first = true;
    // map defined with std::greater<> will be iterated in descending order by the key
    for (auto kv : primaryVertexMap) {
```

src/algorithms/reco/PrimaryVerticesConfig.h
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Daniel Brandenburg
```
veprbl, line 2: Check this

simonge:
> @simonge can you make this a meta factory subset collection?

After a quick look I don't think this is possible at the moment. We would need the BooleanSplit functor from #1587, then getting the magnitude of the components of the position would need to be direct member functions, or extra code of the edm4eic::Vertex rather than getPosition().magnitudeTransverse() (Which looks like itself would need to be extended to accept 4d vectors). That to just filter the vertices.

A new functor would then be needed to sort/reduce the vertices in a secondary factory.

src/algorithms/reco/PrimaryVertices.h
starsdong: Hi Dmitry,

I have seen the example from other EICrecon part, but do you know any one of them is dealing with the SubsetCollection? I tried the setup as you suggested, the codes compile, but crashes when tested with some errors. The difference I can see is that I was following the ScatteredElectron or ElectronReconstruction algorithm which also use SubsetCollection. Any suggestions on this? Thanks

veprbl: Hard to tell without seeing the code. An issue might be because you allocate a collection on your end, like you do on line 47 above.

veprbl: Submitted a conversion of the original code https://github.com/eic/EICrecon/pull/1615

src/algorithms/reco/PrimaryVertices.cc
```diff
@@ -0,0 +1,102 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Daniel Brandenburg
```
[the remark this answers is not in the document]
starsdong: fixed now

src/algorithms/reco/PrimaryVertices.h
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Daniel Brandenburg
```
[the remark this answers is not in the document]
starsdong: fixed in the new commits

src/algorithms/reco/PrimaryVertices.h
starsdong: Updated following the instruction to use the algorithms interface. The issue a couple of days ago was due to the null logger pointer (fixed now). Please take a review of the new codes.

src/algorithms/reco/PrimaryVertices.cc
starsdong: Updated. In the meantime, switch to use multimap to avoid vertices having the same N_trk number get dropped or replaced.

src/algorithms/reco/PrimaryVertices.cc
starsdong: updated

src/algorithms/reco/PrimaryVerticesConfig.h
```diff
@@ -0,0 +1,23 @@
...
+  struct PrimaryVerticesConfig {
+
+    // For now these are wide open
+    // In the future the cut should depend
+    // on the generator settings
+    float maxVr = 50.0; // mm
+    float maxVz = 500.0; // mm
+    float maxChi2 = 10000.0; //
+    int   minNtrk = 1;  // >=
+    int   maxNtrk = 1000000; // <=
```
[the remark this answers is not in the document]
starsdong: These are post-reconstruction analysis-level selection cuts. The IVF algorithm cuts are embedded in the ACTS algorithm. They are different.

src/algorithms/reco/PrimaryVerticesConfig.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace eicrecon {
+
+  struct PrimaryVerticesConfig {
+
+    // For now these are wide open
+    // In the future the cut should depend
+    // on the generator settings
+    float maxVr = 50.0; // mm
+    float maxVz = 500.0; // mm
```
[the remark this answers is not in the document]
starsdong: see above.

src/algorithms/reco/PrimaryVerticesConfig.h
starsdong: fixed

src/global/reco/PrimaryVertices_factory.h
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Daniel Brandenburg
```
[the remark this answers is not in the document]
starsdong: fixed.

src/algorithms/reco/PrimaryVerticesConfig.h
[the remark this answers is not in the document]
veprbl: Well that is the question, in what way they are different? And if they are, why do we prefer one over another?

I understand, we are discussing trivial facilities, but it would be nice to draw from ACTS experience where possible.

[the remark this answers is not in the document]
starsdong: Those ACTS cuts still apply in the IVF finding/fitting. The cuts listed here at analysis level if there is selection to filter out any vertices. The geometric parameters (i.e. Vr. Vz etc.) should be experiment specific. Other cuts are very loose at this moment, but we include the option here in case users want to apply tighter cuts (e.g. chi2 to remove bad fitted vertices).

veprbl (approved):
No objections raised.

PR #1612, opened 2024-09-11
PIDLookup: avoid complaining about ambiguous associations

veprbl:

This disambiguates associations by selecting one with the largest contribution. As the result, no more need to output pesky warnings:
```
[PFRICH:RICHEndcapNLUTPID] [warning] Found a duplicate association for ReconstructedParticle at index 0
[PFRICH:RICHEndcapNLUTPID] [warning] The previous MCParticle was at 2 and the duplicate is at 0
```

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/pid_lut/PIDLookup.cc
+++ b/src/algorithms/pid_lut/PIDLookup.cc
@@ -8,7 +8,6 @@
 #include <edm4hep/utils/vector_utils.h>
 #include <fmt/core.h>
 #include <podio/podioVersion.h>
-#include <podio/ObjectID.h>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
Looks good to me.

I considered suggesting to make podio-0.17.4 a requirement, but that's used in release 24.05 so not that old.

I wish we could avoid the linear searches for matching associations. Unlikely it matters performance-wise, but we should consider putting the associations in a multimap where we end up having `for (auto& assoc: assocs)` inside another loop.

PR #1616, opened 2024-09-16
Correct xml file being called for different beam settings in CI

simonge:
Fixes xml being used so that it matches the MC sample being run.

It was noted in the Exclusive Diffractive and Tagging meeting that some of the campaigns appear to be being run with the wrong beamline settings. This only address the problem where I can see it in the CI.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Means the beam is being steered correctly for the MC samples being run in the CI

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -401,7 +401,7 @@ jobs:
             detector_config: craterlake_tracking_only
           - beam: 18x275
             minq2: 1000
-            detector_config: craterlake
+            detector_config: craterlake_18x275
     steps:
     - uses: cvmfs-contrib/github-action-cvmfs@v4
     - name: Retrieve simulation files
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -449,7 +449,7 @@ jobs:
         setup: "/opt/detector/epic-${{ env.detector-version }}/bin/thisepic.sh"
         run: |
           url=root://dtn-eic.jlab.org//work/eic2/EPIC/EVGEN/SIDIS/pythia6-eic/1.0.0/${{matrix.beam}}/q2_0to1/pythia_ep_noradcor_${{matrix.beam}}_q2_0.000000001_1.0_run1.ab.hepmc3.tree.root
-          npsim --compactFile ${DETECTOR_PATH}/${DETECTOR}_${{ matrix.detector_config }}.xml -N 100 --inputFiles ${url} --random.seed 1 --outputFile sim_dis_${{matrix.beam}}_minQ2=0_${{ matrix.detector_config }}.edm4hep.root -v WARNING
+          npsim --compactFile ${DETECTOR_PATH}/${DETECTOR}_${{ matrix.detector_config }}_${{matrix.beam}}.xml -N 100 --inputFiles ${url} --random.seed 1 --outputFile sim_dis_${{matrix.beam}}_minQ2=0_${{ matrix.detector_config }}.edm4hep.root -v WARNING
```
veprbl, line 452: I believe, this should not be changed, instead do it in `strategy`:`matrix` above

simonge: I'm not sure how to achieve what we need simply. I can't change `matrix.detector_config` without making changes in many other places and making all the file names even longer (Already problems where I did it for the Q2=1000 events).

.github/workflows/linux-eic-shell.yml
```diff
@@ -871,7 +875,7 @@ jobs:
         - CC: clang
           beam: 18x275
           minq2: 1000
-          detector_config: craterlake
+          detector_config: craterlake_18x275
```
simonge, line 878: If this is the preferred approach I'm happy. Although almost all of the benefits of having the matrix expansion will disappear as this needs to be expanded for the Q2=0 and change the detector config for the 5x41 to `craterlake_5x41` also. Maybe it's not a bad thing to be explicit with the configurations.

simonge:
Is this the script which is run for all of the hepmc3 campaigns?
https://github.com/eic/simulation_campaign_hepmc3/blob/main/scripts/run.sh

veprbl:
> Is this the script which is run for all of the hepmc3 campaigns? https://github.com/eic/simulation_campaign_hepmc3/blob/main/scripts/run.sh

I believe so, yes. cc @rahmans1

veprbl (approved):

veprbl:
I assumed that was because the files with those configurations were never produced on main.

veprbl:
(Long way to say "Yes")

PR #1623, opened 2024-09-30
update dynamic range of B0 ECAL

michael-pitt:
This RP solves the truncation in energy response of the B0 calorimeter as reported in slide 11 [here](https://indico.bnl.gov/event/24923/#12-b0-emcal). The energy threshold based on the simulated light yields of calorimeter crystals interfaces with the readout (slide 12 from the presentation above)

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: update thresholds in reconstruction

no 
restore correct energy response for B0 ECAL.


ajentsch (approved):

wdconinc (approved):
These changes are fine, but can you set realistic digitization thresholds? You currently have:
```
            .thresholdFactor = 0.0,
            .thresholdValue = 0.0,
```
This leads to a zero threshold which is not realistic. One of these should probably be non-zero, depending on whether you set a threshold based on a factor of pedestal widths above the pedestal mean, or by explicit ADC channel.

src/detectors/B0ECAL/B0ECAL.cc
```diff
@@ -33,11 +33,11 @@ extern "C" {
           {"B0ECalRawHits"},
 #endif
           {
-            .eRes = {0.0 * sqrt(dd4hep::GeV), 0.02, 0.0 * dd4hep::GeV},
+            .eRes = {0.0326 * sqrt(dd4hep::GeV), 0.02, 0.0 * dd4hep::GeV},
             .tRes = 0.0 * dd4hep::ns,
             .threshold= 5.0 * dd4hep::MeV,
-            .capADC = 16384,
-            .dyRangeADC = 20 * dd4hep::GeV,
+            .capADC = 160000,
```
wdconinc, line 39: We should also use a realistic ADC capacity. The 14 bit was ok, but this is not a realistic ADC capacity. Especially relevant since it is likely too large which will cause you to think you have better performance than warranted.

michael-pitt: We assume 6x6 mm2 SiPM with 15mircon pitch - resulting in (6/0.015)^2 = 160000 pixels per sensors. If 160000 photons in the sensor corresponds to 170GeV deposited energy (base on the simulation studies) we obtain:
- Stochastic noise of sqrt(170/160000) sqrt(GeV) = 0.0326 sqrt(GeV)

The threshold was also set with respect to number of photons in the sensor (at least 10 photons - at least 10 pixels with a signal).

I will change the ADC capacity to 14bits, 1 ADC count will be corresponding to 10MeV (instead of 1 MeV) and I will reduce the threshold to 1 (instead of 10)

veprbl: @michael-pitt Would you mind adding your nice explanations as a comment to the code? Thank you.

veprbl: Regarding the stochastic noise, it's not implemented yet, we might use https://github.com/eic/EICrecon/pull/1064, or ad-hoc smearing can be added in the digitzation.

michael-pitt: I added explanations in https://github.com/eic/EICrecon/pull/1623/commits/4a7bfabe069a6b369f9614317a8934eb339dd489:
https://github.com/eic/EICrecon/blob/4a7bfabe069a6b369f9614317a8934eb339dd489/src/detectors/B0ECAL/B0ECAL.cc#L36

https://github.com/eic/EICrecon/blob/4a7bfabe069a6b369f9614317a8934eb339dd489/src/detectors/B0ECAL/B0ECAL.cc#L59

src/detectors/B0ECAL/B0ECAL.cc
```diff
@@ -33,11 +33,12 @@ extern "C" {
           {"B0ECalRawHits"},
 #endif
           {
-            .eRes = {0.0 * sqrt(dd4hep::GeV), 0.02, 0.0 * dd4hep::GeV},
+                         // The stocastic term is set using light yield in PbOW4 of N_photons = 145.75 / GeV / mm, for 6x6 mm2 sensors with PDE=0.18 (a=1/sqrt(145.75*36*0.18))
+            .eRes = {0.0326 * sqrt(dd4hep::GeV), 0.02, 0.0 * dd4hep::GeV},
```
veprbl, line 37: I'm pretty sure you can remove the 2% term, it's probably was copied from Athena code for another calorimeter and has no meaning here.

veprbl (dismissed):

src/detectors/B0ECAL/B0ECAL.cc
veprbl: cc @Chao1009

veprbl: Also #1633

michael-pitt: thank you for your suggestion, I removed the 2% term. 

also cc'ing @FriederikeBock

veprbl (approved):

src/detectors/B0ECAL/B0ECAL.cc
Chao1009: Yes, the smearing is better to be always zero unless there are some unknown sources for the resolution.

src/detectors/B0ECAL/B0ECAL.cc
```diff
@@ -33,11 +33,12 @@ extern "C" {
           {"B0ECalRawHits"},
 #endif
           {
-            .eRes = {0.0 * sqrt(dd4hep::GeV), 0.02, 0.0 * dd4hep::GeV},
+                         // The stocastic term is set using light yield in PbOW4 of N_photons = 145.75 / GeV / mm, for 6x6 mm2 sensors with PDE=0.18 (a=1/sqrt(145.75*36*0.18))
```
veprbl, line 36: I'm now realizing, I don't understand this $145 ph/GeV$ number, this is too high, especially per $mm^2$. The light yield for PbWO4 is understood to be no more than few hundreds of photons per $MeV$ (and that is not always reached).
@michael-pitt Can you comment?

veprbl: So $145 * 1000 / 400$ (1000 GeV to MeV conversion factor, 400 = 20x20 mm face) is $362 ph/MeV$ this is too much even for raw yield, not including any absorption.

Issue #1341, opened 2024-03-20
More Robust Neutral Identification in Reconstruction


ruse-traveler:
**Is your feature request related to a problem? Please describe.**
One of the identified TDR wants from the calorimetry side was more robust neutral identification in EICrecon. Currently, neutral particle reconstruction is exclusively handled in the `MatchClusters` algorithm, which can certainly be improved.

**Describe the solution you'd like**
Ultimately, this is the role Particle Flow (PF) will play in our reconstruction, and being able to clearly identify neutral hadrons is the hallmark of any robust PF algorithm.

**Describe alternatives you've considered**
However, Particle Flow was identified as _not_ a TDR priority. Furthermore, while any similar functionality will need to be integrated into EICrecon in the future, such studies can be carried out post-facto on current EICrecon output.

PR #1627, opened 2024-10-08
Implement Track-Cluster Subtraction (PFA1a)

### LINKED ISSUE #900: Reconstruction: **[Particle Flow]** Implementation of PF algorithm + factories

### LINKED ISSUE #2127: Implement PFA1(a): Track-Cluster Subtractor
The 2nd stage of PFAlpha, track-cluster subtraction, is needed to separate calorimeter energy from charged vs. neutral particles.

ruse-traveler:

This PR implements the `TrackClusterSubtraction` algorithm, which ingests clusters from a specified calorimeter, any matched tracks, and subtracts off the expected energy deposited of the sum of tracks pointing to each cluster. This is the second stage of of the baseline particle flow routine, _PFAlpha_

This algorithm outputs two collections of clusters:
1. **Remnant clusters,** which consist of un-matched clusters and any cluster with nonzero energy left after the subtraction; and
2. **Expected clusters,** which consist of matched clusters with any remnant energy subtracted from them.

PFA1_VisualExplainer

In both cases, all properties of the cluster before subtraction are copied to the remnant and expected clusters with energies updated accordingly. The remnant clusters will be used to reconstruct neutral particles downstream, and the expected clusters will be used in the kinematic reconstruction of charged particles downstream.

Users are provided two methods to check if the leftover energy is consistent with 0:
1. A **basic method,** which checks if the leftover energy is less than 0 or within floating point precision of 0; or a
2. **Resolution method,** which checks if the leftover energy is less than 0 or within some number of standard deviations of 0 based on the resolution of the tracker and calorimeter.

For the latter method, the number of sigma to use and the resolutions of the tracker/calorimeter are configurable parameters. 

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1341 #2127)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### To-do
Will follow up with a benchmark for this stage of PFAlpha with a benchmark.


src/algorithms/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.16)
 add_subdirectory(interfaces)
 add_subdirectory(calorimetry)
 add_subdirectory(tracking)
+add_subdirectory(particle)
```
wdconinc: For consideration, isn't `particle` a bit too generic and might we use `particle_flow` here?

src/algorithms/particle/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+# The macro grabs sources as *.cc *.cpp *.c and headers as *.h *.hh *.hpp Then
+# correctly sets sources for ${_name}_plugin and ${_name}_library targets Adds
+# headers to the correct installation directory
+plugin_glob_all(${PLUGIN_NAME})
+
+# Find dependencies
+plugin_add_algorithms(${PLUGIN_NAME})
+plugin_add_dd4hep(${PLUGIN_NAME})
+plugin_add_event_model(${PLUGIN_NAME})
+plugin_add_eigen3(${PLUGIN_NAME})
```
wdconinc: Appears to be unused.

src/global/particle/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+cmake_minimum_required(VERSION 3.16)
```
wdconinc: Inherit from top level.

src/algorithms/particle/TrackClusterSubtractorConfig.h
```diff
@@ -0,0 +1,33 @@
...
+// Copyright (C) 2025 Derek Anderson
+
+#pragma once
+
+namespace eicrecon {
+
+struct TrackClusterSubtractorConfig {
+
+  ///! fraction of track energy to subtract
+  double fracEnergyToSub = 1.0;
```
wdconinc: Don't use abbreviations; it affects clarity.

src/algorithms/particle/TrackClusterSubtractorConfig.h
```diff
@@ -0,0 +1,33 @@
...
+
+  ///! default mass to use for track energy
```
wdconinc: Not mass but pdg.

src/algorithms/particle/TrackClusterSubtractorConfig.h
```diff
@@ -0,0 +1,33 @@
...
+  int32_t defaultMassPdg = 211;
+
+  ///! index of surface to use for measuring momentum
+  uint64_t surfaceToUse = 1;
```
wdconinc: This is poorly defined; where do these surfaces come from and do they start counting from 0? Do we expect uint64_t many surfaces?

src/algorithms/particle/TrackClusterSubtractorConfig.h
```diff
@@ -0,0 +1,33 @@
...
+
+  ///! turn on/off checking against resolutions
+  bool doNSigmaCut = false;
+
+  ///! max no. of sigma to be consistent w/ zero
+  uint32_t nSigmaMax = 1;
+
+  ///! tracking momentum resolution to use
+  double trkReso = 1.0;
```
wdconinc: Avoid abbreviations.

src/algorithms/particle/TrackClusterSubtractor.h
```diff
@@ -0,0 +1,90 @@
...
+      if (lhs.getObjectID().collectionID == rhs.getObjectID().collectionID) {
+        return (lhs.getObjectID().index < rhs.getObjectID().index);
+      } else {
+        return (lhs.getObjectID().collectionID < rhs.getObjectID().collectionID);
+      }
+    }
+  };
+
+  ///! Alias for vectors of clusters
+  using VecClust = std::vector<edm4eic::Cluster>;
```
wdconinc: Avoid public types if only needed privately.

src/algorithms/particle/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,215 @@
...
+ *          a remnant cluster with the subtracted energy
+ *       b. And create an expected cluster with energy equal
+ *          to the difference between the total and the remnant.
+ *    4. Flag any un-matched clusters as remnants.
+ */
+void TrackClusterSubtractor::process(const TrackClusterSubtractor::Input& input,
+                                     const TrackClusterSubtractor::Output& output) const {
+
+  // grab inputs/outputs
+  const auto [in_match, in_cluster, in_project] = input;
```
wdconinc: Variable names plural for collections.

src/algorithms/particle/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,215 @@
...
+bool TrackClusterSubtractor::is_zero(const double difference) const {
+
+  // if < 0, automatically return true
+  if (difference < 0) {
+    return true;
+  }
+
+  // calculate nSigma
+  const double totReso =
+      std::sqrt((m_cfg.trkReso * m_cfg.trkReso) + (m_cfg.calReso * m_cfg.calReso));
```
wdconinc: You could work in variances instead of taking sqrt.

src/algorithms/particle/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,215 @@
...
+      out_remnant->push_back(remain_clust);
+      trace("Created remnant cluster with {} GeV", remain_clust.getEnergy());
+    }
+
+    // ------------------------------------------------------------------------
+    // 3(b). Create cluster with energy equal to eTotal - eRemnant and match
+    // ------------------------------------------------------------------------
+    auto expect_clust = cluster.clone();
+    expect_clust.setEnergy(cluster.getEnergy() - eSubToUse);
+    out_expect->push_back(expect_clust);
```
wdconinc: Avoid clone and push_back; use create.

src/algorithms/particle/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,215 @@
...
+    return;
+  }
+
+  // emit debugging message if no matched tracks in collection
+  if (in_match->size() == 0) {
+    debug("No matched tracks in collection.");
+  }
+
+  // --------------------------------------------------------------------------
+  // 1. Build map of clusters onto projections
```
wdconinc: This first step seems to indicate we are missing a collection of associations that should be determined ahead of this and would avoid a search.

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,229 @@
...
+  for (const auto& [cluster, projections] : mapClusterToProjections) {
+
+    // do subtraction
+    const double eToSubtract = m_cfg.energyFractionToSubtract * sum_track_energy(projections);
+    const double eSubtracted = cluster.getEnergy() - eToSubtract;
+    trace("Subtracted {} GeV from cluster with {} GeV", eToSubtract, cluster.getEnergy());
+
+    // check if consistent with zero,
+    // set eSub accordingly
+    const bool isZero             = is_zero(eSubtracted);
```
mandrenguyen: Apologies, if my comments are all quite high level. 
The function name `is_zero` hides a bit the fact that if the track "energy" is greater than that of the linked calo clusters it will treat them as if they are compatible. 

I believe you we are inevitably going to have tracks for which we don't want to directly use their kinematics "as is".  I imagine we are always going to want to store tracks with a reasonably high efficiency selection, admitting that some will be poorly measured.  Depending on the working point, you might choose to accept all tracks on the condition of calo compatibility (or not).  However, in the case when the we don't find enough calo energy, I think it's inevitable that one will want to apply a tightened set of track selection criteria to try to reject some fakes.  Of course some of them won't be fakes at all,  but rather muons. But I think that one nevertheless has to leverage the calo compatibility in particle flow against fake tracks.   

This may be more of a comment for the future if the track selection, detector conditions, etc.  is not sufficiently well known yet.   You might consider some place holders though.   Or maybe at this point just a renaming of this function to make it a bit more transparent, and a little comment to indicate what's happening. 

A related comment is that you can also partially handle this in the momentum assignment.  If you don't think the track parameters are trustworthy (e.g., few hits, bad chi2, etc.) you can choose to weight towards the calo measurement, compared to the one from the tracker.

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,229 @@
...
+  if (difference < 0) {
+    return true;
+  }
+
+  // do appropriate comparison
+  bool isZero = false;
+  if (m_cfg.doNSigmaCut) {
+
+    // calculate n sigma squared
+    const double resolution2 = (m_cfg.trackResolution * m_cfg.trackResolution) +
```
mandrenguyen: Are we treating trackResolution as a single external parameter?  Can't we grab the uncertainty directly from the Kalman fit?

src/algorithms/particle_flow/TrackClusterSubtractor.cc
ruse-traveler: Ah! That's a great point! Truth be told, my focus here was more on getting machinery in to do track-cluster subtraction, so I hadn't given much thought to how we would handle fake tracks. But I agree the calo compatibility is a really important metric!

For now, I'd be happy to rename the function to `is_track_energy_greater_than_calo` (or similar),  add some comments, and flag this as a important direction for future development!

I wonder what this might look like in implementation? I could imagine an algorithm that filters tracks based on their compatibility, and the "real" tracks are passed downstream to be converted to particles. Alternatively, I could also see this happening in the particle conversion step...

ruse-traveler: > A related comment is that you can also partially handle this in the momentum assignment. If you don't think the track parameters are trustworthy (e.g., few hits, bad chi2, etc.) you can choose to weight towards the calo measurement, compared to the one from the tracker.

Side note: we actually have something gesturing in this direction in PFA3 currently, where you can take the resolution-weighted average of the track and cluster when calculating the particle energy!

src/algorithms/particle_flow/TrackClusterSubtractor.cc
ruse-traveler: Ah! True! We do store the covariance matrix with the other track parameters, so we could pull the track uncertainty from there. The only catch is that we currently don't propagate any uncertainties on the calorimeter clusters, and I'm not sure how close we are to reasonably calculating them 🫤 

So the rationale behind treating the track resolution as an external parameter was mostly for simplicity, and also to treat it and the calorimeter resolution on similar footing...

src/global/particle_flow/particle_flow.cc
```diff
@@ -0,0 +1,159 @@
...
+  // PFAlpha: baseline PF implementation
+  // ====================================================================
+
+  std::vector<std::pair<double, double>> thetaRanges{
+      {3.0 * dd4hep::degree, 50.0 * dd4hep::degree},
+      {45.0 * dd4hep::degree, 135.0 * dd4hep::degree},
+      {130.0 * dd4hep::degree, 177.0 * dd4hep::degree}};
+
+  auto subDivideByTheta =
+      [&thetaRanges](const edm4eic::TrackSegment& projection) -> std::vector<int> {
```
ruse-traveler:
```suggestion
      [thetaRanges](const edm4eic::TrackSegment& projection) -> std::vector<int> {
```

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,215 @@
...
+    return;
+  }
+
+  // emit debugging message if no matched tracks in collection
+  if (in_match->size() == 0) {
+    debug("No matched tracks in collection.");
+  }
+
+  // --------------------------------------------------------------------------
+  // 1. Build map of clusters onto projections
```
ruse-traveler: I think this would be a good direction to explore, but I'd also like to run it past the PWGs & DSCs since it would have implications for analyses... Would you be okay flagging this as a follow-up item?

src/algorithms/particle_flow/TrackClusterSubtractor.cc
ruse-traveler: Just pushed some changes renaming `is_zero` to be a little more descriptive of the actual check!

src/algorithms/particle_flow/TrackClusterSubtractor.cc
ruse-traveler: Just pushed some changes which remove `trackResolution` and instead use the covariance on the projection momentum directly!

src/factories/particle_flow/particle_flow.cc
```diff
@@ -0,0 +1,93 @@
...
+  // ====================================================================
+
+  // backward -----------------------------------------------------------
+
+  app->Add(new JOmniFactoryGeneratorT<TrackClusterSubtractor_factory>(
+      "EcalEndcapNRemnantClusters",
+      {"EcalEndcapNTrackClusterMatches", "EcalEndcapNClusters", "CalorimeterTrackProjections"},
+      {"EcalEndcapNRemnantClusters", "EcalEndcapNExpectedClusters",
+       "EcalEndcapNTrackExpectedClusterMatches"},
+      {.fracEnergyToSub = 1.0, .defaultMassPdg = 211, .surfaceToUse = 1},
```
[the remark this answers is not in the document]
ruse-traveler: In the latest round of commits, I pushed some changes that filter the projections by theta to reduce some of the combinatorics!

src/algorithms/particle_flow/TrackClusterSubtractorConfig.h
```diff
@@ -0,0 +1,33 @@
...
+struct TrackClusterSubtractorConfig {
+
+  ///! fraction of track energy to subtract
+  double fracEnergyToSub = 1.0;
+
+  ///! default mass to use for track energy
+  int32_t defaultMassPdg = 211;
+
+  ///! index of surface to use for measuring momentum
+  uint64_t surfaceToUse = 1;
```
ruse-traveler: In the latest round of commits I reduced the `unit64_t` to a `unint8_t` and pointed out the algorithm! When testing out some stuff, I also confirmed that the projections start at 1 (e.g. in the attached screenshot)...

SampleProjections

ruse-traveler:
All right! I think this is ready for another round of reviewing! The biggest changes in the latest round of commits are:

1. Projections are now separated by theta ahead of the algorithm in an effort to mitigate combinatorics; and
2. I've removed the `trackResolution` parameter in favor of working directly with the projection covariance matrix.

I think the combinatorics can be further mitigated by introducing a link between projections and clusters, but I feel that that is a little out-of-scope of this PR.

And -- much further down the road -- it would be preferable to treat the clusters and tracks on equal footing and use the variances from both rather than treating the calorimeter resolution as an external parameter. But I think we're a long ways away from calculating/using the cluster variances...

veprbl:
```
In file included from /home/runner/work/EICrecon/EICrecon/src/global/particle_flow/particle_flow.cc:21:
/home/runner/work/EICrecon/EICrecon/src/factories/particle_flow/TrackClusterSubtractor_factory.h:47:76: error: no member named 'trackResolution' in 'eicrecon::TrackClusterSubtractorConfig'
   47 |   ParameterRef<double> m_trackResolution{this, "trackResolution", config().trackResolution};
      |                                                                   ~~~~~~~~ ^
1 error generated.
```

src/factories/particle_flow/TrackClusterSubtractor_factory.h
```diff
@@ -0,0 +1,76 @@
...
+#endif
+
+  // parameter bindings
+  ParameterRef<double> m_energyFractionToSubtract{this, "energyFractionToSubtract",
+                                                  config().energyFractionToSubtract};
+  ParameterRef<int32_t> m_defaultPDG{this, "defaultPDG", config().defaultPDG};
+  ParameterRef<uint8_t> m_surfaceToUse{this, "surfaceToUse", config().surfaceToUse};
+  ParameterRef<bool> m_doNSigmaCut{this, "doNSigmaCut", config().doNSigmaCut};
+  ParameterRef<uint32_t> m_nSigmaMax{this, "nSigmaMax", config().nSigmaMax};
+  ParameterRef<double> m_trackResolution{this, "trackResolution", config().trackResolution};
```
ruse-traveler, line 46:
```suggestion
```

ruse-traveler:
> ```
> In file included from /home/runner/work/EICrecon/EICrecon/src/global/particle_flow/particle_flow.cc:21:
> /home/runner/work/EICrecon/EICrecon/src/factories/particle_flow/TrackClusterSubtractor_factory.h:47:76: error: no member named 'trackResolution' in 'eicrecon::TrackClusterSubtractorConfig'
>    47 |   ParameterRef<double> m_trackResolution{this, "trackResolution", config().trackResolution};
>       |                                                                   ~~~~~~~~ ^
> 1 error generated.
> ```

😩 Whoops! Just fixed it...

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,276 @@
...
+  ///! lambda to compute momentum variance for a track projected to a point
+  auto getVariance = [&getRightProduct](const edm4eic::TrackPoint& point) {
+    double numerator   = 0.0;
+    double denominator = 0.0;
+    for (std::size_t i = 0; i < 3; ++i) {
+      numerator += point.momentum[i] * getRightProduct(point, i);
+      denominator += point.momentum[i] * point.momentum[i];
+    }
+    return denominator < std::numeric_limits<double>::epsilon() ? 0.0 : numerator / denominator;
+  };
```
veprbl: This is *not* how you calculate the variance of a Eucledian norm if you know vector itself and its covariance.

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,276 @@
...
+
+    // measure momentum at specified surface
+    double momentum         = 0.0;
+    double momentumVariance = 0.0;
+    for (const auto& point : project.getPoints()) {
+      if (point.surface != m_cfg.surfaceToUse) {
+        continue;
+      } else {
+        momentum         = edm4hep::utils::magnitude(point.momentum);
+        momentumVariance = getVariance(point);
```
veprbl: Here it is assumed that momentum is a 3 vector, like it says in
https://github.com/eic/EDM4eic/blob/7d7705d9bbd4794b1b875c04258e7d986aea8a47/edm4eic.yaml#L166-L167
Well, that documentation lies to you, here is what it really is:
https://github.com/eic/EICrecon/blob/a48dc9371f3459fe3c04ec866b8a38b319cffd71/src/algorithms/tracking/TrackProjector.cc#L133-L136
Perhaps, you only want modulo of the first component.

src/global/particle_flow/particle_flow.cc
```diff
@@ -24,6 +35,128 @@ void InitPlugin(JApplication* app) {
...
+      {.energyFractionToSubtract = 1.0, .defaultPDG = 211, .surfaceToUse = 1},
+      app // TODO: remove me once fixed
+      ));
+
+  app->Add(new JOmniFactoryGeneratorT<TrackClusterSubtractor_factory>(
+      "HcalBarrelRemnantClusters",
+      {"HcalBarrelTrackClusterMatches", "HcalBarrelClusters", "BarrelCalorimeterTrackProjections"},
+      {"HcalBarrelRemnantClusters", "HcalBarrelExpectedClusters",
+#if EDM4EIC_BUILD_VERSION >= EDM4EIC_VERSION(8, 7, 0)
+       "HcalBarrelTrackExpectedClusterLink",
```
veprbl:
```suggestion
       "HcalBarrelTrackExpectedClusterLinks",
```

src/factories/particle_flow/TrackClusterSubtractor_factory.h
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2025 Derek Anderson
```
veprbl:
```suggestion
// Copyright (C) 2026 Derek Anderson
```

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,276 @@
...
+#include <edm4eic/Cov3f.h>
+#include <edm4eic/Track.h>
+#include <edm4eic/TrackPoint.h>
+#include <edm4hep/Vector3f.h>
+#include <edm4hep/utils/vector_utils.h>
+#include <podio/ObjectID.h>
+#include <podio/RelationRange.h>
+#include <podio/detail/Link.h>
+#include <podio/detail/LinkCollectionImpl.h>
+#include <stdint.h>
```
veprbl:
```suggestion
#include <cstdint>
```
I hope, IWYU will be happy with that too.

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,276 @@
...
+
+    // calculate n sigma squared
+    const double totalVariance =
+        variance + (m_cfg.calorimeterResolution * m_cfg.calorimeterResolution);
+    const uint32_t nSigma2 =
+        static_cast<uint32_t>(std::floor((difference * difference) / totalVariance));
+    const uint32_t nSigmaMax2 = m_cfg.nSigmaMax * m_cfg.nSigmaMax;
+
+    isGreaterThan = (nSigma2 < nSigmaMax2);
+    trace("Within {} NSigma^2, track energy sum greater than calorimeter cluster: difference = {} "
```
veprbl: Did you want to have `if (isGreaterThan) {` here?

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,276 @@
...
+      trace("Created remnant cluster with {} GeV", remain_cluster.getEnergy());
+    }
+
+    // ------------------------------------------------------------------------
+    // 3(b). Create cluster with energy equal to eTotal - eRemnant and match
+    // ------------------------------------------------------------------------
+    auto expect_cluster = cluster.clone();
+    expect_cluster.setEnergy(cluster.getEnergy() - eSubtractedToUse);
+    out_expectants->push_back(expect_cluster);
+    trace("Created subtracted cluster with {} GeV (originally {} GeV)", expect_cluster.getEnergy(),
```
veprbl:
```suggestion
    trace("Created expected cluster with {} GeV (originally {} GeV)", expect_cluster.getEnergy(),
```
Would be in agreement to messages below.

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,276 @@
...
+            project.getTrack().getObjectID().index);
+    }
+  } // end cluster-to-projections loop
+  debug("Finished subtraction, {} remnant clusters and {} expected clusters", out_remnants->size(),
+        out_expectants->size());
+
+  // --------------------------------------------------------------------------
+  // 4. Any unmatched clusters are remnants by definition
+  // --------------------------------------------------------------------------
+  for (const auto& cluster : vecNoMatchClust) {
```
veprbl: There is no other use to vecNoMatchClust, why not merge this and previous loop?

src/global/particle_flow/particle_flow.cc
```diff
@@ -24,6 +35,128 @@ void InitPlugin(JApplication* app) {
...
   // ====================================================================
 
+  // --------------------------------------------------------------------
+  // PFA (1a) arbitration: apply track correction to clusters
+  // --------------------------------------------------------------------
+
+  std::vector<std::pair<double, double>> thetaRanges{
+      {3.0 * dd4hep::degree, 50.0 * dd4hep::degree},
+      {45.0 * dd4hep::degree, 135.0 * dd4hep::degree},
+      {130.0 * dd4hep::degree, 177.0 * dd4hep::degree}};
```
veprbl: These numbers will require some tweaking later. We need to file an issue for that.

src/global/particle_flow/particle_flow.cc
```diff
@@ -24,6 +35,128 @@ void InitPlugin(JApplication* app) {
...
+      [thetaRanges](const edm4eic::TrackSegment& projection) -> std::vector<int> {
+    const double theta = edm4hep::utils::anglePolar(projection.getTrack().getMomentum());
+    std::vector<int> binIndices;
+    for (std::size_t iTheta = 0; iTheta < thetaRanges.size(); ++iTheta) {
+      if ((theta >= thetaRanges[iTheta].first) && (theta < thetaRanges[iTheta].second)) {
+        binIndices.push_back(iTheta);
+      }
+    }
+    return binIndices;
+  };
```
veprbl: Why subdivide by theta, should it not be by system IDs (encoded as "extra" in acts surface id's)?

veprbl (commented):
Nice piece of functionality and is long overdue. I have some comments.

src/algorithms/particle_flow/TrackClusterSubtractor.cc
ruse-traveler: Ah! Good point!

src/algorithms/particle_flow/TrackClusterSubtractor.cc
ruse-traveler: Whoops! Thank you for catching that!

src/global/particle_flow/particle_flow.cc
ruse-traveler: True! System IDs would also work here! I went with theta initially just because I was thinking to divide up the projections by hemisphere, and there was an already existing example of using theta with the `FilterMatching` algorithm.

We could also get the system IDs from the points, too. Maybe something like this?
```c++
auto subdivideBySystemID = [systemIDs](const edm4eic::TrackSegment& projection) -> std::vector<int> {
  const int id = projection.getPoints()[0].system
  std::vector<int> idIndices;
  for (std::size_t iID = 0; iID < systemIDs.size(); ++iID) {
    if (id == systemIDs[iID]) {
      idIndices.push_back(iID);
    }
  }
  return idIndices;
};
```

veprbl: Looks good to me.

src/algorithms/particle_flow/TrackClusterSubtractor.cc
ruse-traveler: Following up on the discussion in the Reco WG yesterday, I've reverted the variance calculation back to the "dumb" approach of treating the resolution as a parameter. I've opened up a issue to flag this as an area to improve on:

https://github.com/eic/EICrecon/issues/2687

Echoing my thoughts from the meeting: this most likely isn't difficult to implement, but I don't trust myself to be able to work though it & validate it quick enough to make it in the release. And I think at this stage even the less sophisticated version is still a good step forward!

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,237 @@
...
+    isGreaterThan = (nSigma2 < nSigmaMax2);
+    if (isGreaterThan) {
+      trace(
+          "Within {} NSigma^2, track energy sum greater than calorimeter cluster: difference = {} "
+          "GeV, nSigma^2 = {}",
+          nSigmaMax2, difference, nSigma2);
+    }
+  } else {
+    isGreaterThan = difference < std::numeric_limits<double>::epsilon();
+    trace("Track energy sum greater than calorimeter cluster: difference = {} GeV", difference);
```
veprbl:
```suggestion
    if (isGreaterThan) {
      trace("Track energy sum greater than calorimeter cluster: difference = {} GeV", difference);
    }
```

src/global/particle_flow/particle_flow.cc
```diff
@@ -101,6 +110,139 @@ void InitPlugin(JApplication* app) {
...
+       "LFHCALTrackExpectedClusterLinks",
+#endif
+       "LFHCALTrackExpectedClusterMatches"},
+      {.energyFractionToSubtract = 1.0, .defaultPDG = 211, .surfaceToUse = 1},
+      app // TODO: remove me once fixed
+      ));
+
+  app->Add(new JOmniFactoryGeneratorT<TrackClusterSubtractor_factory>(
+      "HcalEndcapPInsertRemnantClusters",
+      {"HcalEndcapPInsertTrackSplitMergeClusterMatches", "HcalEndcapPInsertClusters",
```
veprbl:
```

Error: JEventProcessorPODIO] [error] Omitting PODIO collection 'HcalEndcapPInsertExpectedClusters' due to exception: JOmniFactory: Failed to get collection HcalEndcapPInsertTrackSplitMergeClusterMatches: Missing databundle with uniquename 'HcalEndcapPInsertTrackSplitMergeClusterMatches'.
Error: JEventProcessorPODIO] [error] Omitting PODIO collection 'HcalEndcapPInsertRemnantClusters' due to exception: JOmniFactory: Failed to get collection HcalEndcapPInsertTrackSplitMergeClusterMatches: Missing databundle with uniquename 'HcalEndcapPInsertTrackSplitMergeClusterMatches'.
Error: JEventProcessorPODIO] [error] Omitting PODIO collection 'HcalEndcapPInsertTrackExpectedClusterLinks' due to exception: JOmniFactory: Failed to get collection HcalEndcapPInsertTrackSplitMergeClusterMatches: Missing databundle with uniquename 'HcalEndcapPInsertTrackSplitMergeClusterMatches'.
Error: JEventProcessorPODIO] [error] Omitting PODIO collection 'HcalEndcapPInsertTrackExpectedClusterMatches' due to exception: JOmniFactory: Failed to get collection HcalEndcapPInsertTrackSplitMergeClusterMatches: Missing databundle with uniquename 'HcalEndcapPInsertTrackSplitMergeClusterMatches'.
```
https://github.com/eic/EICrecon/actions/runs/26891856212/job/79320942404#step:7:643

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,237 @@
...
+    // measure momentum at specified surface
+    double momentum = 0.0;
+    for (const auto& point : project.getPoints()) {
+      if (point.surface != m_cfg.surfaceToUse) {
+        continue;
+      } else {
+        momentum = edm4hep::utils::magnitude(point.momentum);
+        break;
+      }
+    }
```
veprbl:
```suggestion
    double momentum = 0.0;
    bool momentum_valid = false;
    for (const auto& point : project.getPoints()) {
      if (point.surface != m_cfg.surfaceToUse) {
        continue;
      } else {
        momentum = edm4hep::utils::magnitude(point.momentum);
        momentum_valid = true;
        break;
      }
    }
    if (!momentum_valid) {
      continue;
    }
```

src/global/particle_flow/particle_flow.cc
ruse-traveler: Whoops! Forgot that we don't pass the insert matches through the split-merger yet 😅

ruse-traveler:
Thanks for the feedback @veprbl ! Just pushed a round of commits integrating the suggestions!

src/algorithms/particle_flow/TrackClusterSubtractor.cc
```diff
@@ -0,0 +1,244 @@
...
+
+    // create a track-cluster match for expected clusters
+    for (const auto& project : projections) {
+#if EDM4EIC_BUILD_VERSION >= EDM4EIC_VERSION(8, 7, 0)
+      edm4eic::MutableTrackClusterLink link = out_links->create();
+      link.setFrom(expect_cluster);
+      link.setTo(project.getTrack());
+      link.setWeight(1.0); // FIXME placeholder
+      trace("Matched expected cluster {} to track {}", expect_cluster.getObjectID().index,
+            project.getTrack().getObjectID().index);
```
veprbl:
```suggestion
```

veprbl: duplicated trace

ruse-traveler: Whoops! Good catch!

veprbl (approved):
LGTM

PR #1629, opened 2024-10-10
fix(ci): avoid python3.10 reference; don't install bokeh, pygithub

wdconinc:
We are using python3.11 now, so the `pip install` would install to `~/.local/lib/python3.11/site-packages` if it did anything, and the hardcoded `PYTHONPATH` prepend would then fail to find the module. That this continued to work demonstrates that we don't need to install anything locally, and can get rid of the `PYTHONPATH` modifications too.

This PR removes the `PYTHONPATH` additions and the custom `pip install`. If an update to the bokeh and pygithub packages are needed, then this is better handled by updating what is in the container, now that we have reached a more stable capybara stage of development.

- [x] Bug fix (issue: hardcoded reference to python3.10)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -769,7 +769,6 @@ jobs:
         platform-release: "${{ env.platform }}:${{ env.release }}"
         setup: "/opt/detector/epic-${{ env.detector-version }}/bin/thisepic.sh"
         run: |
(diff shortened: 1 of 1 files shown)


veprbl (commented):
No objections, of course

veprbl (approved):

PR #1630, opened 2024-10-10
fix(ci): always run capybara, even when ref is not found

wdconinc:
When, for some reason, the reference artifacts on `main` are not available (typically recovering from a failing `main`), then we don't generate capybara output and the `build-docs` job fails when trying to merge the capy files. E.g. https://github.com/eic/EICrecon/actions/runs/11282255676/job/31379983435. There is no gracious "don't fail when you can't merge anything" option.

This PR ensures that we always have capy files generated, even when we don't actually have a reference to compare to. The only thing that will be included if there are is reference is the new file. That's still useful information to plot.

- [x] Bug fix (issue https://github.com/eic/EICrecon/actions/runs/11282255676/job/31379983435)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -764,7 +764,6 @@ jobs:
         if_no_artifact_found: warn
     - name: Compare to previous artifacts
       uses: eic/run-cvmfs-osg-eic-shell@main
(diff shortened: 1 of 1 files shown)


veprbl (approved):
This should work

PR #1633, opened 2024-10-15
EEEMCal: remove 2% smearing

veprbl:
This comes from Athena times where all calorimeters were smeared by 2%. We don't expect this term, we want to have something more motivated like #1064, instead. Submitting separately to confirm the effect.


Chao1009 (approved):
Yes, the smearing is better to be always zero unless there are some unknown sources for the resolution.

PR #1634, opened 2024-10-15
Fix RP reco bug in p_y

ajentsch:

This PR introduces a bug fix for the p_y reconstruction in the Roman pots, and also switches the reco to global coordinates. There are some issues with the basic RP concept and how local coordinates are calculated using the cellIDs. It means an additional set of shifts would need to be calculated to do a proper transformation to the coordinate system local to the proton orbit. Using global coordinates simplifies this to the same procedure as before, but with different offsets (the matrices themselves don't actually change).

- [ xx] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x ] Tests for the changes have been added

See attached slides for some discussion and comparison plots.

No, it fixes the default momentum reconstruction to be correct (within the limits of how the static matrix approach works). Users will only need to change the axis limits on their 2D plots of the hits on the Roman pots (see the Roman pots geometry definition to pick a more-reasonable range).

Yes, it improves the pT reconstruction for the Roman pots protons.
[ePIC_roman_pots_reco_matrix_bug_fix_10_15_2024.pdf](https://github.com/user-attachments/files/17382021/ePIC_roman_pots_reco_matrix_bug_fix_10_15_2024.pdf)

### DIFF
--- a/src/algorithms/fardetectors/MatrixTransferStatic.cc
+++ b/src/algorithms/fardetectors/MatrixTransferStatic.cc
@@ -252,6 +252,8 @@ void eicrecon::MatrixTransferStatic::process(
         }
       }
 
+	  Yip[1] = Yrp[0]/aY[0][1];
+
       // convert polar angles to radians
       double rsx = Xip[1] * dd4hep::mrad;
       double rsy = Yip[1] * dd4hep::mrad;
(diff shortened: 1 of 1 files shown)


veprbl (approved):
Alex says this is tested

ajentsch:
> Alex says this is tested

Yes, and there should be a PDF attached which details how the bug was uncovered, and some plots showing the comparisons, etc.

epic-capybara:
Successfully created backport PR for `v1.18`:
- #1636

Issue #1659, opened 2024-10-30
 Last tracker hits not used in CKF


ShujieL:
### Environment: (where does this bug occur, have you tried other environments)
main branch. v24.09
### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. open a single particle reconstruction file, e.g. 5 GeV pion+ such that for every event, the particle should go through all tracking layers. 
2. check the number of events with at least one valid tracker digi hit on the last tracker layer (outer MPGD barrel, or MPGD/AC-LGAD disk). 
3. pull out the good measurements that are associated to each track, and check if there is at least one measurement on the last layer.

### Expected Result: (what do you expect when you execute the steps above)
The number of events with digitized hits on the last layer should equal (up to the efficiency) to the number of tracks reconstructed with a measurement on the last layer

### Actual Result: (what do you get when you execute the steps above)
Attached are the comparison of sim hits, digi hits, and measurement on each layer from
EPIC/RECO/24.09.0/epic_craterlake/SINGLE/pi+/10GeV/45to135deg/pi+_10GeV_45to135deg.0001.eicrecon.tree.edm4eic.root
While most of the MPGD hits are digitized after the threshold change at https://github.com/eic/epic/issues/774, only half of those hits are used in track recon.

<img width="600" alt="Screenshot 2024-10-29 at 21 05 03" src="https://github.com/user-attachments/assets/e389ef0c-1552-4224-b7c7-984e2260850c">

ShujieL:
@simonge suggested me to check the phi distribution of TOF. While that looks alright,  the outer MPGD missing hits definitely have some pattern.

@mposik1983 any thoughts? 
<img width="593" alt="Screenshot 2024-10-30 at 22 48 51" src="https://github.com/user-attachments/assets/aa22520e-7f75-44e8-b02a-2693a5b1367a">

mposik1983:
Hi Shujie, thanks for looking at this. Do you happen to have the comparisons for theta, x,y, and z distributions?

________________________________
From: ShujieL ***@***.***>
Sent: Thursday, October 31, 2024 1:52:07 AM
To: eic/EICrecon ***@***.***>
Cc: Matthew Posik ***@***.***>; Mention ***@***.***>
Subject: [External] Re: [eic/EICrecon] Last tracker hits not used in CKF (Issue #1659)


[...]

PR #1660, opened 2024-10-30
Fix naming of BTOF hits

simonge:
Changes the name of the BTOF hit collections to use Hits instead of Hit. This was inconsistent between the name output from the hit reconstruction and collection collector input.

- [x] Bug fix (issue #1659 probably some of the cause at least)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Hopefully adds the BTOF into the tracking


simonge:
Turns out this does nothing functional because the discrepancy between `TOFBarrelRawHits` and `TOFBarrelRawHit` in
https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L53
isn't actually used.

veprbl (approved):

PR #1662, opened 2024-11-02
TrackParamTruthInitConfig: maxEtaForward=10 for B0 coverage

veprbl:
We should cover down to 5.5 mrad.
The new value eta=10 covers that:
```
>>> import numpy as np
>>> np.rad2deg(2*np.arctan(np.exp(-10)))
0.005202448727587163
```

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/tracking/TrackParamTruthInitConfig.h
+++ b/src/algorithms/tracking/TrackParamTruthInitConfig.h
@@ -12,7 +12,7 @@ struct TrackParamTruthInitConfig {
     double maxVertexY       = 80  * dd4hep::mm;
     double maxVertexZ       = 200 * dd4hep::mm;
     double minMomentum      = 100 * dd4hep::MeV;
-    double maxEtaForward    = 6.0;
+    double maxEtaForward    = 10.0;
     double maxEtaBackward   = 4.1;
     double momentumSmear    = 0.1;
 


michael-pitt:
I need to test it, but I think the covered eta is up to 6 (we usually report eta<5.9):
```
>>>-np.log(np.tan(5.5e-3/2))
5.896151846465876
```
The inner radius of the tracking layers defines this. If the tracking layer is 3.5cm from the beam center, which at the distance of 630cm gives this opening angle of `np.arctan(3.5/630)`.

ajentsch:
Going to an eta of 10 is putting us in the vicinity of the Roman pots and the OMD, which are both silicon trackers, and would potentially allow for very unphysical tracks to be reconstructed with ACTS. I would recommend only going to eta of 6.0.

veprbl:
Yeah, I think there are couple mistakes in my estimate, I target 5.5 mdeg instead of 5.5 mrad. And also it doesn’t account for the crossing angle. Yet it works to at least have some seeds get produced, instead of not having any.

veprbl:
Based on what's discussed on Mattermost, it sounds like this isn't useful.

PR #1663, opened 2024-11-02
CKFTracking: gracefully handle failures when extrapolating to perigee surface

veprbl:
Starting with Acts 34.x the CKF fitting had been divorced from propagation to the target surface https://github.com/acts-project/acts/commit/d9f775f4155f1a13e316aa142e939ef7178ff665, so we now have to do it ourselves:
https://github.com/eic/EICrecon/blob/5e39885644bdf867aa2060ccc46da2500678c1da/src/algorithms/tracking/CKFTracking.cc#L278-L296
However the catch is that B0 CKF tracks may be affected by failures to extrapolate explicitly. The failure currently leads to a null pointer dereference. This PR changes that to a mere warning, the track is omitted from the output in the process. We would be able to upgrade to Acts 34 with this change.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -336,6 +336,15 @@ namespace eicrecon {
 
         std::optional<unsigned int> lastSeed;
         for (const auto& track : constTracks) {
+#if Acts_VERSION_MAJOR >= 34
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
Looks good.

PR #1668, opened 2024-11-06
Enable calorimeter hit merging by functions

ruse-traveler:

This PR extends the functionality of the `CalorimeterHitsMerger` algorithm. Previously, reconstructed hits could only be merged across a given field of the readout of a calorimeter. This presents a challenge for calorimeters such as the Barrel HCal where

1. our readout has no segmentation beyond just eta  & phi, and
2. it could be useful to study the response of the detector as a function of how many readout channels we gang together after reconstruction.

This PR addresses this point by utilizing the `EvaluatorSvc` in a manner similar to the `adjacencyMatrix`, `peakNeighbourhoodMatrix` of the `CalorimeterIslandClustering` and the `sampFrac` of `CalorimeterHitReco`. Now the user has the ability to specify an (almost) arbitrarily complex transformation for a specific field of the readout via the `fieldTransformations` parameter, which defines both the field to transform and the function to map the indices of that field onto the desired reference indices.

For example:

```
app->Add(new JOmniFactoryGeneratorT<CalorimeterHitsMerger_factory>(
  "HcalBarrelMergedHits", {"HcalBarrelRecHits"}, {"HcalBarrelMergedHits"},
  {
    .readout = "HcalBarrelHits",
    .fieldTransformations = {"phi:phi-(5*((phi/5)-floor(phi/5)))"}
  },
));
```

Here, the `HcalBarrelMergedHits` collection will merge 5 hits (i.e. scintillator tiles for the BHCal) adjacent in phi into a one with the position and cellID of the 1st  of the 5, and no hits will be merged along eta.

The previous behavior of the algorithm can be recovered by simply specifying the index to be mapped onto. For example:
```
app->Add(new JOmniFactoryGeneratorT<CalorimeterHitsMerger_factory>(
  "HcalEndcapNMergedHits", {"HcalEndcapNRecHits"}, {"HcalEndcapNMergedHits"},
  {
    .readout = "HcalEndcapNHits",
    .fieldTransformations = {"layer:4", "slice:0"}
  },
));
```

An example script of to change a transformation (and how to update the adjacency matrix accordingly) from the command-line is provided in the snippets repo [here](https://github.com/eic/snippets/blob/main/Calorimetery/CaloDebugTools/UtilityScripts/RunEICReconWithTileMerging.rb).

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1669 )
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
@@ -43,6 +43,25 @@ void CalorimeterHitsMerger::init() {
         return;
     }
 
+    // check that input field, mask, and transformation vectors
+    // are the same length
+    if (m_cfg.fields.size() != m_cfg.fieldRefs.size()) {
+      error(
+        "Size of field and reference mask vectors are different ({} vs. {}).",
+        m_cfg.fields.size(),
+        m_cfg.fieldRefs.size()
+      );
+      return;
+    }
+    if (m_cfg.fields.size() != m_cfg.fieldTransformations.size()) {
+      error(
+        "Size of field and transformation vectors are different ({} vs. {}).",
+        m_cfg.fields.size(),
+        m_cfg.fieldTransformations.size()
+      );
+      return;
+    }
+
     // initialize descriptor + decoders
     try {
       id_desc = m_detector->readout(m_cfg.readout).idSpec();
@@ -56,46 +75,42 @@ void CalorimeterHitsMerger::init() {
 //        throw std::runtime_error(mess);
     }
 
-
-/* TO REMOVE
-    // if no field-by-field transformations provided, initialize relevant bitmasks
-    // otherwise intialize relevant functionals
-    if (m_cfg.fieldTransformations.empty()) {
-      id_mask = 0;
-      std::vector<RefField> ref_fields;
-      for (size_t i = 0; i < m_cfg.fields.size(); ++i) {
-          id_mask |= id_desc.field(m_cfg.fields[i])->mask();
-          // use the provided id number to find ref cell, or use 0
-          int ref = i < m_cfg.fieldRefs.size() ? m_cfg.fieldRefs[i] : 0;
-          ref_fields.emplace_back(m_cfg.fields[i], ref);
+    // lambda to translate IDDescriptor fields into function parameters
+    std::function hit_transform = [this](const edm4eic::CalorimeterHit& hit) {
+      std::unordered_map<std::string, double> params;
+      for (const auto& name_field : id_desc.fields()) {
+        params.emplace(name_field.first, name_field.second->value(hit.getCellID()));
+        trace("{} = {}", name_field.first, name_field.second->value(hit.getCellID()));
       }
-      ref_mask = id_desc.encode(ref_fields);
-      id_mask = ~id_mask;
-      debug("ID mask in {:s}: {:#064b}", m_cfg.readout, id_mask);
-    } else {
+      return params;
+    };
 
-      // lambda to translate IDDescriptor fields into function parameters
-      std::function hit_to_map = [this](const edm4eic::CalorimeterHit& hit) {
-        std::unordered_map<std::string, double> params;
-        for (const auto& name_field : id_desc.fields()) {
-          params.emplace(name_field.first, name_field.second->value(hit.getCellID()));
-          trace("{} = {}", name_field.first, name_field.second->value(hit.getCellID()));
-        }
-        return params;
-      };
+    // loop through provided readout fields
+    auto& svc = algorithms::ServiceSvc::instance();
+    for (std::size_t iField = 0; std::string& field : m_cfg.fields) {
 
-      // intialize functions
-      //   - NOTE this assumes provided fields are 1-to-1!
-      auto& svc = algorithms::ServiceSvc::instance();
-      for (std::size_t iMap = 0; const auto& mapping : m_cfg.fieldTransformations) {
-        if (iMap < m_cfg.fields.size()) {
-          ref_maps[m_cfg.fields.at(iMap)] = svc.service<EvaluatorSvc>("EvaluatorSvc")->compile(mapping, hit_to_map);
-          trace("Mapping for field {} = {}", m_cfg.fields.at(iMap), mapping);
-        }
-        ++iMap;
-      }  // end loop over fieldTransformations
-    }
-*/
+      // grab provided field reference masks
+      // and transformations
+      uint64_t field_mask = m_cfg.fieldRefs.at(iField);
+      std::string field_transfrom = m_cfg.fieldTransformations.at(iField);
+
+      // grab name and value of provided field
+      auto name_field = id_desc.field(field);
+
+      // set mask or transformation for each field
+      //   - if no transformation provided, reference
+      //     mask will be used
+      if (field_transform.empty()) {
+        ref_maps[field] = [field_mask, name_field](const edm4eic::CalorimterHit& hit) -> int {
+          return (name_field->value(hit.cellID()) & ~field_mask) | field_mask;
+        };
+        trace("{}: using mask {:#064b}", field, field_mask);
+      } else {
+        ref_maps[field] = svc.service<EvaluatorSvc>("EvaluatorSvc")->compile(field_transform, hit_transform);
+        trace("{}: using transformation '{}'", field, field_transform);
+      }
+      ++iField;
+    }  // end field loop
 
 }
 
@@ -186,35 +201,6 @@ void CalorimeterHitsMerger::build_map_via_funcs(
   MergeMap& merge_map
 ) const {
 
-  // throw error if field and transformation vectors
-  // are different sizes 
-  if (m_cfg.fields.size() != m_fieldTransformations.size()) {
-    error(
-      "field and transformation vectors are different sizes ({} vs. {}). Vectors should be 1-to-1.",
-      m_cfg.fields.size(),
-      m_fieldTransformations.size()
-    );
-    return;
-  }
-
-  // loop over hits
-  std::vector<RefField> ref_fields;
-  for (std::size_t iHit = 0; const auto& hit : *in_hits) {
-
-    // loop through readout fields
-    for (std::size_t iField = 0; const auto& name_field : id_desc.fields()) {
-
-
-
-    }  // end field loop
-
-    // add hit to appropriate group
-    merge_map[ref_id].push_back(iHit);
-    ++iHit;
-
-  }  // end hit loop
-
-/* TO REMOVE
   std::vector<RefField> ref_fields;
   for (std::size_t iHit = 0; const auto& hit : *in_hits) {
 
@@ -245,7 +231,6 @@ void CalorimeterHitsMerger::build_map_via_funcs(
     // add hit to appropriate group
     merge_map[ref_id].push_back(iHit);
     ++iHit;
-*/
 
   }  // end hit loop
 
--- a/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
@@ -56,15 +56,15 @@ void CalorimeterHitsMerger::init() {
 //        throw std::runtime_error(mess);
     }
 
-    // if no field-by-field mappings provided, initialize relevant bitmasks
+    // if no field-by-field transformations provided, initialize relevant bitmasks
     // otherwise intialize relevant functionals
-    if (m_cfg.mappings.empty()) {
+    if (m_cfg.fieldTransformations.empty()) {
       id_mask = 0;
       std::vector<RefField> ref_fields;
       for (size_t i = 0; i < m_cfg.fields.size(); ++i) {
           id_mask |= id_desc.field(m_cfg.fields[i])->mask();
           // use the provided id number to find ref cell, or use 0
-          int ref = i < m_cfg.refs.size() ? m_cfg.refs[i] : 0;
+          int ref = i < m_cfg.fieldRefs.size() ? m_cfg.fieldRefs[i] : 0;
           ref_fields.emplace_back(m_cfg.fields[i], ref);
       }
       ref_mask = id_desc.encode(ref_fields);
@@ -85,13 +85,13 @@ void CalorimeterHitsMerger::init() {
       // intialize functions
       //   - NOTE this assumes provided fields are 1-to-1!
       auto& svc = algorithms::ServiceSvc::instance();
-      for (std::size_t iMap = 0; const auto& mapping : m_cfg.mappings) {
+      for (std::size_t iMap = 0; const auto& mapping : m_cfg.fieldTransformations) {
         if (iMap < m_cfg.fields.size()) {
           ref_maps[m_cfg.fields.at(iMap)] = svc.service<EvaluatorSvc>("EvaluatorSvc")->compile(mapping, hit_to_map);
           trace("Mapping for field {} = {}", m_cfg.fields.at(iMap), mapping);
         }
         ++iMap;
-      }  // end loop over mappings
+      }  // end loop over fieldTransformations
     }
 
 }
@@ -105,7 +105,7 @@ void CalorimeterHitsMerger::process(
 
     // find the hits that belong to the same group (for merging)
     MergeMap merge_map;
-    if (m_cfg.mappings.empty()) {
+    if (m_cfg.fieldTransformations.empty()) {
       for (std::size_t ix = 0; const auto &h : *in_hits) {
         uint64_t id = h.getCellID() & id_mask;
         merge_map[id].push_back(ix);
--- a/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitsMerger.cc
@@ -12,16 +12,17 @@
 
 #include <DD4hep/Alignments.h>
 #include <DD4hep/DetElement.h>
-#include <DD4hep/IDDescriptor.h>
 #include <DD4hep/Objects.h>
 #include <DD4hep/Readout.h>
 #include <DD4hep/VolumeManager.h>
 #include <DDSegmentation/BitFieldCoder.h>
+#include <edm4eic/CalorimeterHit.h>  // NEW
 #include <Evaluator/DD4hepUnits.h>
 #include <Math/GenVector/Cartesian3D.h>
 #include <Math/GenVector/DisplacementVector3D.h>
 #include <fmt/core.h>
 #include <algorithm>
+#include <algorithms/service.h>  // NEW
 #include <cmath>
 #include <cstddef>
 #include <gsl/pointers>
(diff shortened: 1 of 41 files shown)


ruse-traveler:
I think this is ready to go. I'll post some plots here showing the test case in the BHCal at work shortly...

ruse-traveler:
Here are some (eta, phi) maps of the tiles before and after merging. I tried out two cases for merging (using the example map from the PR description): once with the no. of tiles to merge set to 1 (so in other words, don't merge the tiles, and once with the no. set to 5 (merge into an sPHENIX-style tower).

Broadly, things look like I would expect them to. Except for the giant hole in the hit map...
mergedHitPhiVsEta_nMerge5 e10pim d14m11y2024
recoHitPhiVsEta e10pim d14m11y2024
mergedHitPhiVsEta_nMerge1 e10pim d14m11y2024

ruse-traveler:
Then another odd feature I noticed is that even when the no. to merge is 1, there are still slight changes in the eta, phi distributions of the tiles even though it should just be copying one collection onto another...
recoVsMergedHitPhi_nMerge1 e10pim d14m11y2024
recoVsMergedHitEta_nMerge1 e10pim d14m11y2024

src/factories/calorimetry/CalorimeterHitsMerger_factory.h
```diff
@@ -22,6 +22,7 @@ class CalorimeterHitsMerger_factory : public JOmniFactory<CalorimeterHitsMerger_
     ParameterRef<std::string> m_readout {this, "readout", config().readout};
     ParameterRef<std::vector<std::string>> m_fields {this, "fields", config().fields};
     ParameterRef<std::vector<int>> m_refs {this, "refs", config().refs};
+    ParameterRef<std::vector<std::string>> m_mappings {this, "mappings", config().mappings};
```
veprbl, line 25: Mappings sounds a bit generic to me, and it doesn't self-describe that there should be as many mappings as fields. Should we call it something like field transformations or rewrites?
```suggestion
    ParameterRef<std::vector<std::string>> m_field_transformations {this, "fieldTransformations", config().field_transformations};
```

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
```diff
@@ -41,24 +43,57 @@ void CalorimeterHitsMerger::init() {
...
+      std::vector<RefField> ref_fields;
+      for (size_t i = 0; i < m_cfg.fields.size(); ++i) {
+          id_mask |= id_desc.field(m_cfg.fields[i])->mask();
+          // use the provided id number to find ref cell, or use 0
+          int ref = i < m_cfg.refs.size() ? m_cfg.refs[i] : 0;
+          ref_fields.emplace_back(m_cfg.fields[i], ref);
+      }
+      ref_mask = id_desc.encode(ref_fields);
+      id_mask = ~id_mask;
+      debug("ID mask in {:s}: {:#064b}", m_cfg.readout, id_mask);
```
veprbl, line 72: Can we put mask operations as closures to the same ref_maps, so that the application of transofrmation is uniform?

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
```diff
@@ -140,4 +178,44 @@ void CalorimeterHitsMerger::process(
...
+
+  std::vector<RefField> ref_fields;
+  for (std::size_t iHit = 0; const auto& hit : *in_hits) {
+
+    // make sure vector is clear
+    ref_fields.clear();
+    for (std::size_t iField = 0; const auto& name_field : id_desc.fields()) {
+
+      // check if field has associated mapping
+      const bool foundMapping = (
```
veprbl, line 194: This looks a bit complicated, and I doubt that it actually works for mixing of "mappings" transformations and "refs" bitmasks. Could it be simplified assuming 1-to-1 loop?

veprbl (commented):
I'm wondering if we should just have only one function that can map any cellID to any cellID, and to that function we'd provide the precomputed bit masks. That would be most general and simple implementation, at the cost of complexitry. E.g. your example could look like: `(cellID & ~_phi_mask) | ((((cellID & _phi_mask) >> _phi_mask_shift) / 5) << _phi_mask_shift)` and we could introduce some helper functions to not have to deal with most raw C++. Just throwing this in as an idea.

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
```diff
@@ -41,24 +43,57 @@ void CalorimeterHitsMerger::init() {
...
+          trace("{} = {}", name_field.first, name_field.second->value(hit.getCellID()));
+        }
+        return params;
+      };
+
+      // intialize functions
+      //   - NOTE this assumes provided fields are 1-to-1!
+      auto& svc = algorithms::ServiceSvc::instance();
+      for (std::size_t iMap = 0; const auto& mapping : m_cfg.mappings) {
+        if (iMap < m_cfg.fields.size()) {
```
veprbl, line 89: This also breaks my brain a bit. A check for matching length and a 1-to-1 loop would be nicer.

ruse-traveler:
> I'm wondering if we should just have only one function that can map any cellID to any cellID, and to that function we'd provide the precomputed bit masks. That would be most general and simple implementation, at the cost of complexitry. E.g. your example could look like: `(cellID & ~_phi_mask) | ((((cellID & _phi_mask) >> _phi_mask_shift) / 5) << _phi_mask_shift)` and we could introduce some helper functions to not have to deal with most raw C++. Just throwing this in as an idea.

Definitely an option! And it might be nice to have that functionality available even if we keep the field-specific mappings...

To be frank, I still find bitwise operations a bit mystifying and opaque, which is part of why I shied away from using them here 😅 
One thing I was hoping to accomplish with this was to provide a more "user friendly" a way to express mappings, which is another reason why I opted for the more "functional" expressions...

src/detectors/BHCAL/BHCAL.cc
```diff
@@ -12,10 +12,66 @@
...
 #include "factories/calorimetry/CalorimeterIslandCluster_factory.h"
 #include "factories/calorimetry/CalorimeterTruthClustering_factory.h"
 #include "factories/calorimetry/TrackClusterMergeSplitter_factory.h"
 
+namespace eicrecon {
+
+  // --------------------------------------------------------------------------
+  // Helper method to generate appropriate phi map
+  // --------------------------------------------------------------------------
+  std::string MakePhiMap(const std::size_t nMerge = 1) {
```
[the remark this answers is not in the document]
ruse-traveler: Fair! Part of adding this was just to provide an example of how to wire this into a complete workflow, but these can be easily moved downstream into some analysis scripts for the BHCal DSC.

(Though it would be pretty cool to be able to provide std::function's as parameters 👀)

src/detectors/BHCAL/BHCAL.cc
```diff
@@ -12,10 +12,66 @@
...
+    // set up checks at wraparound.
+    // magic constants:
+    //   320 - number of tiles per row
+    const std::string sMerge    = std::to_string(nMerge);
+    const std::string wrapDef   = "(abs(phi_1 - phi_2) == (320 - 1))";
+    const std::string wrapMerge = "(abs(320 - abs(phi_1 - phi_2)) <= " + sMerge + ")";
+
+    // combine strings into horizontal adjacency conditions
+    const std::string phiAdjacent  = "(abs(phi_1 - phi_2) == " + sMerge + ")";
+    const std::string wrapAdjacent = nMerge > 1 ? wrapMerge : wrapDef;
```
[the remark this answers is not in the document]
ruse-traveler: True! This was more just me being over-cautious and wanting to _definitely_ use the defaults when not doing any merging.

src/factories/calorimetry/CalorimeterHitsMerger_factory.h
ruse-traveler: Good point! I also like that name!

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
ruse-traveler: I don't quite follow... So by closure do you mean if a user provides both a bitmask and a mapping, they both get applied?

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
ruse-traveler: Yes, as-is it's set up so that if a user provides a mapping, it defaults to using that and will ignore any provided bitmasks. But it can definitely be just turned into a 1-to-1 loop where it checks for the mapping or the bitmask.

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
ruse-traveler: Sure! This was trying to work in the above design choice that providing a map makes the algorithm ignores any provided bitmasks.

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
veprbl: Allow me to clarify. I meant that this code can be replaced with something like
```
for (size_t ix = 0; std::string &field : m_cfg.fields) {
  uint64_t ref = m_cfg.refs.at(ix);
  ref_maps[field] = [ref](const edm4eic::CalorimeterHit&) -> int {
    return ref;
  };
  ix++;
}
```
Then we can remove all the different logic for handling ref_mask and id_mask.

Chao1009:
> To be frank, I still find bitwise operations a bit mystifying and opaque, which is part of why I shied away from using them here 😅 One thing I was hoping to accomplish with this was to provide a more "user friendly" a way to express mappings, which is another reason why I opted for the more "functional" expressions...

Agreed.
Instead of masking the merging fields, which rely on the bitwise operation, we may extract the values of the non-merging ones.
For example, if the encoded cellID has three fields - "sector", "layer", "slice", and we would like to combine signals from all "slices".
We can extract a vector of ints for ["sector", "layer"] for each event and then group those with the same "sector" and "layer." This would improve the readability but (probably) at the cost of performance.

veprbl:
@Chao1009 Is that not the suggestion from https://github.com/eic/EICrecon/pull/1668#discussion_r1849028494 ?

Chao1009:
> @Chao1009 Is that not the suggestion from [#1668 (comment)](https://github.com/eic/EICrecon/pull/1668#discussion_r1849028494) ?

It's similar. The only difference is that I suggest using a single input for the fields and mappings (or field_transformations). For example, it accepts two types of inputs like
`{"phi", "theta:<transformation_string>"}`
The former input ("phi") means a simple merge of the "phi" field, while the latter does some transformation for "theta".
So we can avoid one-to-one matching of the fields and mappings in the code.

veprbl:
I think the current implementation in this PR already allows to provide arbitrary functions of arbitrary other fields to re-define fields. The old merging behavior was to just mask out bits of the original values and set them to all binary 1.

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
veprbl: I think this it's more like:
```
for (size_t ix = 0; std::string &field : m_cfg.fields) {
  uint64_t ref_mask = m_cfg.refs.at(ix);
  auto field_element = id_desc.field(m_cfg.fields[i]);
  ref_maps[field] = [ref_mask,field_element](const edm4eic::CalorimeterHit& hit) -> int {
    return (field_element->value(hit.cellID()) & ~ref_mask) | ref_mask;
  };
  ix++;
}
```

ruse-traveler: I see! And agreed! That would help to simplify things... I'll work through the implications for the rest of the algorithm...

ruse-traveler:
🥳 Happy new year!

I've implemented @veprbl's suggestions, which helped to greatly simplify the changes! I have to admit, though, that it feels a bit awkward to have both `fieldRefs` (which do the older, bit-masking operation) and `fieldTransformations` (which utilize the evaluator service) floating around...

How would you both feel about dropping the `fieldRefs`? Similar behavior to the masks could be achieved using a transformation like:
```
  .fields = {"eta"},
  .fieldTransformations = {"5"}  // merge all tiles in an eta column into the 5th one...
  ```
  This would get close to your suggestion, @Chao1009, albeit with the fields and transformations in 2 separate parameters...

**Comment by veprbl** [comment]
Happy New Year! I'm fine with dropping ref masks.

File: src/algorithms/calorimetry/CalorimeterHitsMerger.cc [hunk 7]
```diff
@@ -41,24 +43,75 @@ void CalorimeterHitsMerger::init() {
...
+      return;
+    }
+    if (m_cfg.fields.size() != m_cfg.fieldTransformations.size()) {
+      error(
+        "Size of field and transformation vectors are different ({} vs. {}).",
+        m_cfg.fields.size(),
+        m_cfg.fieldTransformations.size()
+      );
+      return;
+    }
```
veprbl, line 63: You require both to be set, but then ignore fieldRefs if transformations are defined. That is slightly confusing.

ruse-traveler: True! I think dropping `fieldRefs` entirely will help remove the confusion...

Chao1009:
> 🥳 Happy new year!
> 
> I've implemented @veprbl's suggestions, which helped to greatly simplify the changes! I have to admit, though, that it feels a bit awkward to have both `fieldRefs` (which do the older, bit-masking operation) and `fieldTransformations` (which utilize the evaluator service) floating around...
> 
> How would you both feel about dropping the `fieldRefs`? Similar behavior to the masks could be achieved using a transformation like:
> 
> ```
>   .fields = {"eta"},
>   .fieldTransformations = {"5"}  // merge all tiles in an eta column into the 5th one...
> ```
> 
> This would get close to your suggestion, @Chao1009, albeit with the fields and transformations in 2 separate parameters...

Although I think further merging them (`fields `and `transformations`) would provide better readability, my key request is to remove `refs`, which redundantly does a simple `transformation` in the code. So I'm happy with the current changes.

ruse-traveler:
> Although I think further merging them (fields and fieldTransformations) would provide better readability, my key request is to remove fieldRefs, which redundantly does a simple fieldTransformation in the code. So I'm happy with dropping fieldRefs.

Hmmm... It would also reduce the chance that one of the vectors are different lengths or orders. It won't be much more work to merge the parameters, so let me check and see if anything breaks!

ruse-traveler:
> Although I think further merging them (`fields `and `fieldTransformations`) would provide better readability, my key request is to remove `fieldRefs`, which redundantly does a simple `fieldTransformation` in the code. So I'm happy with dropping `fieldRefs`.

It works! I've gone ahead and implemented changes.

veprbl:
Did this break HcalEndcapPInsertMergedHits?

https://eicrecon.epic-eic.org/pr/1668/capybara/rec_dis_5x41_minQ2=0_craterlake_5x41/index.html#HcalEndcapPInsertMergedHits

ruse-traveler:
> Did this break HcalEndcapPInsertMergedHits?
> 
> https://eicrecon.epic-eic.org/pr/1668/capybara/rec_dis_5x41_minQ2=0_craterlake_5x41/index.html#HcalEndcapPInsertMergedHits

Uh oh, looks like it 😩 
Let me take a closer look...

ruse-traveler:
> Did this break HcalEndcapPInsertMergedHits?
> 
> https://eicrecon.epic-eic.org/pr/1668/capybara/rec_dis_5x41_minQ2=0_craterlake_5x41/index.html#HcalEndcapPInsertMergedHits

Looks like it was a typo (second field should've been "slices" not "layers") 😵

src/detectors/BHCAL/BHCAL.cc
```diff
@@ -83,12 +85,27 @@ extern "C" {
...
+        ));
+
         app->Add(new JOmniFactoryGeneratorT<CalorimeterTruthClustering_factory>(
           "HcalBarrelTruthProtoClusters", {"HcalBarrelRecHits", "HcalBarrelHits"}, {"HcalBarrelTruthProtoClusters"},
           app   // TODO: Remove me once fixed
         ));
+
         app->Add(new JOmniFactoryGeneratorT<CalorimeterIslandCluster_factory>(
-          "HcalBarrelIslandProtoClusters", {"HcalBarrelRecHits"}, {"HcalBarrelIslandProtoClusters"},
+          "HcalBarrelIslandProtoClusters", {"HcalBarrelMergedHits"}, {"HcalBarrelIslandProtoClusters"},
```
veprbl, line 108:
```suggestion
          "HcalBarrelIslandProtoClusters", {"HcalBarrelRecHits"}, {"HcalBarrelIslandProtoClusters"},
```
This is not ready yet. Let's merge without this so that we can fix regression in HcalBarrelClusterAssociations in a separate PR.

src/algorithms/calorimetry/CalorimeterHitsMerger.cc
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Chao Peng, Jihee Kim, Sylvester Joosten, Whitney Armstrong, Wouter Deconinck, David Lawrence
+// Copyright (C) 2022 Chao Peng, Jihee Kim, Sylvester Joosten, Whitney Armstrong, Wouter Deconinck, David Lawrence, Derek Anderson
```
veprbl, line 2:
```suggestion
// Copyright (C) 2022 - 2025 Chao Peng, Jihee Kim, Sylvester Joosten, Whitney Armstrong, Wouter Deconinck, David Lawrence, Derek Anderson
```

src/detectors/BHCAL/BHCAL.cc
ruse-traveler: Sounds good. Should I also remove the MergedHits factory?

src/algorithms/calorimetry/CalorimeterHitsMerger.h
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Chao Peng, Jihee Kim, Sylvester Joosten, Whitney Armstrong, Wouter Deconinck, David Lawrence
+// Copyright (C) 2022 Chao Peng, Jihee Kim, Sylvester Joosten, Whitney Armstrong, Wouter Deconinck, David Lawrence, Derek Anderson
```
veprbl, line 2:
```suggestion
// Copyright (C) 2022 - 2025 Chao Peng, Jihee Kim, Sylvester Joosten, Whitney Armstrong, Wouter Deconinck, David Lawrence, Derek Anderson
```

veprbl (dismissed):
Diff LGTM

src/detectors/BHCAL/BHCAL.cc
veprbl: It doesn't cause regressions, so we can keep it

ruse-traveler: Gotcha!

veprbl (dismissed):

veprbl (approved):

Chao1009 (approved):

PR #1675, opened 2024-11-19
fix: static_cast<float>(sqrt(...)) to avoid c++-narrowing warning/error

wdconinc:
This PR should fix the eicrecon build error in https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/4175065/raw (root-6.32.06 upgrade).

- [x] Bug fix (issue https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/4175065/raw)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackProjector.cc
+++ b/src/algorithms/tracking/TrackProjector.cc
@@ -153,7 +153,7 @@ namespace eicrecon {
                         static_cast<float>(boundCov(Acts::eBoundPhi, Acts::eBoundQOverP))
                 };
                 const float time{static_cast<float>(boundParams(Acts::eBoundTime))};
-                const float timeError{sqrt(static_cast<float>(boundCov(Acts::eBoundTime, Acts::eBoundTime)))};
+                const float timeError{static_cast<float>(sqrt(static_cast<float>(boundCov(Acts::eBoundTime, Acts::eBoundTime))))};
                 const float theta(boundParams[Acts::eBoundTheta]);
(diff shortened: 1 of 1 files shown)


wdconinc:
Not sure why this became an issue with the ROOT upgrade.

src/algorithms/tracking/TrackProjector.cc
```diff
@@ -153,7 +153,7 @@ namespace eicrecon {
                         static_cast<float>(boundCov(Acts::eBoundPhi, Acts::eBoundQOverP))
                 };
                 const float time{static_cast<float>(boundParams(Acts::eBoundTime))};
-                const float timeError{sqrt(static_cast<float>(boundCov(Acts::eBoundTime, Acts::eBoundTime)))};
+                const float timeError{static_cast<float>(sqrt(static_cast<float>(boundCov(Acts::eBoundTime, Acts::eBoundTime))))};
```
veprbl, line 156: I don't get why return type does not match, but okay. We might as well let sqrt to be calculated on the original type.
```suggestion
                const float timeError{static_cast<float>(sqrt(boundCov(Acts::eBoundTime, Acts::eBoundTime)))};
```

wdconinc: Good suggestion, but I'm not yet sure why this suddenly came up.

veprbl (approved):

PR #1676, opened 2024-11-21
feat: support EDM4hep v0.99, with TrackerHit as interface type

wdconinc:
EDM4hep v0.99 [introduced](https://github.com/key4hep/EDM4hep/pull/252) TrackerHit as an interface type, and the datatype that used to be called TrackerHit is now called TrackerHit3D. We use this in a few places, and this PR disambiguates the TrackerHit datatype. To minimize impact on the code and limit the annoying code effects to the preamble, this defines an alias of the new name.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue: https://github.com/key4hep/EDM4hep/pull/252)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/fardetectors/FarDetectorLinearTracking.cc
+++ b/src/algorithms/fardetectors/FarDetectorLinearTracking.cc
@@ -6,7 +6,15 @@
 #include <edm4eic/TrackPoint.h>
 #include <edm4eic/TrackSegmentCollection.h>
 #include <edm4eic/vector_utils.h>
+#include <edm4hep/EDM4hepVersion.h>
+#if EDM4HEP_BUILD_VERSION < EDM4HEP_VERSION(0, 99, 0)
 #include <edm4hep/TrackerHitCollection.h>
+namespace edm4hep {
+  using TrackerHit3DCollection = TrackerHitCollection;
+}
+#else
+#include <edm4hep/TrackerHit3DCollection.h>
+#endif
 #include <edm4hep/Vector3d.h>
 #include <edm4hep/Vector3f.h>
 #include <edm4hep/utils/vector_utils.h>
@@ -71,7 +79,7 @@ namespace eicrecon {
 
     void FarDetectorLinearTracking::buildMatrixRecursive(int level,
                                                         Eigen::MatrixXd* hitMatrix,
-                                                        const std::vector<gsl::not_null<const edm4hep::TrackerHitCollection*>>& hits,
+                                                        const std::vector<gsl::not_null<const edm4hep::TrackerHit3DCollection*>>& hits,
                                                         gsl::not_null<edm4eic::TrackSegmentCollection*> outputTracks ) const {
 
       // Iterate over hits in this layer
--- a/src/algorithms/fardetectors/FarDetectorLinearTracking.h
+++ b/src/algorithms/fardetectors/FarDetectorLinearTracking.h
@@ -26,7 +26,7 @@ namespace edm4hep {
 namespace eicrecon {
 
 using FarDetectorLinearTrackingAlgorithm =
-    algorithms::Algorithm<algorithms::Input<std::vector<edm4hep::TrackerHitCollection>>,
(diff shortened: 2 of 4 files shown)


src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -6,7 +6,15 @@
 #include <edm4eic/TrackPoint.h>
 #include <edm4eic/TrackSegmentCollection.h>
 #include <edm4eic/vector_utils.h>
+#include <edm4hep/EDM4hepVersion.h>
+#if EDM4HEP_BUILD_VERSION < EDM4HEP_VERSION(0, 99, 0)
```
wdconinc, line 10: Alt
```suggestion
#if __has_include("edm4hep/TrackerHit3DCollection.h")
```

veprbl: I always assumed we are meant to use `edm4hep_VERSION` and `EDM4HEP_BUILD_VERSION` is for internal use.

wdconinc: Who knows.

simonge:
I believe it is now possible to use edm4eic tracker hits rather than the edm4hep model, this should fix a couple of other things and avoid unnecessarily having to add this complexity.

wdconinc:
> I believe it is now possible to use edm4eic tracker hits rather than the edm4hep model, this should fix a couple of other things and avoid unnecessarily having to add this complexity.

It seems we have one to one relations to raw hits in edm4eic so it's not a trivial change.

simonge:
> > I believe it is now possible to use edm4eic tracker hits rather than the edm4hep model, this should fix a couple of other things and avoid unnecessarily having to add this complexity.
> 
> It seems we have one to one relations to raw hits in edm4eic so it's not a trivial change.

Yes, just been looking through the options again now and being reminded of all the subtle changes (in my opinion) need addressing.

The edm4hep TrackerHit has changed significantly in v0.99 from our current version but isn't really what is required either.

simonge:
Rather than continuing to use the edm4hep version, I would be in favour of adding the Measurement3D to edm4eic where the surface from Measurement2D is replaced by the 3rd spatial dimension and covariance.

Is there no point during the acts tracking procedure where the data is in this format? Every time I try and look through it, I find the acts coding style slightly impenetrable.

wdconinc:
> Rather than continuing to use the edm4hep version, I would be in favour of adding the Measurement3D to edm4eic where the surface from Measurement2D is replaced by the 3rd spatial dimension and covariance.

I think we should try to move closer to EDM4hep where we can, not move further away. I think this makes sense especially for data types that are before typical reconstruction steps such as clustering or tracking. We are already much farther away from EDM4hep than I wish we were.

simonge:
The problem as usual is keeping associations so that they can be traced through the reconstruction.

The pre-v0.10 version of edm4hep::TrackerHit had this convenient but probably misguided vector of ObjectIDs which allowed a link to any podio object, in this case between the TrackerHit and an edm4eic::RawTrackerHit. At very least this will have to be replaced with a edm4hep::TrackerHitSimTrackerHitLink, linking back to the simulated rather than raw hit through our edm4eic::MCRecoTrackerHitAssociation.

edm4hep v0.99 doesn't appear to have very good support for our approach of converting everything into the raw tracker/calorimeter hits with just digitized time and charge values and no links/relations/associations between them. There is no RawTrackerHit type but admittedly this would just be identical the the RawCalorimeterHit.

Is the the only block from updating to v0.99?

veprbl:
I see this:
```
src/algorithms/fardetectors/FarDetectorTrackerCluster.cc:196:14: error: no member named 'addToRawHits' in 'edm4hep::MutableTrackerHit3D'
      hitPos.addToRawHits(hit);
      ~~~~~~ ^
```

wdconinc:
Proposal: let's go with modifying edm4eic to look like the former edm4hep functionality, see https://github.com/eic/EDM4eic/pull/105.

veprbl:
This is not needed anymore.

PR #1682, opened 2024-11-29
fix: static_cast<float>(sqrt(...)) to avoid c++-narrowing warning/error (again)

wdconinc:
#1675 again see there.


veprbl (approved):
Right, there is a bit of code duplication. We need to change both instances.

PR #1687, opened 2024-12-03
feat(ci): runs-on: ubuntu-latest -> ubuntu-22.04

wdconinc:
This PR updates the target operating system for CI jobs, from ubuntu-latest to a more explicit ubuntu-22.04. This is in part to avoid the automatic update of implicit ubuntu-latest that will break everything (https://github.com/eic/EICrecon/pull/1686), and in part because it makes sense to be explicit about the target operating system.

- [ ] Bug fix (issue #__)
- [x] New feature (issue https://github.com/actions/runner-images/issues/10636)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -49,7 +49,7 @@ jobs:
   env:
     # env cannot be used in matrix, but outputs can
     # so this job turns env into outputs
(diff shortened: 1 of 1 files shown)


veprbl (approved):
No objections from me.

PR #1690, opened 2024-12-18
Allow xrootd files as inputs to eicrecon

simonge:
Currently eicrecon has two checks for files existing, neither of which allow the file to be on a remote xrootd server. This PR adapts JEventSourcePODIO.cc so that the first check allows for remote files and (currently) removes the second check.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

tbd

Allows remote files to be passed directly to eicrecon

### DIFF
--- a/src/services/io/podio/JEventSourcePODIO.cc
+++ b/src/services/io/podio/JEventSourcePODIO.cc
@@ -131,17 +131,6 @@ void JEventSourcePODIO::Open() {
     // Open primary events file
     try {
 
-        // Verify file exists
-        // if( ! std::filesystem::exists(GetResourceName()) ){
-        //     // Here we go against the standard practice of throwing an error and print
-        //     // the message and exit immediately. This is because we want the last message
-        //     // on the screen to be that the file doesn't exist.
-        //     auto mess = fmt::format(fmt::emphasis::bold | fg(fmt::color::red),"ERROR: ");
-        //     mess += fmt::format(fmt::emphasis::bold, "file: {} does not exist!",  GetResourceName());
-        //     std::cerr << std::endl << std::endl << mess << std::endl << std::endl;
-        //     std::_Exit(EXIT_FAILURE);
-        // }
-
         m_reader.openFile( GetResourceName() );
 
         auto version = m_reader.currentFileVersion();
--- a/src/services/io/podio/JEventSourcePODIO.cc
+++ b/src/services/io/podio/JEventSourcePODIO.cc
@@ -14,16 +14,12 @@
 #include <JANA/Utils/JTypeInfo.h>
 #include <TFile.h>
 #include <TObject.h>
-#include <fmt/color.h>
 #include <fmt/core.h>
-#include <fmt/format.h>
 #include <podio/CollectionBase.h>
 #include <podio/Frame.h>
 #include <podio/podioVersion.h>
 #include <algorithm>
-#include <cstdlib>
 #include <exception>
-#include <filesystem>
 #include <iostream>
 #include <map>
 #include <utility>
--- a/src/services/io/podio/JEventSourcePODIO.cc
+++ b/src/services/io/podio/JEventSourcePODIO.cc
@@ -132,15 +132,15 @@ void JEventSourcePODIO::Open() {
     try {
 
         // Verify file exists
-        if( ! std::filesystem::exists(GetResourceName()) ){
-            // Here we go against the standard practice of throwing an error and print
-            // the message and exit immediately. This is because we want the last message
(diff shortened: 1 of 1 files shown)


simonge:
Questions:
Is there a situation in which both checks are required?
Can Open() just do the second check with CheckOpenable()?

src/services/io/podio/JEventSourcePODIO.cc
```diff
@@ -267,7 +267,7 @@ double JEventSourceGeneratorT<JEventSourcePODIO>::CheckOpenable(std::string reso
 
     // PODIO FrameReader segfaults on legacy input files, so we use ROOT to validate beforehand. Of course,
     // we can't validate if ROOT can't read the file.
-    std::unique_ptr<TFile> file = std::make_unique<TFile>(resource_name.c_str());
+    std::unique_ptr<TFile> file = std::unique_ptr<TFile>{TFile::Open(resource_name.c_str())};
```
veprbl, line 270: I don't see how TFile::Open will be different here. From what I understand this function isn't meant to check if file exists, but check it it's a compatible format. There is also a legacy PODIO version of the source that can be chosen instead.

src/services/io/podio/JEventSourcePODIO.cc
```diff
@@ -132,15 +132,15 @@ void JEventSourcePODIO::Open() {
...
-        }
+        // if( ! std::filesystem::exists(GetResourceName()) ){
+        //     // Here we go against the standard practice of throwing an error and print
+        //     // the message and exit immediately. This is because we want the last message
+        //     // on the screen to be that the file doesn't exist.
+        //     auto mess = fmt::format(fmt::emphasis::bold | fg(fmt::color::red),"ERROR: ");
+        //     mess += fmt::format(fmt::emphasis::bold, "file: {} does not exist!",  GetResourceName());
+        //     std::cerr << std::endl << std::endl << mess << std::endl << std::endl;
+        //     std::_Exit(EXIT_FAILURE);
+        // }
```
veprbl, line 143: Should not be a problem to remove this. How bad is the error message?

If the new message is unintelligible, we need to keep the check. Perhaps logic like in https://github.com/AIDASoft/DD4hep/pull/1283/files#diff-c3daa04c0212b4cc1ba7523f1952d41583fe46dd73aee5ec8c052061c6bbbc7fR570 using a regex like `^[A-Za-z][A-Za-z0-9+-.]*:` (based on my understanding of section 3.1 of [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986)) should be enough.

src/services/io/podio/JEventSourcePODIO.cc
simonge: There is a very useful error (and JANA) from root before making this change:
```
Error in <TFile::TFile>: please use TFile::Open to access remote files:
        auto f = std::unique_ptr<TFile>{TFile::Open("root://dtn-eic.jlab.org//work/eic2/EPIC/RECO/24.08.1/epic_craterlake/SIDIS/pythia6-eic/1.0.0/10x100/q2_0to1/pythia_ep_noradcor_10x100_q2_0.000000001_1.0_run9.ab.0544.eicrecon.tree.edm4eic.root")};
----------------------------------------------------------
JException
  Message:  Unable to open event source "root://dtn-eic.jlab.org//work/eic2/EPIC/RECO/24.08.1/epic_craterlake/SIDIS/pythia6-eic/1.0.0/10x100/q2_0to1/pythia_ep_noradcor_10x100_q2_0.000000001_1.0_run9.ab.0544.eicrecon.tree.edm4eic.root": No suitable generator found!
```

src/services/io/podio/JEventSourcePODIO.cc
simonge: The error without this is just the error thrown by the CheckOpenable function as it already identifies that the file doesn't exist and stops.

veprbl:
After testing this on my side, I think, I understand this better now. The check being removed is seemingly impossible to even trigger as TFile's exception in CheckOpenable happen before it and are not being caught. Thus this change does not modify the current behaviour and does provide promised functionality (streaming from XRootD).

veprbl (dismissed):
LGTM

veprbl (approved):

simonge:
Should a test be added for this or is the Jlab XRootD server and file locations potentially too volatile to want to rely on?

veprbl:
You are welcome to add a test. I don't think full simulations are published anymore. There will be some old ones, and reconstruction can be restricted to few detectors that don't rely on [a consistent] geometry being used.

PR #1695, opened 2025-01-07
2D-strip digitization for CyMBaL and Outer MPGDs

ybedfer:
- More realistic, 2D-strip readout of `CyMBaL` and `Outer` MPGDs.
- This is done at the digitization level, elaborating on a new description of these detectors in `epic` (commit `#efc19d119`).
- A run-time option (`MPGD:SiFactoryPattern`) allows to disable 2D-strip and go back to pixel readout on a per MPGD-detector basis. 
- If more realistic, the digitization is still simplistic: single-hit clusters.

- [x] Other: New feature.

No.

Yes. 

### Residuals `TrackerHit - SimTrackerHit`
[CyMBaL.Zphi.singleHit.pdf](https://github.com/user-attachments/files/18335335/CyMBaL.Zphi.singleHit.pdf)
[OuterRes.UV.singleHit.pdf](https://github.com/user-attachments/files/18335258/OuterRes.UV.singleHit.pdf)


src/algorithms/digi/MPGDTrackerDigi.cc
```diff
@@ -0,0 +1,238 @@
...
+        CellID cIDn = m_seg->cellID(lpos,dummy,vIDn);
+        sim2IDs.push_back({cIDp,cIDn}); // Remember cellIDs.
+        // ***** DEBUGGING INFO
+        if(level() >= algorithms::LogLevel::kDebug) {
+            CellID hIDp = cIDp>>32, sIDp = cIDp>>30&0x3;
+            debug("--------------------");
+            debug("Hit cellIDp  = 0x{:08x}, 0x{:08x} 0x{:02x}", hIDp, vIDp, sIDp);
+            CellID hIDn = cIDn>>32, sIDn = cIDn>>30&0x3;
+            debug("Hit cellIDn  = 0x{:08x}, 0x{:08x} 0x{:02x}", hIDn, vIDn, sIDn);
+            //#define MPGDDigi_DEBUG
```
veprbl, line 147:
```suggestion
```

src/algorithms/digi/MPGDTrackerDigi.cc
```diff
@@ -0,0 +1,238 @@
...
+            // ...strip field: =0: pixels, =1: 1st coord., =2: 2nd coord.
+            // ...the 32 bits of the hitID field are equally shared by the two
+            //   coordinates,
+            // ...segmentations @ reconstruction and simulation time are identical.
+            CellID hID =  cID>>32,  sID = vID>>30&0x3;
+            debug("Hit cellID   = 0x{:08x}, 0x{:08x} 0x{:02x}", hID,  vID,  sID);
+            CellID xid = hID&0xffff,     xidp = hIDp&0xffff;
+            CellID yid = hID>>16&0xffff, yidn = hIDn>>16&0xffff;
+            if (xid!=xidp || yid!=yidn)
+                printf("** MPGDTrackerDigi: Strip segmentation inconsistency: m_seg(0x%4lx,0x%4lx) != sim_hit(0x%4lx,0x%4lx)\n",xidp,xid,yidn,yid);
```
veprbl, line 160: No printf, please. Ideally we would not have debug `#ifdef`s too, check debug levels should be sufficient.

src/algorithms/digi/MPGDTrackerDigi.h
```diff
@@ -0,0 +1,60 @@
...
+  std::function<double()> m_gauss;
+  // FIXME replace with standard random engine
+  // std::default_random_engine generator; // TODO: need something more appropriate here
+  // std::normal_distribution<double> m_normDist; // defaults to mean=0, sigma=1
+
+  // algorithms::Generator m_rng = algorithms::RandomSvc::instance().generator();
+
+  /** Segmentation */
+  const dd4hep::Detector* m_detector{nullptr};
+  const dd4hep::rec::CellIDPositionConverter* m_cellid_converter{nullptr};
```
veprbl, line 56:
```suggestion
```
Unused?

src/detectors/MPGD/MPGD.cc
```diff
@@ -10,15 +10,27 @@
 #include "algorithms/interfaces/WithPodConfig.h"
 #include "extensions/jana/JOmniFactoryGeneratorT.h"
 #include "factories/digi/SiliconTrackerDigi_factory.h"
+#include "factories/digi/MPGDTrackerDigi_factory.h"
 #include "factories/tracking/TrackerHitReconstruction_factory.h"
 
+// ***** NOTA BENE:
+// 2D-STRIP DIGITIZATION
+// - Is prodided by "MPGDTrackerDigi".
```
veprbl, line 18:
```suggestion
// - Is produced by "MPGDTrackerDigi".
```

src/detectors/MPGD/MPGD.cc
```diff
@@ -10,15 +10,27 @@
...
+// - Requires "MultiSegmentation" <readout> in geometry file.
```
veprbl, line 19:
```suggestion
// - Relies on "MultiSegmentation" <readout> in "compact" geometry file.
```

src/detectors/MPGD/MPGD.cc
```diff
@@ -10,15 +10,27 @@
...
+// PIXEL DIGITIZATION
+// - Is prodided by "SiliconTrackerDigi".
+// TEMPORARILY PROVIDE FOR SWITCHING BACK TO PIXEL DIGITIZATION
+#define Si_FACTORY_PATTERN 0 // Bit pattern: 0x1=CyMBaL, 0x2=Outer,...
```
veprbl, line 23: Would be nice to make it into an option `MPGD:MPGDBarrel:SiFactoryPattern`.

src/algorithms/digi/MPGDTrackerDigi.cc
```diff
@@ -0,0 +1,238 @@
...
+        const CellID volMask = 0xffffffff;
+        CellID cID = sim_hit.getCellID(), vID = cID&volMask /* not necessary but cleaner*/;
+        DetElement local = volman.lookupDetElement(vID);
+        const auto lpos = local.nominal().worldToLocal(gpos);
+        // p "strip"
+        CellID stripBitp = ((CellID)0x1)<<30, vIDp = vID|stripBitp;
+        CellID cIDp = m_seg->cellID(lpos,dummy,vIDp);
+        // n "strip"
+        CellID stripBitn = ((CellID)0x2)<<30, vIDn = vID|stripBitn;
+        CellID cIDn = m_seg->cellID(lpos,dummy,vIDn);
```
veprbl, line 138: I'm confused here. If you use `dummy` position, then `cIDp` and `cIDn` will not differ beyond the bits in the `strip` field? From your description above I took that the hit position would be used to produce coordinates in a new segmentation.

src/algorithms/digi/MPGDTrackerDigi.cc
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    // ***** raw_hit INSTANTIATION AND raw<-sim_hit's ASSOCIATION
+    for (auto item : cell_hit_map) {
+        raw_hits->push_back(item.second);
+        Sim2IDs::const_iterator sim_it = sim2IDs.cbegin();
+        for (const auto& sim_hit : *sim_hits) {
+            CellIDs cIDs = *sim_it++;
+            for (CellID cID : {cIDs.first,cIDs.second}) {
+                if (item.first == cID) {
```
veprbl, line 221: Would it make sense to strip the outer loop and instead do a lookup of `cID` in cell_hit_map here?

src/algorithms/digi/MPGDTrackerDigi.cc
```diff
@@ -0,0 +1,238 @@
...
+    std::unordered_map<std::uint64_t, edm4eic::MutableRawTrackerHit> cell_hit_map;
+    // Prepare for strip segmentation
+    const Position dummy(0,0,0);
+    const VolumeManager& volman = m_detector->volumeManager();
+
+    using CellIDs = std::pair<CellID,CellID>;
+    using Sim2IDs = std::vector<CellIDs>; Sim2IDs sim2IDs;
+    for (const edm4hep::SimTrackerHit& sim_hit : *sim_hits) {
+
+        // ***** TIME SMEARING
```
veprbl, line 112: This essentially does a digi step (as you say, derived from `SiliconTrackerDigi`) together with a simulation step (produce hits in strips). I don't see how those are inherently convoluted. Can you keep re-using `SiliconTrackerDigi`? Then this algorithm can focus just on mangling cellIDs for the MPGD raw tracker hits.

veprbl (commented):
Thank you for your submission. I have a few comments and questions on this.

src/algorithms/digi/MPGDTrackerDigi.cc
```diff
@@ -0,0 +1,238 @@
...
+
+        // ***** SEGMENTATION
+        // - The two cellID's, w/ "strip" setting of 0x1 (called 'p') and 0x2
+        //  (called 'n') are evaluated based on "sim_hit" Cartesian coordinates.
+        // - To get the new cellID's, we need the local position.
+        const edm4hep::Vector3d &pos = sim_hit.getPosition();
+        const double &mm = dd4hep::mm;
+        Position gpos(pos.x*mm,pos.y*mm,pos.z*mm);
+        const CellID volMask = 0xffffffff;
+        CellID cID = sim_hit.getCellID(), vID = cID&volMask /* not necessary but cleaner*/;
```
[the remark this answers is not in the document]
veprbl: Some links to documentation:
* https://github.com/AIDASoft/DD4hep/blob/874fe2829843bf0684134bc8854c4cc6300368b5/DDCore/include/DDSegmentation/BitFieldCoder.h
* https://github.com/AIDASoft/DD4hep/blob/874fe2829843bf0684134bc8854c4cc6300368b5/DDCore/src/IDDescriptor.cpp

src/algorithms/digi/MPGDTrackerDigi.cc
ybedfer: In the `dd4hep::Segmentation::cellID` method, the 2nd argument is the global position, see:
```
      /// Determine the cell ID based on the position
      virtual CellID cellID(const Vector3D& localPosition, const Vector3D& globalPosition,
                            const VolumeID& volumeID) const = 0;
```
Which _global_ position is disregarded in the segmentation classes foreseen for MPGDs (`CartesianGrid...` and `CylindricalGridPhiZ`), and in fact in most segmentation classes.

`cIDp` and `cIDn` do differ, thanks to the discrimination based on the `strip` field in the `MultiSegmentation` context.

src/detectors/MPGD/MPGD.cc
ybedfer: I don't think so. The `#define` is meant to be temporary (as said in the comment). And it is also foreseen to be enabled only on rare occasions, and probably only by me, when a problem linked to strip-, as opposed to pixel-, hits shows up anywhere in the reconstruction. I used it myself to debug my code producing the residual distributions attached to the PR. When the whole thing is finalized (meaning including also the end-cap MPGDs), I will remove it altogether.

veprbl: We usually don't take in commented code as it's not compiled and not ran/tested by our workflows. The "TODO" comment is appreciated, but not sufficient. Let's file an issue to this repo documenting the change to be done so that we can follow up on this so this is removed or converted to an option, as I suggested.

src/algorithms/digi/MPGDTrackerDigi.cc
veprbl: > Which _global_ position is disregarded in the segmentation classes foreseen for MPGDs (`CartesianGrid...` and `CylindricalGridPhiZ`), and in fact in most segmentation classes.

Thank you for the explanation. Could you add this sentence as a code comment for future readers?

ybedfer: I modified the already existing comment:
```
        // ***** SEGMENTATION
        // - The two cellID's are encoded via a "dd4hep::MultiSegmentation"
        //  discriminating on the strip field, w/ "strip" setting of 0x1 (
        //  called 'p') and 0x2 (called 'n').
        // - They are evaluatedd based on "sim_hit" Cartesian coordinates
        //  positions
        //   Given that all the segmentation classes foreseen for MPGDs (
        //  "CartesianGrid.." for Outer and EndCaps, "CylindricalGridPhiZ" for
        //  "CyMBaL") disregard the _global_ position argument to
        //  "dd4hep::Segmentation::cellID", we need the _local_ position and
        //  only that.
```

src/algorithms/digi/MPGDTrackerDigi.cc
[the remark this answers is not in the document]
ybedfer: I simplified out the magical numbers:
```
        using dd4hep::mm;
        Position gpos(pos.x*mm,pos.y*mm,pos.z*mm);
        CellID vID = // Note: Only the bits corresponding to the volumeID will
          // be used. The rest, encoding the segmentation stored in "sim_hit",
          // being disregared.
          sim_hit.getCellID();
```
As explained in the comment, the `cellID` method retains by itself the volumeID part and disregards the rest.

src/algorithms/digi/MPGDTrackerDigi.cc
ybedfer: I removed the `#ifdef` blocks.

src/algorithms/digi/MPGDTrackerDigi.cc
ybedfer: Later on, I intend to modify the digitization, as opposed to simulation, part. Timing will be specific to this 2D-strip setup, w/ a time smearing receiving two different contributions: one common to both strips (originating in the ionization process) and one distinct (from the electronics and charge spreading). Same for the amplitude. I can then indeed split the method into two parts: simulation and then digitization. But I cannot see how I could use `SiliconTrackerDigi` for the latter.
I also need to make the simulation part more complex, w/ multi-hit clusters along each strip coordinate.

src/algorithms/digi/MPGDTrackerDigi.cc
ybedfer: - The problem is that `cID` in `sim_hit` and `cell_hit_map` are not of the same kind: pixelID _vs_. stripID.
- Yet pixelID contains all the information: it's the concatenation of the two stripIDs (this, provided that the same segmentation is used in the production of input data and in `EICrecon` (which must usually be the case)).
-  => Therefore indeed, the association block could be made simpler.
         And not only the association, but the whole algorithm could be made simpler: no need to go through retrieving the
       local position; just split the input pixelID and add two distinct strip fields and that's it.
         And it also makes more sense: in my present implementation, the cellID stored in `sim_hit` is useless.
         Note that historically, the idea to have a 3-valued MultiSegmentation (pixel, strip1, strip2) came to my mind lately:
       as a way to do the PR in two steps: `epic` (while maintaining compatibility w/ old `EICrecon`) and then `EICrecon`.
 - But it's more complicated than this. Later, as I already said elsewhere, I need to implement multi-hit clusters. Then
  most  probably, I will be happy to have at hand the local position, in order to handle edge effects and god knows what.
- => Therefore I am reluctant to do the modification. May also be that I am reluctant simply because it would be changing
 the whole file thoroughly. Yet, to be on the safe side, I would keep it as is. And I could file an issue to the repo, as you
 are suggesting yourself elsewhere, stating that it might be possible to proceed in a much simpler way. And add a 
 detailed comment to the file, explaining all the above. I am preparing it.
- Bottom line: I don't know how to file an issue.  Could you tell it to me in short? Or point the relevant documentation?

src/detectors/MPGD/MPGD.cc
ybedfer: - Giving it a second thought: yes indeed, having an option `SiliconTrackerDigi` or `MPGDTrackerDigi` would be simpler.
- It was even suggested to me by colleagues from the MPGD group.
- Problem: I don't know how to do this! Is there a class to which I can add the option as a data member? And how do I
 set the option at running time?

src/algorithms/digi/MPGDTrackerDigi.cc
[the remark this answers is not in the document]
veprbl: Looks better, thanks. There are some residual `>> 30) | 0x3` that you could address, but it's already much more readable.

src/detectors/MPGD/MPGD.cc
veprbl: Just an example:
https://github.com/eic/EICrecon/blob/d96cbff18ba2b6e258791ffd6daf94d4b0562ba8/src/utilities/dump_flags/DumpFlags_processor.cc#L34-L35
this adds an option that gets loaded into `m_python_file_name` variable (converted to an appropriate type)..

src/algorithms/digi/MPGDTrackerDigi.cc
veprbl: This was just a suggestion on my part. Ideally, we build our reconstruction modules that are most simple to understand. If there is any piece that we can simplify, this is utmost welcome thing to do.

The issues simply go to https://github.com/eic/EICrecon/issues, there is a button to create one. There are some templates, but you are not expected to follow them strictly.

src/detectors/MPGD/MPGD.cc
ybedfer: I have committed a version w/ option and w/o `#ifdef`, see #8fd77ef3.

src/algorithms/digi/MPGDTrackerDigi.cc
ybedfer: I have added a general description of the `MPGDTrackerDigi` class in comment:
```
  Digitization specific to MPGDs.
  - What's special in MPGDs is their 2D-strip readout: i.e. simultaneous
   registering along two sets of coordinate strips.
  - The "process" method involves a combination of segmentation, simulation and
   digitization.
  - The segmentation done when producing "sim_hits", and stored as a pixel
   cellID, is overwritten by strip cellIDs, via "dd4hep::MultiSegmentation".
  - The simulation will eventually involve simulating the amplitude and timing
   correlation of the two coordinates, and the spreading of the charge on
   adjacent strips producing multi-hit clusters.
    The preliminary version is rudimentary: single-hit clusters, with identical
   timing and uncorrelated amplitudes.
  - The digitization follows the standard steps, only that it needs to be
   convoluted with the simulation.
  NOTA BENE: The file could be simplified, see issue #1702.
```
This, with a reference to issue #1702 ("https://github.com/eic/EICrecon/issues/1702").

veprbl:
If I look at benchmarks, this seems to break tracking. E.g. see how histograms are empty in
https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/4486371/artifacts/file/results/tracking_performances_dis/epic_craterlake_tracking_only/pythia8NCDIS_18x275_minQ2=1_combined_5/plots.pdf
And this is how they look on main:
https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/4480093/artifacts/file/results/tracking_performances_dis/epic_craterlake_tracking_only/pythia8NCDIS_18x275_minQ2=1_combined_5/plots.pdf

ybedfer:
I set the default = no `MPGDTrackerDigi_factory`.
And let's try again.
Anyway I have absolutely no clue what can be the reason for the empty histos. `RawTrackHit` is as it should. The association with `SimTrackerHit`  is also as it should. See the histos I put in the PR.

veprbl:
Works again: https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/4488939/artifacts/file/results/tracking_performances_dis/epic_craterlake_tracking_only/pythia8NCDIS_18x275_minQ2=1_combined_5/plots.pdf

ybedfer:
Then, could it be that:
 1 The seeding of tracks is based on MPGDs?
 2 The seeding algorithm is relying on MPGD hit being pixels and fails when it gets strips instead?

If so, we could agree to merge the 2D-strip as it is now: meaning
 - w/o `MPGDTrackerDigi_factory` by default, and then working,
 - w/ still the option `-PMPGD:SiFactoryPattern=0` in order to investigate the seeding (could also be `=0x2`: 2D-strip in CyMBaL alone, or `=0x1`: 2D-strip in Outer alone).

veprbl:
> Then, could it be that:
> 
> 1. The seeding of tracks is based on MPGDs?
> 2. The seeding algorithm is relying on MPGD hits being pixels and fails when it gets strips instead?

I believe we don't have pixels or strips, just hits. For seeding, I would imagine, it does matter what are their positions. Is there supposed to be an additional algorithm to reconstruct intersections of strips?

cc @bschmookler

ybedfer:
> Is there supposed to be an additional algorithm to reconstruct intersections of strips?
- I don't know. But if the seeding hypothesis, _i.e._ seeding-in-MPGDs-fails-when-strips, is confirmed, it could be a way out.
- What's striking is the 100% failure rate. => Something very basic must be missing. 
- Can you tell me which piece(s) of software are involved in the seeding? So that I can investigate. Can be that seeding is not the issue. But it's anyway the starting point, from where investigations can develop. 

cc @bschmookler

bschmookler:
The seeding algorithm takes in all the digitized tracker hits in global coordinates (CentralTrackingRecHits):

https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L153-L159

It then only searches for seeds in the SVT volume in the barrel: 

https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/OrthogonalTrackSeedingConfig.h#L17

<br/><br/>

Does the problem you are seeing also occur when you look at the truth-seeded tracking? In that case, the seeding on using the MC particle information to seed the Combinatorial Kalman Filter, rather that any tracking detector info.

ybedfer:
I cannot reproduce the problem.
I privately executed the `tracking_performances_dis` benchmark, taken from:
```
https://github.com/eic/detector_benchmarks/tree/master/benchmarks/tracking_performances_dis
```
successively w/ option `MPGD:SiFactoryPattern` set =0x3 (`SiliconTrackerDigi`) and = 0x0 (`MPGDTrackerDigi`).
=> No difference other than second order ones. See attached pdf.
[MPGD.vs.Si.TrackerDigi.pdf](https://github.com/user-attachments/files/18466104/MPGD.vs.Si.TrackerDigi.pdf)

I am the going to switch back to  `MPGDTrackerDigi` by default. So that the check is started again. This, presuming that the previously observed problem of empty histos was due to some glitch. Let's see...

If the problem happens again, the ideal would be to have in a next iteration a podio output of the benchmark including MPGD hits.

ybedfer:
I don't quite understand the reason for the failure of `eicweb/eic_container`.
- There's an error concerning `bench:lfhcal_campaigns`: `Job's log exceeded limit of 4194304 bytes.` which seems to be unrelated to my PR.
- Else a failure in `bench:tracking_performance_campaigns` which happens to be more in line w/ the empty-histos previous problem but seems to be due to file handling by the computer system:
```
Error in rule fetch_epic:
    jobid: 22
    output: EPIC/RECO/23.10.0/epic_craterlake/SINGLE/pi-/10GeV/130to177deg/pi-_10GeV_130to177deg.0000.eicrecon.tree.edm4eic.root
    shell:
        
xrdcp --debug 2 root://dtn-eic.jlab.org//work/eic2/EPIC/RECO/23.10.0/epic_craterlake/SINGLE/pi-/10GeV/130to177deg/pi-_10GeV_130to177deg.0000.eicrecon.tree.edm4eic.root EPIC/RECO/23.10.0/epic_craterlake/SINGLE/pi-/10GeV/130to177deg/pi-_10GeV_130to177deg.0000.eicrecon.tree.edm4eic.root
        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)
```

veprbl:
The failures are known, and explained by a change to external states (new PyPi packages published, old campaigns deleted), that is to be fixed by https://github.com/eic/detector_benchmarks/pull/128.

The latest tracking_perfromances_dis looks fine now
https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/4546314/artifacts/file/results/tracking_performances_dis/epic_craterlake_tracking_only/pythia8NCDIS_18x275_minQ2=1_combined_5/plots.pdf
which is what I can not explain yet. Let's assume that was a transient issue.

veprbl (approved):
The strip re-encoding algorithm looks good to me. It's unfortunate that this is based on a copy of `SiliconTrackerDigi` instead of keeping digitization as a common separate algorithm. Test results look good to me.

ybedfer:
> keeping digitization as a common separate algorithm.
As is, `SiliconTrackerDigi` cannot serve for the digitization of MPGDs, because it's dealing w/ independent input hits, see _e.g._:
```
    for (const auto& sim_hit : *sim_hits) {

        // time smearing
        double time_smearing = m_gauss();
        double result_time = sim_hit.getTime() + time_smearing;
        auto hit_time_stamp = (std::int32_t) (result_time * 1e3);
```
While MPGD hits are are correlated: a single ionization process, w/ its energy deposit and timing, gives rise to two hits.

PR #1698, opened 2025-01-08
Change EventHeader check to be non fatal

### LINKED ISSUE #1696: Make EventHeader collection non-essential
EICrecon will only run when an EventHeader collection is present in the input file, the simulation and EICrecon itself by default save this to their output files but is a list of collections is specified when running EICrecon on the command line this needs to manually be added if the output collections.

The problem comes when running in multiple stages, investigating the effects of changing configuration/algorithms from later in the analysis without rerunning the full reconstruction.

All it appears to do is set the JEvent event and run numbers, which as far as I can see aren't explicitly used anywhere at the moment. Removing the check entirely still works allowing files to be analysed without an EventHeader collection.

Allow files to be analysed without an EventHeader collection, adding a warning that it wasn't present.

**Describe alternatives you've considered**
Add EventHeader to a list of collections which is output independently of what is requested. This could be done in conjunction with the other solution.

simonge:
Allows running of EICrecon on files without an EventHeader collection. Files with a single/small subset of output collections can have analysis continued without needing to manually persist EventHeader.

A warning has been added.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1696)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/services/io/podio/JEventSourcePODIO.cc
+++ b/src/services/io/podio/JEventSourcePODIO.cc
@@ -199,12 +199,17 @@ void JEventSourcePODIO::GetEvent(std::shared_ptr<JEvent> _event) {
     auto frame_data = m_reader.readEntry("events", Nevents_read);
     auto frame = std::make_unique<podio::Frame>(std::move(frame_data));
 
-    const auto& event_headers = frame->get<edm4hep::EventHeaderCollection>("EventHeader"); // TODO: What is the collection name?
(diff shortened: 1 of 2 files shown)


veprbl (approved):
This is probably okay for now, but only because we don't rely on those numbers. Once conditions DBs are implemented, this will have to be provided somehow.

Issue #1645, opened 2024-10-28
Update Track-Cluster Merge/Splitter to output Track-Cluster Associations


ruse-traveler:
Now that we have formal edm4eic::TrackClusterAssociations (c.f. [edm4eic#52](https://github.com/eic/EDM4eic/pull/52)). The TrackClusterMergeSplitter algorithm should produce track-cluster associations alongside merged/split clusters.

**Describe the solution you'd like**
n/a

**Describe alternatives you've considered**
n/a

**Additional context**
This falls into the reconstruction WG's eID/PFA development program.

PR #1699, opened 2025-01-09
Update TrackClusterMergeSplitter to output track-cluster associations (PFA0a)

### LINKED ISSUE #900: Reconstruction: **[Particle Flow]** Implementation of PF algorithm + factories

### LINKED ISSUE #2125: Implement PFA0(a): Updated Track-Cluster Merge/Splitter
The existing track-cluster merge/splitter needs updates to work as a component of PFAlpha.

ruse-traveler:

This PR updates the `TrackClusterMergeSplitter` algorithm to output both `edm4eic::TrackClusterMatch` and MC particle-cluster associations. In this process, it reaps what was sown by originally writing the algorithm to operate on protoclusters rather than clusters: the algorithm will now ingest fully formed cluster and update relevant quantities.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1645 #2125 )
- [ ] Documentation update
- [ ] Other: __

No.

Yes. Track-cluster and MC particle-cluster associations will now be produced by the algorithm.


src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -374,76 +434,311 @@ namespace eicrecon {
...
+
+
+  // --------------------------------------------------------------------------
+  //! Calculate cluster shape parameters
+  // --------------------------------------------------------------------------
+  /*! Calculation originally written by Chao Peng, Dhevan Gangadharan,
+   *  and Sebouh Paul.  Code is copied from the `CalorimeterClusterRecoCoG`
+   *  algorithm.
+   */
+  void TrackClusterMergeSplitter::calculate_shape_parameters(edm4eic::MutableCluster& clust) const {
```
veprbl, line 535: If this factorizes so well, we might as well move it to a separate factory.

src/detectors/EHCAL/EHCAL.cc
```diff
@@ -136,10 +136,20 @@ extern "C" {
...
-            "HcalEndcapNSplitMergeProtoClusters",
-            {"HcalEndcapNIslandProtoClusters",
-             "CalorimeterTrackProjections"},
-            {"HcalEndcapNSplitMergeProtoClusters"},
+            "HcalEndcapNSplitMergeClusters",
+            {"HcalEndcapNClusters",
+             "CalorimeterTrackProjections",
+             "HcalEndcapNClusterAssociations",
+#if EDM4EIC_VERSION_MAJOR >= 7
+             "HcalEndcapNRawHitAssocitions"},
```
veprbl, line 144: looks like a typo

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
```diff
@@ -26,33 +38,37 @@
...
-  };  // end CompareObjectID
+  };  // end CompareCluster
 
 
 
   // --------------------------------------------------------------------------
   //! Convenience types
   // --------------------------------------------------------------------------
+  using MatrixF = std::vector<std::vector<float>>;
+  using VecMatrix = std::vector<MatrixF>;
```
veprbl, line 64: Looks like, in some places you try to use `map`s, and here you give up and just rely on array indices? Is it true that every hit has a weight related to every cluster and every projection?

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
```diff
@@ -26,33 +38,37 @@
...
+    bool operator() (const edm4eic::Cluster& lhs, const edm4eic::Cluster& rhs) const {
       if (lhs.getObjectID().collectionID == rhs.getObjectID().collectionID) {
         return (lhs.getObjectID().index < rhs.getObjectID().index);
       } else {
         return (lhs.getObjectID().collectionID < rhs.getObjectID().collectionID);
       }
     }
 
-  };  // end CompareObjectID
+  };  // end CompareCluster
```
veprbl, line 56: Can we index by podio::ObjectID? Then only one implementation would be needed (I wonder if it already exists upstream...)

veprbl: I guess, having explicit object types has a value. Can we make this into a templated comparator? Basically write `template<typename T>` in front and replace `edm4eic::Cluster` with `T`.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
veprbl: (This is already a big PR. If you decide to go along with my suggestion, you could refactor existing code separately, and then rebase this to re-use that new facility)

ruse-traveler: I'd be very much in favor of that! Keeps it nice and modular 😉 

But how would this work in the data model? Would we have one collection without shape parameters filled in (likely not saved by default) and then one with them filled in (saved by default)?

src/detectors/EHCAL/EHCAL.cc
ruse-traveler: Oh shoot... Good catch!

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
ruse-traveler: Ah! Good suggestion! I'd definitely be in favor of moving it up upstream or making it available across several algorithms, but would it go into some sort of "utils" namespace?

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
ruse-traveler: So I do think we need a weight that's related to every cluster and every projection pointing to the merged cluster:
- the former is because the merged cluster will (naturally) be composed of all of the hits of the clusters being merged,
- and the latter is because we'll need to know the sum of all projections' momenta for the weight, so that summing over the split clusters should give the merged cluster energy before splitting
(Unless I'm misunderstanding how the splitting should work!)

All that being siad, I probably could rework it to use maps instead of indices. I was thinking of this map of weights (`hit_weight = hit_weight(cluster, projection)`) as a "matrix", and this was an easy way of implementing a dynamic 2D matrix of floats 🤷

veprbl: > So I do think we need a weight that's related to every cluster and every projection pointing to the merged cluster:

I think you're right. Looks like this only involves hits from the cluster group.

> All that being siad, I probably could rework it to use maps instead of indices. I was thinking of this map of weights (`hit_weight = hit_weight(cluster, projection)`) as a "matrix", and this was an easy way of implementing a dynamic 2D matrix of floats 🤷

Ideally we write the code to maximize clarity. Having a consistent use of `map`s could help with that, if there is a neat way to use those. Using maps also provides basic sanity checks to the code.

If you have a preference to use tensors here, then just remember `std::vector<std::vector<std::vector<` has triple dereferences/allocations and spoils data locality, so it won't win that much in terms of performance. We already use `eigen` library, that could be better for your case, perhaps.

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
veprbl: Let's keep the code internal first. But it would be okay to make it generic already.

Also, I've checked, podio's `ObjectID` (https://github.com/AIDASoft/podio/blob/master/include/podio/ObjectID.h) doesn't seem to implement order relation (needed for `std::map`) or hash (needed for `std::unordered_map`). We could definitely contribute the latter.

ruse-traveler: Gotcha! Sounds good! Just pushed a commit to template the comparator!

Should I open an issue on the PODIO repository?

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
ruse-traveler: I see! I'll give it some thought and let you know which route (`map` or `eigen`) I go with!

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
veprbl: Probably too minor for a full issue, let's explain it with code https://github.com/AIDASoft/podio/pull/733

ruse-traveler: Sounds good!

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
ruse-traveler: After mulling it over, I think it a natural way might be to actually use a mix of `map`s and `eigen`: for each projection, we know how many clusters and hits we're going to need so we could easily define a map of projections onto `MatrixXd`! Let me give it a shot...

ruse-traveler: I stand corrected: it was much easier to just use `std::map<edm4eic::CalorimeterHit, double>`. Just pushed some changes that rewrite the splitting calculation to use that (it's a little cleaner now).

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Rebase done! One new wrinkle I can immediately spot is how to handle the track-cluster matches: those will be pointing back to the split/merged clusters _without_ cluster shapes...

A proposal: what if we had a new `meta` algorithm for copying associations? This could be useful for other algorithms that are one-to-one transformations. I have some ideas about how this might be implemented...

veprbl: I don't get it. Cluster-track matching will be presented with the final clusters *with shapes*.

ruse-traveler: Agreed! It should be! But as-is the cluster shape calculation won't copy the track-cluster matches (unless I'm misunderstanding how associations works). So that means the clusters in the `*MergeSplitTrackClusterMatches` collections are the ones _without_ shapes...

veprbl: I thought, you were talking about the common track cluster matcher, and didn't know that we export track-cluster matches from a "private" matcher that you have. Indeed, a generic algorithms to rewrite relations would be useful.  Maybe, for now we could disable writing of those matches, or that's undesirable?

ruse-traveler: Ahhh, I see! I would definitely prefer to keep the matches for a couple reasons:
- The 1st is that these matches (rather than the common matcher) will be inputs into downstream PF components, namely the subtractor and charged candidate former;
- And the 2nd is that my hope is that the output of the merge/splitter is useful for studies beyond just particle flow (having merged clusters will be useful for, e.g., HCal studies), and so having updated matches for the merge/split clusters would be useful!

All that being said, I should note that for the PF baseline we won't need any cluster shapes, so we can factorize the PF development and working on a generic relation re-writer!

veprbl: Regarding general studies using the matches, aren't these a bit special? In the sense that they were used to construct resulting clusters, but they might be not what a track-cluster matcher would find by looking at those clusters.

I think, we can be optimistic about our ability to follow up with a relation re-writer.

ruse-traveler: > Regarding general studies using the matches, aren't these a bit special? In the sense that they were used to construct resulting clusters, but they might be not what a track-cluster matcher would find by looking at those clusters.

True! The matching procedure here does differ in that regard! But it's not clear to me how we would propagate the track-merged cluster relation downstream without outputting updated track-cluster matches...

veprbl: I'm only talking about what is being written to the file. IIRC "dangling" relations are not allowed (there is a bug in PODIO which allows for OneToOne relations to not get checked, but that may be fixed at some point). To fix those, we would need to also save clusters without the shapes.

ruse-traveler: Ahhhh! I understand now! 😅 
I'm absolutely good with disabling them in the output! 👍

ruse-traveler: Just pushed changes to disable the track-cluster associations!

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -373,77 +427,202 @@ namespace eicrecon {
...
+        ++iHit;
+      }  // end hit loop
+      trace("Merged input cluster {} into output cluster {}", old_clust.getObjectID().index, new_clust.getObjectID().index);
+    }  // end cluster loop
+
+    // update cluster position by taking energy-weighted
+    // average of positions of clusters to merge
+    edm4hep::Vector3f rClust = new_clust.getPosition();
+    for (const auto& old_clust : old_clusts) {
+      rClust = rClust + ((old_clust.getEnergy() / eClust) * old_clust.getPosition());
```
veprbl, line 501: Does this not completely ignore split_weights? Why not do it in the same loop over hits above?

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
```diff
@@ -61,11 +80,21 @@ namespace eicrecon {
...
+      std::optional<edm4eic::MCRecoClusterParticleAssociationCollection>,
+#if EDM4EIC_VERSION_MAJOR >= 7
+      std::optional<edm4eic::MCRecoCalorimeterHitAssociationCollection>
+#else
+      std::optional<edm4hep::SimCalorimeterHitCollection>
+#endif
     >,
     algorithms::Output<
-      edm4eic::ProtoClusterCollection
+      edm4eic::ClusterCollection,
```
veprbl, line 93: So, we were discussing that this should not re-implement shape calculation, and that was factorized out of RecoCoG, but a copy of RecoCoG still becomes part of this algorithm? And we don't actually output associations without shapes anymore, which, I believe, was the original point of this PR. This makes me think that we should have just made a factory to propagate associations from protoclusters to clusters instead of doing this change.

veprbl (commented):
A few comments here.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -373,77 +427,202 @@ namespace eicrecon {
...
+#else
+    const edm4hep::SimCalorimeterHitCollection* old_sim_hits,
+#endif
+    edm4eic::MCRecoClusterParticleAssociationCollection* new_clust_assocs
+  ) const {
 
-    // grab total energy
-    const float eClust = get_cluster_energy(clust) * m_cfg.sampFrac;
+    // lambda to compare MCParticles
+    auto compare = [](const edm4hep::MCParticle& lhs, const edm4hep::MCParticle& rhs) {
```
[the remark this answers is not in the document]
ruse-traveler: Whoops! Good catch!

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
ruse-traveler: So the primary point of this PR was to output specifically track-cluster matches. To do this, this requires updating the algorithm to operate on clusters rather than protoclusters since -- quite reasonably in my opinion! -- we don't have track-protocluster matches. I don't think an association propagator would help us in this context (esp. since it would necessitate a data-model change).

The partial duplication of RecoCoG then followed from that switch to clusters. My intent wasn't so much to completely duplicate the algorithm but to update various quantities in a way that's consistent with what RecoCoG does, including the handling of particle-cluster associations. My thinking is that the merging functionality here is useful beyond PF reconstruction, and so I would prefer that the produced clusters here are consistent with those that RecoCoG produces.

That being said, in the interest of keeping PF development moving, I would be willing to excise the RecoCoG bits and go with a "bare-bones" energy/position reconstruction for the time being. That should be good enough for implementing the rest of the baseline. But we should revisit this topic, though, since this certainly won't be the last reclustering algorithm we write!

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: Ahhh good catch!

src/algorithms/calorimetry/TrackClusterMergeSplitter.h
ruse-traveler: Reading my response back, I feel that not only was my tone unprofessional and dismissive, I did not clearly articulate (or articulate at all) some of my questions and concerns, and for that I apologize.

The approachI took here was informed by the intent to eventually move this algorithm downstream of a centralized track-cluster matching algorithm. As such, this algorithm reclusters already constructed clusters. This PR was my first thought as to how to do such a reclustering and I sincerely believed it was aligned with our framework. However, it clearly isn't and certainly isn't the best approach.

The points you raised make sense, and it didn't occur to me to go back through the protoclusters. I'm not completely clear on how this might look, though, so help me understand. My first thought is that such an algorithm flow might look something like this:

```
<start>-->[protoclusters]--(Reco CoG)-->[clusters]----(track matching)-->[cluster matches]--
--(convert clusters)-->[protoclusters]----(copy matches)-->[protocluster matches]--
--(merge/split)-->[protoclusters + matches]-->(Reco CoG)-->[clusters]--
--(copy matches)-->[updated matches]--><end>
```

Is this in the right direction?

veprbl: > Reading my response back, I feel that not only was my tone unprofessional and dismissive, I did not clearly articulate (or articulate at all) some of my questions and concerns, and for that I apologize.

No, it was all good points and I was able to finally understand some of these things better. I did have to push back on merge splitter being able to do its thing in isolation.

> The approach took here was informed by the intent to eventually move this algorithm downstream of a centralized track-cluster matching algorithm. As such, this algorithm reclusters already constructed clusters. This PR was my first thought as to how to do such a reclustering and I sincerely believed it was aligned with our framework. However, it clearly isn't and certainly isn't the best approach.

I'm starting to think that `edm4eic::ProtoCluster` is not such a great type. `edm4eic::Cluster` is already a superset of it, and we could have used just that one type. 

> The points you raised make sense, and it didn't occur to me to go back through the protoclusters. I'm not completely clear on how this might look, though, so help me understand. My first thought is that such an algorithm flow might look something like this:
> 
> ```
> <start>-->[protoclusters]--(Reco CoG)-->[clusters]----(track matching)-->[cluster matches]--
> --(convert clusters)-->[protoclusters]----(copy matches)-->[protocluster matches]--
> --(merge/split)-->[protoclusters + matches]-->(Reco CoG)-->[clusters]--
> --(copy matches)-->[updated matches]--><end>
> ```
> 
> Is this in the right direction?

To be honest, I can not tell with one-dimensional graph. This is what I think could work and provide maximal reuse of existing facilities:

```mermaid
flowchart TD
    RecHits--> CalorimeterIslandCluster:::algo --> IslandProtoClusters
    IslandProtoClusters --> CalorimeterClusterRecoCoG:::algo --> ClustersWithoutShapes
    --> CalorimeterClusterShape:::algo
    CalorimeterClusterRecoCoG:::algo --> ClusterAssociationsWithoutShapes
    --> CalorimeterClusterShape:::algo
    --> Clusters
    CalorimeterClusterShape:::algo --> ClusterAssociations
    Clusters --> TrackClusterMergeSplitter:::algo
    ClusterAssociations --> TrackClusterMergeSplitter:::algo
    CalorimeterTrackProjections --> TrackClusterMergeSplitter:::algo
    --> SplitMergeProtoClusters --> CalorimeterClusterRecoCoG':::algo
    TrackClusterMergeSplitter:::algo -->|remove?| SplitMergeProtoClustersAssociations
    TrackClusterMergeSplitter:::algo --> TrackSplitMergeProtoClusterMatches

    CalorimeterClusterRecoCoG':::algo --> SplitMergeClustersWithoutShapes
    --> CalorimeterClusterShape':::algo
    CalorimeterClusterRecoCoG':::algo --> SplitMergeClusterAssociationsWithoutShapes
    --> CalorimeterClusterShape':::algo
    --> SplitMergeClusters
    CalorimeterClusterShape':::algo --> SplitMergeClusterAssociations

    TrackSplitMergeProtoClusterMatches --> CalorimeterProtoClusterMatchPromotion:::algo --> SplitMergeClusterTrackMatches

    SplitMergeClusters --> ...
    SplitMergeClusterAssociations --> ...
    SplitMergeClusterTrackMatches --> ...

    classDef algo fill:#f96
```

We would need a new "CalorimeterProtoClusterMatchPromotion" algorithm operating on a new type for "TrackSplitMergeProtoClusterMatches".

(you can click quote my message to see source code of mermaid diagram and modify, if you want to adjust it)

ruse-traveler: I see! Thanks! The diagram is extremely helpful!!

> I did have to push back on merge splitter being able to do its thing in isolation.

Completely fair! I was falling into bad habits of writing big, monolithic pieces of code while working on this PR. The above topology looks a _lot_ more workable!

> I'm starting to think that edm4eic::ProtoCluster is not such a great type. edm4eic::Cluster is already a superset of it, and we could have used just that one type.

If we stick with protoclusters for the time being, would we also need an algorithm to "demote" clusters into protoclusters that could be fed into the merge/splitter?

veprbl: > If we stick with protoclusters for the time being, would we also need an algorithm to "demote" clusters into protoclusters that could be fed into the merge/splitter?

The diagram above has `TrackClusterMergeSplitter` consuming a collection of `edm4eic::Cluster`s (don't you need cluster positions to do the matching?) and outputting `edm4eic::ProtoCluster`s.

veprbl: ... so demotion happens inside TrackClusterMergeSplitter, at least to what's drawn.

ruse-traveler: Ahhhh I see! Makes sense! That would make it pretty straightforward to keep track of the protocluster energies and positions without having to duplicate `getEnergy()` and `getPosition()` as functions.

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -363,10 +363,20 @@ extern "C" {
...
+             "CalorimeterTrackProjections",
+             "LFHCALClusterAssociations",
+#if EDM4EIC_VERSION_MAJOR >= 7
+             "LFHCALRawHitAssocitions"},
+#else
+             "LFHCALHit"},
+#endif
+            {"LFHCALSplitMergeClustersWithoutShapes",
+#if EDM4EIC_VERSION_MAJOR >= 8
+             "LFHCALTrackSplitMergeClusterMatches",
```
veprbl, line 377:
```suggestion
             "LFHCALSplitMergeClusterMatches",
```
naming consistency here and elsewhere?

ruse-traveler:
Note for those following this PR: following discussion in the [09.02.2025 PF Technical Discussion](https://indico.bnl.gov/event/29615), this will be paused until the collaboration is ready to upgrade to [JANA2 2.4.3](https://github.com/JeffersonLab/JANA2/releases/tag/v2.4.3). If workforce is available, development on upstream PFAlpha components will continue.

ruse-traveler:
> Note for those following this PR: following discussion in the [09.02.2025 PF Technical Discussion](https://indico.bnl.gov/event/29615), this will be paused until the collaboration is ready to upgrade to [JANA2 2.4.3](https://github.com/JeffersonLab/JANA2/releases/tag/v2.4.3). If workforce is available, development on upstream PFAlpha components will continue.

As a potential workaround to allow PFA0 development to continue, I have opened up a PR to introduce a parallel `edm4eic::TrackProtoClusterMatch` at [EDM4eic#130](https://github.com/eic/EDM4eic/pull/130). This would an association parallel to the track-protocluster link.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -153,251 +149,173 @@ void TrackClusterMergeSplitter::process(const TrackClusterMergeSplitter::Input&
...
+  // 3. Create an output protocluster for each merged cluster
+  //    and for each track pointing to merged cluster
   // ------------------------------------------------------------------------
-  for (auto& [clustSeed, vecClustToMerge] : mapClustToMerge) {
-    merge_and_split_clusters(vecClustToMerge, mapProjToSplit[clustSeed], out_protoclusters);
+  for (auto& [clust_seed, vecClustToMerge] : mapClustToMerge) {
+
+    // create a cluster for each projection to merged cluster
+    VecProto new_protos;
+    for (const auto& project : mapProjToSplit[clust_seed]) {
```
veprbl:
```suggestion
    for (const auto& [[maybe_unused]] project : mapProjToSplit[clust_seed]) {
```
But maybe this is a sign that the creation of proto's belongs somewhere else.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -153,251 +149,173 @@ void TrackClusterMergeSplitter::process(const TrackClusterMergeSplitter::Input&
...
-  std::vector<edm4eic::MutableProtoCluster> new_clusters;
-  for (const auto& proj [[maybe_unused]] : to_split) {
-    new_clusters.push_back(out_protoclusters->create());
-  }
   trace("Splitting merged cluster across {} tracks", to_split.size());
 
-  // loop over all hits from all clusters to merge
-  std::vector<float> weights(to_split.size(), 1.);
+  // calculate weights for splitting
+  VecWeights weights(to_split.size());
```
veprbl: What is the point of hiding the type? This makes code unreadable to non-sPHENIX members.

veprbl (dismissed):
Capybara report shows some differences, I'm not sure if those are for the better or not. The part for outputting the matches looks correct though.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
ruse-traveler: TBH this is completely a "me" thing: my thinking was that this would be more expressive than just a `std::vector`, but I see your point here. There's no real benefit to hiding the type in this case.

I've opted to remove all aliases for types that are only used once (i.e. most of the maps) and retains the ones that appear multiple times in the code (e.g. the vectors of clusters, etc.). In line with some of the comments on [the PFA0a PR], I've also opted to use lower case + `_t` for all of the aliases.

veprbl:
But also you get 2 tracks more often than 1? Also why is $|\eta|<1$ cut has such a poor acceptance? I don't really understand how to interpret the plots.

mandrenguyen (commented):
A couple of quick comments.  Maybe more coming after I've had a bit more time to digest.

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -1,147 +1,145 @@
...
-    const float eProjSeed  = m_cfg.avgEP * edm4hep::utils::magnitude(projSeed.momentum);
+    const float eClustSeed = clust_seed.getEnergy();
+    const float eProjSeed  = m_cfg.avgEP * edm4hep::utils::magnitude(project_seed.value().momentum);
 
     // ----------------------------------------------------------------------
-    // 3.i. Calculate significance
+    // 2(a). Calculate significance
     // ----------------------------------------------------------------------
     const float sigSeed = (eClustSeed - eProjSeed) / m_cfg.sigEP;
     trace("Seed energy = {}, expected energy = {}, significance = {}", eClustSeed, eProjSeed,
```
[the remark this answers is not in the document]
ruse-traveler: Ah! This is a good point to clarify: the goal here is to identify clusters whose E/p is smaller than expected by `m_config.nSigEP` sigmas. If it is, then we merge all clusters in `m_config.drAdd` of the linked track-cluster!

For cases where the energy is too large wrt to the sum of tracks, then we'll propagate them as-is and create additional particles from the excess in PFA1!

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
```diff
@@ -153,251 +151,190 @@ void TrackClusterMergeSplitter::process(const TrackClusterMergeSplitter::Input&
       if (drToSeed > m_cfg.drAdd) {
```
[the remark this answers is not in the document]
ruse-traveler: Just pushed to some changes to remove `std::hypot` from the merging calculation! I opted to keep using `dist` during the splitting calculation since I was leery of changing the functional shape of the weights. But, if we need to, I think that might be a good target for optimization.

ruse-traveler:
> Hmmm.... Now that I'm looking at it with fresh eyes, I don't understand it either 😓 Let me take another stab at it :/
> 
> In the meantime, I did track down what was going on with the doubled clusters last week (I was creating an entry in the merge map even when there was nothing to merge) and pushed a change to fix it!

Following up on this, here are some updated plots. These are showing the no. of particles, tracks, clusters (_before_ merge/splitting), and track-cluster matches for 1K single pi- events. This all seems in line with what I would expect: there is almost always 1 track per event with ~98% efficiency, ~4 BHCal clusters on average, and a ~96% track-cluster matching effiency (i.e. 96% of tracks were matched to a BHCal cluster).

I haven't updated the installation of EICrecon I've been using to test in a bit, so that's why the momentum, phi, and eta of the tracks (and match E/p) are 0.

checkMatcherPerformance_genPars e10th45135nevt100pim
edd-436e-a46c-320c1de2dea6" />
checkMatcherPerformance_bhcalIslandClusters e10th45135nevt100pim
checkMatcherPerformance_barrelTrackClusterMatches e10th45135nevt100pim

ruse-traveler:
Then, to get a slightly finer-grained idea of how the algorithm was performing was behaving, I compared the distribution of clusters before (in black)/after (in green/red) the merge/splitting for `Eclust < 5 GeV` and `Eclust > 5 GeV` (attached here).

I would assume that -- to 0th order -- we would usually end up with 1 or 2 hard clusters and a lot of soft clusters, then the algorithm should merge those into one closer to the particle energy. That would mean we should see a steeper drop in soft clusters (e.g. below 5 GeV) before vs. after the merge/splitter, and a relatively smaller drop in hard clusters before vs. after. Which we sort of do!

I'm still a little surprised that there's still such a big reduction in number at the harder end, but I think that's something we can dig into during the benchmarking phase!

checkMatcherPerformance_bhcalIslandVsMergeClusters_eneL5 e10th45135nevt100pim
checkMatcherPerformance_islandVsMergeClust_eneG5 e10th45135nevt100pim d6m4y2026

veprbl (dismissed):
LGTM

veprbl (dismissed):

veprbl:
```
Message:  JOmniFactory 'EEMC:EcalEndcapNTruthClustersWithoutShapes': Wrong number of input collection names: 3 expected, 2 found.
```

wdconinc:
> ```
> Message:  JOmniFactory 'EEMC:EcalEndcapNTruthClustersWithoutShapes': Wrong number of input collection names: 3 expected, 2 found.
> ```

"Why do you use c++?"

"Because it gives the errors at compile time instead of run time."

"..."

ruse-traveler:
> ```
> Message:  JOmniFactory 'EEMC:EcalEndcapNTruthClustersWithoutShapes': Wrong number of input collection names: 3 expected, 2 found.
> ```

Hmmm... So this does seem odd:
```c++
using CalorimeterClusterRecoCoGAlgorithm = algorithms::Algorithm<
    algorithms::Input<edm4eic::ProtoClusterCollection,
#if EDM4EIC_BUILD_VERSION >= EDM4EIC_VERSION(8, 7, 0)
                      std::optional<edm4eic::MCRecoCalorimeterHitLinkCollection>,
#endif
                      std::optional<edm4eic::MCRecoCalorimeterHitAssociationCollection>>,
    algorithms::Output<edm4eic::ClusterCollection,
#if EDM4EIC_BUILD_VERSION >= EDM4EIC_VERSION(8, 7, 0)
                       std::optional<edm4eic::MCRecoClusterParticleLinkCollection>,
#endif
                       std::optional<edm4eic::MCRecoClusterParticleAssociationCollection>>>;
```
 Shouldn't the `std::optional<edm4eic::MCRecoCalorimeterHitAssociationCollection>>` be wrappend in a `#else`? But this is _also_ in main...

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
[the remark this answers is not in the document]
mandrenguyen: Thanks.  I guess that given that we're going to anyway create new particles from the excess, I'm not sure I really understand why one would filter at this stage.  Imagine you have a collinear charged and neutral hadron that share an HCAL cell.  You wouldn't want to veto the HCAL cell because the energy is incompatible with the track at this stage.  That would cause the resulting PF neutral hadron to "eat" the (true) charged particle's energy.  If you instead create the two PF particles downstream without such a veto then I would imagine you would assign only part of the HCAL energy to the PF charged hadron, such that the resulting PF neutral hadron energy would be less.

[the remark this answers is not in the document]
ruse-traveler: Agreed! The way this stage should treat a case like this is that here the HCAL cell (assuming it's in a cluster by itself) should have an `E/p` greater than expected for the charged hadron, and thus won't perform any merging on it. And since there's only 1 track pointing to the cell, no splitting will be attempted. The HCAL cell will then be copied to the output unmodified, where downstream PFA1 will divide up the charged & neutral energies in the cell!

This is a good edge case to check, though! I'll add a unit test to make sure this is behaving as expected in this case (likely in a follow-up PR).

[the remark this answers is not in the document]
ruse-traveler: Also, all that being said, if we find the performance is better, then we can easily swap the order of PFA0 and PFA1 so that this algorithm's operating on track-cluster links after excess neutral energy has been removed!

ruse-traveler:
I'm also seeing some ominous warnings about indirect memory leaks 😬 
```
Indirect leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x55f9ede05b01 in operator new(unsigned long) (/home/runner/work/EICrecon/EICrecon/install/bin/eicrecon+0x129b01) (BuildId: 6789ff755341963810b48dbf45f76e76108e7ceb)
    #1 0x7b7a8fec7d9e in std::__new_allocator<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*>::allocate(unsigned long, void const*) /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/new_allocator.h:151:27
    #2 0x7b7a8fec7d9e in std::allocator<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*>::allocate(unsigned long) /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/allocator.h:196:32
    #3 0x7b7a8fec7d9e in std::allocator_traits<std::allocator<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*>>::allocate(std::allocator<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*>&, unsigned long) /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/alloc_traits.h:515:20
    #4 0x7b7a8fec7d9e in std::_Vector_base<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*, std::allocator<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*>>::_M_allocate(unsigned long) /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/stl_vector.h:380:20
    #5 0x7b7a8fec7d9e in void std::vector<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*, std::allocator<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*>>::_M_realloc_append<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase* const&>(JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase* const&) /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/vector.tcc:596:33
    #6 0x7b7a8fec7981 in std::vector<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*, std::allocator<JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*>>::push_back(JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase* const&) /usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/stl_vector.h:1294:4
    #7 0x7b7a8fec7981 in JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::RegisterService(JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::ServiceBase*) /home/runner/work/EICrecon/EICrecon/src/extensions/jana/JOmniFactory.h:364:59
    #8 0x7b7a8fe71a83 in JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>::Service<AlgorithmsInit_service>::Service(JOmniFactory<eicrecon::CalorimeterClusterRecoCoG_factory, eicrecon::CalorimeterClusterRecoCoGConfig>*) /home/runner/work/EICrecon/EICrecon/src/extensions/jana/JOmniFactory.h:354:43
    #9 0x7b7a8fe71a83 in eicrecon::CalorimeterClusterRecoCoG_factory::CalorimeterClusterRecoCoG_factory() /home/runner/work/EICrecon/EICrecon/src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h:43:51
    #10 0x7b7a8fe6f4c8 in JOmniFactoryGeneratorT<eicrecon::CalorimeterClusterRecoCoG_factory>::GenerateFactories(JFactorySet*) /home/runner/work/EICrecon/EICrecon/src/extensions/jana/JOmniFactoryGeneratorT.h:80:31
    #11 0x7f7ab9e837ab in JFactorySet::JFactorySet(std::vector<JFactoryGenerator*, std::allocator<JFactoryGenerator*>> const&) (/opt/local/lib/libJANA.so+0x6f7ab) (BuildId: b3cddca4c48070bc5a328c0572f0712b4712b3fa)
```

(e.g. from [eicrecon-benchmarks-plugins (clang++, e, craterlake, lfhcal_studies)](https://github.com/eic/EICrecon/actions/runs/24057226148/job/70170507276?pr=1699#logs))

wdconinc:
> Shouldn't the `std::optional<edm4eic::MCRecoCalorimeterHitAssociationCollection>>` be wrappend in a `#else`? But this is _also_ in main...

Current approach is to ALSO write out Links in addition to Associations. We can drop Associations due to our 6-month backward compatibilty guarantee, so we have to switch this gradually by introducing links, switching stuff over to using links, deprecating associations, then dropping associations.

ruse-traveler:
> Current approach is to ALSO write out Links in addition to Associations. We can drop Associations due to our 6-month backward compatibilty guarantee, so we have to switch this gradually by introducing links, switching stuff over to using links, deprecating associations, then dropping associations.

Ahhhhh I see! Thanks!

src/algorithms/calorimetry/TrackClusterMergeSplitter.cc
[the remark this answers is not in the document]
mandrenguyen: Let me make sure I have it correct.  If the E/p is very large, it wouldn't invalidate the link between the track and the HCAL cluster.  Instead it would just "lock" that HCAL cluster against cluster merging?   

I think I'm missing something very basic about what cluster merging actually does in this context.  What difference will it make downstream whether a track is linked to two clusters or a single merged one?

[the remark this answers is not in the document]
ruse-traveler: Apologies for the delayed response, but that's correct! The track-cluster won't be invalidated if E/p is very large, and   no merging will be done _on_ that cluster! Though, there is an edge case here, which I'll describe below.

The goal behind this merging step is to correct split clusters, which are identified here by a very low E/p! In these cases, we should see a cluster (or clusters) near the low E/p track-cluster link which _don't_ have a linked track. If we were to pass the clusters as-is to the subtraction step, then these cluster fragments would be incorrectly treated as neutral energy!

**Edge Case:** in the event that a _very low_ E/p track-cluster link is within `drAdd` of a _very large_ E/p track-cluster link, then the very large cluster will be merged into the very low cluster. This results in a single cluster with 2 tracks linked to it. The splitting step, where this single cluster is split between tracks the linked to it, is intended to mitigate this effect!

veprbl:
To complete this PR, one needs to repeat my last change for other detectors.

ruse-traveler:
> To complete this PR, one needs to repeat my last change for other detectors.

Gotcha! I'll implement those changes ASAP

src/detectors/BHCAL/BHCAL.cc
```diff
@@ -204,7 +210,7 @@ void InitPlugin(JApplication* app) {
 #if EDM4EIC_BUILD_VERSION >= EDM4EIC_VERSION(8, 7, 0)
           "HcalBarrelRawHitLinks", // edm4eic::MCRecoCalorimeterHitLink
 #endif
-          "HcalBarrelRawHitAssociations" // edm4eic::MCRecoCalorimeterHitAssociationCollection
+          "HcalBarrelRawHitAssociations" //edm4eic::MCRecoCalorimeterHitAssociationCollection
```
ruse-traveler:
```suggestion
          "HcalBarrelRawHitAssociations" // edm4eic::MCRecoCalorimeterHitAssociationCollection
```

src/detectors/FHCAL/FHCAL.cc
```diff
@@ -362,14 +368,19 @@ void InitPlugin(JApplication* app) {
 #if EDM4EIC_BUILD_VERSION >= EDM4EIC_VERSION(8, 7, 0)
           "LFHCALRawHitLinks", // edm4eic::MCRecoCalorimeterHitLink
 #endif
-          "LFHCALRawHitAssociations" // edm4hep::MCRecoCalorimeterHitAssociationCollection
+          "LFHCALRawHitAssociations" // edm4eic::MCRecoCalorimeterHitAssociationCollection
```
ruse-traveler:
```suggestion
          "LFHCALRawHitAssociations" // edm4hep::MCRecoCalorimeterHitAssociationCollection
```

src/detectors/FEMC/FEMC.cc
```diff
@@ -234,22 +234,27 @@ void InitPlugin(JApplication* app) {
...
+       .surfaceToUse                 = 1,
        .transverseEnergyProfileScale = 1.0},
       app // TODO: remove me once fixed
       ));
 
   app->Add(new JOmniFactoryGeneratorT<CalorimeterClusterRecoCoG_factory>(
       "EcalEndcapPSplitMergeClustersWithoutShapes",
       {
-          "EcalEndcapPSplitMergeProtoClusters", // edm4eic::ProtoClusterCollection
+          "EcalEndcapPSplitMergeProtoClusters", //edm4eic::ProtoClusterCollection
```
ruse-traveler:
```suggestion
          "EcalEndcapPSplitMergeProtoClusters", // edm4eic::ProtoClusterCollection
```

veprbl (approved):
Ship it

PR #1703, opened 2025-01-14
fix: avoid integer abs in MatrixTransferStatic.cc and calorimetry_HEXPLIT.cc

wdconinc:
`abs` without `std` results in using integer absolute value, which the compiler warns about:
```
warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
```

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (dismissed):

veprbl (dismissed):

veprbl:
How fun, it only became visible with std::abs. #1751

wdconinc:
> How fun, it only became visible with std::abs. #1751

What's up with the placement of those warnings though :-)

veprbl:
> > How fun, it only became visible with std::abs. #1751
> 
> What's up with the placement of those warnings though :-)

Yeah... I've convinced myself, they are off by one.

veprbl (approved):

veprbl (approved):

PR #1712, opened 2025-01-15
fix: rm unused options in IterativeVertexFinder.cc

wdconinc:
### Briefly, what does this PR introduce?
This PR removes an entirely unused options variable (and type alias) in IterativeVertexFinder.

### What kind of change does this PR introduce?
- [x] Bug fix (issue: unused code)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1525, opened 2024-07-02
feat: fix compilation for Acts versions from v33 up to v35

wdconinc:
### Briefly, what does this PR introduce?
This adapts our tracking algorithms to any Acts version between v31 and v36 (to be released, so additional refactors may hit the Acts main branch that would need porting).

The commits in this branch are topologically ordered with reference to the commit message title, Acts PR number, and version validity. All commits compile individually, but no event-for-event validation has been performed.

The support for mulitple versions is done by somewhat extensive use of preprocessor directives based on `Acts_VERSION_MAJOR`.

Builds with:
- v32.1.0: https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/3481513#L3103
- v33.1.0: https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/3482111#L3549
- v34.1.0: https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/3482134#L3699
- v35.0.0: locally, but not yet in eic_container since this requires #1520 and related eic-spack PRs first

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue: Acts versions are released faster than we can keep up with)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1677, opened 2024-11-24
CKFTracking: remove Acts tracks that failed to extrapolate, not just the trajectories

veprbl:
### Briefly, what does this PR introduce?

This is re-implements #1663 in such a way that downstream algorithms, such as the AmbiguitySolver, that operate on Acts EDM do not see the failed tracks.
Resolves: #1672

### What kind of change does this PR introduce?
- [x] Bug fix (issue #1672)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No

### Does this PR change default behavior?
No

PR #1715, opened 2025-01-16
feat: fix compilation for Acts versions from v36 up to v39

wdconinc:
This adapts our tracking algorithms to any Acts version between v36 and v38.

The commits in this branch are topologically ordered with reference to the commit message title, Acts PR number, and version validity. All commits compile individually, but no event-for-event validation has been performed. See also https://github.com/eic/EICrecon/pull/1525.

The support for mulitple versions is done by somewhat extensive use of preprocessor directives based on `Acts_VERSION_MAJOR` and `Acts_VERSION_MINOR`.

Builds with: 35.2.0 36.0.0 36.1.0 36.2.0 36.3.2 37.4.0 38.2.0 39.1.0

Needs:
- [x] #1711 
- [x] #1712

- [ ] Bug fix (issue #__)
- [x] New feature (issue: Acts versions are released faster than we can keep up with)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -76,7 +76,7 @@
 #include <list>
 #include <optional>
 #include <ostream>
-#include <ranges>
+#include <set>
 #include <system_error>
 #include <utility>
 
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -11,7 +11,6 @@
 #include <Acts/EventData/TrackStateProxy.hpp>
 #include <Acts/EventData/Types.hpp>
 #include <Acts/Geometry/Layer.hpp>
-#include <Acts/Propagator/ActionList.hpp>
 #include <boost/container/vector.hpp>
 #include <boost/move/utility_core.hpp>
 #if Acts_VERSION_MAJOR < 36
@@ -28,12 +27,20 @@
 
 #include "Acts/EventData/ProxyAccessor.hpp"
 #if Acts_VERSION_MAJOR >= 34
+#if Acts_VERSION_MAJOR >= 37
+#include "Acts/Propagator/ActorList.hpp"
+#else
 #include "Acts/Propagator/AbortList.hpp"
+#include <Acts/Propagator/ActionList.hpp>
+#endif
 #include "Acts/Propagator/EigenStepper.hpp"
 #include "Acts/Propagator/MaterialInteractor.hpp"
 #include "Acts/Propagator/Navigator.hpp"
 #endif
 #include <Acts/Propagator/Propagator.hpp>
+#if Acts_VERSION_MAJOR >= 36
+#include "Acts/Propagator/PropagatorOptions.hpp"
+#endif
 #if Acts_VERSION_MAJOR >= 34
 #include "Acts/Propagator/StandardAborters.hpp"
 #endif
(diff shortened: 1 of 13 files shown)


src/algorithms/tracking/CKFTrackingFunction.cc
```diff
@@ -38,8 +38,13 @@ namespace eicrecon{
...
+#if Acts_VERSION_MAJOR >= 36
+  using CKF =
+      Acts::CombinatorialKalmanFilter<Propagator, ActsExamples::TrackContainer>;
+#else
   using CKF =
       Acts::CombinatorialKalmanFilter<Propagator, Acts::VectorMultiTrajectory>;
+#endif
 
   using TrackContainer =
       Acts::TrackContainer<Acts::VectorTrackContainer,
```
wdconinc, line 50: I think `TrackContainer` is now unused for Acts_VERSION_MAJOR >= 36.

wdconinc: Likewise, `#include <Acts/EventData/VectorMultiTrajectory.hpp>` is probably unnecessary for Acts_VERSION_MAJOR >= 36.

wdconinc:
# ACTS 36.2.0

https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/4693815/artifacts/file/results/tracking_performances/local/mom_resol_-1.0_eta_1.0.png

# ACTS 37.4.0

https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/4694190/artifacts/file/results/tracking_performances/local/mom_resol_-1.0_eta_1.0.png

# ACTS 38.2.0

https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/4694543/artifacts/file/results/tracking_performances/local/mom_resol_-1.0_eta_1.0.png

veprbl:
https://veprbl.github.io/capybara-reports/960fcece9a90344c1b230c23d1f10bec/

src/algorithms/tracking/CKFTracking.cc
```diff
@@ -371,6 +451,11 @@ namespace eicrecon {
 
           lastSeed = constSeedNumber(track);
 
+          if (!track.hasReferenceSurface()) {
+            ACTS_WARNING("Track has no reference surface.");
+            continue;
+          }
```
veprbl, line 457: Where did this come from? Upstream code only has a `ACTS_ERROR("Track has no reference surface");` but in a different context.

veprbl: Also, to simply continue here is not correct. We tried that before, and this is what was implemented instead: https://github.com/eic/EICrecon/pull/1677

wdconinc: Thanks for the careful review. Segfaults happen two lines below this even with the #1677 treatment included (which has always been in this branch for testing since it well predates it). I guess that means even with a successful extrapolation we now have referenceSurfacePtrs that are null and the check in #1677 is insufficient.

Suggestion: revert this, and I'll dig into it more.

wdconinc: Actually, it may be that in #1677 we should have had the addition to failed tracks also in the smoothing step.

wdconinc: At least this is an easy to reproduce issue, since it segfaults out on the very first event (and presumably every event).

wdconinc: #1741 doesn't address the underlying issue, but is a sensible fix anyway.

veprbl:
So there were two crash-inducing bugs. The second one has been worked around, and reported as https://github.com/acts-project/acts/issues/4166.

veprbl (approved):
Looks like this allows to start testing with Acts up to 39.1.0. It also allows to update Acts to 36.0.0 without regressions: https://veprbl.github.io/capybara-reports/ed0ce6b7e234b7d1786b388532d040bf/

PR #1728, opened 2025-02-05
BUGFIX: PIDLookup: fix proton sampling threshold

veprbl:
Electron threshold was used for sampling protons due to what looks like some copy-paste error. The threshold reported in ParticleID for 2212 is correct, however. Basically, this is only issue for `ReconstructedParticles.PDG`.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes

### DIFF
--- a/src/algorithms/pid_lut/PIDLookup.cc
+++ b/src/algorithms/pid_lut/PIDLookup.cc
@@ -126,7 +126,7 @@ void PIDLookup::process(const Input& input, const Output& output) const {
         identified_pdg = 321; // kaon
         recopart.setParticleIDUsed((*partids_out)[partids_out->size() - 2]);
       } else if (random_unit_interval < (entry->prob_electron + entry->prob_pion +
-                                         entry->prob_kaon + entry->prob_electron)) {
+                                         entry->prob_kaon + entry->prob_proton)) {
         identified_pdg = 2212; // proton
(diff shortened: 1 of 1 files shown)


wdconinc (approved):

src/algorithms/pid_lut/PIDLookup.cc
```diff
@@ -126,7 +126,7 @@ void PIDLookup::process(const Input& input, const Output& output) const {
         identified_pdg = 321; // kaon
         recopart.setParticleIDUsed((*partids_out)[partids_out->size() - 2]);
       } else if (random_unit_interval < (entry->prob_electron + entry->prob_pion +
-                                         entry->prob_kaon + entry->prob_electron)) {
+                                         entry->prob_kaon + entry->prob_proton)) {
```
wdconinc, line 129: The underlying problem here is probably that we encourage copy & paste. We should not have `entry` be a struct with fields per particle type. Can this be treated as an array with a more programmatic approach that is sure to not lead to copy/paste errors?

veprbl: Probably. Even if that's the case, there is no point in future-proofing PID lookup codebase.

PR #1731, opened 2025-02-07
Lambda reconstruction in ZDC

sebouh137:
Reconstruction of Lambda baryons using the ZDC.  Also includes functionality to reconstruction photons and neutrons in the ZDC and distinguish between them.  

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no
Yes.  Adds two new algorithms (one for photon and neutron recon in the ZDC, and a second to find/fit lambda candidates from these particles).


src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -5,7 +5,7 @@
 #include <edm4eic/ReconstructedParticleCollection.h>
 #include <edm4hep/Vector3f.h>
 #include <edm4hep/utils/vector_utils.h>
-#include <cmath.h>
+#include <math.h>
```
veprbl, line 8: `<cmath.h>` is not a thing `<cmath>` is.

sebouh137: thanks!

sebouh137:
Dear @veprbl , 
Let me know if there's anything else I need do in order to get your approval for this pull request.  
Thanks,
Sebouh

wdconinc:
Can you properly rebase so no unrelated changes show up here?

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+      xn.RotateY(-m_cfg.rot_y);
+      x1.RotateY(-m_cfg.rot_y);
+      x2.RotateY(-m_cfg.rot_y);
+
+
+      /*std::cout << "nx recon " << xn.X() << " g1x recon " << x1.X() << " g2x recon " << x2.X() << std::endl;
+
+        std::cout << "nz recon " << xn.Z() << " g1z recon " << x1.Z() << " g2z recon " << x2.Z() << " z face=" << m_cfg.zmax<< std::endl;*/
+
+
```
veprbl, line 60:
```suggestion
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+          g1.Boost(b);
+          g2.Boost(b);
+
+          //convert vertex to mm:
+          vtx=vtx*(1/dd4hep::mm);
+
+          auto rec_part = out_lambdas->create();
+          rec_part.setPDG(3122);
+          //edm4hep::Vector3f position(vtx.X(), vtx.Y(), vtx.Z());
+          //edm4hep::Vector3f momentum(lambda.X(), lambda.Y(), lambda.Z());
```
veprbl, line 102:
```suggestion
```

src/algorithms/reco/FarForwardLambdaReconstruction.h
```diff
@@ -0,0 +1,51 @@
...
+                                        "Reconstructs lambda candidates and their decay products from the reconstructed neutrons and photons"} {}
+
+         void init() final;
+         void process(const Input&, const Output&) const final;
+
+    private:
+        std::shared_ptr<spdlog::logger> m_log;
+        double m_neutron{0.93956542052};
+      double m_pi0{0.1349768277676847};
+      double m_lambda{1.115683138712051};
```
veprbl, line 47: Please use `ParticleSvc` to retrieve these.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+      double l1=sqrt(cluster.getShapeParameters(4))*dd4hep::mm;
+      double l2=sqrt(cluster.getShapeParameters(5))*dd4hep::mm;
+      double l3=sqrt(cluster.getShapeParameters(6))*dd4hep::mm;
+
+      //z in the local coordinates
+      double z= (cluster.getPosition().z*cos(m_cfg.rot_y)+cluster.getPosition().x*sin(m_cfg.rot_y))*dd4hep::mm;
+
+      /*std::cout << "zcut=" << m_cfg.gamma_zmax << " z recon=" << z << std::endl;
+      std::cout << "l1=" << l1 << " l2=" << l2 << " l3=" << l3 << std::endl;
+      std::cout << "max length=" << m_cfg.gamma_max_length << " max width=" << m_cfg.gamma_max_width << std::endl;*/
```
veprbl, line 57:
```suggestion
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+      }
+
+      double Etot=Etot_hcal;
+      if (Etot > 0 && Emax > 0){
+          auto rec_part = out_neutrons->create();
+          double corr=calc_corr(Etot,m_cfg.n_scale_corr_coeff_hcal);
+          Etot_hcal=Etot_hcal/(1+corr);
+          //corr=calc_corr(Etot,m_cfg.scale_corr_coeff_ecal);
+          //Etot_ecal=Etot_ecal/(1+corr);
+          Etot=Etot_hcal;//+Etot_ecal;
```
veprbl, line 113:
```suggestion
          Etot=Etot_hcal;
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+          rec_part.setCharge(0);
+          rec_part.setMass(m_neutron);
+          for (const auto& cluster : *clustersHcal){
+            rec_part.addToClusters(cluster);
+          }
+          /*for (const auto& cluster : *clustersEcal){
+            rec_part.addToClusters(cluster);
+            }*/
+      }
+        //m_log->debug("Found {} neutron candidates", out_neutrons->size());
```
veprbl, line 130:
```suggestion
      }
```

src/factories/reco/FarForwardNeutralsReconstruction_factory.h
```diff
@@ -0,0 +1,49 @@
...
+    PodioInput<edm4eic::Cluster> m_clusters_hcal_input {this};
+    PodioOutput<edm4eic::ReconstructedParticle> m_neutrons_output {this};
+    PodioOutput<edm4eic::ReconstructedParticle> m_gammas_output {this};
+    ParameterRef<std::vector<double>> m_n_scale_corr_coeff_hcal     {this, "n_scale_corr_coeff_hcal",          config().n_scale_corr_coeff_hcal};
+    ParameterRef<std::vector<double>> m_gamma_scale_corr_coeff_hcal     {this, "gamma_scale_corr_coeff_hcal",          config().gamma_scale_corr_coeff_hcal};
+    ParameterRef<double> m_rot_y     {this, "rot_y",          config().rot_y};
+    ParameterRef<double> m_gamma_zmax     {this, "gamma_zmax",          config().gamma_zmax};
+
+    ParameterRef<double> m_gamma_max_length     {this, "gamma_max_length",          config().gamma_max_length};
+    ParameterRef<double> m_gamma_max_width     {this, "gamma_max_width",          config().gamma_max_width};
```
veprbl, line 28:
```suggestion
    ParameterRef<std::vector<double>> m_n_scale_corr_coeff_hcal     {this, "nScaleCorrCoeffHcal",          config().n_scale_corr_coeff_hcal};
    ParameterRef<std::vector<double>> m_gamma_scale_corr_coeff_hcal     {this, "gammaScaleCorrCoeffHcal",          config().gamma_scale_corr_coeff_hcal};
    ParameterRef<double> m_rot_y     {this, "rotY",          config().rot_y};
    ParameterRef<double> m_gamma_zmax     {this, "gammaZMax",          config().gamma_zmax};

    ParameterRef<double> m_gamma_max_length     {this, "gammaMaxLength",          config().gamma_max_length};
    ParameterRef<double> m_gamma_max_width     {this, "gammaMaxWidth",          config().gamma_max_width};
```
Config names exposed to JANA2 are camelCase.

src/global/reco/reco.cc
```diff
@@ -310,15 +312,42 @@ void InitPlugin(JApplication *app) {
...
+    ));
+    app->Add(new JOmniFactoryGeneratorT<FarForwardNeutralsReconstruction_factory>(
+           "ReconstructedFarForwardZDCNeutrons",
+           {"HcalFarForwardZDCClusters"},  // edm4eic::ClusterCollection
+           {"ReconstructedFarForwardZDCNeutrons", "ReconstructedFarForwardZDCGammas"}, // edm4eic::ReconstrutedParticleCollection,
+           {
+            .n_scale_corr_coeff_hcal={-0.11, -1.5, 0},
+            .gamma_scale_corr_coeff_hcal={0, -.13, 0},
+            .rot_y=-0.025,
+            .gamma_zmax=(300+35800)*dd4hep::mm,
```
veprbl, line 332:
```suggestion
            .gamma_zmax=(300+35800)*dd4hep::mm,
```
Is it possible to retrieve these from the geometry?

src/global/reco/reco.cc
```diff
@@ -310,15 +312,42 @@ void InitPlugin(JApplication *app) {
...
+            },
           app   // TODO: Remove me once fixed
     ));
+    app->Add(new JOmniFactoryGeneratorT<FarForwardLambdaReconstruction_factory>(
+           "ReconstructedFarForwardZDCLambdas",
+           {"ReconstructedFarForwardZDCNeutrons", "ReconstructedFarForwardZDCGammas"}, // edm4eic::ReconstrutedParticleCollection,
+           {"ReconstructedFarForwardZDCLambdas", "ReconstructedFarForwardZDCLambdaDecayProducts"}, // edm4eic::ReconstrutedParticleCollection,
+           {
+             .rot_y=-0.025,
+             .zmax=35800*dd4hep::mm,
```
veprbl, line 344: Same question.

src/algorithms/reco/FarForwardNeutralsReconstruction.h
```diff
@@ -0,0 +1,53 @@
...
+                  {"outputNeutrons", "outputGammas"},
+                                        "Merges all HCAL clusters in a collection into a neutron candidate and photon candidates "} {}
+
+         void init() final;
+         void process(const Input&, const Output&) const final;
+         double calc_corr(double Etot, const std::vector<double>&) const;
+         bool isGamma(const edm4eic::Cluster& cluster) const;
+    private:
+        std::shared_ptr<spdlog::logger> m_log;
+        double m_neutron{0.93956542052*dd4hep::GeV};
```
veprbl, line 48: Also `ParticleSvc`.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+
+          Etot_hcal += E;
+          if (E > Emax){
+            Emax = E;
+            n_position = cluster.getPosition();
+          }
+      }
+
+      double Etot=Etot_hcal;
+      if (Etot > 0 && Emax > 0){
```
ruse-traveler, line 107: Correct me if I'm wrong, but the neutron reconstruction here is identical to the one in the existing `FarForwardNeutronReconstruction` algorithm, right? If so, we need clarify how the two algorithms relate to each other.

Personally, I'm in favor of moving the reconstruction to this algorithm: the fact that you identify gammas ahead of time means you can clean up potential background to your neutron reconstruction! But then I would strongly suggest to remove the neutron-only algorithm

src/factories/reco/FarForwardNeutralsReconstruction_factory.h
ruse-traveler, line 28: Formatting:
```suggestion
    ParameterRef<std::vector<double>> m_n_scale_corr_coeff_hcal {this, "n_scale_corr_coeff_hcal", config().n_scale_corr_coeff_hcal};
    ParameterRef<std::vector<double>> m_gamma_scale_corr_coeff_hcal {this, "gamma_scale_corr_coeff_hcal", config().gamma_scale_corr_coeff_hcal};
    ParameterRef<double> m_rot_y {this, "rot_y", config().rot_y};
    ParameterRef<double> m_gamma_zmax {this, "gamma_zmax", config().gamma_zmax};
    ParameterRef<double> m_gamma_max_length {this, "gamma_max_length", config().gamma_max_length};
    ParameterRef<double> m_gamma_max_width {this, "gamma_max_width", config().gamma_max_width};
```

src/factories/reco/FarForwardNeutralsReconstruction_factory.h
```diff
@@ -0,0 +1,49 @@
...
+#include "extensions/jana/JOmniFactory.h"
+
+
+namespace eicrecon {
+
+  class FarForwardNeutralsReconstruction_factory : public JOmniFactory<FarForwardNeutralsReconstruction_factory,FarForwardNeutralsReconstructionConfig> {
+
+   using AlgoT = eicrecon::FarForwardNeutralsReconstruction;
+     private:
+         std::unique_ptr<AlgoT> m_algo;
```
ruse-traveler, line 18: Formatting
```suggestion
    public:
      using AlgoT = eicrecon::FarForwardNeutralsReconstruction;
    private:
      std::unique_ptr<AlgoT> m_algo;
```

src/factories/reco/FarForwardLambdaReconstruction_factory.h
```diff
@@ -0,0 +1,47 @@
...
+    PodioInput<edm4eic::ReconstructedParticle> m_neutrons_input {this};
+    PodioInput<edm4eic::ReconstructedParticle> m_gammas_input {this};
+    PodioOutput<edm4eic::ReconstructedParticle> m_lambda_output {this};
+    PodioOutput<edm4eic::ReconstructedParticle> m_decay_products_output {this};
+
+    ParameterRef<double> m_rot_y     {this, "rot_y",          config().rot_y};
+    ParameterRef<double> m_zmax     {this, "zmax",          config().zmax};
+    ParameterRef<double> m_lambda_max_mass_dev     {this, "lambda_max_mass_dev",          config().lambda_max_mass_dev};
+    ParameterRef<int> m_iterations     {this, "iterations",          config().iterations};
+    Service<AlgorithmsInit_service> m_algorithmsInit {this};
```
ruse-traveler, line 28: Formatting
```suggestion
    ParameterRef<double> m_rot_y {this, "rot_y", config().rot_y};
    ParameterRef<double> m_zmax {this, "zmax", config().zmax};
    ParameterRef<double> m_lambda_max_mass_dev {this, "lambda_max_mass_dev", config().lambda_max_mass_dev};
    ParameterRef<int> m_iterations {this, "iterations", config().iterations};

    Service<AlgorithmsInit_service> m_algorithmsInit {this};
```

src/factories/reco/FarForwardLambdaReconstruction_factory.h
```diff
@@ -0,0 +1,47 @@
...
+#include "extensions/jana/JOmniFactory.h"
+
+
+namespace eicrecon {
+
+  class FarForwardLambdaReconstruction_factory : public JOmniFactory<FarForwardLambdaReconstruction_factory,FarForwardLambdaReconstructionConfig> {
+
+   using AlgoT = eicrecon::FarForwardLambdaReconstruction;
+     private:
+         std::unique_ptr<AlgoT> m_algo;
```
ruse-traveler, line 18: Formatting
```suggestion
    public:
      using AlgoT = eicrecon::FarForwardLambdaReconstruction;
    private:
      std::unique_ptr<AlgoT> m_algo;
```

src/factories/calorimetry/HEXPLIT_factory.h
```diff
@@ -20,6 +20,7 @@ class HEXPLIT_factory : public JOmniFactory<HEXPLIT_factory, HEXPLITConfig> {
 
     ParameterRef<double> m_MIP      {this, "MIP",           config().MIP};
     ParameterRef<double> m_Emin_in_MIPs     {this, "Emin_in_MIPs",          config().Emin_in_MIPs};
+    ParameterRef<double> m_delta_in_MIPs     {this, "delta_in_MIPs",          config().delta_in_MIPs};
     ParameterRef<double> m_tmax     {this, "tmax",          config().tmax};
```
ruse-traveler, line 24: Formatting
```suggestion
    ParameterRef<double> m_MIP {this, "MIP", config().MIP};
    ParameterRef<double> m_Emin_in_MIPs {this, "Emin_in_MIPs", config().Emin_in_MIPs};
    ParameterRef<double> m_delta_in_MIPs {this, "delta_in_MIPs", config().delta_in_MIPs};
    ParameterRef<double> m_tmax {this, "tmax", config().tmax};
```

(I know I'm being tedious, but small things like formatting really help us in the long run!)

src/algorithms/reco/FarForwardNeutralsReconstruction.h
```diff
@@ -0,0 +1,53 @@
...
+    >;
+    class FarForwardNeutralsReconstruction :
+       public FarForwardNeutralsReconstructionAlgorithm,
+       public WithPodConfig<FarForwardNeutralsReconstructionConfig> {
+       public:
+         FarForwardNeutralsReconstruction(std::string_view name)
+                  : FarForwardNeutralsReconstructionAlgorithm{name,
+                                        {"inputClustersHcal"},
+                  {"outputNeutrons", "outputGammas"},
+                                        "Merges all HCAL clusters in a collection into a neutron candidate and photon candidates "} {}
```
ruse-traveler, line 40: Formatting
```suggestion
         FarForwardNeutralsReconstruction(std::string_view name)
                  : FarForwardNeutralsReconstructionAlgorithm{
                      name,
                      {"inputClustersHcal"},
                      {"outputNeutrons", "outputGammas"},
                      "Merges all HCAL clusters in a collection into a neutron candidate and photon candidates "}
                  {}
```

src/algorithms/reco/FarForwardNeutralsReconstruction.h
```diff
@@ -0,0 +1,53 @@
...
+
+         void init() final;
+         void process(const Input&, const Output&) const final;
+         double calc_corr(double Etot, const std::vector<double>&) const;
+         bool isGamma(const edm4eic::Cluster& cluster) const;
+    private:
+        std::shared_ptr<spdlog::logger> m_log;
```
ruse-traveler, line 47: These need to be private (they can't be part of the algorithm's API)
```suggestion
    private:
         double calc_corr(double Etot, const std::vector<double>&) const;
         bool isGamma(const edm4eic::Cluster& cluster) const;

        std::shared_ptr<spdlog::logger> m_log;
```

src/algorithms/reco/FarForwardNeutralsReconstruction.h
ruse-traveler, line 48: As above, I'd suggest using `ParticleSvc` instead of hard coding PDG values.
```suggestion
    const algorithms::ParticleSvc& m_particleSvc = algorithms::ParticleSvc::instance();
```

src/algorithms/reco/FarForwardNeutralsReconstruction.h
ruse-traveler: Also, super minor point: I'd suggest being consistent with camel-vs.-snake case (e.g. `calc_corr` vs. `isGamma`). It's small, but helps with readability!

src/algorithms/reco/FarForwardLambdaReconstructionConfig.h
```diff
@@ -0,0 +1,21 @@
...
+#pragma once
+#include <float.h>
+#include <DD4hep/Detector.h>
+
+
+namespace eicrecon {
+
+  struct FarForwardLambdaReconstructionConfig {
+    /** transformation from global coordinates to proton-frame coordinates*/
+    double rot_y=-0.025;
```
ruse-traveler, line 12: I'd strongly suggest a more descriptive name here, especially considering that a parameter with the same name is used for a different rotation in the neutral reconstruction!
```suggestion
    double global_to_proton_rotation=-0.025;
```

src/algorithms/reco/FarForwardNeutralsReconstructionConfig.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace eicrecon {
+
+  struct FarForwardNeutralsReconstructionConfig {
+    /** Correction factors for neutrons in the Hcal */
+    std::vector<double>      n_scale_corr_coeff_hcal={-0.11, -1.5, 0};
+    /** Correction factors for gammas in the Hcal */
+    std::vector<double>      gamma_scale_corr_coeff_hcal={0, -0.13, 0};
+    /** rotation from global to local coordinates */
+    double rot_y=-0.025;
```
ruse-traveler, line 16: Like above, I'd suggest a more descriptive name here!
```suggestion
    double global_to_local_rotation=-0.025;
```

ruse-traveler (changes requested):
Hi Sebouh, this is very nice to see! Thank you for the hard work on this! Just a heads-up: this is a big PR and is probably going to take a couple iterations, so please bear with me.

I think we need to clarify a couple major points, namely:
1. How do these two algorithms relate to each other (`NeutralReconstruction` to `LambdaReconstruction`)?
2. How does this new `NeutralReconstruction` algorithm relate to the old `FarForwardNeutronReconstruction` algorithm?

Towards point 1, something I immediately noticed is that there are 4 `edm4eic::ReconstructedParticle` collections created (and saved by default) across these 2 algorithms, where 1 is a subset of another. This could be reduced to just 2 by making use of features we already have built into our data model: gammas/neutrons can be distinguished by PID, and resonance-to-decay connections can be encoded via relations.

And towards point 2, I noticed the neutron reconstruction functionality in the new `NeutralReconstruction` algorithm seems to duplicate the functionality of the existing `NeutronReconstruction`, which raises the question of what the two separate neutron collections are _for_? Is there a reason the neutrons from the new algorithm won't also suffice for more general analyses?

I've sketched out several suggestions towards both of these points in my comments!

src/algorithms/reco/FarForwardNeutralsReconstruction.h
veprbl: m_log should not be defined in the algorithm. All the methods like `debug()` are already inherited.

sebouh137:
@wdconinc @veprbl @ruse-traveler , thanks for your feedback; I will make sure these changes are made.

src/algorithms/reco/FarForwardNeutralsReconstruction.h
sebouh137: done.

src/algorithms/reco/FarForwardLambdaReconstruction.h
sebouh137: done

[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+      /*std::cout << "zcut=" << m_cfg.gamma_zmax << " z recon=" << z << std::endl;
+      std::cout << "l1=" << l1 << " l2=" << l2 << " l3=" << l3 << std::endl;
+      std::cout << "max length=" << m_cfg.gamma_max_length << " max width=" << m_cfg.gamma_max_width << std::endl;*/
+      if (z> m_cfg.gamma_zmax)
+        return false;
+      if (l1> m_cfg.gamma_max_length || l2> m_cfg.gamma_max_length || l3 > m_cfg.gamma_max_length)
+        return false;
+      if ((l1> m_cfg.gamma_max_width) + (l2> m_cfg.gamma_max_width) + (l3 > m_cfg.gamma_max_width)>2)
+        return false;
+      return true;
```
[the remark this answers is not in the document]
sebouh137: Actually, what I meant was >=2.   The second largest of l1, l2, and l3 should be less than m_cfg.gamma_max_width.  Fixed.

src/algorithms/reco/FarForwardNeutralsReconstruction.h
sebouh137: done

sebouh137:
> Hi Sebouh, this is very nice to see! Thank you for the hard work on this! Just a heads-up: this is a big PR and is probably going to take a couple iterations, so please bear with me.
> 
> I think we need to clarify a couple major points, namely:
> 
> 1. How do these two algorithms relate to each other (`NeutralReconstruction` to `LambdaReconstruction`)?
Neutral reconstruction identifies which showers belong to neutrons and to photons, and creates reconstructed particles from those.   Lambda reconstruction takes the output of Neutral reconstruction and outputs the lambda reconstructed particle, and also the decay products in the lambda CM frame.  
> 2. How does this new `NeutralReconstruction` algorithm relate to the old `FarForwardNeutronReconstruction` algorithm?
Neutral is an upgraded version of the old neutron algorithm, in that it separates out the clusters identified as photons, and merges the other clusters to form a neutron candidate, whereas in the old neutron reconstruction, all clusters were merged together into one neutron candidate.  It would still be possible to get the same results as the older neutron reconstruction if you set the max length of the photons to zero (in that case, none of the clusters would be identified as photons, and everything would be merged into the neutron).  So I don't think the original neutron reconstruction is still needed.  

> 
> Towards point 1, something I immediately noticed is that there are 4 `edm4eic::ReconstructedParticle` collections created (and saved by default) across these 2 algorithms, where 1 is a subset of another. This could be reduced to just 2 by making use of features we already have built into our data model: gammas/neutrons can be distinguished by PID, and resonance-to-decay connections can be encoded via relations.
Seems reasonable.  
> 
> And towards point 2, I noticed the neutron reconstruction functionality in the new `NeutralReconstruction` algorithm seems to duplicate the functionality of the existing `NeutronReconstruction`, which raises the question of what the two separate neutron collections are _for_? Is there a reason the neutrons from the new algorithm won't also suffice for more general analyses?
> 
The new algorithm, in principle, should suffice for general analyses.  So I think I should drop the previous one, and just have one algorithm for making 
> I've sketched out several suggestions towards both of these points in my comments!
Thanks, I'll take a look at these

src/factories/reco/FarForwardNeutralsReconstruction_factory.h
sebouh137: converted to camelCase

src/global/reco/reco.cc
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "algorithms/interfaces/WithPodConfig.h"
+#include "algorithms/reco/FarForwardLambdaReconstructionConfig.h"
+
+namespace eicrecon {
+
+using FarForwardLambdaReconstructionAlgorithm = algorithms::Algorithm<
+   algorithms::Input<
+      const edm4eic::ReconstructedParticleCollection,
+      const edm4eic::ReconstructedParticleCollection
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.h
```diff
@@ -0,0 +1,51 @@
...
+    >,
+    algorithms::Output<
+      edm4eic::ReconstructedParticleCollection,
+      edm4eic::ReconstructedParticleCollection
```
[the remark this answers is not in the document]
sebouh137: I'd have to think about this.  The way this works right now is that we have a collection of neutrons and photons (which I have now merged) in the lab frame, and the output is a collection of lambdas in the lab frame, and a collection of lambda decay products in the lambda CM frame.  Is it possible to have the lambda reconstruction modify the particles in the input collection (ie, change their reference point to where the reconstructed lambda decayed, and the momentum direction to point to this point)?  Still, I would like some way to store the direction of the neutron in the lambda's CM frame, since that's important for polarization measurements ... not sure the best way to do this.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+
+    }
+
+
+
+    void FarForwardNeutralsReconstruction::process(const FarForwardNeutralsReconstruction::Input& input,
+                      const FarForwardNeutralsReconstruction::Output& output) const {
+      const auto [clustersHcal] = input;
+      auto [out_neutrons, out_gammas] = output;
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+          rec_part.setMomentum(momentum);
+          rec_part.setReferencePoint(n_position);
+          rec_part.setCharge(0);
+          rec_part.setMass(m_neutron);
+          for (const auto& cluster : *clustersHcal){
+            rec_part.addToClusters(cluster);
+          }
+          /*for (const auto& cluster : *clustersEcal){
+            rec_part.addToClusters(cluster);
+            }*/
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,133 @@
...
+          }
+      }
+
+      double Etot=Etot_hcal;
+      if (Etot > 0 && Emax > 0){
+          auto rec_part = out_neutrons->create();
+          double corr=calc_corr(Etot,m_cfg.n_scale_corr_coeff_hcal);
+          Etot_hcal=Etot_hcal/(1+corr);
+          //corr=calc_corr(Etot,m_cfg.scale_corr_coeff_ecal);
+          //Etot_ecal=Etot_ecal/(1+corr);
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
sebouh137: removed the old neutron reconstruction, as requested

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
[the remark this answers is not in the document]
sebouh137: done

src/factories/reco/FarForwardNeutralsReconstruction_factory.h
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+      toTVector3(x2,(*gammas)[1].getReferencePoint()*dd4hep::mm);
+
+      xn.RotateY(-m_cfg.rot_y);
+      x1.RotateY(-m_cfg.rot_y);
+      x2.RotateY(-m_cfg.rot_y);
+
+
+      /*std::cout << "nx recon " << xn.X() << " g1x recon " << x1.X() << " g2x recon " << x2.X() << std::endl;
+
+        std::cout << "nz recon " << xn.Z() << " g1z recon " << x1.Z() << " g2z recon " << x2.Z() << " z face=" << m_cfg.zmax<< std::endl;*/
```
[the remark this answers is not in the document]
sebouh137: done

src/factories/calorimetry/HEXPLIT_factory.h
sebouh137: done, thanks

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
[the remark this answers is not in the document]
sebouh137: done.

src/factories/reco/FarForwardLambdaReconstruction_factory.h
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+
+
+      TVector3 vtx(0,0,0);
+      double f=0;
+      double df=0.5;
+      double theta_open_expected=2*asin(m_pi0/(2*sqrt(E1*E2)));
```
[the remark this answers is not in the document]
sebouh137: I've changed this so that the local variable m_pi0 is set using the particle service.  Thanks

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,148 @@
...
+          gammas.push_back(part);
+        }
+      }
+
+
+      if (neutrons.size()!=1 || gammas.size()!=2)
+        return;
+      double m_neutron=m_particleSvc.particle(2112).mass;
+      double m_pi0=m_particleSvc.particle(111).mass;
+      double m_lambda=m_particleSvc.particle(3122).mass;
```
veprbl, line 52:
```suggestion
      static const double m_neutron = m_particleSvc.particle(2112).mass;
      static const double m_pi0 = m_particleSvc.particle(111).mass;
      static const double m_lambda = m_particleSvc.particle(3122).mass;
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -0,0 +1,131 @@
...
+
+          Etot_hcal += E;
+          if (E > Emax){
+            Emax = E;
+            n_position = cluster.getPosition();
+          }
+      }
+
+      double Etot=Etot_hcal;
+      double m_neutron=m_particleSvc.particle(2112).mass;
```
veprbl, line 105:
```suggestion
      static const double m_neutron=m_particleSvc.particle(2112).mass;
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+
+    void FarForwardLambdaReconstruction::process(const FarForwardLambdaReconstruction::Input& input,
+                      const FarForwardLambdaReconstruction::Output& output) const {
+      const auto [neutrons, gammas] = input;
+      auto [out_lambdas, out_decay_products] = output;
+
+      if (neutrons->size()!=1 || gammas->size()!=2)
+        return;
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+        else if (theta_open<theta_open_expected)
+          f+=df;
+
+        vtx=lambda.Vect()*(f*m_cfg.zmax/lambda.Z());
+        df/=2;
+
+        if (i==m_cfg.iterations-1){
+          double mass_rec=lambda.M();
+          if (abs(mass_rec-m_lambda)>m_cfg.lambda_max_mass_dev)
+            return;
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+        TLorentzVector lambda=n+g1+g2;
+        if (theta_open>theta_open_expected)
+          f-=df;
+        else if (theta_open<theta_open_expected)
+          f+=df;
+
+        vtx=lambda.Vect()*(f*m_cfg.zmax/lambda.Z());
+        df/=2;
+
+        if (i==m_cfg.iterations-1){
```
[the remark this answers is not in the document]
sebouh137: At the moment, I am not sure of any.  As this is a binary search algorithm, the minimum step size on the vertex position is proportional to 2^(-# of iterations), so we could in principal finetune this number by determining the minimum number of iterations before we reach diminishing returns on the vertex position resolution.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+          double mass_rec=lambda.M();
+          if (abs(mass_rec-m_lambda)>m_cfg.lambda_max_mass_dev)
+            return;
+
+          // rotate everything back to the lab coordinates.
+          vtx.RotateY(m_cfg.rot_y);
+          lambda.RotateY(m_cfg.rot_y);
+          n.RotateY(m_cfg.rot_y);
+          g1.RotateY(m_cfg.rot_y);
+          g2.RotateY(m_cfg.rot_y);
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+
+          rec_part = out_decay_products->create();
+          rec_part.setPDG(22);
+          rec_part.setEnergy(g2.E());
+          rec_part.setMomentum({static_cast<float>(g2.X()), static_cast<float>(g2.Y()), static_cast<float>(g2.Z())});
+          rec_part.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()), static_cast<float>(vtx.Z())});
+          rec_part.setCharge(0);
+          rec_part.setMass(0);
+
+          return;
```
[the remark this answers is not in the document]
sebouh137: That's a good suggestion, but I think this could be done in a future pull request.

src/algorithms/reco/FarForwardLambdaReconstruction.h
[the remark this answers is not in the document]
ruse-traveler: Ahhh, I see. Unfortunately, you won't be able to modify the input collections since collections are immutable. But I think there's a couple of ways we could go about this!
- On one hand, as long as the production vertex is saved, the boosting can always be done after reconstruction in analysis since you'll have the decay products identified via the relations. I think this might be the simplest approach here, since you have immediate access to the decay products via relations...
- Another approach could be to add a new algorithm that boosts a collection of particles into the CM frame of a specified resonance. This is similar to the approach we took for the [Breit frame transformation](https://github.com/eic/EICrecon/blob/c24bc15fa7e9f315db9d0f002460f70219fc1182/src/algorithms/reco/TransformBreitFrame.h). This might get a little tricky for analysis, though, since you only need a subset of the reconstructed neutrals in the ZDC...

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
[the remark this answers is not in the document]
ruse-traveler: Ahhhh, I see! But if the second largest of the 3 is less then `gamma_max_width`, wouldn't the smallest also be less than `gamma_max_width` and `(l1 > max_width) + (l2 > max_width) + (l3 > max_width)` evaluate to 1?

src/algorithms/reco/FarForwardLambdaReconstruction.cc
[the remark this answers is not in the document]
ruse-traveler: Gotcha! Makes sense! I think for now it's fine as-is, but that might be something to explore down the road! Maybe it could cut out after the step size dips beneath the position resolution or something to that effect...

src/algorithms/reco/FarForwardLambdaReconstruction.cc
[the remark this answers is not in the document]
ruse-traveler: True! For the time being, it can always be sorted out in analysis!

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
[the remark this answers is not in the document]
sebouh137: Correct.  Fixed this in the code.

src/algorithms/reco/FarForwardLambdaReconstruction.h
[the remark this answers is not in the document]
sebouh137: The kinematics of the decay products are not the same as the ones from the neutrals reconstruction, since the vertex position is no longer assumed to be at the origin (so the direction of their momenta changes when doing this).  

I think I have found a work-around here.  I am storing the CM-frame decay products in the same collection as the lambdas, and then adding these particles to the list of particles associated with the lambda.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+  /* converts one type of vector format to another */
+  void toTVector3(TVector3& v1, const edm4hep::Vector3f& v2){
+    v1.SetXYZ(v2.x,v2.y,v2.z);
+  }
+
+
+    void FarForwardLambdaReconstruction::process(const FarForwardLambdaReconstruction::Input& input,
+                      const FarForwardLambdaReconstruction::Output& output) const {
+      const auto [neutrons, gammas] = input;
+      auto [out_lambdas, out_decay_products] = output;
```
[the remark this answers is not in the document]
sebouh137: done

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,139 @@
...
+          rec_part.setCharge(0);
+          rec_part.setMass(0);
+
+          rec_part = out_decay_products->create();
+          rec_part.setPDG(22);
+          rec_part.setEnergy(g2.E());
+          rec_part.setMomentum({static_cast<float>(g2.X()), static_cast<float>(g2.Y()), static_cast<float>(g2.Z())});
+          rec_part.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()), static_cast<float>(vtx.Z())});
+          rec_part.setCharge(0);
+          rec_part.setMass(0);
```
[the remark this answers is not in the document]
sebouh137: done

sebouh137:
@wdconinc @veprbl @ruse-traveler , thanks for the feedback.  If you are satisfied, I would like to get this pull request finished so that I can follow through with a pull request on detector_benchmarks which is dependent on these changes. 
 https://github.com/eic/detector_benchmarks/pull/138
Thanks,
Sebouh

src/algorithms/reco/FarForwardLambdaReconstruction.h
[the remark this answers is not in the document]
ruse-traveler: Please don't do that: that runs counter to all other reconstructed particle collections, and will make things confusing by mixing different levels of information in a single branch (the resonance vs. the decay products).

If you have the production vertex, the original momenta (wrt. the origin) of the decay products, and the lambda momentum, wouldn't it be straightforward enough to do downstream? Either in an analysis script or in subsequent algorithm?

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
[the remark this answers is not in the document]
ruse-traveler: Ah! I see my confusion: this needs to be NOT true in order for a cluster to be labeled a photon! I'm okay with this as-is in this case.

src/algorithms/pid_lut/PIDLookup.cc
```diff
@@ -5,6 +5,7 @@
 #include <edm4eic/MCRecoParticleAssociationCollection.h>
 #include <edm4eic/ReconstructedParticleCollection.h>
 #include <edm4hep/MCParticleCollection.h>
+#include <edm4hep/Vector3f.h>
```
ruse-traveler, line 8: Hmmm... This might need another rebase...

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,158 @@
...
+      std::vector<edm4eic::ReconstructedParticle> gammas={};
+      for (auto part: *neutrals){
+        if (part.getPDG()==2112){
+          neutrons.push_back(part);
+        }
+        if (part.getPDG()==22){
+          gammas.push_back(part);
+        }
+      }
+
```
ruse-traveler, line 47:
```suggestion
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,158 @@
...
+            for(int i = 0; i<m_cfg.iterations; i++){
+              n={pn*(xn-vtx).Unit(), En};
+              g1={E1*(x1-vtx).Unit(), E1};
+              g2={E2*(x2-vtx).Unit(), E2};
+              double theta_open=g1.Angle(g2.Vect());
+              lambda=n+g1+g2;
+              if (theta_open>theta_open_expected)
+                f-=df;
+              else if (theta_open<theta_open_expected)
+                f+=df;
```
ruse-traveler, line 93: Super minor suggestion (I get a little nervous when see multi-line conditionals without curly braces) 
```suggestion
              if (theta_open>theta_open_expected) {
                f-=df;
              } else if (theta_open<theta_open_expected) {
                f+=df;
              }
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,158 @@
...
+
+    void FarForwardLambdaReconstruction::init() {
+
+    }
+
+  /* converts one type of vector format to another */
+  void toTVector3(TVector3& v1, const edm4hep::Vector3f& v2){
+    v1.SetXYZ(v2.x,v2.y,v2.z);
+  }
+
```
ruse-traveler, line 31: For consistency...
```suggestion
```

src/algorithms/reco/FarForwardLambdaReconstruction.h
```diff
@@ -0,0 +1,48 @@
...
+                                                            {"outputLambdaAndDecayProducts"},
+                                        "Reconstructs lambda candidates and their decay products from the reconstructed neutrons and photons"} {}
+
+         void init() final;
+         void process(const Input&, const Output&) const final;
+
+    private:
+        std::shared_ptr<spdlog::logger> m_log;
+       const algorithms::ParticleSvc& m_particleSvc = algorithms::ParticleSvc::instance();
+        const dd4hep::Detector* m_detector{algorithms::GeoSvc::instance().detector()};
```
ruse-traveler, line 45: It looks like the geometry isn't used anywhere in this algorithm now? If that's the case then you can safely remove it and the associated headers!
```suggestion
```

ruse-traveler (commented):
This is looking good! We still need to settle how to handle the decay products, but beyond that I only have a couple of super minor suggestions.

So once the decay product question is resolved, I think this will be ready to merge!

src/algorithms/reco/FarForwardLambdaReconstruction.h
sebouh137: Correct, it's being accessed in the ZDC.cc, so removing this line makes sense

src/algorithms/pid_lut/PIDLookup.cc
sebouh137: Iwyu told me to make this change if I remember correctly

ruse-traveler: Ah! Gotcha! IWYU continues to surprise...

src/algorithms/reco/FarForwardLambdaReconstruction.h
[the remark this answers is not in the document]
ruse-traveler: So I've been mulling it over, and wanted to clarify my thoughts here:
- While I would prefer the boosting be moved to either a downstream algorithm or analysis code, I am willing to compromise on leaving it in this algorithm for now and make factorizing this part of the algorithm a to-do
- If we do leave in this algorithm, though, I think it would be best to write out the lambdas and boosted decay products in separate collections, since
  - This will make it easier to factorize later,
  - And makes sure we aren't mixing levels of information in each collection!
- The one-to-many relations require some care, though...
  - So far, relations almost always point _backwards_ to inputs (e.g. the cluster-hit relations) or _horizontally_ to objects grouped together (e.g. the `PseudoCluster`)
  - This raises the question of how the boosted particles should relate to lambdas and the input particles... My thinking so far is:
    1. The lambda one-to-many relations point backwards to the (unboosted) input particles used to reconstruct the lambda,
    2. And the boosted particle one-to-many relations point backwards to the original, unboosted particle that got boosted (so "many" = 1 here)...
  - The rationale for this scheme is to preserve the algorithmic flow being done here...

What are your thoughts, @veprbl and @wdconinc? How do the FCC folks typically handle situations like this?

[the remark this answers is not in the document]
sebouh137: @ruse-traveler ,  If we do not hear back from @wdconinc and @veprbl by tomorrow morning, I will do exactly as you have suggested, and then we can get this pull request accepted.  I'm in a bit of a time crunch to get this pull request through.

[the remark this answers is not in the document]
veprbl: The advice is probably not to mix in the same collection reconstructed particles and reconstructed daughters for ease of analysis, and to not mix particles with momenta given in different reference frames to avoid confusion. It's fine to have particles point to each other to preserve those relationships.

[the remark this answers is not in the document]
sebouh137: ok done.  I have also made connections from the CM particles to the lambda candidates as well.

ruse-traveler (dismissed):
This looks great! My final suggestion is to not put a relation between the lambda and the boosted decay products (I provide the rationale in the comments), but other than that I think this is ready to ship. Thank you for all of the hard work on this!

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,166 @@
...
+            neutron_cm.setEnergy(n.E());
+            neutron_cm.setMomentum({static_cast<float>(n.X()), static_cast<float>(n.Y()), static_cast<float>(n.Z())});
+            neutron_cm.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()), static_cast<float>(vtx.Z())});
+            neutron_cm.setCharge(0);
+            neutron_cm.setMass(m_neutron);
+            //link the reconstructed lambda to the input neutron,
+            // the cm neutron to the reconstructed lambda,
+            // and the cm neutron to the input neutron
+            rec_lambda.addToParticles(neutrons[i_n]);
+            neutron_cm.addToParticles(rec_lambda);
```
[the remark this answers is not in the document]
sebouh137: ok thanks!

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -0,0 +1,166 @@
...
+
+            auto gamma2_cm = out_decay_products->create();
+            gamma2_cm.setPDG(22);
+            gamma2_cm.setEnergy(g2.E());
+            gamma2_cm.setMomentum({static_cast<float>(g2.X()), static_cast<float>(g2.Y()), static_cast<float>(g2.Z())});
+            gamma2_cm.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()), static_cast<float>(vtx.Z())});
+            gamma2_cm.setCharge(0);
+            gamma2_cm.setMass(0);
+            rec_lambda.addToParticles(gammas[i_2]);
+            gamma2_cm.addToParticles(rec_lambda);
```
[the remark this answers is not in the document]
sebouh137: done

sebouh137:
Hi @ruse-traveler , 
I have made the changes you have requested; now I just need you to redo the approve thing, and then it will be merged.  Next we can go through the related change in detector_benchmarks, https://github.com/eic/detector_benchmarks/pull/138 , which I want to push by the end of the week.

sebouh137:
Dear @wdconinc,
all of the convos between me and you, @ruse-traveler and @veprbl  have been resolved.  @ruse-traveler has given his approval above, with the condition that I made a few changes which I have now made.  The only check that is failing is the eicweb/eic-container (which appears to have nothing to do with my pull request).  Please accept this pull request as soon as possible.  
Thanks, 
Sebouh

sebouh137:
> Dear @wdconinc, all of the convos between me and you, @ruse-traveler and @veprbl have been resolved. @ruse-traveler has given his approval above, with the condition that I made a few changes which I have now made. The only check that is failing is the eicweb/eic-container (which appears to have nothing to do with my pull request). Please accept this pull request as soon as possible. Thanks, Sebouh

@wdconinc I will need you to click "approve" so that this will go into the merge queue and then the change will propagate over to the checks in the detector_benchmarks project.  Thanks

ruse-traveler (approved):

wdconinc:
> @wdconinc I will need you to click "approve" so that this will go into the merge queue and then the change will propagate over to the checks in the detector_benchmarks project.  Thanks

I don't have any special powers and I haven't kept up with his PR like @ruse-traveler and @veprbl have, so if you really want me to review it from scratch instead of deferring to them it will take longer... Both @ruse-traveler and @veprbl are in the 'line of command' on these topics, and I'm not (would be different if this was an ops topic).

Issue #1705, opened 2025-01-15
Retire support for compiling in old eic containers (February 2025)


veprbl:
We currently support unreasonable amount of backwards compatibility
https://github.com/eic/EICrecon/blob/d96cbff18ba2b6e258791ffd6daf94d4b0562ba8/.github/workflows/linux-eic-shell.yml#L89-L95

Suggestion is to look into updating those versions and removing some of the `#if`s that are not needed after that.

wdconinc:
6 months?

veprbl:
#### Software versions in the containers

|           Tag|    acts|  dd4hep| edm4eic| edm4hep|   jana2|   podio|
|-|-|-|-|-|-|-|
|24.04.0-stable|  31.2.0|    1.28|   5.0.0|  0.10.3|2.1.2|  0.17.3|
|24.05.2-stable|  31.2.0|    1.28|   5.0.0|  0.10.3|2.1.2|  0.17.3|
|24.06.0-stable|  31.2.0|    1.28|   6.0.1|  0.10.3|2.2.1-rc1|    0.99|
|24.07.0-stable|  31.2.0|    1.29|   6.0.1|  0.10.5|2.2.1-rc1|    0.99|
|24.08.1-stable|  31.2.0|    1.29|   7.0.0|  0.10.5|   2.3.1|    0.99|
|24.09.0-stable|  33.1.0|    1.29|   7.0.0|  0.10.5|   2.3.1|    0.99|
|24.10.1-stable|  33.1.0|    1.30|   7.0.0|  0.10.5|   2.3.2|    0.99|
|24.11.2-stable|  33.1.0|    1.30|   8.0.0|  0.10.5|   2.3.2|   1.0.1|
|24.12.0-stable|  33.1.0|    1.30|   8.0.0|  0.10.5|   2.3.2|     1.1|
|25.01.1-stable|  34.1.0|    1.30|   8.0.0|  0.10.5|   2.3.2|     1.1|
|25.02.0-stable|  35.2.0|    1.30|   8.0.0|  0.10.5|   2.4.0|     1.1|

#### Work plan

 - [ ] Drop old containers from the matrix, add new to the matrix
 - [ ] Drop support for ACTS 31 and older
 - [ ] Drop support for DD4hep 1.28 and older
 - [ ] Assume edm4eic 7+
 - [ ] Drop support for PODIO 0.17.3

#### Code
<details>
  <summary>versions.sh</summary>
<pre>
[...]

PR #1735, opened 2025-02-12
linux-eic-shell.yml: update list of supported releases

veprbl:

As discussed in #1705, this is dropping a check for whether EICrecon compiles on `24.04.0-stable`. The earliest checked version will now be `24.09.0-stable`.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc (commented):
One side effect of more test platforms is that we are more likely to be unable to take advantage of ccache. I think we can't run this many platforms without evicting the first ccache by the time we write the last platforms's ccache ...

veprbl:
@wdconinc I've limited to two select releases. I don't think there is a way to automate if we have to do such cherry-picking.

wdconinc (approved):

Issue #1730, opened 2025-02-05
CentralTrackSegments have wrong associations with Track-s


DraTeots:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released):
- Which version (or `HEAD` for the most recent on git):
- Any specific OS or system where the issue occurs?

I see it in any reconstructed edm4eic files from the latest campaign available. E.g. 

```bash
 xrdcp "root://dtn-eic.jlab.org//work/eic2/EPIC/RECO/24.12.0/epic_craterlake/DIS/NC/18x275/minQ2=10/pythia8NCDIS_18x275_minQ2=10_beamEffects_xAngle=-0.025_hiDiv_5.1748.eicrecon.tree.edm4eic.root" ./
```

### The problem:

According to EDM4EIC model, edm4eic::TrackSegment has a link to corresponding edm4eic::Track information (and it could be then used to get get track parameters, etc.) 

```yaml
  edm4eic::TrackSegment:
    Description: "A track segment defined by one or more points along a track."
    Author: "S. Joosten"
    Members:
      - float             length            
      - float             lengthError    
    OneToOneRelations:
      - edm4eic::Track    track             // <==  Track used for this projection
    VectorMembers:
      - edm4eic::TrackPoint points   
```

On low level, besides "CentralTrackSegments" the root tree has "_CentralTrackSegments_track" collection which has track collection id and index for each track corresponding to a segment. The promel is that for  CentralTrackSegments they are not set: 

```
branch_name=CentralTrackSegments
track_objid_index -2
track_objid_index -2
track_objid_index -2
track_objid_index -2
track_objid_index -2
track_objid_coll 0
track_objid_coll 0
track_objid_coll 0
track_objid_coll 0
track_objid_coll 0
```

I believe it is to be EicRecon problem as these values are set correctly for other TrackSegment collections such as CalorimeterTrackProjections or DRICHAerogelTracks. That is how it looks when its working: 

```
branch_name=DRICHAerogelTracks
track_objid_index 2
track_objid_index 6
track_objid_index 8
track_objid_coll 530999115
track_objid_coll 530999115
track_objid_coll 530999115
```

For CalorimeterTrackProjections

```
branch_name=CalorimeterTrackProjections
track_objid_index 0
track_objid_index 1
track_objid_index 2
track_objid_index 3
track_objid_index 4
track_objid_index 5
track_objid_index 6
track_objid_index 7
track_objid_index 8
[...]

PR #1736, opened 2025-02-13
TrackProjector: adopt algorithms interface

veprbl:
Implementing additional associations will require adding an additional input #1730. The algorithm has not yet been converted to comply with the algorithms interface, so we do that here.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


DraTeots (approved):
I don't have detailed understanding of the current EICrecon codebase at the moment. Still AFAIK this PR seems inline with moving everything towards algorithms API and nicely converts this piece of code. I also appreciate that this PR gets EICrecon closer for fixing #1730

PR #1739, opened 2025-02-19
TrackProjector: fill relation from segment to track

veprbl:
The `TrackProjector` records Acts TrackStates as `edm4eic::TrackSegment`s to be saved as "CentralTrackSegments" collection. The current implementation is lacking associations from those segments to the full tracks. This PR implements the same logic as in https://github.com/eic/EICrecon/blob/d072a3ea944b7e4f171b03237de95ba5176c24c5/src/algorithms/tracking/TrackPropagation.cc#L143 to provide this information.

Fixes: #1730

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes

### DIFF
--- a/src/algorithms/tracking/TrackProjector.cc
+++ b/src/algorithms/tracking/TrackProjector.cc
@@ -20,7 +20,7 @@
 #include <edm4hep/utils/vector_utils.h>
 #include <fmt/core.h>
 #include <fmt/ostream.h>
-#include <stdint.h>
+#include <cstdint>
 #include <cmath>
 #include <gsl/pointers>
 #include <iterator>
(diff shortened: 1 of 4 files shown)


src/algorithms/tracking/TrackProjector.cc
```diff
@@ -41,13 +42,13 @@ namespace eicrecon {
...
-        const auto [trajectories] = input;
+        const auto [acts_trajectories, tracks] = input;
         auto [track_segments] = output;
 
-        debug("Track projector event process. Num of input trajectories: {}", std::size(trajectories));
+        debug("Track projector event process. Num of input trajectories: {}", std::size(acts_trajectories));
 
         // Loop over the trajectories
-        for (const auto &traj: trajectories) {
+        for (size_t i = 0; const auto &traj: acts_trajectories) {
```
wdconinc, line 51: Avoid `stddef.h` C-header by using C++ type?
```suggestion
        for (std::size_t i = 0; const auto &traj: acts_trajectories) {
```

DraTeots (approved):

PR #1741, opened 2025-02-21
fix: also add tracks that failed smoothing to failed_tracks

wdconinc:
Since #1677 we remove tracks that failed extrapolation. We don't do that with tracks that failed smoothing. This PR also applies the same treatment to tracks that failed smoothing.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -291,6 +291,7 @@ namespace eicrecon {
                     ACTS_ERROR("Smoothing for seed "
                         << iseed << " and track " << track.index()
                         << " failed with error " << smoothingResult.error());
+                    failed_tracks.push_back(track.index());
                     continue;
                 }
 


veprbl:
I don't think we should be "hiding" tracks that failed smothing without a clear reason. If the tracks constituted by background, or have unreasonable parameters estimated, then we could reconsider. Would you agree?

veprbl:
> I thought that ship had sailed when we decided to hide tracks that fail extrapolation? How is smoothing different?

Tracks that fail extrapolation do not have a usable set of parameters, so we can't store them.

wdconinc:
> > I thought that ship had sailed when we decided to hide tracks that fail extrapolation? How is smoothing different?
> 
> Tracks that fail extrapolation do not have a usable set of parameters, so we can't store them.

But we don't even attempt extrapolation when smoothing fails. If we want to allow tracks that failed smoothing, then we should still attempt extrapolation.

veprbl:
> > > I thought that ship had sailed when we decided to hide tracks that fail extrapolation? How is smoothing different?
> > 
> > 
> > Tracks that fail extrapolation do not have a usable set of parameters, so we can't store them.
> 
> But we don't even attempt extrapolation when smoothing fails. If we want to allow tracks that failed smoothing, then we should still attempt extrapolation.

Right, so in the current setup, smoothing has never ever failed, because if it would, that'd cause a segfault. If we apply your suggested change, if will be allowed to fail without interrupting anything.

veprbl:
Should we pivot towards making this into a "fatal" exception?

veprbl:
Fatal as in this was never expected to happen and user is forced to report/confront this.

veprbl:
I misunderstood your comment in https://github.com/eic/EICrecon/pull/1715#discussion_r1966172195 to mean that this does not fix the issue because smoothing never fails. Now I can reproduce that smoothing does fail, so a correct workaround is, indeed, to add to `failed_tracks`, like you are suggesting here. Removing the `continue;` might also work as an alternative.

veprbl (approved):

PR #1742, opened 2025-02-24
feat: output InitTrackParams (truth seeding results)

wdconinc:
Apparently we don't write out InitTrackParams by default. Probably wouldn't be bad to give it a better name too... but that's backwards incompatible.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl:
Can we give it a better name before it becomes public?

wdconinc:
> Can we give it a better name before it becomes public?

`CentralTrackTruthSeedingResults` would be the natural candidate but: it's neither seeding nor central tracking.

I'm open to suggestions.

veprbl (approved):

PR #1744, opened 2025-02-26
feat: split B0 from central detector tracking

wdconinc:
There is no way a relevant track can have both hits in the forward endcaps and the B0 tracker. We can therefore separate the tracking of both, and by doing so gain in the combinatoric search space and allow for parallelization.

This PR splits the B0 tracker seeding and tracking from the central detector seeding and tracking. After separate tracking, the tracks are merged again in the `ReconstructedChargedWithoutPIDParticles` collection.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


src/global/tracking/tracking.cc
```diff
@@ -275,50 +277,208 @@ void InitPlugin(JApplication *app) {
...
+    std::vector<std::string> input_track_collections, input_track_assoc_collections;
     //Check size of input_rec_collections to determine if CentralCKFTracks should be added to the input_track_collections
     if (input_rec_collections.size() > 0) {
         input_track_collections.push_back("CentralCKFTracks");
+        input_track_assoc_collections.push_back("CentralCKFTrackAssociations");
+    }
+    //Check if the B0Tracker readout is present in the current configuration
+    if (readouts.find("B0TrackerHits") != readouts.end()) {
+        input_track_collections.push_back("B0TrackerCKFTracks");
+        input_track_assoc_collections.push_back("B0TrackerCKFTrackAssociations");
```
veprbl, line 433: Since seeding doesn't work for B0 yet, people are fishing in 
`*TruthSeeded*` collection to get their B0 tracks.  We need to mix `B0TrackerCKFTruthSeededTracks` into `ReconstructedTruthSeededChargedWithoutPIDParticles` to not break current analyses.

simonge:
Ideally would this happen after the PID or even cluster matching as none of the tracks should be able to match the central PID or calorimeter detectors?

One thing this change could adversely effect is the important contribution of the B0 to vertexing.

src/global/tracking/tracking.cc
wdconinc: But that wasn't what was happening, right?

wdconinc:
> One thing this change could adversely effect is the important contribution of the B0 to vertexing.

The B0 tracks are recombined with central tracks before we do vertexing.

src/global/tracking/tracking.cc
veprbl: Before this, B0 hits were mixed to both realistic and truth seeding. The proposed code only mixes in real seeded tracks (far backward doesn't ever use truth seeding).

wdconinc: Ok, I see what you mean.

wdconinc: 20c61be317d1c900a20af9ed45343ff077a73c54

simonge:
> > One thing this change could adversely effect is the important contribution of the B0 to vertexing.
> 
> The B0 tracks are recombined with central tracks before we do vertexing.

Do you not need to combine the Central and B0 CKFActsTrajectories when passed into the iterative vertex finder?

https://github.com/eic/EICrecon/pull/1744/files#diff-e093c4fb16d0a0c3255aa88be403f9ea77ce7ac00a8120f47b34c8de55b30bd9L232-L238

wdconinc:
> Do you not need to combine the Central and B0 CKFActsTrajectories when passed into the iterative vertex finder?

Yeah, and that's hard. Showstopper for now.

simonge:
> > Do you not need to combine the Central and B0 CKFActsTrajectories when passed into the iterative vertex finder?
> 
> Yeah, and that's hard. Showstopper for now.

Is creating a collection which adds both vectors of trajectories not as simple with Acts internal types? Would it instead be possible to make the iterative vertex finder input variadic and pass both collections.

wdconinc:
(This is still not ready for review)

wdconinc:
The number of tracks now found in the B0 chain is consistent with the decrease in the Central chain.
Screenshot_20250415-203431 Screenshot_20250415-203447

wdconinc:
Merged main into branch to resolve conflicts (rebase was too painful after #1843).

wdconinc:
Latest four commits are cherry picks from #1824.

src/global/tracking/tracking.cc
```diff
@@ -166,9 +186,27 @@ void InitPlugin(JApplication* app) {
                                                               },
                                                               app));
 
+  app->Add(new JOmniFactoryGeneratorT<ActsTrajectoriesMerger_factory>(
+      "CentralB0CKFActsTrajectories",
+      {
+          "CentralCKFActsTrajectories",
+          "B0TrackerCKFActsTrajectories",
```
ShujieL, line 193: better move this down to after where B0TrackerCKFActsTrajectories is defined.

src/global/tracking/tracking.cc
```diff
@@ -166,9 +186,27 @@ void InitPlugin(JApplication* app) {
...
+          "CentralAndB0TrackerCKFActsTrajectories",
+      },
+      app));
+
   app->Add(new JOmniFactoryGeneratorT<IterativeVertexFinder_factory>(
-      "CentralTrackVertices", {"CentralCKFActsTrajectories", "ReconstructedChargedParticles"},
-      {"CentralTrackVertices"}, {}, app));
+      "CentralTrackVertices",
+      {
+          "CentralAndB0TrackerCKFActsTrajectories",
```
ShujieL, line 203: same here, can you move this block down, and rename the output from "CentralTrackVertices" to "CentralAndB0TrackVertices"?

wdconinc: The point here is that you can't constrain the vertices well without the B0.

src/global/tracking/tracking.cc
veprbl: I agree, pushed a commit to implement this.

src/global/tracking/tracking.cc
veprbl: (for posterity) The current state of `CentralTrackVertices` is 100% agreement with the original reference in epic-capybara.

PR #1745, opened 2025-02-27
Move PID look up tables into global pid_lut file

simonge:
Moves all of the PID LUT factories out of their detector plugin into the global pid plugin. Also moves PID factory files into the factories directory.

Maybe a matter of personal taste but this would make the data flow through the PID algorithms more transparent.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: Refactoring

No
No


src/global/pid/pid.cc
```diff
@@ -7,8 +7,11 @@
```
veprbl, line 7: Moving everything to the same file clarifies data flow for LUT PID. Actual implementations for detectors can not be as uniform and probably not going to co-exist as well. That was my original reasoning.

veprbl (commented):
This should be fine to do.

src/global/pid/pid.cc
simonge: Is there ever going to be a situation where the PID from the LUT from some detectors are mixed with real PID from others?
I don't know if it's always possible but it certainly feels nicer not to jump back and forth between plugins.

veprbl: I'd say, yes, as soon as any PID reconstruction is production ready (works on full DIS events), we then don't keep its LUT around.

wdconinc: What decade do you estimate that happening in? ;-)

wdconinc:
What's the plan for this PR? Since the conflicts are likely due to the clang-format or clang-tidy PR, I can offer to resolve them, but only if there's a plan to merge this.

simonge:
It's not a lot of work so I can just create a new clean branch and PR to avoid the mess. Unless you have a better idea.

simonge:
Thanks, hopefully that's much better

src/global/pid_lut/pid_lut.cc
```diff
@@ -0,0 +1,247 @@
...
+  int ForwardRICH_ID = 0;
+  try {
+    auto detector  = app->GetService<DD4hep_service>()->detector();
+    ForwardRICH_ID = detector->constant<int>("ForwardRICH_ID");
+  } catch (const std::runtime_error&) {
+    // Nothing
+  }
+  PIDLookupConfig drich_pid_cfg{
+      .filename                 = "calibrations/drich.lut",
+      .system                   = ForwardRICH_ID,
```
wdconinc, line 205: (OT) Not for this PR, but at some point I would like to remove these constructs (which require plugins to be loaded in a certain order that we are unable to enforce) and move it into algorithm initialization:
```suggestion
  PIDLookupConfig drich_pid_cfg{
      .filename                 = "calibrations/drich.lut",
      .system                   = "ForwardRICH_ID",
```
Doesn't work in all cases, but it does work in several of them.

wdconinc (dismissed):
Looks good to me. Waiting for diffs.

simonge:
Should I ignore iwyu asking for both:
```
#include <JANA/JApplicationFwd.h>
```

I'll wait for the diffs before drawing conclusions about the tests and benchmarks not using the right libraries

wdconinc:
> I'll wait for the diffs before drawing conclusions about the tests and benchmarks not using the right libraries

Diff looks good.

wdconinc (approved):

src/global/pid_lut/pid_lut.cc
simonge: I'll try and do this next.

PR #1746, opened 2025-02-27
More deprecation-related fixes

nathanwbrei:
- Fixes deprecation warning inside JANA2
- Fixes an include file which was coming from JANA2
- Fixes a warning relating to using abs() instead of fabs() inside a test case

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no

no

### DIFF
--- a/src/tests/algorithms_test/calorimetry_HEXPLIT.cc
+++ b/src/tests/algorithms_test/calorimetry_HEXPLIT.cc
@@ -88,7 +88,7 @@ TEST_CASE( "the subcell-splitting algorithm runs", "[HEXPLIT]" ) {
     Esum+=subcell.getEnergy();
     i++;
     if (i%12==0){
-      REQUIRE(abs(Esum-E[i/12-1])/E[i/12-1]<tol);
+      REQUIRE(fabs(Esum-E[i/12-1])/E[i/12-1]<tol);
       Esum=0;
     }
   }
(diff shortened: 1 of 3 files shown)


src/utilities/eicrecon/eicrecon_cli.cpp
```diff
@@ -320,8 +320,9 @@ void PrintPodioCollections(JApplication* app) {
       for (auto* event_source : event_sources) {
         //                    std::cout << event_source->GetPluginName() << std::endl;  // podio.so
         //                    std::cout << event_source->GetResourceName() << std::endl;
-        if (event_source->GetPluginName().find("podio") != std::string::npos)
-          event_source->DoInitialize();
+        if (event_source->GetPluginName().find("podio") != std::string::npos) {
+          event_source->DoOpen();
+        }
```
veprbl, line 325: Will this work in 2.3.1?
If we merge https://github.com/eic/EICrecon/pull/1735 that would be the earliest tested, according to https://github.com/eic/EICrecon/issues/1705#issuecomment-2655044372

src/tests/algorithms_test/calorimetry_HEXPLIT.cc
```diff
@@ -88,7 +88,7 @@ TEST_CASE( "the subcell-splitting algorithm runs", "[HEXPLIT]" ) {
     Esum+=subcell.getEnergy();
     i++;
     if (i%12==0){
-      REQUIRE(abs(Esum-E[i/12-1])/E[i/12-1]<tol);
+      REQUIRE(fabs(Esum-E[i/12-1])/E[i/12-1]<tol);
```
veprbl, line 91:
```suggestion
      REQUIRE(std::abs(Esum-E[i/12-1])/E[i/12-1]<tol);
```
We can use the c++ function

veprbl (commented):
LGTM, but may need another PR approved first

src/tests/algorithms_test/calorimetry_HEXPLIT.cc
wdconinc: can -> should

veprbl (approved):
This is good

PR #1756, opened 2025-03-11
treewide: drop support for PODIO 0.17.x and older

veprbl:
Removes some outdated conditionals.

- [x] Bug fix (issue #1705)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


ruse-traveler (commented):
Looks good! But I am seeing an odd regression in `TaggerTrackerTracks.time`... Is that to be expected?

veprbl:
image

The values in the green histogram can't be right either. But, yeah, it just can't be because of this. cc @simonge

simonge:
Most likely culprit is not initializing `time=0`
https://github.com/eic/EICrecon/blob/main/src/algorithms/fardetectors/FarDetectorMLReconstruction.cc#L105

Not sure why/how it would have gone wrong in the reference.

This was planned to be implemented with a real time down the line but for now I'll try and fix it as it is.

ruse-traveler (approved):
LGTM!

PR #1762, opened 2025-03-19
Combine pulses in the same CellID

simonge:
Follow on from https://github.com/eic/EICrecon/pull/1758

Allows a linear sum of pulses across time for each cell within a certain period of each other.

Example plots come from combining the first 5 hits in the whole of the Low-Q2 tagger into one channel.

**Combined pulses**

PulseCombined

**Pulses which were combined into one**

Pulses

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Adds TaggerTrackerCombinedPulses and TaggerTrackerCombinedPulsesWithNoise to the output.


src/algorithms/digi/PulseCombiner.cc
```diff
@@ -0,0 +1,153 @@
...
+    int endStep = startStep + pulseSize;
+    for(int i = 0; i < newPulse.size(); i++) {
+      // Add pulse values to new pulse
+      float contribution = 0;
+      if(i >= startStep && i < endStep){
+        contribution = pulse.getAmplitude()[i - startStep];
+      } else { // Interpolate first and last two values to extrapolate over 0
+        if(i < startStep) {
+          contribution = pulse.getAmplitude()[0] + (pulse.getAmplitude()[1] - pulse.getAmplitude()[0])*(i - startStep);
+          if(std::signbit(contribution)!=std::signbit(pulse.getAmplitude()[0])) continue;
```
veprbl, line 139: This looks a bit dangerous. If the numerical derivative comes in at a wrong sign, this will be never triggered and waveform will be filled with junk. What is the purpose of this interpolation?

src/algorithms/digi/PulseCombiner.cc
```diff
@@ -0,0 +1,153 @@
...
+        cluster_pulses.back().push_back(pulse);
+        clusterEndTime = std::max(clusterEndTime, pulseEndTime);
+      } else {
+        isNewPulse = true;
+      }
+    }
+    if(isNewPulse) {
+      cluster_pulses.push_back({pulse});
+      clusterEndTime = pulseEndTime;
+      isNewPulse = false;
```
veprbl, line 105:
```suggestion
  bool makeNewPulse = true;
  // Create clusters of pulse indices which overlap with at least the minimum separation
  for (auto pulse: ordered_pulses) {
    float pulseStartTime = pulse.getTime();
    float pulseEndTime = pulse.getTime() + pulse.getInterval()*pulse.getAmplitude().size();
    if(!makeNewPulse) {
      if (pulseStartTime < clusterEndTime + m_minimum_separation) {
        cluster_pulses.back().push_back(pulse);
        clusterEndTime = std::max(clusterEndTime, pulseEndTime);
      } else {
        makeNewPulse = true;
      }
    }
    if(makeNewPulse) {
      cluster_pulses.push_back({pulse});
      clusterEndTime = pulseEndTime;
      makeNewPulse = false;
```
Maybe it's just me, but this name helps to better understand the code when reading from top to bottom.

src/algorithms/digi/PulseCombiner.cc
simonge: If the derivative is wrong, this should mean the pulse is moving away from 0 so either the length of the original pulse should be longer to capture the length, or noise has already been included. 

If it's never triggered the values should be 0 from that particular pulse rather than junk until the last time step from all of the pulses being added.

It could be removed but obviously brings the simulation a little closer to reality, there won't be any sudden drops to 0 from the minimum contribution included from the pulse generator. I will add a configuration to turn it off for the moment anyway.

src/algorithms/digi/PulseCombiner.cc
```diff
@@ -0,0 +1,153 @@
...
+  m_minimum_separation = m_cfg.minimum_separation;
+
+  // Get the detector readout and set bit shift if set
+  if(!(m_cfg.readout.empty() && m_cfg.combine_field.empty())) {
+    try {
+      auto detector = algorithms::GeoSvc::instance().detector();
+      auto id_dec = detector->readout(m_cfg.readout).idSpec().decoder();
+      // Get the bitshift for the detector
+      int bitshift = id_dec->highestBit() - (*id_dec)[m_cfg.combine_field].offset() - (*id_dec)[m_cfg.combine_field].width();
+      m_detector_bitmask = m_detector_bitmask >> bitshift;
```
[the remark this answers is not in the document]
simonge: I've updated the mask construction to use the logic you suggested and checked it is still consistent with what was working before. (Now it doesn't include 1's where there aren't associated bit fields in the encoder.

src/algorithms/digi/PulseCombiner.cc
veprbl: Sounds like something that should be addressed in the waveform generator. It should have more padding for the pulse tails to be faithfully represented.

src/algorithms/digi/PulseCombiner.cc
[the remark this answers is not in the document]
veprbl: Thanks.

src/algorithms/digi/PulseCombiner.cc
simonge: Ok, I have removed the interpolation functionality

veprbl (dismissed):

veprbl (approved):

src/algorithms/digi/PulseCombiner.cc
veprbl: Great, thanks!

wdconinc:
This PR broke main since the TaggerTrackerCluster still expects TrackerHits but receives RawTrackerHits now. Due to the use of TaggerTrackerTracks in the algorithmic branch that results in ReconstructedChargedParticles, this results in nothing in those branches either.

A quick attempt at fixing this by simply modifying the types in FarDetectorTrackerCluster indicates that least the eDep and time fields need to be swapped for something that's available in RawTrackerHit.

I'm going to be reverting this merge tomorrow night so it doesn't break things all next week (all ci workflows are failing too). I will try a few more things to fix it before then. If it can be fixed, no need to revert.

Two other comments.
1. How was this not noticed during review? Are we running CI against the actual code or are we picking up the plugins in /opt/local again?
2. This is ultimately hidden by the absence of exception warnings.

wdconinc:
> A quick attempt at fixing this by simply modifying the types in FarDetectorTrackerCluster indicates that least the eDep and time fields need to be swapped for something that's available in RawTrackerHit.

This is turning out to cascade quite a bit into other changes, not a simple type change. Another quick fix might be to re-add TrackerHitReconstruction factories after the TaggerTrackerSplitHits factory.

simonge:
> > A quick attempt at fixing this by simply modifying the types in FarDetectorTrackerCluster indicates that least the eDep and time fields need to be swapped for something that's available in RawTrackerHit.
> 
> This is turning out to cascade quite a bit into other changes, not a simple type change. Another quick fix might be to re-add TrackerHitReconstruction factories after the TaggerTrackerSplitHits factory.

I don't think the Raw and Rec hits should have been changed at all in LOWQ2.cc, I'm not sure why/when it was added. This wasn't meant to effect the current workflow at all yet, I will have a look later unless you can fix it in the meantime.

PR #1763, opened 2025-03-21
Low-Q2 Charge Sharing

simonge:
Introduces sensor charge sharing for hits in the Low-Q2 Tagger dividing a single SimTrackerHit into several.

The method originally introduced for LGADs has been simplified and optimized to make it more accessible for future users, making more use of functions provided by dd4hep.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Adds the TaggerTrackerSharedHits to the output and uses them as the input to the digitization.


veprbl:
Need a rebase.

simonge:
There are some changes in the BTOF charge sharing I need to understand/fix and discus whether the changes are actually improvements or not.

ssedd1123:
> There are some changes in the BTOF charge sharing I need to understand/fix and discus whether the changes are actually improvements or not.

What's the problem you're facing, if you don't mind me asking? I think it is a nice improvement over the original LGADChargeSharing. I've always considered testing cellID with position converter and see if it throws error is a bit of a hack. I did not know m_seg.neighbours exist at all.

simonge:
> What's the problem you're facing, if you don't mind me asking? I think it is a nice improvement over the original LGADChargeSharing. I've always considered testing cellID with position converter and see if it throws error is a bit of a hack. I did not know m_seg.neighbours exist at all.

It's the differences in the capybara I don't quite understand:
https://eicrecon.epic-eic.org/pr/1763/capybara/rec_dis_10x100_minQ2=0_craterlake/index.html#TOFBarrelADCTDC

I thought that the changes shouldn't make a difference to the number or charge of hits if the charge limit is set to 0. But it looks like there are significantly fewer hits being created.

The neighbours will only work if the sensor uses a segmentation in the geometry which may not always be the case (but probably always should be...). I don't believe this will account for the edges of sensors either, so a hit in the detector can probably spread outside the sensor.

simonge:
Looks like the problem comes from the MultiSegmentation used for the BTOF. The dd4hep neighbours function doesn't return any cellIDs when it is used. If it is possible to recover which sub-segmentation was used for the specific CellID this might be fixable but I haven't looked yet.

simonge:
Any help with the following compilation error would be appreciated:

```
/usr/bin/ld: ../../algorithms/digi/libalgorithms_digi.so: undefined reference to `dd4hep::SegmentationObject* dd4hep::detail::safe_cast<dd4hep::SegmentationObject>::cast<dd4hep::DDSegmentation::Segmentation>(dd4hep::DDSegmentation::Segmentation const*)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [src/tests/algorithms_test/CMakeFiles/algorithms_test.dir/build.make:388: src/tests/algorithms_test/algorithms_test] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2959: src/tests/algorithms_test/CMakeFiles/algorithms_test.dir/all] Error 2
```

If I exclude the algorithms_test from the top level CMakeLists.txt it works fine despite algorithms_digi being included in other plugins.

src/detectors/BTOF/BTOF.cc
```diff
@@ -52,13 +52,14 @@ void InitPlugin(JApplication* app) {
...
-       .sigma_sharingy        = 0.5 * dd4hep::cm,
-       .readout               = "TOFBarrelHits",
-       .same_sensor_condition = "sensor_1 == sensor_2",
-       .neighbor_fields       = {"x", "y"}},
+  app->Add(new JOmniFactoryGeneratorT<SiliconChargeSharing_factory>(
+      "TOFBarrelChargeSharing", {"TOFBarrelHits"}, {"TOFBarrelSharedHits"},
+      {
+          .sigma_sharingx = 0.1 * dd4hep::cm,
+          .sigma_sharingy = 0.5 * dd4hep::cm,
+          .min_edep       = 0.0,
```
veprbl, line 60:
```suggestion
          .min_edep       = 0.0 * edm4eic::unit::GeV,
```

src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -45,9 +47,20 @@ void InitPlugin(JApplication* app) {
...
   using namespace eicrecon;
 
+  std::string readout = "TaggerTrackerHits";
+
+  app->Add(new JOmniFactoryGeneratorT<SiliconChargeSharing_factory>(
+      "TaggerTrackerChargeSharing", {"TaggerTrackerHits"}, {"TaggerTrackerSharedHits"},
+      {
+          .sigma_sharingx = 15 * dd4hep::um,
+          .sigma_sharingy = 15 * dd4hep::um,
+          .min_edep       = 1.0e-7,
```
veprbl, line 57:
```suggestion
          .min_edep       = 0.1 * edm4eic::unit::keV,
```

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -0,0 +1,212 @@
...
+                                      globalHitPos.z * dd4hep::mm),
+                     m_transform_map[element]);
+
+    std::unordered_set<dd4hep::rec::CellID> tested_cells;
+    std::vector<std::pair<dd4hep::rec::CellID, float>> cell_charge;
+    findAllNeighborsInSensor(cellID, tested_cells, cell_charge, edep, hitPos, element);
+
+    //cout number of neighbors found
+    // error("Found {} neighbors for cellID {}", cell_charge.size(), cellID);
+
```
veprbl, line 85:
```suggestion
```
Please remove or uncomment and use `trace()`.

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -0,0 +1,212 @@
...
+  // Store cellID and deposited energy
+  cell_charge.push_back(std::make_pair(testCellID, edepCell));
+
+  // As there is charge in the cell, test the neighbors too
+  std::set<dd4hep::rec::CellID> testCellNeighbours;
+  segmentation->neighbours(testCellID, testCellNeighbours);
+
+  for (const auto& neighbourCell : testCellNeighbours) {
+    // error("Testing neighbour cellID {}", neighbourCell);
```
veprbl, line 136:
```suggestion
```
Uncomment and use `trace()` or remove.

veprbl (commented):
Looks good. You dug out some interesting functionality for DD4hep's segmentations, although, it may not work to find neighbours this way in detectors with pixel volumes.

simonge:
I don't understand where the capybara diffs in `TaggerTrackerHitPulses` time and position come from.

The distributions in `TaggerTrackerHits` and `TaggerTrackerSharedHits` look correct and almost identical so simply swapping the input to the shared hits shouldn't effect the values this much.

Additionally the reference distribution for the `TaggerTrackerHitPulses` time is different to those run against other PRs.

ssedd1123 (commented):
I can't comment on the TaggerTracker issue, but I verify that this new charge sharing code works on BTOF sensors. The new class looks good from my side.

veprbl:
> I don't understand where the capybara diffs in `TaggerTrackerHitPulses` time and position come from.

I also wasn't sure, but explained by the fact that you were feeding `TaggerTrackerSharedHits` to digitization instead of `TaggerTrackerHits`. I see you reverted it now. That is temporary until you have clustering working?

simonge:
> > I don't understand where the capybara diffs in `TaggerTrackerHitPulses` time and position come from.
> 
> I also wasn't sure, but explained by the fact that you were feeding `TaggerTrackerSharedHits` to digitization instead of `TaggerTrackerHits`. I see you reverted it now. That is temporary until you have clustering working?

It was more a test to try and see if the positions and times were corrupted downstream in other algorithms. My other thought was that cloning the hit and editing the mutable version might cause issues but is appears the pulses still have the wrong positions and times reported in capybara. I will have to do some local tests to see if this happens. The difference in the reference distribution from this and other PRs surely shouldn't happen.

The clustering for LowQ2 hits should be working already, I just wanted to avoid introducing something that is more realistic but not fully calibrated yet until the real detector response is better understood. I will swap it back in for a future PR when it's been tuned a bit better, otherwise I might just forget and leave it.

simonge:
Looks like the position was a unit error.

I'm still not sure what could be happening with the time in the reference spectrum, all other PRs have it effectively matching what the new values are.
image

src/detectors/BTOF/BTOF.cc
```diff
@@ -52,13 +52,14 @@ void InitPlugin(JApplication* app) {
...
 
-  app->Add(new JOmniFactoryGeneratorT<LGADChargeSharing_factory>(
-      "LGADChargeSharing", {"TOFBarrelHits"}, {"TOFBarrelSharedHits"},
-      {.sigma_sharingx        = 0.1 * dd4hep::cm,
-       .sigma_sharingy        = 0.5 * dd4hep::cm,
-       .readout               = "TOFBarrelHits",
-       .same_sensor_condition = "sensor_1 == sensor_2",
-       .neighbor_fields       = {"x", "y"}},
+  app->Add(new JOmniFactoryGeneratorT<SiliconChargeSharing_factory>(
+      "TOFBarrelChargeSharing", {"TOFBarrelHits"}, {"TOFBarrelSharedHits"},
```
wdconinc, line 56: Many factories (for historical reasons) use the primary output collection as the name. Consistency helps when changing factory properties.
```suggestion
      "TOFBarrelSharedHits", {"TOFBarrelHits"}, {"TOFBarrelSharedHits"},
```

src/algorithms/digi/SiliconChargeSharing.h
```diff
@@ -0,0 +1,68 @@
...
+using SiliconChargeSharingAlgorithm =
+    algorithms::Algorithm<algorithms::Input<edm4hep::SimTrackerHitCollection>,
+                          algorithms::Output<edm4hep::SimTrackerHitCollection>>;
+
+class SiliconChargeSharing : public SiliconChargeSharingAlgorithm,
+                             public WithPodConfig<SiliconChargeSharingConfig> {
+
+public:
+  SiliconChargeSharing(std::string_view name)
+      : SiliconChargeSharingAlgorithm{name, {"TOFBarrelHits"}, {"TOFBarrelSharedHits"}, ""} {};
```
veprbl, line 37:
```suggestion
      : SiliconChargeSharingAlgorithm{name, {"inputHits"}, {"outputSharedHits"}, ""} {};
```

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -0,0 +1,209 @@
...
+  // Store cellID and deposited energy
+  cell_charge.push_back(std::make_pair(testCellID, edepCell));
+
+  // As there is charge in the cell, test the neighbors too
+  std::set<dd4hep::rec::CellID> testCellNeighbours;
+  segmentation->neighbours(testCellID, testCellNeighbours);
+
+  for (const auto& neighbourCell : testCellNeighbours) {
+    if (tested_cells.find(neighbourCell) == tested_cells.end()) {
+      findAllNeighborsInSensor(neighbourCell, tested_cells, cell_charge, edep, hitPos, element);
```
wdconinc, line 135: Since this is recursive, can you add a comment on how boundedness is ensured, what worst case performance conditions are, and what typical performance scaling is? We've run into O(n^3) inefficiencies when a different detector adopted a generic algorithm without thinking about the increase in detector segmentation.

wdconinc: Note that we spend 20% of the reconstruction time on TOF digitization now... https://github.com/eic/EICrecon/actions/runs/14943453065/job/41984210764?pr=1763#step:7:1271

wdconinc: Addressing performance can be a future PR, but the comment about scaling would still be useful before merging this.

ssedd1123: This is dynamic programming. Every cell that has been searched will be stored in `tested_cells`, and once a cellID appears in `test_cells`, the recursion on line 135 will not execute. Therefore, the number of times `findAllNeighborsInSensor` is called should equals to number of cells in a sensors (+ boundaries if you want to be precise). `tested_cells` is unordered_map, so searching complexity is ~O(1). Therefore, complexity of findAllNeighborsInSensor should be O(n), where n is the number of neighbors (or total number of cells in a sensor).

wdconinc: I'm assuming neighbors scales with dimensions only, O(d).

wdconinc: Maybe the impact of digitization on performance is then mostly due to the time / 1024 subsampling of the pulse shapes. The janatop output isn't always exclusive.

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -0,0 +1,209 @@
...
+    m_transform_map.try_emplace(element, &element->nominal().worldTransformation());
+    m_segmentation_map.try_emplace(element, getLocalSegmentation(cellID));
+
+    // Try and get a box of the detectorElement solid
+    try {
+      dd4hep::Box box = element->solid();
+
+      if (box) {
+        m_x_range_map.try_emplace(element, box->GetDX());
+        m_y_range_map.try_emplace(element, box->GetDY());
```
wdconinc, line 66: Can/should the map be from element to pair {x,y}?

wdconinc: And should there not be a condition that this is CartesianGridXY, since for CartesianGridXZ the ranges would be incorrect?

simonge: I have swapped to a {x,y} pair.
I believe the getLocalSegmentation should currently restrict it to XY segmentation, this could be generalized if necessary in the future.

src/algorithms/digi/SiliconChargeSharing.cc
simonge: I have just pushed some updates which should improve the performance, primarily not adding hits with 0 deposited energy across the whole detector. I had delayed this so that the BTOF diffs could be checked without the cut as it was before. This was mostly fine for the BTOF as the number of cells isn't huge but larger detectors it would have caused significant issues.

wdconinc: Thanks!

simonge:
@wdconinc Any idea on this new memory leak I've introduced or how to debug it?

wdconinc:
> @wdconinc Any idea on this new memory leak I've introduced or how to debug it?

I'm guessing it came in through https://github.com/eic/epic/pull/842. I'm looking into it.

simonge:
I'm happier with how this looks now, if this can be approved.

Looks like the TOF digitization is closer to 10% now.
A bit off topic but is converting hits into measurements taking 35% of the computation or has that been combined with the CKF itself running somehow?

The change in the epic geometry changed the reference sample between PRs which caused me some confusion but did highlight the bugs in the pulse generation.

wdconinc:
> A bit off topic but is converting hits into measurements taking 35% of the computation or has that been combined with the CKF itself running somehow?

It's not always clear or consistent. JANA2 works by producing inputs for factories just-in-time, through calls to other factories. So, if a factory needs to get another factory to do something, then it will sit there and the extra time will be counted against it. If the inputs are already available when a next factory runs, it will not get that extra cost in running time. This all depends on the order in which factories are activated, and janatop has not enough info in its structures to subtract the time spent in supplier factories.

Looking at https://github.com/eic/EICrecon/actions/runs/14996201580/job/42131377031?pr=1763#step:7:1365, there are a few unambiguous lines possible:
- any digitization step that takes a non-negligible amount of time will show us as a link from raw hits to sim hits, and will be exclusive timing in the digitization itself,
- any link that starts from the JEventProcessorPodio is (at worst) the total time it takes to produce that output (but some time can have been accounted already in other links).

So, from that output, I'd agree that 10% is spent in creating TOFBarrelADCTDC, and all its inputs. It seems like 7.4% of time is spent inside the factory that created TOFBarrelPulses and all its inputs, and 2.3% is then spent inside TOFBarrelSmoothPulses (by which time the input will already have been created, apparently).

veprbl (approved):
Diff looks good, and capybara looks reasonable.

PR #1765, opened 2025-03-24
feat: support edm4eic::TrackerHit::rawHits plural

wdconinc:
This PR adds support for edm4eic::TrackerHit::rawHits as a OneToManyRelation. See https://github.com/eic/EDM4eic/pull/105.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

When more than one RawTrackerHit is combined in one TrackerHit, we may need to think about how to do the weights of the associations correctly.


.vscode/settings.json
wdconinc, line 1: This one snuck in via the GitHub IDE interface, but I'll leave it in because it prevents a pop-up (and potential mistake) due to a past PR from main into a feature branch.

veprbl:
Looks like this is now resolved by https://github.com/eic/EICrecon/pull/1770

PR #1704, opened 2025-01-14
services/io/podio: remove unused datamodel_LinkDef.h

veprbl:
### Briefly, what does this PR introduce?

Removing it as it should be irrelevant for recent PODIO with frames format.
Resolves: #1693

### What kind of change does this PR introduce?
- [x] Bug fix (issue ##1693)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
Not supposed to

### Does this PR change default behavior?
No

PR #1772, opened 2025-03-25
jana_plugin.cmake: explicitly depend plugins on podio_datamodel_glue

veprbl:

https://github.com/eic/EICrecon/pull/1704 broke multiprocess builds (e.g. `make -j10`). The issue is that cmake header scanner ignores rule that generates datamodel_glue.h, which leads to race conditions:
```
src/services/io/podio/JEventSourcePODIO.cc:30:10: fatal error: services/io/podio/datamodel_glue.h: No such file or directory
   30 | #include "services/io/podio/datamodel_glue.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This PR introduces a target-level dependency. The case of external plugin is handled by the fact that the headers are installed along with EICrecon.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/cmake/jana_plugin.cmake
+++ b/cmake/jana_plugin.cmake
@@ -1,6 +1,34 @@
 # Ensure GNU filesystem layout
 include(GNUInstallDirs)
 
+# Sets up properties common for plugin "library" library or "plugin" library
+macro(_plugin_common_target_properties _target)
+  target_include_directories(
+    ${_target}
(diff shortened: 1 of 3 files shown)


wdconinc (approved):
Looks good to me. Nice simplification of common properties, and thanks for moving the generated include to a target.

Issue #1657, opened 2024-10-29
Fix Low-Q2 tagger truth associations 


simonge:
#1589 broke the tagger truth associations.

This can either be fixed by making a CombinedTrackAssociations collection or moving the Tagger hits to a separate reconstruction pathway as suggested in #1571. Avoiding unnecessary attempts to match tracks to clusters and PID information and removing the (only possible) ID as an electron from the record.

The introduction of TaggerTrackerTrackAssociations should be done either way.

PR #1775, opened 2025-03-31
Add Lowq2 track association

simonge:
Reorganization of the FarDetectorLinearTracking to allow Track-Particle associations to be produced for the Low-Q2 Tagger.

Not quite the whole picture but this PR should stand alone before the associations are propagated to the Post-ML tracks and injected into the common tracking flow.

There are lots of alternative approaches to this, the current one here felt like the least ugly (assuming it works).

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1657)
- [ ] Documentation update
- [ ] Other: __

No
Changes the output of FarDetectorLinearTracking to an edm4eic::Track rather than edm4eic::TrackSegment. At a later date both might be produced as the TrackSegment contains TrackPoints which are equivalent to a missing 3D measurement.


src/detectors/LOWQ2/LOWQ2.cc
```diff
@@ -154,9 +151,15 @@ void InitPlugin(JApplication* app) {
   }
 
   // Combine the tracks from each module into one collection
-  app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::TrackSegment>>(
+  app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::Track>>(
       "TaggerTrackerTrackSegments", outputTrackTags, {"TaggerTrackerTrackSegments"}, app));
```
veprbl, line 155:
```suggestion
      "TaggerTrackerTracks", outputTrackTags, {"TaggerTrackerTracks"}, app));
```

src/algorithms/fardetectors/FarDetectorLinearProjection.h
```diff
@@ -17,18 +17,16 @@
 namespace eicrecon {
 
 using FarDetectorLinearProjectionAlgorithm =
-    algorithms::Algorithm<algorithms::Input<edm4eic::TrackSegmentCollection>,
+    algorithms::Algorithm<algorithms::Input<edm4eic::TrackCollection>,
                           algorithms::Output<edm4eic::TrackParametersCollection>>;
```
veprbl, line 21: (Just a thought)
For Acts tracking, we actually output projections as TrackSegments.

src/factories/fardetectors/FarDetectorLinearProjection_factory.h
```diff
@@ -20,7 +21,7 @@ class FarDetectorLinearProjection_factory
 private:
   std::unique_ptr<AlgoT> m_algo;
 
-  PodioInput<edm4eic::TrackSegment> m_hits_input{this};
+  PodioInput<edm4eic::Track> m_hits_input{this};
```
veprbl, line 24:
```suggestion
  PodioInput<edm4eic::Track> m_tracks_input{this};
```

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -74,42 +80,65 @@ void FarDetectorLinearTracking::process(const FarDetectorLinearTracking::Input&
...
+  // Create vector to store indexes of hits in the track
+  std::vector<int> layerHitIndex(m_cfg.n_layer, 0);
 
-  // Iterate over hits in this layer
-  for (auto hit : hits[level]) {
+  int layer = 0;
 
-    hitMatrix->col(level) << hit;
+  // Iterate over all combinations of measurements in the layers without recursion
+  while (layer >= 0) {
```
veprbl, line 95:
```suggestion
  while (true) {
```
Might as well document that it's up to `break` to terminate this loop.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -74,42 +80,65 @@ void FarDetectorLinearTracking::process(const FarDetectorLinearTracking::Input&
...
+    hitMatrix.col(layer) << convertedHits[layer][layerHitIndex[layer]];
```
veprbl, line 96: So the reason why we have to recompute SVD in `checkHitCombination` is because `hitMatrix` is appended here. But it never is reset? Does this mean that algorithm depends on how hits are sorted in the layers?

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
```diff
@@ -74,42 +80,65 @@ void FarDetectorLinearTracking::process(const FarDetectorLinearTracking::Input&
...
+    if (layer > 0 && m_cfg.restrict_direction) {
+      isValid = checkHitPair(hitMatrix.col(layer - 1), hitMatrix.col(layer));
+    }
+
+    // If valid hit combination, move to the next layer or check the combination
+    if (isValid) {
+      if (layer == m_cfg.n_layer - 1) {
+        // Check the combination, if chi2 limit is passed, add the track to the output
+        checkHitCombination(&hitMatrix, outputTracks, assocTracks, inputhits, assocParts,
+                            layerHitIndex);
```
veprbl, line 109: Do we need to ensure that hits are not used again for another track?

veprbl (commented):
Don't see any problems with this, although, I admit, I understand only so much of this. Would be nice to have a unit test for this to check some basic cases. Other than that, there are few comments and questions from me.

src/detectors/LOWQ2/LOWQ2.cc
veprbl: And it would match TaggerTrackerTrackAssociations below. Right?

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
veprbl: Maybe I'm confused, but shouldn't hitMatrix.col(layer) add an extra row (of verctors) every time? It's not an update, is it?

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
veprbl: Okay.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
simonge: `hitMatrix.col(layer)` selects the column of the Eigen matrix relevant to the layer of the detector, then `<<` sets the values of that column to the values in the appropriate `convertedHits` vector.

The Eigen use of `<<` doesn't feel entirely consistent with the rest of C++.

src/detectors/LOWQ2/LOWQ2.cc
veprbl: Looks resolved now.

src/algorithms/fardetectors/FarDetectorLinearTracking.cc
veprbl: Thanks for explaining. I believe, one can push (not replace) individual values with `operator<<` in Eigen, that's why I asked.

veprbl (approved):
LGTM

PR #1777, opened 2025-04-01
Use lowq2 track associations

simonge:
Follow on from https://github.com/eic/EICrecon/pull/1775

Propagates the associations into the tracking workflow so the MCParticle best associated with a track in the Low-Q2 tagger should be identifiable.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1657)
- [ ] Documentation update
- [ ] Other: __

No
Additional MC to Reconstructed Particle associations will be output


src/algorithms/fardetectors/FarDetectorMLReconstruction.h
```diff
@@ -24,19 +25,22 @@ enum FarDetectorMLNNIndexIn { PosY, PosZ, DirX, DirY };
...
 public:
   FarDetectorMLReconstruction(std::string_view name)
-      : FarDetectorMLReconstructionAlgorithm{name,
-                                             {"TrackParameters", "BeamElectrons"},
-                                             {"Trajectory", "TrackParameters", "Track"},
-                                             "Reconstruct track parameters using ML method."} {}
+      : FarDetectorMLReconstructionAlgorithm{
+            name,
+            {"TrackParameters", "BeamElectrons", "FittedTracks", "FittedTrackAssociations"},
+            {"Trajectory", "TrackParameters", "Track", "PropegatedTrackAssociations"},
```
veprbl, line 42:
```suggestion
            {"Trajectory", "TrackParameters", "Track", "PropagatedTrackAssociations"},
```

veprbl:
Do you understand why, according to capybara report, all the inclusive algorithms and breit frame particles are so excited by this change?

simonge:
> Do you understand why, according to capybara report, all the inclusive algorithms and breit frame particles are so excited by this change?

It comes from requiring truth matching of the first scattered electron to a reconstructed particle in the ScatteredElectronTruth and HadronicFinalState algorithms. For the minQ2=0 most of these were missing as the electrons go into the Low-Q2 Tagger.
https://github.com/eic/EICrecon/blob/main/src/algorithms/reco/ScatteredElectronsTruth.cc#L67
https://github.com/eic/EICrecon/blob/main/src/algorithms/reco/HadronicFinalState.cc#L73

For the HFS this seems a little bit counter productive as you will be restricting the kinematic range of the reconstruction which use it to areas which other approaches work better.

veprbl (approved):
LGTM

PR #1778, opened 2025-04-01
LGADPulseGeneration -> SiliconPulseGeneration, LGADPulseDigitization -> EICROCDigitization

ssedd1123:
### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?

LGADPulseGeneration class is removed in favor of SiliconPulseGeneration class. Added a SiliconPulseDiscretization class to convert the smooth pulse from SiliconPulseGeneration to it's discretized form with size of time bin set to be equal to that of BTOF LGAD clock cycle. 

LGADPulseDigitization is renamed to SiliconPulseDigitization for a unified naming scheme.

### Does this PR change default behavior?

BTOF now uses SiliconPulseGeneration.

PR #1779, opened 2025-04-01
Pr/lgad clustering

ssedd1123:
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

TOFBarrelRecHits is now generated from the new class "LGADHitReconstruction". This class takes the digitized ADC and TDC values from each strips within the BTOF sensor to reconstruct the hit location, hit time and energy deposition.

Currently, this class is just a skeleton with position calculated by a simple weighted average, energy is just the sum of ADC values inside a sensor and time is just the earliest TDC value within a sensor, transformed linearly to get a rough time and EDep. There is no  time walk correction and this class is unable to distinguish multiple hits in the same sensor if they arrive within the same EICROC cycle (25 ns). However, our group is developing a more sophisticated clustering algorithm and the time walk correction will be implemented in the future in another pull request. This is just a skeletal class for our group to develop further. 

Name of the class is still open for discussion. See #1778

TOFBarrelRecHits is now generated by LGADHitReconstruction. If TOFBarrelRecHits are used by downstream classes for track reconstruction, I believe it will lower the tracking efficiency because digitization class discard hits with too low energy threshold. The lack of time walk correction and energy loss through the boundaries of BTOF sensor also contributes to the worsening of position resolution, which may or may not have huge ramifications. 

Part of the reason I make this pull request is precisely because I don't fully understand it's downstream effect. Please feel free to let me know if any part of the performance is significantly degraded.

### DIFF
--- a/src/algorithms/reco/LGADHitReconstruction.cc
+++ b/src/algorithms/reco/LGADHitReconstruction.cc
@@ -1,0 +1,120 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Chun Yuen Tsang
+
+#include "LGADHitReconstruction.h"
+
+#include <Evaluator/DD4hepUnits.h>
+#include <Math/GenVector/Cartesian3D.h>
+#include <Math/GenVector/DisplacementVector3D.h>
+#include <edm4eic/Cov3f.h>
+#include <edm4eic/CovDiag3f.h>
+#include <edm4hep/Vector3f.h>
+#include <fmt/core.h>
+#include <iterator>
+#include <spdlog/common.h>
+#include <stddef.h>
+#include <utility>
+#include <vector>
+
+#include "TMatrixT.h"
+
+namespace eicrecon {
+
+void LGADHitReconstruction::init() {
+
+  m_converter = algorithms::GeoSvc::instance().cellIDPositionConverter();
+  m_detector  = algorithms::GeoSvc::instance().detector();
+  auto seg = m_detector -> readout(m_cfg.readout).segmentation();
+  auto type = seg.type();
+  m_decoder = seg.decoder();
+}
+
+dd4hep::rec::CellID LGADHitReconstruction::getSensorInfos(const dd4hep::rec::CellID& id) const {
+  // CellID for BarrelTOF is composed of 6 parts
+  // system, layer, module, sensor, x, y
+  // If we fix x and y to zero, what remains will be the sensor information only
+  auto id_return = id;
+  m_decoder->set(id_return, "x", 0);
+  m_decoder->set(id_return, "y", 0);
+  return id_return;
+}
+
+void LGADHitReconstruction::process(const LGADHitReconstruction::Input& input,
+		                    const LGADHitReconstruction::Output& output) const {
+  using dd4hep::mm;
+
+  const auto [TDCADC_hits] = input;
+  auto [rec_hits] = output;
+
+  // collection of ADC values from all sensors and group them by sensor 
+  std::unordered_map<dd4hep::rec::CellID, std::vector<HitInfo>> hitsBySensors;
+
+  for (const auto& TDCADC_hit : *TDCADC_hits) {
+
+    auto id = TDCADC_hit.getCellID();
+
+    // Get position and dimension
+    auto pos = m_converter->position(id);
+    // Get sensors info
+    auto sensorID = this->getSensorInfos(id);
+    hitsBySensors[sensorID].emplace_back(pos.x(), pos.y(), pos.z(), 
+		                         int(TDCADC_hit.getCharge()),
+                                         int(TDCADC_hit.getTimeStamp()), id);
+  }
+
+  for (const auto& sensor : hitsBySensors) {
+    // Right now the clustering algorithm a simple average over all hits in a sensors
+    // Will be problematic near the edges, but it's just an illustration
+    double ave_x = 0, ave_y = 0, ave_z = 0;
+    double tot_charge = 0;
+    const auto& hits  = sensor.second;
+    if(hits.size() == 0) continue;
+    // find cellID for the cell with maximum ADC value within a sensor
+    auto id         = hits[0].id;
+    auto curr_adc   = hits[0].adc;
+    auto first_tdc  = hits[0].tdc;
+
+    for (const auto& hit : hits) {
+      // weigh all hits by ADC value
+      ave_x += hit.adc * hit.x;
+      ave_y += hit.adc * hit.y;
+      ave_z += hit.adc * hit.z;
+
+      tot_charge += hit.adc;
+      if (hit.adc > curr_adc) {
+        curr_adc = hit.adc;
+        id       = hit.id;
+      }
+      first_tdc = std::min(first_tdc, hit.tdc);
+    }
+
+    ave_x /= tot_charge;
+    ave_y /= tot_charge;
+    ave_z /= tot_charge;
+
+    // adc to charge
+    float charge = tot_charge * m_cfg.c_slope + m_cfg.c_intercept;
+    // TDC to time
+    float time = first_tdc * m_cfg.t_slope + m_cfg.t_intercept;
+
+    auto cellSize = m_converter->cellDimensions(id);
+    double varX = cellSize[0] / mm;
+    varX *= varX;
+    double varY = cellSize[1] / mm;
+    varY *= varY;
+    double varZ = cellSize.size() > 2? cellSize[2] / mm : 0;
+    varZ *= varZ;
+
+    rec_hits->create(id,
+                     edm4hep::Vector3f{static_cast<float>(ave_x / mm),
+                                       static_cast<float>(ave_y / mm),
+                                       static_cast<float>(ave_z / mm)}, // mm
+                     edm4eic::CovDiag3f{varX, varY, varZ}, // should be the covariance of position
+                     time,                                 // ns
+                     0.0F,                                 // covariance of time
+                     charge,                               // total ADC sum
+                     0.0F);                                // Error on the energy
+  }
+}
+
+} // namespace eicrecon
--- a/src/algorithms/reco/LGADHitReconstruction.h
+++ b/src/algorithms/reco/LGADHitReconstruction.h
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Chun Yuen Tsang
+
+#pragma once
+
+#include <DD4hep/Detector.h>
+#include <DDRec/CellIDPositionConverter.h>
+#include <edm4eic/RawTrackerHitCollection.h>
+#include <edm4eic/TrackerHitCollection.h>
+#include <algorithms/algorithm.h>
+#include <memory>
+#include <spdlog/logger.h>
+#include <algorithms/geo.h>
+#include <algorithms/service.h>
+
+#include "LGADHitReconstructionConfig.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+using LGADHitReconstructionAlgorithm = 
+	algorithms::Algorithm<algorithms::Input<edm4eic::RawTrackerHitCollection>,
+	                      algorithms::Output<edm4eic::TrackerHitCollection>>;
+
+class LGADHitReconstruction : public LGADHitReconstructionAlgorithm,
+	                      public WithPodConfig<LGADHitReconstructionConfig> {
+
+public:
+  LGADHitReconstruction(std::string_view name) 
+    : LGADHitReconstructionAlgorithm{name, {"TOFBarrelADCTDC"}, {"TOFBarrelRecHit"}, ""} {};
+
+  void init() final;
+  void process(const Input&, const Output&) const final;
+
+private:
+  struct HitInfo {
+    double x, y, z;
+    int adc, tdc;
+    dd4hep::rec::CellID id;
+  };
+  dd4hep::rec::CellID getSensorInfos(const dd4hep::rec::CellID& id) const;
+  /** algorithm logger */
+  std::shared_ptr<spdlog::logger> m_log;
+
+  /// Cell ID position converter
+  const dd4hep::rec::CellIDPositionConverter* m_converter = nullptr;
+
+  /// fetch sensor information from cellID
+  const dd4hep::DDSegmentation::BitFieldCoder* m_decoder = nullptr;
+
+  const dd4hep::Detector* m_detector = nullptr;
+};
+} // namespace eicrecon
--- a/src/algorithms/reco/LGADHitReconstructionConfig.h
+++ b/src/algorithms/reco/LGADHitReconstructionConfig.h
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Chun Yuen Tsang
+
+#pragma once
+
+namespace eicrecon {
+    struct LGADHitReconstructionConfig {
+        // parameters that convert ADC to EDep
+        double c_slope = 3.86976e-7, c_intercept=2.42716e-5;
+        // parameters that convert TDC to hit time (ns)
+        double t_slope = 0.0197305, t_intercept = 0.208047;
+	std::string readout = "TOFBarrelHits";
+    };
+}
--- a/src/detectors/BTOF/BTOF.cc
+++ b/src/detectors/BTOF/BTOF.cc
@@ -21,6 +21,7 @@
 #include "factories/digi/LGADChargeSharing_factory.h"
 #include "factories/digi/LGADPulseGeneration_factory.h"
 #include "factories/digi/LGADPulseDigitization_factory.h"
+#include "factories/reco/LGADHitReconstruction_factory.h"
 #include "global/pid_lut/PIDLookup_factory.h"
 #include "services/geometry/dd4hep/DD4hep_service.h"
 
@@ -48,13 +49,11 @@ void InitPlugin(JApplication* app) {
   ));
 
   // Convert raw digitized hits into hits with geometry info (ready for tracking)
-  app->Add(new JOmniFactoryGeneratorT<TrackerHitReconstruction_factory>(
+  app->Add(new JOmniFactoryGeneratorT<LGADHitReconstruction_factory>(
       "TOFBarrelRecHits",
-      {"TOFBarrelRawHits"},    // Input data collection tags
+      {"TOFBarrelADCTDC"},    // Input data collection tags
       {"TOFBarrelRecHits"},    // Output data tag
-      {
-          .timeResolution = 10,
-      },
+      {},
       app
   ));         // Hit reco default config for factories
 
--- a/src/factories/reco/LGADHitReconstruction_factory.h
+++ b/src/factories/reco/LGADHitReconstruction_factory.h
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2023 Wouter Deconinck
+
+#pragma once
+
+#include "algorithms/reco/LGADHitReconstruction.h"
+#include "extensions/jana/JOmniFactory.h"
+#include "services/geometry/dd4hep/DD4hep_service.h"
+
+namespace eicrecon {
+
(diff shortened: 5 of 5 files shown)


simonge:
I think here you will want to split the algorithm into two parts again:

1. Calibration of the RawTrackerHit into a TrackerHit.
2. Clustering of the reconstructed hits into a Measurement2D.

The calibration of an individual hit can be very simple at the moment as there is no variation in the gain, delay, noise etc. from channel to channel but basic timewalk can be added there.

The TrackerHit only has a single relation to a RawTrackerHit while the Measurement2D is supposed to allow for clustering. I have a similar algorithm to what you're developing, which requires pre-divided collections, here:
https://github.com/eic/EICrecon/blob/main/src/algorithms/fardetectors/FarDetectorTrackerCluster.cc

A basic weighted clustering algorithm could easily be shared between tracking detectors but I wouldn't necessarily advise adopting mine.

This will require a bit of a change of where the Measurements are included into the rest of the tracking pipeline

simonge:
> I can certainly output the cluster information as Measurement2D, but I am not sure if the tracking group is using it. As a matter of fact, I don't know what the downstream track reconstruction class takes as input. I know that before I was asked to develop TOF class, the final output from BTOF.cc is "TOFBarrelRecHits", which is a TrackerHit class, so I just assumed that ACTS downstream takes "TOFBarrelRecHits" as input for their track reconstruction. Can you confirm if downstream classes will still work correctly if I switch "TOFBarrelRecHits" to a Measurement2D class please? Because at this stage, I do want to see how significantly track reconstruction is affacted by digitization.

The Measurement2D is the interface required by acts. At the moment all of the other TrackerHits are converted into Measurement2Ds, simply without clustering. You will need to make sure you get the correct surface from the acts geometry so it gets used correctly:
https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackerMeasurementFromHits.cc

ssedd1123:
TofEfficiency_process complains that TOFBarrelRecHits is no longer of type TrackerHit. I modified it to accept Measurement2D, but I don't know if I did it correctly as I did not know how to test TofEfficiency_process locally. Please double check my modifications, and if you can tell me how to run it locally or give me some relevant references, I would greatly appreciate that.

ssedd1123:
All physics_benchmarks failed. I did some investigation and, apparently, it's caused by the failure of Measurement2D to be associated with the corresponding raw_hit. I tried to revert the output of LGADHitClustering back to TrackerHit, do some very rudimentary hit matching and assign it to the output TrackerHit with TrackerHit::setRawHit. I don't know if the resultant physics_benchmarks outputs make sense, but at least snakemake is not throwing any error when I run it locally.

I tried to call TrackerHit::setRawHit on all the tracker hits before adding them to Measurement2D, but the physics_benchmarks still fail. Did I miss a step here? Is there a equivalent function in Measurement2D for TrackerHit::setRawHit?

veprbl:
I believe the issue was fixed in #1792. I've rebased to fix conflicts due to  recent introduction of forced clang-format. Let's see how this runs (may take time, there is a bit of job backlog currently)

veprbl:
Somehow an issue crept into the previous merge: https://github.com/eic/EICrecon/compare/49dbca98c2027fa823766080b2914c40457d5f93..a1d28b172db6746afd9eeac66cb4fd69ee5fcecf where formatted code from main branch got duplicated.

ssedd1123:
> @ssedd1123 Could you point to a failed benchmark CI result?

If I don't use ClusterAssociation class, it will fail at generating physics_benchmarks/results/epic_craterlake/dis/5on41/minQ2=1/kinematics_correlations

That's not the only failure, but that's what I was using to diagnose my code. 

Below is a screenshot of what I would see if I run that benchmark locally without cluster association,

image

veprbl:
Okay, that one benchmakr is a bit cryptic about its failures, but common cause is that something is wrong with tracking.

Looking at https://eicrecon.epic-eic.org/pr/1779/capybara/rec_dis_5x41_minQ2=0_craterlake_5x41/index.html#CentralCKFTracks
This appears to reduce number of hits on track:
image
image

ssedd1123:
A reduction in BTOF efficiency is expected because the pulse amplitude needs to pass threshold set in `EICROCDigitization` class. Position and timing resolution are going to get worse too because there is no time walk correction.

However, `LGADHitClusterAssociation` do nothing to boost the efficiency back up. All it does is to convert the Measurement2D back to TrackerHit and associate it with the corresponding raw hit, and somehow that eliminates errors from the benchmarks. 

I don't know the details of that benchmark, but it seems to me that the lack of efficiency is not the only reason why that benchmark fails.

ssedd1123:
It looks like the loss of BTOF efficiency isn't that significant. I passed all tests, but only with LGADHitClusterAssociation class enabled. I can't pass the test for physics_benchmarks/results/epic_craterlake/dis/5on41/minQ2=1/kinematics_correlations without it. @simonge If I am missing a step or use Measurement2D incorrectly, please let me know. 

image

simonge:
From what I can see, you shouldn't need the `LGADHitClusterAssociation` algorithm creating `TOFBarrelRecHits` at all. This will just get passed into the `TrackerMeasurementFromHits` algorithm which does a very simple conversion back into `Measurement2Ds` again.

Instead you want to use the `TOFBarrelClusterHits` now directly.

So remove the following line to take the BTOF hits out of the other tracking workflow:
https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L52

Then you will need to add a new `Measurement2D` `CollectionCollector_factory` which combines the other measurements with those of the BTOF before passing them on the the acts CKF tracking factories.
https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L93

ssedd1123:
> From what I can see, you shouldn't need the `LGADHitClusterAssociation` algorithm creating `TOFBarrelRecHits` at all. This will just get passed into the `TrackerMeasurementFromHits` algorithm which does a very simple conversion back into `Measurement2Ds` again.
> 
> Instead you want to use the `TOFBarrelClusterHits` now directly.
> 
> So remove the following line to take the BTOF hits out of the other tracking workflow: https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L52
> 
> Then you will need to add a new `Measurement2D` `CollectionCollector_factory` which combines the other measurements with those of the BTOF before passing them on the the acts CKF tracking factories. https://github.com/eic/EICrecon/blob/main/src/global/tracking/tracking.cc#L93

I think I get it now. CollectionCollector is just collecting various `TrackerHit`s, but since `TOFBarrelRecHits` is now `Measurement2D`, it won't work. I added a new `CollectionCollector` for measurement2D as you have advised. Now the tests work locally. Let's see if it passes the workflow. Thanks for the suggestions. 

I renamed the output of type `Measurement2D `from BTOF to "TOFBarrelClusterHits". I still keep the `LGADHitClusterAssociation` class to generate `TOFBarrelRecHits` of type `TrackerHit`. I don't know if any other downstream algorithms need `TOFBarrelRecHits` of type `TrackerHit`. `TofEfficiency_process` for instance only works with `TrackerHit`. I keep the association class to keep things from breaking, but if we confirm there is no other side-effect when `TOFBarrelRecHits` is switched to type `Measurement2D`, then I can remove `LGADHitClusterAssociation`.

src/algorithms/reco/LGADHitCalibrationConfig.h
```diff
@@ -0,0 +1,13 @@
...
+// Copyright (C) 2025 Chun Yuen Tsang
+
+#pragma once
+
+namespace eicrecon {
+struct LGADHitCalibrationConfig {
+  // parameters that convert ADC to EDep
+  double c_slope = 3.86976e-7, c_intercept = 2.42716e-5;
+  // parameters that convert TDC to hit time (ns)
+  double t_slope = 0.0197305, t_intercept = 0.208047;
```
simonge, line 11: Could you comment on where these values come from? I would have expected `t_slope=25/(2^10)=0.024414`. Then without any timewalk calibration `t_intercept` would be half of that to move to the middle of the bin.

ssedd1123: Excellent point. My numbers were empirically found to be the optimal by plotting `TOFBarrelHits.time `vs. `TOFBarrelRecHits.time` when I first made the pull request, but I see now that those numbers were incorrect for the latest commit. I confirmed that the slope and intercept you've suggested is much more accurate. Here's the `TOFBarrelHits.time `vs. `TOFBarrelRecHits.time` with your suggested calibration parameters,

image

 The command to plot it from podio_output.root is `events->Draw("TOFBarrelHits.time:TOFBarrelRecHits.time", "TOFBarrelHits@.size()==1")`. It's easier to focus on event with only one hit, hence `"TOFBarrelHits@.size()==1"`. Those parameters will be updated.

simonge:
Please can you remind me what the purpose of the LGADHitClusterAssociation algorithm is. If this is just to appease some tests and benchmarks it would be much better to address the problem with those rather than add this extra algorithm which converts the measurements back into hits.

src/global/tracking/tracking.cc
```diff
@@ -57,7 +58,13 @@ void InitPlugin(JApplication* app) {
       app));
 
   app->Add(new JOmniFactoryGeneratorT<TrackerMeasurementFromHits_factory>(
-      "CentralTrackerMeasurements", {"CentralTrackingRecHits"}, {"CentralTrackerMeasurements"},
+      "CentralTrackerMeasurements", {"CentralTrackingRecHits"}, {"CentralTrackerRecMeasurements"},
```
simonge, line 61: Could we be more explicit in the naming here, something like:
```suggestion
      "CentralTrackerNoTOFBarrelMeasurements", {"CentralTrackingRecHits"}, {"CentralTrackerNoTOFBarrelMeasurements"},
```
This collection should also be added to the output file.

src/global/tracking/tracking.cc
```diff
@@ -57,7 +58,13 @@ void InitPlugin(JApplication* app) {
...
+      app));
+
+  // include the collection from previous collector
+  app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::Measurement2D>>(
+      "CentralTrackingMeas2D", {"CentralTrackerRecMeasurements", "TOFBarrelClusterHits"},
```
simonge, line 66:
```suggestion
      "CentralTrackerMeasurements", {"CentralTrackerNoTOFBarrelMeasurements", "TOFBarrelClusterHits"},
```

src/algorithms/reco/LGADHitClustering.cc
```diff
@@ -0,0 +1,141 @@
...
+    // NOT the cell with MAX ADC value
+    const auto* context = m_converter->findContext(id);
+    auto gPos           = this->_local2Global(context, locPos);
+    id                  = m_converter->cellID(gPos);
+
+    if (!id)
+      error("Failed to find cellID for cluster center! Does your detector have hole in the middle "
+            "or somehow concave in shape? This class doesn't work with concave detector.");
+
+    cluster.setSurface(id);
```
simonge, line 134: From my understanding this should be the acts surface id rather than the dd4hep cell id (They might often but not always be the same thing). e.g.
https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/TrackerMeasurementFromHits.cc#L134

src/algorithms/reco/LGADHitClustering.cc
```diff
@@ -0,0 +1,141 @@
...
+    auto id = calibrated_hit.getCellID();
+    // Get sensors info
+    auto sensorID = this->getSensorInfos(id);
+    hitsBySensors[sensorID].emplace_back(calibrated_hit);
+  }
+
+  for (const auto& sensor : hitsBySensors) {
+    auto cluster = clusters->create();
+    // Right now the clustering algorithm a simple average over all hits in a sensors
+    // Will be problematic near the edges, but it's just an illustration
```
simonge, line 84: This is probably a bit too simplistic at the moment. There is no check if the hits are in neighbouring cells or within an appropriate time window from each other.

ssedd1123: It is implemented. Only hits within a time window given by `timeWindow` in `algorithm/reco/LGADHitClusteringConfig.h` will be grouped into one cluster.

simonge:
> @bschmookler can you also comment on how this algorithm will change the track to particle association?

I was just looking at the track-particle association and it just counts the number of hits over all contributing measurements coming from a particular MCParticle. To keep this equivalent and not over bias through large clusters I believe you would just need to swap to adding the weight of the association rather than +1 for every hit. There might be further technicalities I didn't look at.

src/algorithms/reco/LGADHitClustering.cc
simonge: The timing window looks good but there is no check for neighbouring cells. If hits occurred in opposite corners of the sensor it would reconstruct a hit in the middle.

ssedd1123: Hmmmmm it's a bit too sophisticated for this, don't you think? I am envisioning this class to just be a platform for future development of clustering algorithm, instead of it being the final product.

In any case, that shouldn't be too hard. I can implement union find algorithm for that. I am working on that.

simonge: I would mostly consider this the core feature of clustering. The extra complexity I thought you were alluding to would come in e.g. detector specific weighting scheme to locate the centroid, compensating for edge effects, separating overlapping clusters into separate measurements.

simonge:
We should probably understand and fix why this can propagate to impossible values in the InclusiveKinematicsKB collection.
https://eicrecon.epic-eic.org/pr/1779/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#InclusiveKinematicsJB

src/algorithms/reco/LGADHitClustering.cc
ssedd1123: I updated the clustering algorithm. I implemented union-find algorithm to group hits to clusters. Now hits are associated with each other if they are neighbor of each other AND time difference between them is smaller than `m_cfg.deltaT`, which is 1 ns in this case. 

The time complexity is O(k*n*log(n)*m^2*alpha(p)), where k is average number of cells that are hit in an event, n is number of neighbors for each cell, and m is the average number of hits in a cell, p is average number of distinct cluster in an event and alpha is the inverse Ackermann function. The alpha term comes from the merging method of union-find. I assumed O(n*log(n)) is the complexity of finding neighbors.

ShujieL:
Seems that the` segmentation.neighbours()` function returns two closest cells along each direction,e.g. the 2D x-y grid will have 4 neighbours, the diagonal cells are not included. Please make sure it serves your purpose. 

Here's the script I used for this test for your reference:
```import dd4hep
from dd4hep import DetElement, VolumeManager, CellID, segmentation
import ROOT
import DDRec

xml_path = <your_epic_detector_xml>
cellID = <valid_cell_id>
detector_name = "VertexBarrel"

dd4hep.dd4hep.initialize()
detector = dd4hep.Detector.getInstance()
detector.fromCompact(xml_path)
for i,j in detector.sensitiveDetectors():
	print(i)
sens = detector.sensitiveDetector(detector_name)
neighbours = ROOT.std.set("dd4hep::CellID")()
sens.readout().segmentation().neighbours(cellID, neighbours)

converter = DDRec.CellIDPositionConverter(detector)
for nn in neighbours:
	converter.position(nn)
	print(f"Cell ID {nn}, Position:",pos.x(),pos.y(),pos.z())```

**Comment by ssedd1123** [comment]
> Seems that the` segmentation.neighbours()` function returns two closest cells along each direction,e.g. the 2D x-y grid will have 4 neighbours, the diagonal cells are not included. Please make sure it serves your purpose.
> 
> Here's the script I used for this test for your reference:
> 
> ```
> from dd4hep import DetElement, VolumeManager, CellID, segmentation
> import ROOT
> import DDRec
> 
> xml_path = <your_epic_detector_xml>
> cellID = <valid_cell_id>
> detector_name = "VertexBarrel"
> 
> dd4hep.dd4hep.initialize()
> detector = dd4hep.Detector.getInstance()
> detector.fromCompact(xml_path)
> for i,j in detector.sensitiveDetectors():
> 	print(i)
> sens = detector.sensitiveDetector(detector_name)
> neighbours = ROOT.std.set("dd4hep::CellID")()
> sens.readout().segmentation().neighbours(cellID, neighbours)
> 
> converter = DDRec.CellIDPositionConverter(detector)
> for nn in neighbours:
> 	converter.position(nn)
> 	print(f"Cell ID {nn}, Position:",pos.x(),pos.y(),pos.z())```
> ```

Thanks for the finding, but I believe my algorithm doesn't need diagonal neighbors to work. The diagonal elements will be grouped by association. e.g. if (0, 0) and (0, 1) are neighbors and (0, 1) and (1, 1) are neighbors, then (0, 0), (0, 1) and (1, 1) will all belong to the same clusters by association. The association part is done via union-find algorithm.

**Comment by simonge** [comment]
@ssedd1123 Please can you separate out the work on the CFDROC Digitization to a different branch. It would be nice to finish the clustering bits.

**Comment by simonge** [comment]
> > @ssedd1123 Please can you separate out the work on the CFDROC Digitization to a different branch. It would be nice to finish the clustering bits.
> 
> In a twist of event, it's actually EICROC that needs to be removed.
> 
> I had a discussion with the hardware people at BNL who has the LGAD prototype. They told me that BTOF uses constant fraction discriminator but ETOF uses constant threshold (EICROC). Since I focus on BTOF here, we need to use constant fraction discriminator (CFD).

Ok good to know. It will be useful to have models for both in the code base but I would still suggest moving the addition of the CFD digitization into a separate PR.

**Comment by ssedd1123** [comment]
> > > @ssedd1123 Please can you separate out the work on the CFDROC Digitization to a different branch. It would be nice to finish the clustering bits.
> > 
> > 
> > In a twist of event, it's actually EICROC that needs to be removed.
> > I had a discussion with the hardware people at BNL who has the LGAD prototype. They told me that BTOF uses constant fraction discriminator but ETOF uses constant threshold (EICROC). Since I focus on BTOF here, we need to use constant fraction discriminator (CFD).
> 
> Ok good to know. It will be useful to have models for both in the code base but I would still suggest moving the addition of the CFD digitization into a separate PR.

Alright okay I will work on it.

**Comment by ssedd1123** [comment]
All the failed tests failed with "Cannot connect to the Docker daemon". How can we restart those runs? Or is it caused by something on my side?

**Comment by veprbl** [comment]
> All the failed tests failed with "Cannot connect to the Docker daemon". How can we restart those runs? Or is it caused by something on my side?

I've restarted manually. Currently, there is no wide access to granular restarting. In a pinch you could restart trigger_container job on GitHub, but that reruns everything, so if an outage is ongoing they may fail again.

Also note that eicweb jobs are not required for merging PRs for EICrecon.

**Comment by simonge** [comment]
Looks like your revert of tracking.cc might have come from an out of data local main. Please recover the B0 changes.

**Comment by ssedd1123** [comment]
Sorry for my mistake. `tracking.cc` is now updated to that from the main branch.

File: src/algorithms/fardetectors/FarDetectorMLReconstructionConfig.h [hunk 6]
```diff
@@ -11,6 +11,6 @@ struct FarDetectorMLReconstructionConfig {
   std::string modelPath;
   std::string methodName;
 
-  bool requireBeamElectron{true};
+  bool requireBeamElectron{false};
```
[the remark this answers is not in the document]
ssedd1123: I didn't make any modifications to FarDetectorMLReconstructionConfig.h. I don't think we should touch that file in this pull request.

src/algorithms/tracking/LGADHitClustering.cc
```diff
@@ -0,0 +1,253 @@
...
+  for (size_t id = 0; id < hits.size(); ++id) {
+    const auto& hit = hits[id];
+    if (hit.getTime() < earliest_time) {
+      earliest_time = hit.getTime();
+      time_err      = hit.getTimeError();
+    }
+    // weigh all hits by ADC value
+    auto pos = m_seg->position(hit.getCellID());
+    if (hit.getEdep() < 0)
+      error("Edep for hit at cellID{} is negative. Abort!.", hit.getCellID());
```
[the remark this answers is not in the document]
ssedd1123: The error message is modified.

veprbl (approved):

PR #948, opened 2023-09-12
Add Attenuation in CalorimeterHitDigi

Chao1009:
### Briefly, what does this PR introduce?
Add an attenuation factor and a light collection efficiency in CalorimeterHitDigi

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators @mariakzurek @sly2j @wdconinc 

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No, by default this attenuation is turned off (attenuation = 0.)

### Does this PR change default behavior?
No, by default this attenuation is turned off (attenuation = 0.)

Issue #1776, opened 2025-04-01
Implementation of the BIC attenuated hits


mhkim-anl:
This issue is for implementing the BIC attenuated hits.

[will be updated in detail...]

veprbl:
Consider setting gains based on evaluator expressions, like in https://github.com/eic/EICrecon/pull/1349

mhkim-anl:
@veprbl This branch doesn't include the gain yet, but I'll refer to the expressions in #1349 when I start implementing the gains.

PR #1783, opened 2025-04-05
1776 implementation of the bic attenuated hits

mhkim-anl:

This PR adds new factory, output, plugin, algorithm for
1. Regroup the `SimCalorimeterHit` by primary MC particle.
2. Merge the `SimCalorimeterHit` with respect to a common quantity, e.g., same z-segmentation.
3. Attenuate the energy of the `SimCalorimeterHit`.
4. Store the processed `SimCalorimeterHit` and corresponding `CaloHitContribution` in the output.

For detectors that don't need 2. or 3., `is_merge` and `is_attenuation` can be used to include or omit them. Because the algorithm has multiple functions for `SimCalorimeterHit`, we named it `SimCalorimeterHitProcessor`.

The Barrel Imaging Calorimeter (BIC) will use all the above three functions. Therefore, the followings have been added or updated accordingly.

* `src/factories/calorimetry/SimCalorimeterHitProcessor_factory.h`

* `src/services/io/podio/JEventProcessorPODIO.cc`
`EcalBarrelScFiPAttenuatedHits` (for readout at p-going side), `EcalBarrelScFiNAttenuatedHits` (for readout at e-going side), `EcalBarrelScFiPAttenuatedHitContributions`, and `EcalBarrelScFiNAttenuatedHitContributions` were added.

* `src/detectors/BEMC/BEMC.cc`
Constants and plugins for `EcalBarrelScFiPAttenuatedHits`, `EcalBarrelScFiNAttenuatedHits`, `EcalBarrelScFiPAttenuatedHitContributions`, and `EcalBarrelScFiNAttenuatedHitContributions` were added.

* `src/algorithms/calorimetry/SimCalorimeterHitProcessorConfig.h`
* `src/algorithms/calorimetry/SimCalorimeterHitProcessor.h`
* `src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc`

Two types of the energy distributions below were compared to cross-check the new algorithm and they showed the same distributions.
1. Analyze an edm4hep file directly using a standalone code.
2. Analyze the eicrecon file that was generated from the edm4hep through EICrecon that includes this PR.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1776)
- [ ] Documentation update
- [ ] Other: __

No
No

Closes: #948


wdconinc:
Can you remove the duplicate files at the project top level please?

mhkim-anl:
@wdconinc I'm sorry I realized that I committed the codes in the wrong location. I removed the duplicate files and all the codes presented in the above PR request were replaced by most recent ones.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -259,6 +258,8 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "EcalBarrelImagingRecHits",
             "EcalBarrelImagingClusters",
             "EcalBarrelImagingClusterAssociations",
+            "EcalBarrelScFiPAttenuatedHits",
+            "EcalBarrelScFiNAttenuatedHits"
             "EcalBarrelScFiRawHits",
             "EcalBarrelScFiRecHits",
             "EcalBarrelScFiClusters",
```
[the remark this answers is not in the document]
wdconinc: This indeed needs fixing.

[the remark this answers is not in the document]
mhkim-anl: @wdconinc Thanks for pointing that out. I added the comma and updated the code.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -346,10 +347,6 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
             "EcalFarForwardZDCRawHitAssociations",
             "HcalFarForwardZDCRawHitAssociations",
 #endif
-#if EDM4EIC_VERSION_MAJOR >= 8
-            "TrackClusterMatches",
-#endif
-
```
wdconinc, line 352: Please review why this is removed.

mhkim-anl: I'm sorry, I don't understand why that part was removed. I guess I removed it by mistake while adding the new outputs. I cloned the "JEventProcessorPODIO.cc" from main branch again and added the "EcalBarrelScFiPAttenuatedHits" and "EcalBarrelScFiNAttenuatedHits".

mhkim-anl:
The following codes have been updated for the final review. The top comments that present the net information have also been updated.

* `src/factories/calorimetry/CalorimeterHitAttenuation_factory.h`
* `src/services/io/podio/JEventProcessorPODIO.cc`
* `src/detectors/BEMC/BEMC.cc`
* `src/algorithms/calorimetry/CalorimeterHitAttenuationConfig.h`
* `src/algorithms/calorimetry/CalorimeterHitAttenuation.h`
* `src/algorithms/calorimetry/CalorimeterHitAttenuation.cc`

ruse-traveler:
> The following codes have been updated for the final review. The top comments that present the net information have also been updated.
> 
> * `src/factories/calorimetry/CalorimeterHitAttenuation_factory.h`
> * `src/services/io/podio/JEventProcessorPODIO.cc`
> * `src/detectors/BEMC/BEMC.cc`
> * `src/algorithms/calorimetry/CalorimeterHitAttenuationConfig.h`
> * `src/algorithms/calorimetry/CalorimeterHitAttenuation.h`
> * `src/algorithms/calorimetry/CalorimeterHitAttenuation.cc`

BTW don't forget to mark this as ready for review!

mhkim-anl:
Referring to Derek's comments and after discussing with Sylvester and Derek, the `CalorimeterHitAttenuation` has been updated so that it also works for a detector that only needs the regrouping. Since the algorithm works for `SimCalorimeterHit` and has more than one function (regroup, merge, and attenuation) the name has been changed to `SimCalorimeterHitProcessor`. You can find more details on the top explanation. Now, this PR is ready for review.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,189 @@
...
+					out_hit_contrib.setParticle(par);
+
+					auto out_hit = out_hits->create();
+					out_hit.setCellID(leading_hit.getCellID());
+					out_hit.setEnergy(static_cast<float>(edepSum*attFactor));
+					out_hit.setPosition(leading_hit.getPosition());
+					out_hit.addToContributions(out_hit_contrib);
+				}
+			}
+			else{
```
ruse-traveler, line 152:
```suggestion
			} else {
```

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,189 @@
...
+					uint64_t hid = ahit.getCellID() & m_id_mask;
+
+					trace("org cell ID in {:s}: {:#064b}", m_cfg.readout, ahit.getCellID());
+					trace("new cell ID in {:s}: {:#064b}", m_cfg.readout, hid);
+
+					merge_map[hid].push_back(ix);
+					ix++;
+				}
+
+				for(const auto &[id, ixs] : merge_map){
```
ruse-traveler, line 115:
```suggestion
				for (const auto &[id, ixs] : merge_map) {
```

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,189 @@
...
+					float edepSum = 0;
+					float timeEar = std::numeric_limits<double>::max();
+					auto leading_hit = hits[ixs[0]];
+					auto leading_contrib = hits[ixs[0]].getContributions(0);
+
+					for(size_t i=0; i<ixs.size(); ++i){
+						auto hit = hits[ixs[i]];
+						edepSum += hit.getEnergy();
+
+						for (const auto& c : hit.getContributions()){
```
ruse-traveler, line 125:
```suggestion
						for (const auto& c : hit.getContributions()) {
```

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,189 @@
...
+					ix++;
+				}
+
+				for(const auto &[id, ixs] : merge_map){
+					float edepSum = 0;
+					float timeEar = std::numeric_limits<double>::max();
+					auto leading_hit = hits[ixs[0]];
+					auto leading_contrib = hits[ixs[0]].getContributions(0);
+
+					for(size_t i=0; i<ixs.size(); ++i){
```
ruse-traveler, line 121:
```suggestion
					for (size_t i=0; i<ixs.size(); ++i) {
```

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,189 @@
...
+
+		const auto [in_hits] = input;
+		auto [out_hits, out_hit_contribs] = output;
+
+		// map for regrouping
+		std::map<edm4hep::MCParticle, std::vector<edm4hep::SimCalorimeterHit>> mapMCParToSimCalHit;
+
+		// regroup the sim hits by mc particle
+		for(const auto &ih : *in_hits){
+			for(const auto &contrib : ih.getContributions()){
```
ruse-traveler, line 77:
```suggestion
		for (const auto &ih : *in_hits) {
			for (const auto &contrib : ih.getContributions()) {
```

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,189 @@
...
+                // get id_mask for adding up hits that have the same z-segmentation
+		if (!m_cfg.mergeField.empty()) {
+			is_merge = true;
+			id_inverse_mask |= m_id_spec.field(m_cfg.mergeField)->mask();
+			debug("ID mask in {:s}: {:#064b}", m_cfg.readout, m_id_mask);
+                }
+                m_id_mask = ~id_inverse_mask;
+
+		// get attenuationMode for attenuating hits
+		if(!m_cfg.attenuationField.empty()) {
```
ruse-traveler, line 60:
```suggestion
		if (!m_cfg.attenuationField.empty()) {
```

ruse-traveler (dismissed):
This is looking great! I have a couple extremely minor formatting suggestions, and there are some merge conflicts to fix. But beyond that I think this is ready to go!

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
wdconinc: Please rebase/merge updates and let the code formatter do its thing.

mhkim-anl:
Derek and Wouter, thanks for your comments. Referring to your comments, I've merged the latest changes from main into this branch and code formatting has also been done. Please let me know if any of these don't seem to be applied correctly. Thanks.

wdconinc:
Note the compilation failures in https://github.com/eic/EICrecon/actions/runs/15100378521/job/42440476764?pr=1783#step:7:688.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+  // get decoders
+  try {
+    m_id_spec = m_geo.detector()->readout(m_cfg.readout).idSpec();
+  } catch (...) {
+    debug("Failed to load ID decoder for {}", m_cfg.readout);
+    throw std::runtime_error(fmt::format("Failed to load ID decoder for {}", m_cfg.readout));
+  }
+
+  uint64_t id_inverse_mask = 0;
+  // get id_mask for adding up hits that have the same z-segmentation
```
wdconinc, line 51: Remove references to "z-segmentation" in comments and code. This class is supposed to be generic. so this should refer generically to "dimensions that are merged over" or something.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+  if (!m_cfg.mergeField.empty()) {
+    is_merge = true;
+    id_inverse_mask |= m_id_spec.field(m_cfg.mergeField)->mask();
+    debug("ID mask in {:s}: {:#064b}", m_cfg.readout, m_id_mask);
```
wdconinc, line 55: How can this print `m_id_mask` when it is only set outside and after this conditional block?

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+  }
+  m_id_mask = ~id_inverse_mask;
+
+  // get attenuationMode for attenuating hits
+  if (!m_cfg.attenuationField.empty()) {
```
wdconinc, line 60: In DD4hep, "fields" are referring to parts of the CellID (i.e. segmentation) bit patternl. Here you are using "field" to refer to a distance. The name of the config should then make clear it is an attenuation length.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+    is_attenuation    = true;
+    m_attenuationMode = m_geo.detector()->constant<double>(m_cfg.attenuationField) / dd4hep::mm;
```
wdconinc, line 62: You are dividing out dimensionality without making it clear in the variable that this is a number in mm. If this is to move from one unit system (DD4hep) to another unit system (edm4hep), there should be a corresponding multiplication, so it looks like `* edm4hep::mm / dd4hep::mm`.

wdconinc: This variable is also confusingly named. This is not a "mode", it is again a distance.

wdconinc: Finally, if this is an optional quantity, it should be `std::optional` and avoid the `is_attenuation` companion boolean.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+    for (const auto& contrib : ih.getContributions()) {
+      edm4hep::MCParticle primary = get_primary(contrib);
+
+      edm4hep::MutableSimCalorimeterHit simhit;
+      simhit.setCellID(ih.getCellID());
+      simhit.setEnergy(contrib.getEnergy());
+      simhit.setPosition(ih.getPosition());
+      simhit.addToContributions(contrib);
+
+      mapMCParToSimCalHit[primary].push_back(simhit);
```
wdconinc, line 86: Instead of creating a temporary and copying it into the map, it makes more sense to `emplace_back` right away.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+
+  const auto [in_hits]              = input;
+  auto [out_hits, out_hit_contribs] = output;
+
+  // map for regrouping
+  std::map<edm4hep::MCParticle, std::vector<edm4hep::SimCalorimeterHit>> mapMCParToSimCalHit;
+
+  // regroup the sim hits by mc particle
+  for (const auto& ih : *in_hits) {
+    for (const auto& contrib : ih.getContributions()) {
```
wdconinc, line 77: This grows potentially large. Can you provide some comment on how large you expect this to grow in worst case conditions? Does it affect memory limits on our running of reconstruction?

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
wdconinc: As below, this may also be a place to use `std::optional`.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+        trace("new cell ID in {:s}: {:#064b}", m_cfg.readout, hid);
+
+        merge_map[hid].push_back(ix);
+        ix++;
+      }
+
+      for (const auto& [id, ixs] : merge_map) {
+        float edepSum        = 0;
+        float timeEar        = std::numeric_limits<double>::max();
+        auto leading_hit     = hits[ixs[0]];
```
wdconinc, line 116: Is the presence of element 0 guaranteed?

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+
+        for (size_t i = 0; i < ixs.size(); ++i) {
+          auto hit = hits[ixs[i]];
+          edepSum += hit.getEnergy();
+
+          for (const auto& c : hit.getContributions()) {
+            if (c.getTime() <= timeEar) {
+              timeEar = c.getTime();
+            }
+          }
```
wdconinc, line 127: Can we use a standard min/max algorithm here?

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+
+      for (const auto& [id, ixs] : merge_map) {
+        float edepSum        = 0;
+        float timeEar        = std::numeric_limits<double>::max();
+        auto leading_hit     = hits[ixs[0]];
+        auto leading_contrib = hits[ixs[0]].getContributions(0);
+
+        for (size_t i = 0; i < ixs.size(); ++i) {
+          auto hit = hits[ixs[i]];
+          edepSum += hit.getEnergy();
```
wdconinc, line 121: Can we use an accumulate or reduce algorithm here?

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+        ix++;
+      }
+
+      for (const auto& [id, ixs] : merge_map) {
+        float edepSum        = 0;
+        float timeEar        = std::numeric_limits<double>::max();
+        auto leading_hit     = hits[ixs[0]];
+        auto leading_contrib = hits[ixs[0]].getContributions(0);
+
+        for (size_t i = 0; i < ixs.size(); ++i) {
```
wdconinc, line 119: Can we use a range-for loop?

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,190 @@
...
+        out_hit_contrib.setTime(contrib.getTime());
+        out_hit_contrib.setStepPosition(contrib.getStepPosition());
+        out_hit_contrib.setParticle(par);
+
+        auto out_hit = out_hits->create();
+        out_hit.setCellID(hit.getCellID());
+        out_hit.setEnergy(hit.getEnergy());
+        out_hit.setPosition(hit.getPosition());
+        out_hit.addToContributions(out_hit_contrib);
+      }
```
wdconinc, line 166: This looks like code duplication that should be avoided.

src/algorithms/calorimetry/SimCalorimeterHitProcessorConfig.h
```diff
@@ -0,0 +1,22 @@
...
+
+struct SimCalorimeterHitProcessorConfig {
+
+  // parameters for attenuation function
+  std::vector<double> attPars;
+
+  // fields for adding up energies and attenuate them
+  std::string readout{""};
+  std::string attenuationField{""};
+  std::string mergeField{""};
```
wdconinc, line 19: As describe above, rename these config fields.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.h
```diff
@@ -0,0 +1,63 @@
...
+
+private:
+  uint64_t m_id_mask{0};
+
+  dd4hep::IDDescriptor m_id_spec;
+
+  const algorithms::GeoSvc& m_geo = algorithms::GeoSvc::instance();
+
+  // a reference value for attenuation
+  double m_attenuationMode;
```
wdconinc, line 50: As described above, rename this member field.

src/detectors/BEMC/BEMC.cc
```diff
@@ -28,13 +30,39 @@ void InitPlugin(JApplication* app) {
...
   decltype(CalorimeterHitDigiConfig::resolutionTDC) EcalBarrelScFi_resolutionTDC =
       10 * dd4hep::picosecond;
+  app->Add(new JOmniFactoryGeneratorT<SimCalorimeterHitProcessor_factory>(
+      "EcalBarrelScFiPAttenuatedHits", {"EcalBarrelScFiHits"},
+      {"EcalBarrelScFiPAttenuatedHits", "EcalBarrelScFiPAttenuatedHitContributions"},
+      {
+          .attPars          = EcalBarrelScFi_attPars,
+          .readout          = "EcalBarrelScFiHits",
+          .attenuationField = "EcalBarrel_Readout_zmax",
+          .mergeField       = "fiber",
```
wdconinc, line 51: How does merging by `fiber` (instead of in the digitization algorithm that follows it) affect default running?

src/detectors/BEMC/BEMC.cc
```diff
@@ -28,13 +30,39 @@ void InitPlugin(JApplication* app) {
 
   InitJANAPlugin(app);
 
+  // Make sure left and right use the same value
+  decltype(SimCalorimeterHitProcessorConfig::attPars) EcalBarrelScFi_attPars = {0.416212, 747.39875,
```
wdconinc, line 34: Some of the attenuation parameters must be dimensional of some sort. Please add dimensions here. From the formula it seems that `[0]` is dimensionless, but `[1]` and `[2]` are `1/dd4hep::mm` presumably.

src/factories/calorimetry/SimCalorimeterHitProcessor_factory.h
```diff
@@ -0,0 +1,48 @@
...
+  void Configure() {
+    m_algo = std::make_unique<AlgoT>(GetPrefix());
+    m_algo->level(static_cast<algorithms::LogLevel>(logger()->level()));
+    m_algo->applyConfig(config());
+    m_algo->init();
+  }
+
+  void ChangeRun(int64_t run_number) {}
+
+  void Process(int64_t run_number, uint64_t event_number) {
```
wdconinc, line 43: Comment out unused arguments. Use correct type.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
mhkim-anl: Thanks for pointing out my incorrect naming. That variable will be used to get a reference position, e.g., z position of the detector edge, for attenuation. How about naming it "attenuationReference"?

wdconinc: It would be great if the name could refer explicitly to "distance" or "position", so we know it has dimensions of length.

mhkim-anl: @wdconinc Got you. Because the `m_attenuationMode` receives the actual reference position for attenuation, e.g., 191.25*cm, and the `attenuationField` is a name argument to get that value, e.g., "EcalBarrel_Readout_zmax", I changed those names as follows.

`m_attenuationMode` --> `m_attenuationReferencePosition`
`attenuationField` --> `attenuationReferencePositionName`

Please let me know if the name is too long or confusing.

src/algorithms/calorimetry/SimCalorimeterHitProcessorConfig.h
mhkim-anl: @wdconinc Regarding the `mergeField`, it'll be referring to parts of the CellID. For BIC, it will refer to "fiber" to merge hits. In this case, can I leave that variable name as it is?

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
mhkim-anl: @wdconinc From what I understand, using `emplace_back` can't add "contrib" which is `CaloHitContribution` data. Please correct me if I misunderstood.

wdconinc: `emplace_back` returns a reference.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
mhkim-anl: <pre>
\```cpp
for(const auto& ahit : hits){
  uint64_t hid = ahit.getCellID() & m_id_mask;
  merge_map[hid].push_back(ix);
}
\```
</pre>

src/detectors/BEMC/BEMC.cc
mhkim-anl: @wdconinc I'm sorry, I didn't fully catch your question. Could you explain a bit more detail?

wdconinc: I think I was assuming this was included in the default analysis chain, but at this point it is still separate.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
mhkim-anl: @wdconinc The below part
```cpp
m_geo.detector()->constant<double>(m_cfg.attenuationField)
```
gives a cm unit value for BIC, so I tried to convert it to a mm value. I added this motivation as a comment. Please let me know if I misunderstood something.

The type has been changed to `std::optional<double>` and the `is_attenuation` has been removed.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
mhkim-anl: @wdconinc I used the above structure because the numbers of `out_hit` and `out_hit_contrib` and their elements are different depending on whether the `m_id_mask` has a value or not. So it's not a code duplication. Please correct me if I misunderstood some point.

src/detectors/BEMC/BEMC.cc
mhkim-anl: @wdconinc Thanks for your additional comment. The reason the two parts (P and N AttenuatedHits) are separated is because pulses on both sides will be used for realistic energy reconstruction.

src/algorithms/calorimetry/SimCalorimeterHitProcessorConfig.h
wdconinc: Yes, `mergeField` is fine. Here it does refer to the fields of the CellID.

mhkim-anl: Thanks for the confirmation. The `attenuationField` has been changed to `attenuationReferencePositionName` above.

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
mhkim-anl: @wdconinc Declarations of the attenuation-related parameters and variables have been changed so that they follow the same unit system, `edm4eic::unit::mm`. The `m_attenuationReferencePosition` has been changed to `m_attenuationReferencePosition_mm`. Please correct me if I misunderstood something.

src/algorithms/calorimetry/SimCalorimeterHitProcessorConfig.h
```diff
@@ -0,0 +1,24 @@
...
+struct SimCalorimeterHitProcessorConfig {
+
+  // parameters for attenuation function
+  std::vector<double> attPars;
+
+  std::string readout{""};
+  std::string attenuationReferencePositionName{""};
+  // fields for merging hits
+  std::vector<std::string> hitMergeFields{};
+  // fildes for merging contributions
```
wdconinc, line 20:
```suggestion
  // fields for merging contributions
```

src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
```diff
@@ -0,0 +1,236 @@
...
+  float m_min_time{std::numeric_limits<float>::max()};
+  edm4hep::Vector3f m_avg_position{0, 0, 0};
+
+public:
+  void add(const float energy, const float time, const edm4hep::Vector3f& pos) {
+    m_energy += energy;
+    m_avg_time += energy * time;
+    m_avg_position.x += energy * pos.x;
+    m_avg_position.y += energy * pos.y;
+    m_avg_position.z += energy * pos.z;
```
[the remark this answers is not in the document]
sly2j: I'll address this - I forgot that I wrote those library calls :). Also, while I agree it's good to keep in mind numerical stability and dynamic range, in this particular case it is fine to be dominated by high energy hits as our downstream use (i.e. for signal timing) only needs to be accurate up to the ~ cm level.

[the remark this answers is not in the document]
sly2j: addressed in [69b6100](https://github.com/eic/EICrecon/pull/1783/commits/69b6100610ecb999d8650800a9b40d108eb3df31)

wdconinc (approved):

PR #1789, opened 2025-04-12
feat: add duplicate filter on log services (30s timeout)

wdconinc:
This PR adds a duplicate message filter on the logger output. If subsequent identical messages are emitted to the same logger instance within a timeout window (30 seconds by default, settable) they are eaten and replaced with a message `Skipped %u duplicate messages..` (at info level, regrettably not possible at the level of the original message with the spdlog classes though we could extend the dup_filter to achieve that I guess).

- [ ] Bug fix (issue #__)
- [x] New feature (issue: logger duplicate filtering)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, less logger noise.


veprbl:
I don't like the idea of irreproducible log outputs. We should fix the info messages regardless. cc @simonge.

PR #1788, opened 2025-04-12
fix: correct CalorimeterClusterRecoCoG inputs by EDM4eic version

wdconinc:
### Briefly, what does this PR introduce?
This PR fixes the inputs used for various CalorimeterClusterRecoCoG factories. Depending on the EDM4eic version, the CalorimeterClusterRecoCoG factory uses different inputs, so we need to respect that.

### What kind of change does this PR introduce?
- [x] Bug fix (issue: exceptions in CalorimeterClusterRecoCoG factories)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators
    - @veprbl please evaluate for need for backporting into 1.24.1 and 25.04.1

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
Yes, will again produce outputs that are currently missing.

PR #1790, opened 2025-04-12
feat: log error with algorithm exception before throwing to JANA

wdconinc:
This PR adds printing of exceptions from algorithms to the log stream as errors. Likely to be useful with #1789 since it would otherwise be overwhelming. This will help avoid issues like https://github.com/eic/EICrecon/pull/1762#issuecomment-2798888885, https://github.com/eic/EICrecon/pull/1788, and https://github.com/eic/EICrecon/pull/1795.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: exceptions are silently ignored)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, prints exceptions. Likely a lot for simulation input that does not contain default collections.

### DIFF
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -553,6 +553,7 @@ public:
         output->SetCollection(*this);
       }
     } catch (std::exception& e) {
+      m_logger->error(e.what());
       throw JException(e.what());
     }
   }


wdconinc:
Without log duplicate filter:
image

wdconinc:
This should now already produce a lot fewer false positives. At least for DIS. There's still the OMD which is only defined properly for 130 GeV protons, which we never simulate...

PR #1795, opened 2025-04-13
fix: if no hadronic final state, return instead of throw

wdconinc:
This PR fixes the inclusive kinematics algorithms that need the hadronic final state. Until now these algorithms would assume that `at(0)` returns a valid entry without testing the size first. Now the size is tested and the algorithm returns before throwing an exception in `at(0)`.

See e.g. https://github.com/eic/EICrecon/actions/runs/14424316365/job/40450940599?pr=1790#step:7:1090.

- [x] Bug fix (issue: inclusive kinematics algorithms throw exceptions)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No. Return without action is equivalent to silently ignored exception right now.

### DIFF
--- a/src/algorithms/reco/InclusiveKinematicsDA.cc
+++ b/src/algorithms/reco/InclusiveKinematicsDA.cc
@@ -70,6 +70,10 @@ namespace eicrecon {
     auto theta_e = e_boosted.Theta();
 
     // Get hadronic final state variables
+    if (hfs->size() == 0) {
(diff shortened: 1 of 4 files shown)


veprbl (approved):
Looks fine. `debug` log level is consistent with the "No beam hadron found" error message that can be triggered in the same codes.

PR #1798, opened 2025-04-14
add .git-blame-ignore-revs

veprbl:

To use this, one either has to pass `--ignore-revs-file` or configure (use `--global` as needed):

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


src/global/reco/reco.cc
wdconinc, line 1: Do you want to spin this change off into its own PR and add it to the ignore revs as well?

wdconinc (approved):

PR #1799, opened 2025-04-14
treewide: run clang-format with the new config

veprbl:
A follow up to #1692. It looks like pre-commit-hook CI doesn't apply new config. This was generated using
```
clang-format -i $(git ls-tree -r HEAD --name-only | grep -E "\.(cc|h)$")
```


veprbl:
Okay, hypothesis proven wrong. Closing in favor of #1800.

PR #1803, opened 2025-04-15
fix: `ReconstructedChargedParticleIDs` (unknown) -> `ReconstructedChargedRealPIDParticleIDs` in output collections

wdconinc:
EICrecon tries to write `ReconstructedChargedParticleIDs` to the output file, but that's not a known collection. Remove it from the `output_collections` default value.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -180,7 +180,6 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
     "ReconstructedChargedParticleAssociations",
     "MCScatteredElectronAssociations",    // Remove if/when used internally
     "MCNonScatteredElectronAssociations", // Remove if/when used internally
-    "ReconstructedChargedParticleIDs",
     "ReconstructedBreitFrameParticles",
     "CentralTrackSegments",
     "CentralTrackVertices",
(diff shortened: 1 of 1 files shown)


src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -180,7 +180,6 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
     "ReconstructedChargedParticleAssociations",
     "MCScatteredElectronAssociations",    // Remove if/when used internally
     "MCNonScatteredElectronAssociations", // Remove if/when used internally
-    "ReconstructedChargedParticleIDs",
```
veprbl, line 183: Should we add "ReconstructedChargedRealPIDParticleIDs" instead?
ref. 8f66e97c7ad8c8d9f57dc60cbf1332b95d9d73cd

veprbl: there are still ReconstructedChargedRealPIDParticles above, after all

wdconinc: Sure, it was hard to determine the motivation for why that was there. If that's the reason and it just wasn't changed, then we should indeed add it.

veprbl: Relations to ParticleIDs would be dangling currently. If we export `ReconstructedChargedRealPIDParticles`, ParticleIDs are probably the most important piece.

The `ParticleIDs` for LUT PID are currently spread across things like `DRICHParticleIDs` and other collections, and that's why `ReconstructedChargedParticleIDs` is currently not a thing.

veprbl (approved):
LGTM

PR #1807, opened 2025-04-16
feat: separate PhotoMultiplierHitDigi readout/detector/plugin names

wdconinc:
The PhotoMultiplierHitDigi factory (and algorithm, to lesser extent) assume that plugin name == detector name, and readout == plugin name + "Hits". That's not true for the PFRICH plugin, RICHEndcapN detector, RICHEndcapNHits readout. 

This PR disentangles and corrects the detector and readout passed for the PFRICH, avoiding an exception in the RichGeo service.

Note that this doesn't leave everything in a perfect state (the configuration lambdas that are being set in the factory are non-ideal), but it's the minimum amount of work to sanitize this to avoid the exception and get the PFRICH to get digitized again (at least in theory; I don't know if the PFRICH hits are actually stored).

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/global/digi/PhotoMultiplierHitDigi_factory.h
+++ b/src/global/digi/PhotoMultiplierHitDigi_factory.h
@@ -65,14 +65,20 @@ public:
 
     // Initialize richgeo ReadoutGeo and set random CellID visitor lambda (if a RICH)
     if (GetPluginName() == "DRICH" || GetPluginName() == "PFRICH") {
-      m_RichGeoSvc().GetReadoutGeo(GetPluginName())->SetSeed(config().seed);
+      m_RichGeoSvc()
+          .GetReadoutGeo(config().detectorName, config().readoutClass)
+          ->SetSeed(config().seed);
       m_algo->SetVisitRngCellIDs(
           [this](std::function<void(PhotoMultiplierHitDigi::CellIDType)> lambda, float p) {
-            m_RichGeoSvc().GetReadoutGeo(GetPluginName())->VisitAllRngPixels(lambda, p);
+            m_RichGeoSvc()
+                .GetReadoutGeo(config().detectorName, config().readoutClass)
+                ->VisitAllRngPixels(lambda, p);
           });
       m_algo->SetPixelGapMask(
           [this](PhotoMultiplierHitDigi::CellIDType cellID, dd4hep::Position pos) {
-            return m_RichGeoSvc().GetReadoutGeo(GetPluginName())->PixelGapMask(cellID, pos);
+            return m_RichGeoSvc()
+                .GetReadoutGeo(config().detectorName, config().readoutClass)
+                ->PixelGapMask(cellID, pos);
           });
     }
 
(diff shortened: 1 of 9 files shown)


src/services/geometry/richgeo/ReadoutGeo.cc
```diff
@@ -24,10 +24,11 @@
 #include "services/geometry/richgeo/RichGeo.h"
 
 // constructor
-richgeo::ReadoutGeo::ReadoutGeo(std::string detName_, gsl::not_null<const dd4hep::Detector*> det_,
+richgeo::ReadoutGeo::ReadoutGeo(std::string detName_, std::string readoutClass_,
```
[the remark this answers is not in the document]
wdconinc: Out of scope. Issue filed for follow-up.

src/global/digi/PhotoMultiplierHitDigi_factory.h
```diff
@@ -63,14 +65,20 @@ class PhotoMultiplierHitDigi_factory
...
+                ->VisitAllRngPixels(lambda, p);
           });
       m_algo->SetPixelGapMask(
           [this](PhotoMultiplierHitDigi::CellIDType cellID, dd4hep::Position pos) {
-            return m_RichGeoSvc().GetReadoutGeo(GetPluginName())->PixelGapMask(cellID, pos);
+            return m_RichGeoSvc()
+                .GetReadoutGeo(config().detectorName, config().readoutClass)
+                ->PixelGapMask(cellID, pos);
           });
     }
```
simonge, line 83: Is there a reason this can't/shouldn't be in the algorithm initialization?

simonge:
DRICHRawHits seem to have gone missing.

src/global/digi/PhotoMultiplierHitDigi_factory.h
wdconinc: That requires a more thorough rewrite of the RICH algorithms and services, I'm afraid. The short term goal here is to fix the exceptions thrown by the incorrect names.

wdconinc:
> DRICHRawHits seem to have gone missing.

That was indeed not the goal.

src/services/geometry/richgeo/RichGeo_service.cc
```diff
@@ -76,15 +73,16 @@ richgeo::ActsGeo* RichGeo_service::GetActsGeo(std::string detector_name) {
...
-    auto initialize = [this, &detector_name]() {
+    auto initialize = [this, &detector_name, &readout_class]() {
       if (!m_dd4hepGeo)
         throw JException("RichGeo_service m_dd4hepGeo==null which should never be!");
-      m_readoutGeo =
-          std::make_shared<richgeo::ReadoutGeo>(detector_name, m_dd4hepGeo, m_converter, m_log);
+      m_readoutGeo = std::make_shared<richgeo::ReadoutGeo>(detector_name, readout_class,
+                                                           m_dd4hepGeo, m_converter, m_log);
     };
     std::call_once(m_init_readout, initialize);
```
wdconinc, line 87: So... making the PFRICH work made the DRICH broken because of this `call_once`. Only 1 readout geo per RIchGeo_service...

simonge (approved):
Looks good now.

Is the lock_guard necessary? I thought the call_once would provide the necessary protections?

wdconinc:
> Looks good now.
> 
> Is the lock_guard necessary? I thought the call_once would provide the necessary protections?

The lock_guard protects the map, not the once_flags in it, before the once_flag is passed to the call_once.

PR #1810, opened 2025-04-17
Adopt edm4eic::SimPulse in for current pulse algorithms

simonge:
Allows the new SimPulse to be used by the current developing pulse algorithms

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Outputs edm4eic::SimPulse rather than edm4hep::TimeSeries when the edm4eic version is >=8.1

### DIFF
--- a/src/algorithms/digi/PulseCombiner.cc
+++ b/src/algorithms/digi/PulseCombiner.cc
@@ -73,15 +73,14 @@ void PulseCombiner::process(const PulseCombiner::Input& input,
 #if EDM4EIC_VERSION_MAJOR >= 8 && EDM4EIC_VERSION_MINOR >= 1
         // Sum the pulse array
         float integral = std::accumulate(newPulse.begin(), newPulse.end(), 0.0f);
-        sum_pulse.setIntegral(integral);  
-        sum_pulse.setPosition(edm4hep::Vector3f(cluster[0].getPosition().x,
-                                                cluster[0].getPosition().y,
-                                                cluster[0].getPosition().z));
+        sum_pulse.setIntegral(integral);
+        sum_pulse.setPosition(edm4hep::Vector3f(
+            cluster[0].getPosition().x, cluster[0].getPosition().y, cluster[0].getPosition().z));
         for (auto pulse : cluster) {
           sum_pulse.addToPulses(pulse);
           for (auto particle : pulse.getParticles()) {
             sum_pulse.addToParticles(particle);
-          }         
+          }
           // Not sure if we want/need to keep the hits themselves at this point?
           for (auto hit : pulse.getTrackerHits()) {
             sum_pulse.addToTrackerHits(hit);
@@ -91,7 +90,6 @@ void PulseCombiner::process(const PulseCombiner::Input& input,
           }
         }
 #endif
-
       }
       debug("CellID {} has {} pulses, combined into {} clusters", cellID, pulses.size(),
             clusters.size());
--- a/src/algorithms/digi/PulseCombiner.cc
+++ b/src/algorithms/digi/PulseCombiner.cc
@@ -3,12 +3,27 @@
 //
 // Combine pulses into a larger pulse if they are within a certain time of each other
 
+#include <DD4hep/Detector.h>
+#include <DD4hep/IDDescriptor.h>
+#include <DD4hep/Readout.h>
+#include <DDSegmentation/BitFieldCoder.h>
+#include <algorithms/geo.h>
+#include <edm4hep/MCParticle.h>
+#include <edm4hep/SimCalorimeterHit.h>
+#include <edm4hep/SimTrackerHit.h>
+#include <edm4hep/Vector3f.h>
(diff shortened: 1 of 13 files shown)


src/algorithms/digi/PulseNoise.cc
```diff
@@ -27,11 +27,32 @@ void PulseNoise::process(const PulseNoise::Input& input, const PulseNoise::Outpu
...
       out_pulse.addToAmplitude(pulse.getAmplitude()[i] + noise);
+      integral += pulse.getAmplitude()[i] + noise;
     }
+
+#if EDM4EIC_VERSION_MAJOR >= 8 && EDM4EIC_VERSION_MINOR >= 1
+    out_pulse.setIntegral(integral);
+    out_pulse.setPosition(pulse.getPosition());
+    for (auto subpulse : pulse.getPulses()) {
+      out_pulse.addToPulses(subpulse);
+    }
```
simonge, line 43: I need another opinion on whether this relationship should just include the pulse that noise was added to, or all of the pulses which formed that pulse.

veprbl: I would just point to the pulses from the input collections, but this is fine too.

src/algorithms/digi/PulseNoise.cc
```diff
@@ -27,11 +27,32 @@ void PulseNoise::process(const PulseNoise::Input& input, const PulseNoise::Outpu
     out_pulse.setInterval(pulse.getInterval());
     out_pulse.setTime(pulse.getTime());
 
+    float integral = 0;
     //Add noise to the pulse
     for (int i = 0; i < pulse.getAmplitude().size(); i++) {
       double noise = m_noise(generator) * m_cfg.scale;
       out_pulse.addToAmplitude(pulse.getAmplitude()[i] + noise);
+      integral += pulse.getAmplitude()[i] + noise;
```
veprbl, line 35:
```suggestion
      double amplitude = pulse.getAmplitude()[i] + noise;
      out_pulse.addToAmplitude(amplitude);
      integral += amplitude;
```

src/algorithms/digi/PulseCombiner.cc
```diff
@@ -69,6 +69,27 @@ void PulseCombiner::process(const PulseCombiner::Input& input,
         for (auto pulse : newPulse) {
           sum_pulse.addToAmplitude(pulse);
         }
+
+#if EDM4EIC_VERSION_MAJOR >= 8 && EDM4EIC_VERSION_MINOR >= 1
```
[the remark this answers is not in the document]
simonge: Ah yes. Thanks.

src/algorithms/digi/PulseNoise.cc
simonge: Ok, swapped to that. Should the hits and particle relations both be kept too? The pulse relations is the only one which is really an either or.

veprbl: I would keep those. It doesn't cost us anything to carry them over, and we'd need to add logic to recursively traverse them otherwise.

src/algorithms/digi/PulseCombiner.cc
```diff
@@ -69,6 +69,27 @@ void PulseCombiner::process(const PulseCombiner::Input& input,
...
+        float integral = std::accumulate(newPulse.begin(), newPulse.end(), 0.0f);
+        sum_pulse.setIntegral(integral);
+        sum_pulse.setPosition(edm4hep::Vector3f(
+            cluster[0].getPosition().x, cluster[0].getPosition().y, cluster[0].getPosition().z));
+        for (auto pulse : cluster) {
+          sum_pulse.addToPulses(pulse);
+          for (auto particle : pulse.getParticles()) {
+            sum_pulse.addToParticles(particle);
+          }
+          // Not sure if we want/need to keep the hits themselves at this point?
```
veprbl, line 84:
```suggestion
```

veprbl:
Does this break TOFBarrelADCTDC? I see
image
The TaggerTrackerHitPulses look fine though.

veprbl:
Okay, I've downloaded artifacts manually using `capybara capy 1810`, it looks like the reference file still contains the collection, however it has no entries:
image
I don't have a reasonable explanation for this, but it seems like this is a regression on `main`. And it looks like `capybara bara` will not render comparisons where both collections are empty.

> > Does this break TOFBarrelADCTDC?
> 
> I can't find TOFBarrelADCTDC in any of the capybara reports. What is capybara comparing against? It seems to sometimes not to be the current main branch and instead the eic shell local or similar?

Actually, for GitHub we only download artifacts for main and for the PR, there is no path to execute eicrecon from main implemented. This is how that works:
https://github.com/eic/EICrecon/blob/1d951fd5735020417cf6d6ec0c1881b1f18640b2/.github/workflows/linux-eic-shell.yml#L732-L755

veprbl (approved):
Capybara reports look good, diff looks good..

PR #1811, opened 2025-04-17
fix: avoid geoManager->GetCurrentMatrix in LGADChargeSharing::_global2Local

wdconinc:
This PR avoids a thread-unsafe geoManager->GetCurrentMatrix() call which relies on cached state in the geoManager object.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/digi/LGADChargeSharing.cc
+++ b/src/algorithms/digi/LGADChargeSharing.cc
@@ -69,7 +69,8 @@ void LGADChargeSharing::process(const LGADChargeSharing::Input& input,
   auto [sharedHits]    = output;
 
   for (const auto& hit : *simhits) {
-    auto cellID = hit.getCellID();
+    auto cellID         = hit.getCellID();
+    const auto* context = m_converter->findContext(cellID);
 
     std::unordered_set<dd4hep::rec::CellID> dp;
     std::vector<dd4hep::rec::CellID> neighbors;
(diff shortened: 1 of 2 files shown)


src/algorithms/digi/LGADChargeSharing.cc
```diff
@@ -163,18 +163,18 @@ double LGADChargeSharing::_integralGaus(double mean, double sd, double low_lim,
 }
 
 dd4hep::Position LGADChargeSharing::_cell2LocalPosition(const dd4hep::rec::CellID& cell) const {
```
veprbl, line 165: @ssedd1123 Did this and `_global2Local` come from some other place?

ssedd1123: Do you mean if I use `_global2Local ` in any other classes? I don't think so. If you are asking where I get the inspiration for these codes, I get it by imitating the source code of `CellIDPositionConverter`.

Line 97 - 101 of https://github.com/AIDASoft/DD4hep/blob/ccb7365d16c219c18a2b345551bf8d02f5797662/DDRec/src/CellIDPositionConverter.cpp#L101 is where I learn how to convert between local and global coordinates.

veprbl: If it was taken from somewhere else, we'd want to have it fixed there as well. It looks like we can remove these functions and just use `CellIDPositionConverter::positionNominal`?

wdconinc:
Rebased, conflicts resolved.

veprbl (approved):
Diff LGTM, capybara report looks clean

PR #1813, opened 2025-04-18
fix: iwyu treewide after fixing iwyu.imp refs

wdconinc:
This PR applies IWYU treewide after #1805, to reduce the number of include changes unrelated to changed code that will be suggested on new PRs.

- [x] Bug fix (issue: #1805 followup)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -10,6 +10,7 @@
 #include <Acts/EventData/GenericBoundTrackParameters.hpp>
 #include <Acts/EventData/TrackStateProxy.hpp>
 #include <Acts/EventData/Types.hpp>
+#include <sys/types.h>
 #if Acts_VERSION_MAJOR < 36
 #include <Acts/EventData/Measurement.hpp>
 #endif
@@ -59,7 +60,6 @@
 #include <ActsExamples/EventData/MeasurementCalibration.hpp>
 #include <ActsExamples/EventData/Track.hpp>
 #include <boost/container/vector.hpp>
-#include <boost/move/utility_core.hpp>
 #include <edm4eic/Cov3f.h>
 #include <edm4eic/Cov6f.h>
 #include <edm4eic/Measurement2DCollection.h>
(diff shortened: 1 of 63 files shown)


src/algorithms/tracking/CKFTracking.cc
```diff
@@ -10,6 +10,7 @@
 #include <Acts/EventData/GenericBoundTrackParameters.hpp>
 #include <Acts/EventData/TrackStateProxy.hpp>
 #include <Acts/EventData/Types.hpp>
+#include <sys/types.h>
```
veprbl, line 13: This is not portable. Should not go in, in my opinion.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -8,13 +8,14 @@
...
 #include <Acts/Utilities/Result.hpp>
-
 #include <Evaluator/DD4hepUnits.h>
 #include <edm4hep/Vector3d.h>
 #include <edm4hep/Vector3f.h>
 #include <edm4eic/Cov6f.h>
+#include <edm4hep/Vector3d.h>
 #include <fmt/core.h>
 #include <spdlog/common.h>
+#include <stdlib.h>
```
veprbl, line 18: cstdlib?

src/algorithms/tracking/CKFTracking.cc
wdconinc: This was suggested because of use of ssize_t, which is not a C++ type. Removed in 8b7d37b.

src/algorithms/tracking/TrackParamTruthInit.cc
wdconinc: Replaced abs with std::abs in 008cb96 for this reason.

wdconinc:
Note: I'm ignoring the [iwyu suggestions](https://github.com/eic/EICrecon/actions/runs/14550596667/job/40821670140?pr=1813) of removing Vector3f.h, since they are caused by e.g. `decltype(edm4hep::MCParticleData::momentum)`, so they are edm4hep version dependent.

veprbl (approved):
LGTM

PR #1814, opened 2025-04-20
fix: check scattered electron size in InclusiveKinematics*

wdconinc:
This PR applies #1795 to the scattered electron as well. This fixes e.g. https://github.com/eic/EICrecon/actions/runs/14553024479/job/40826864227?pr=975#step:7:3282.

- [x] Bug fix (issue: more exception thrown, e.g. https://github.com/eic/EICrecon/actions/runs/14553024479/job/40826864227?pr=975#step:7:3282)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/reco/InclusiveKinematicsDA.cc
+++ b/src/algorithms/reco/InclusiveKinematicsDA.cc
@@ -55,6 +55,10 @@ void InclusiveKinematicsDA::process(const InclusiveKinematicsDA::Input& input,
   auto boost = determine_boost(ei, pi);
 
   // Get electron angle
+  if (escat->size() == 0) {
+    debug("No scattered electron found");
+    return;
+  }
   auto kf = escat->at(0);
   PxPyPzEVector e_lab(kf.getMomentum().x, kf.getMomentum().y, kf.getMomentum().z, kf.getEnergy());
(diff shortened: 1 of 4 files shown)


wdconinc:
FYI, checks failing due to cvmfs sync issues.

wdconinc:
Yay, no diff.

wdconinc:
> Yay, no diff.

Wrong. Actually big differences in JB: see e.g. https://eicrecon.epic-eic.org/pr/1814/capybara/rec_dis_10x100_minQ2=0_craterlake/index.html#InclusiveKinematicsJB

Makes sense, I think. We are storing a scattered electron relation, but it is not a requirement for JB. So, previously it would fail for no reason. Now it doesn't fail and we get more JB measurements.

@tylerkutz Does JB require that we have an electron identified, even if we don't use its properties?

wdconinc:
Confirmation at https://chat.epic-eic.org/main/pl/5twkypozgfbkbg1hx4efaqw73a that electron not needed for JB.

mdiefent (approved):
The Jacques-Blondel method uses only measurements of the final state hadronic system. The method was developed for cases in which the scattered lepton cannot be detected. The proposed changes make sense.

PR #1815, opened 2025-04-21
feat: shortcut lambda if float in EvaluatorSvc

wdconinc:
This PR shortcuts the string evaluation to lambda in the CalorimeterHitDigi and CalorimeterHitReco when the correction mean scale or sampling fraction string is a simple float. This avoids having multiple slow evaluator calls in process during multithreaded initialization.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @veprbl 

No.

No.

### DIFF
--- a/src/services/evaluator/EvaluatorSvc.h
+++ b/src/services/evaluator/EvaluatorSvc.h
@@ -2,8 +2,10 @@
 // Copyright (C) 2024 Dmitry Kalinkin
 
 #include <algorithms/logger.h>
+#include <exception>
 #include <functional>
 #include <mutex>
+#include <stdexcept>
 #include <string>
 #include <unordered_map>
 #include <utility>
@@ -29,7 +31,7 @@ public:
 
   /**
    * @brief Compile expression `expr` to std::function
-   * @param expr String expression to compile (e.g. `"a + b"`)
+   * @param expr String expression to compile (e.g. `"0.998"`, `"a + b"`)
    * @param Args Types of arguments for the resulting function
    * @param transform Function providing mapping from Args
    *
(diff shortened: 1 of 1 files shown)


wdconinc:
Yay, no diffs.

src/algorithms/calorimetry/CalorimeterHitReco.cc
```diff
@@ -132,8 +133,17 @@ void CalorimeterHitReco::init() {
...
   };
 
-  auto& serviceSvc = algorithms::ServiceSvc::instance();
-  sampFrac = serviceSvc.service<EvaluatorSvc>("EvaluatorSvc")->compile(m_cfg.sampFrac, hit_to_map);
+  // convert sampFrac string to function that returns sampFrac value
+  try {
+    // trivial function if string is representation of float
+    float sampFrac_value = std::stof(m_cfg.sampFrac);
+    sampFrac = [sampFrac_value](const edm4hep::RawCalorimeterHit&) { return sampFrac_value; };
+  } catch (std::exception& e) {
```
veprbl, line 141: This is wrong. `std::stof("1+1")` will succeed to parse `1.0f` from `"1"` and leave out `"+1"`.

wdconinc: Good catch. See bab408f.

simonge:
Could this be part of the EvaluatorSvc itself rather than reimplemented in every algorithm which might need it?

wdconinc:
> Could this be part of the EvaluatorSvc itself rather than reimplemented in every algorithm which might need it?

Fair point. Shortcutting is now done in the EvaluatorSvc itself.

simonge (approved):
Looks good. A test which validates both options would be a nice addition.
@veprbl might want the last word, confirming the changes you made fix his issue.

wdconinc:
> Looks good. A test which validates both options would be a nice addition. @veprbl might want the last word, confirming the changes you made fix his issue.

Both are tested: "1." in src/tests/algorithms_test/calorimetry_CalorimeterHitDigi.cc, and an expression in src/tests/algorithms_test/digi_SiliconPulseGeneration.cc. There's no dedicated service test, though. Not sure we're well setup for that.

veprbl:
Indeed, there is no setup dedicated for testing services currently. EvaluatorSvc specifically would be relatively easy to test as a unit, not sure why I didn't do it originally.

PR #1816, opened 2025-04-21
fix: improve const-correctness in RichGeo service

wdconinc:
This PR improves const-correctness in the RichGeo service classes. This should make it easier to maintain thread-safety by avoiding the use of services (shared resource) as a way to store state information.

Note: This is the low-hanging fruit only.

- [x] Bug fix (issue: services are supposed to be used with const accessors only since multiple threads use them at the same time)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/detectors/DRICH/DRICH.cc
+++ b/src/detectors/DRICH/DRICH.cc
@@ -74,10 +74,10 @@ void InitPlugin(JApplication* app) {
   TrackPropagationConfig gas_track_cfg;
 
   // get RICH geo service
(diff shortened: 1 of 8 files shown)


wdconinc:
Yay, no diffs.

ruse-traveler (commented):
Changes look good! I am seeing some diffs in the event header (specifically EventHeader.timeStamp) for the 10x100 samples, though... I take it those are unrelated?

ruse-traveler (approved):
🚀

PR #1824, opened 2025-04-23
feat: add ActsTrajectoriesMerger

wdconinc:
This PR pulls the ActsTrajectoriesMerger and non-owning JOmniFactory Output support out of #1744.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1744)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -150,14 +150,12 @@ public:
 
   template <typename T> class Output : public OutputBase {
     std::vector<T*> m_data;
-    bool m_owns_data;
 
   public:
-    Output(JOmniFactory* owner, std::string default_tag_name = "", bool owns_data = true) {
+    Output(JOmniFactory* owner, std::string default_tag_name = "") {
       owner->RegisterOutput(this);
       this->collection_names.push_back(default_tag_name);
-      this->type_name   = JTypeInfo::demangle<T>();
-      this->m_owns_data = owns_data;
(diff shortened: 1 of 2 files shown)


simonge (dismissed):
Looks good.

The B0 PR confirms this actually works. Just to check though, is this creating new Trajectories which point to the data in the two input collections?

If the factory doesn't own the new trajectory, what does?

wdconinc:
> Just to check though, is this creating new Trajectories which point to the data in the two input collections?
> 
> If the factory doesn't own the new trajectory, what does?

Good point.

wdconinc:
> > Just to check though, is this creating new Trajectories which point to the data in the two input collections?
> > If the factory doesn't own the new trajectory, what does?
> 
> Good point.

The reason why I introduced this data ownership policy was to address memory issues that I had attributed to duplicate deletes due to storing the same track pointers in multiple `ActsExamples::Trajectories`. That's why I had naively made it non-owning, since I assume what was happening was that the `ActsExamples::Trajectories` destructor that deleted the tracks.

It turned out to be a different problem entirely: insufficient clearing of the pointer vectors in JOmniFactory (cc @nathanwbrei: sometimes the `std::vector<T*>` has non-zero size when entering `Process`). After fixing that, there is no more need for the data ownership policy boolean, or the underlying functionality in JOmniFactory.

simonge (approved):
Nice, looks much cleaner.

PR #1826, opened 2025-04-23
feat(ci): compile with -Werror -Wall -Wextra

wdconinc:
This PR ensures that CI checks are run with all warnings enabled, and warnings treated as errors. This will flag unused variables, functions, etc.

Note: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562 and https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dbuiltin, in particular 
> Note that sanitizers tend to increase the rate of false positive warnings, most notably those around -Wmaybe-uninitialized. We recommend against combining -Werror and [the use of] sanitizers.

Note: This PR will only be merged when it passes, which will require some other PRs first. Then it will be ensured that code that fails to compile is failing due to the contributed code.

- [x] #1827
- [x] #1828 
- [x] #1829
- [x] #1830
- [x] #1831
- [x] #1832
- [x] #1833
- [x] #1834
- [x] #1835
- [x] #1836
- [x] #1837
- [x] #1838
- [x] #1839
- [x] #1840
- [x] #1844

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @veprbl 

No.

Yes. Code will have to satisfy more criteria.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -72,7 +72,7 @@ jobs:
         CXXFLAGS: [-Werror -Wall -Wextra]
         release: ${{ fromJSON(needs.env.outputs.release_json) }}
         include:
-          # note that entries that overwrite matrix values require a full specific action 
+          # note that entries that overwrite matrix values require a full specific action
           # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations
           # all clang debug jobs get extra flags
           - CXX: clang++
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -71,7 +71,6 @@ jobs:
         CMAKE_BUILD_TYPE: [Release, Debug]
         # FIXME: gcc-12 does not allow ignoring unknown-pragmas with a pragma
         # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431
-        CXXFLAGS: [-Werror -Wall -Wextra -Wno-unknown-pragmas -Wno-error=maybe-uninitialized]
         release: ${{ fromJSON(needs.env.outputs.release_json) }}
         # note that entries that overwrite matrix values require a full specification
         # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations
@@ -80,6 +79,10 @@ jobs:
           - CXX: clang++
             CMAKE_BUILD_TYPE: Debug
         include:
+          - CXX: g++
+            CXXFLAGS: -Werror -Wall -Wextra -Wno-unknown-pragmas -Wno-error=maybe-uninitialized
+          - CXX: clang++
+            CXXFLAGS: -Werror -Wall -Wextra -Wno-unknown-pragmas
           # include clang++ Debug with profile CXXFLAGS
           - CXX: clang++
             CMAKE_BUILD_TYPE: Debug
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -69,6 +69,7 @@ jobs:
       matrix:
         CXX: [g++, clang++]
         CMAKE_BUILD_TYPE: [Release, Debug]
+        CXXFLAGS: [-Werror -Wall -Wextra]
         release: ${{ fromJSON(needs.env.outputs.release_json) }}
         include:
           # all clang debug jobs get extra flags
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -76,6 +76,7 @@ jobs:
           # exclude clang++ Debug with default CXXFLAGS
           - CXX: clang++
             CMAKE_BUILD_TYPE: Debug
+            release: nightly
         include:
           - CXX: g++
             # -Wno-error=maybe-uninitialized for GCC only as it has issues with /usr/include/c++/12/bits/std_function.h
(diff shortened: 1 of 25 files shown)


wdconinc:
Notes to self:
- check that the compilation matrix is still correct
- update the required jobs in settings with new flags

.github/workflows/linux-eic-shell.yml
```diff
@@ -69,12 +69,18 @@ jobs:
       matrix:
         CXX: [g++, clang++]
         CMAKE_BUILD_TYPE: [Release, Debug]
+        # FIXME: gcc-12 does not allow ignoring unknown-pragmas with a pragma
+        # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431
+        CXXFLAGS: [-Werror -Wall -Wextra -Wno-unknown-pragmas]
         release: ${{ fromJSON(needs.env.outputs.release_json) }}
         include:
+          # note that entries that overwrite matrix values require a full specific action
```
wdconinc, line 77:
```suggestion
          # note that entries that overwrite matrix values require a full specification
```

src/algorithms/tracking/ActsGeometryProvider.cc
```diff
@@ -1,6 +1,14 @@
...
 // Copyright (C) 2022 - 2024 Whitney Armstrong, Wouter Deconinck, Dmitry Romanov
 
+// FIXME: gcc-12 affected by https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunknown-pragmas"
+#pragma clang diagnostic ignored "-Wunknown-warning-option"
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#include <functional> // IWYU pragma: keep
+#pragma GCC diagnostic pop
+
```
veprbl, line 11: Just `-Wno-error=unknown-pragmas ...` until we fix this.

src/algorithms/pid/Tools.h
```diff
@@ -30,11 +30,11 @@ namespace Tools {
   // -------------------------------------------------------------------------------------
   // Radiator IDs
 
-  static std::unordered_map<int, std::string> GetRadiatorIDs() {
+  inline std::unordered_map<int, std::string> GetRadiatorIDs() {
```
veprbl, line 33: `inline` keyword doesn't convey anything anymore, we don't need it.

wdconinc: You tried this earlier today and it failed. https://github.com/eic/EICrecon/actions/runs/14823758425/job/41614185550?pr=1826#step:7:560

src/algorithms/tracking/ActsGeometryProvider.cc
wdconinc: What's your suggestion? We already pass `-Wno-unknown-pragmas`, if that's what you are implying.

src/algorithms/pid/Tools.h
veprbl: Oh, I now remember, such issues can reveal (under clang, if I remember correctly). Let's leave it in then.

src/algorithms/tracking/ActsGeometryProvider.cc
veprbl: The context of my comment is:
```
// FIXME: gcc-12 affected by https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunknown-pragmas"
#pragma clang diagnostic ignored "-Wunknown-warning-option"
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
#include <functional> // IWYU pragma: keep
#pragma GCC diagnostic pop
```
We could address it with
`-Wno-error=unknown-pragmas -Wno-error=unknown-warning-option -Wno-error=maybe-uninitialized`

wdconinc: I don't think so. gcc doesn't accept `unknown-warning-option`

Frankly this is a mess on the clang side:
- they parse the GCC pragmas
- there is no way to turn that off
- they fail for some gcc options they don't recognize
- the way to ignore unknown gcc options is with an option that gcc fails on

wdconinc: So, I repeat: what's your suggestion and does it work with gcc and clang both?

.github/workflows/linux-eic-shell.yml
```diff
@@ -69,12 +69,22 @@ jobs:
       matrix:
         CXX: [g++, clang++]
         CMAKE_BUILD_TYPE: [Release, Debug]
+        # FIXME: gcc-12 does not allow ignoring unknown-pragmas with a pragma
+        # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431
+        CXXFLAGS: [-Werror -Wall -Wextra -Wno-unknown-pragmas -Wno-error=maybe-uninitialized]
```
wdconinc, line 74: This isn't right, though. We don't want to disable maybe-uninitialized everywhere just because there is an issue in a gcc header.

wdconinc: I mean, this is exactly where we will find sources of undefined behavior.

veprbl: Sorry, I was testing privately, but pushed to wrong repo.

Yes, I'm thinking to disable this warning. If compilers are buggy, I don't think we should keep the pragmas around.

veprbl: We would not need to disable the same sets of warnings on all compilers, I assume.

wdconinc: The issue in functional for gcc-12 was fixed in gcc-13. It was treated as a legitimate bug. I don't think we have to expect this to happen, any more than having other compiler bugs hitting us.

The warning about false positives when combining sanitizers and initialization warnings is for user code, not targeted at gcc c++ library implementation code.

I think splitting the pipelines so only gcc or only clang runs with certain warnings could make it hard to debug for users.

veprbl (dismissed):
LGTM

veprbl (approved):
Diff LGTM, capybara report clean.

PR #1831, opened 2025-04-24
fix: avoid signed/unsigned comparisons

wdconinc:
This PR avoids signed/unsigned integer comparisons by using size_t for array sizes and explicitly casting unsigned integers into signed equivalents (with inevitable narrowing) where necessary.

- [x] Bug fix (issue: avoid signed and unsigned integer comparisons)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterTruthClustering.cc
+++ b/src/algorithms/calorimetry/CalorimeterTruthClustering.cc
@@ -47,7 +47,8 @@ void CalorimeterTruthClustering::process(const CalorimeterTruthClustering::Input
     // FIXME: to be fixed so proper object tracking can be done without
     // FIXME: requiring Collection classes be used to manage all objects.
     std::size_t mcIndex;
-    if ((hit.getObjectID().index >= 0) && (hit.getObjectID().index < mc->size())) {
+    if ((hit.getObjectID().index >= 0) &&
+        (hit.getObjectID().index < static_cast<long>(mc->size()))) {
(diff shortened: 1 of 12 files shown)


src/algorithms/digi/PulseNoise.h
```diff
@@ -41,8 +41,8 @@ class PulseNoise : public PulseNoiseAlgorithm, public WithPodConfig<PulseNoiseCo
   void process(const Input&, const Output&);
 
 private:
-  std::default_random_engine generator; // TODO: need something more appropriate here
-  dd4hep::detail::FalphaNoise m_noise;
+  mutable std::default_random_engine m_generator; // TODO: need something more appropriate here
+  mutable dd4hep::detail::FalphaNoise m_noise;    // FalphaNoise::operator() is not const
```
simonge, line 45: This might (more) belong in the other PR (https://github.com/eic/EICrecon/pull/1833)
Is calling something mutable a non-ideal way of getting around the const ness of a function?

wdconinc: Correct. That's why that one depends on this PR. The context was overlapping and I didn't think it was that necessary to split them and deal with the conflicts.

wdconinc: Amended and removed this from scope. Will add in #1833 instead.

simonge (approved):
Happy with the changes, was mostly curious.

PR #1833, opened 2025-04-24
fix: ensure algorithms satisfy const process function

wdconinc:
This PR ensures that `process` functions of algorithms are const (enables easier multi-threaded operations). Some initialization that is only possible on first event is marked mutable.

- [x] Bug fix (issue: algorithms process should be const)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/fardetectors/FarDetectorMLReconstruction.h
+++ b/src/algorithms/fardetectors/FarDetectorMLReconstruction.h
@@ -53,7 +53,7 @@ private:
   mutable float m_beamE{10.0};
   mutable std::once_flag m_initBeamE;
 
-  float nnInput[4] = {0.0, 0.0, 0.0, 0.0};
+  mutable float nnInput[4] = {0.0, 0.0, 0.0, 0.0};
 };
 
 } // namespace eicrecon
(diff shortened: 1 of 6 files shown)


src/algorithms/fardetectors/FarDetectorMLReconstruction.h
```diff
@@ -42,16 +42,18 @@ class FarDetectorMLReconstruction : public FarDetectorMLReconstructionAlgorithm,
...
   TMVA::Reader* m_reader{nullptr};
   TMVA::MethodBase* m_method{nullptr};
-  float m_beamE{10.0};
-  std::once_flag m_initBeamE;
-  float nnInput[4] = {0.0, 0.0, 0.0, 0.0};
+
+  mutable float m_beamE{10.0};
+  mutable std::once_flag m_initBeamE;
+
+  mutable float nnInput[4] = {0.0, 0.0, 0.0, 0.0};
```
wdconinc, line 56: FYI, @simonge, this is another use of `mutable`.

wdconinc:
> All good. I'll make sure not to remove the const and use mutable where relevant in the future...

The intention is to make sure you cannot use a non-const process function anymore since it messes with multi-threaded running.

PR #1836, opened 2025-04-24
fix: correct class initializer ordering

wdconinc:
This PR corrects the class initialization ordering to conform to the order in which actual initialization happens.

- [x] Bug fix (issue: compiler warnings/errors)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


simonge (approved):
Looks good too. I didn't even realise this was an issue. Just to check copilot's explanation here, the order of initialization is actually carried out in the order the variables were declared in the class rather than the initialization list order?

PR #1841, opened 2025-04-24
Revert "fix(ci): list more open PRs in GitHub API call"

veprbl:
Looks like this may not work very well due to GitHub limits.

Reverts eic/EICrecon#1818


veprbl:
Pointed privately by @wdconinc , it looks like its just a pages size limit. Closing in favor of https://github.com/eic/epic-capybara/pull/20 .

PR #1843, opened 2025-04-25
Make CollectionCollector collections optional

simonge:
Allows the creation of ReconstructedParticles using configurations which don't contain the calorimeters, e.g. epic_ip6.xml.

The selection of calorimeter hits included will probably need to change in the future.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -52,7 +52,7 @@ public:
       owner->RegisterInput(this);
       this->collection_names.push_back(default_tag);
       this->is_optional = is_optional;
-      this->type_name = JTypeInfo::demangle<T>();
+      this->type_name   = JTypeInfo::demangle<T>();
     }
 
     const std::vector<const T*>& operator()() { return m_data; }
@@ -60,7 +60,9 @@ public:
   private:
     friend class JOmniFactory;
 
-    void GetCollection(const JEvent& event) { m_data = event.Get<T>(this->collection_names[0], !this->is_optional); }
+    void GetCollection(const JEvent& event) {
+      m_data = event.Get<T>(this->collection_names[0], !this->is_optional);
+    }
   };
 
   template <typename PodioT> class PodioInput : public InputBase {
@@ -68,7 +70,8 @@ public:
     const typename PodioTypeMap<PodioT>::collection_t* m_data;
 
   public:
-    PodioInput(JOmniFactory* owner, std::string default_collection_name = "", bool is_optional = false) {
+    PodioInput(JOmniFactory* owner, std::string default_collection_name = "",
+               bool is_optional = false) {
       owner->RegisterInput(this);
       this->collection_names.push_back(default_collection_name);
       this->is_optional = is_optional;
@@ -90,7 +93,8 @@ public:
     std::vector<const typename PodioTypeMap<PodioT>::collection_t*> m_data;
 
   public:
-    VariadicPodioInput(JOmniFactory* owner, std::vector<std::string> default_names = {}, bool is_optional = false) {
+    VariadicPodioInput(JOmniFactory* owner, std::vector<std::string> default_names = {},
+                       bool is_optional = false) {
       owner->RegisterInput(this);
       this->collection_names = default_names;
       this->type_name        = JTypeInfo::demangle<PodioT>();
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -60,7 +60,7 @@ public:
 
     void GetCollection(const JEvent& event) {
       try {
-        m_data = event.Get<T>(this->collection_names[0],!IsOptional);
+        m_data = event.Get<T>(this->collection_names[0], !IsOptional);
       } catch (const std::exception& e) {
         if constexpr (!IsOptional) {
           throw JException("JOmniFactory: Failed to get collection %s: %s",
@@ -88,7 +88,7 @@ public:
 
     void GetCollection(const JEvent& event) {
       try {
-        m_data = event.GetCollection<PodioT>(this->collection_names[0],!IsOptional);
+        m_data = event.GetCollection<PodioT>(this->collection_names[0], !IsOptional);
       } catch (const std::exception& e) {
         if constexpr (!IsOptional) {
           throw JException("JOmniFactory: Failed to get collection %s: %s",
@@ -121,7 +121,7 @@ public:
       m_data.clear();
       for (auto& coll_name : this->collection_names) {
         try {
-          m_data.push_back(event.GetCollection<PodioT>(coll_name,!IsOptional));
+          m_data.push_back(event.GetCollection<PodioT>(coll_name, !IsOptional));
         } catch (const std::exception& e) {
           if constexpr (!IsOptional) {
             throw JException("JOmniFactory: Failed to get collection %s: %s", coll_name.c_str(),
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -61,7 +61,7 @@ public:
     void GetCollection(const JEvent& event) {
       try {
         m_data = event.Get<T>(this->collection_names[0], !IsOptional);
-      } catch (const JExceptionn& e) {
+      } catch (const JException& e) {
         if constexpr (!IsOptional) {
           throw JException("JOmniFactory: Failed to get collection %s: %s",
                            this->collection_names[0].c_str(), e.what());
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -38,6 +38,7 @@ public:
     std::string type_name;
     std::vector<std::string> collection_names;
     bool is_variadic = false;
+    bool is_optional = false;
 
     virtual void GetCollection(const JEvent& event) = 0;
   };
@@ -47,9 +48,10 @@ public:
     std::vector<const T*> m_data;
 
   public:
-    Input(JOmniFactory* owner, std::string default_tag = "") {
+    Input(JOmniFactory* owner, std::string default_tag = "", bool is_optional = false) {
       owner->RegisterInput(this);
       this->collection_names.push_back(default_tag);
+      this->is_optional = is_optional;
       this->type_name = JTypeInfo::demangle<T>();
     }
 
@@ -58,7 +60,7 @@ public:
   private:
     friend class JOmniFactory;
 
-    void GetCollection(const JEvent& event) { m_data = event.Get<T>(this->collection_names[0]); }
+    void GetCollection(const JEvent& event) { m_data = event.Get<T>(this->collection_names[0], !this->is_optional); }
   };
 
   template <typename PodioT> class PodioInput : public InputBase {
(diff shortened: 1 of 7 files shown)


wdconinc:
I'm curious what others think, but I wonder if this adds unmanageable complexity to plugins. They are supposed to be self-contained and "default constructable" but requiring the DD4hep_service means we need to load one plugin before another. We don't resolve this automatically though, so every so often someone will have to manually reorder the plugin list...

Not that I have a better solution here. Maybe we need to finally figure out optional inputs instead of failing when they are absent.

simonge:
> Not that I have a better solution here. Maybe we need to finally figure out optional inputs instead of failing when they are absent.

I'm pretty sure the most recent JANA2 version will allow the optional inputs @nathanwbrei? What was the reason we have our own JOmniFactory rather than directly using the JANA2 versions?

src/factories/meta/CollectionCollector_factory.h
```diff
@@ -15,7 +15,7 @@ class CollectionCollector_factory : public JOmniFactory<CollectionCollector_fact
   std::unique_ptr<AlgoT> m_algo;
 
   typename JOmniFactory<CollectionCollector_factory<T>>::template VariadicPodioInput<T> m_inputs{
-      this};
+      this, std::vector<std::string>{""}, true};
```
wdconinc, line 18: The optional nature of the collection is compile-time information. Maybe after the process of implementing this you have a good idea of what it would take to do this be entirely at compile-time.

Other question: are there no current (or potential) uses of CollectionCollector_factory where we do expect all collections to exist?

simonge: > The optional nature of the collection is compile-time information. Maybe after the process of implementing this you have a good idea of what it would take to do this be entirely at compile-time.

I don't entirely follow what the distinction is here.

> Other question: are there no current (or potential) uses of CollectionCollector_factory where we do expect all collections to exist?

I can't think of anything, not to say it wouldn't arise in some way I haven't considered. I have an alternative implementation which has a SetOptional method in the InputBase which can be set in the CollectionCollector_factory::Configure as true, or via a non-existent CollectionCollectorConfig.

simonge:
> @wdconinc copying across the minimal changes from main JANA2 for optional collections appears to work for me in the small tests I've done.

Might have jumped the gun, I had just been testing the compilation and forgotten to reinstall...

src/factories/meta/CollectionCollector_factory.h
wdconinc: I meant that we don't need to pass a bool variable and then query the bool variable with this->is_optional. This is all already known at compile time, so the bool could be a template parameter, and that can be used in the GetCollection call.

simonge:
I'm pretty sure it works properly now. There may be details beyond my understanding which should be fixed.

veprbl:
Looks like it crashes.

src/factories/meta/CollectionCollector_factory.h
```diff
@@ -6,17 +6,20 @@
 
 namespace eicrecon {
 
-template <class T>
-class CollectionCollector_factory : public JOmniFactory<CollectionCollector_factory<T>> {
+template <class T, bool IsOptional = true>
```
veprbl, line 9:
```suggestion
template <class T, bool IsOptional = false>
```
Opt-out by default? Even if we want optional in most places, it's good to advertise that this is a flag that can be toggled.

src/global/tracking/tracking.cc
```diff
@@ -240,36 +216,25 @@ void InitPlugin(JApplication* app) {
...
-      app));
+      "CombinedTrackAssociations",
+      {"CentralCKFTrackAssociations", "TaggerTrackerTrackAssociations"},
+      {"CombinedTrackAssociations"}, app));
+
   app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::Track>>(
-      "CombinedTruthSeededTracks", input_truth_track_collections, {"CombinedTruthSeededTracks"},
-      app));
+      "CombinedTruthSeededTracks",
+      {"CentralCKFTruthSeededTrackAssociations", "TaggerTrackerTrackAssociations"},
```
veprbl:
```suggestion
      {"CentralCKFTruthSeededTracks", "TaggerTrackerTracks"},
```

veprbl: Crossed with 4582554ce2089b33154717e49a96653d8c42cb48

src/extensions/jana/JOmniFactory.h
```diff
@@ -58,10 +58,19 @@ class JOmniFactory : public JMultifactory {
   private:
     friend class JOmniFactory;
 
-    void GetCollection(const JEvent& event) { m_data = event.Get<T>(this->collection_names[0]); }
+    void GetCollection(const JEvent& event) {
+      try {
+        m_data = event.Get<T>(this->collection_names[0]);
+      } catch (const std::exception& e) {
```
wdconinc, line 64: Can we narrow this to the specific exception when a collection is missing in the input and not the collection's algorithm made a boo-boo? I know, likely it's JException or bust, but I can dream, right?

veprbl: But this is just a rethrowing. I think motivation here was to provide nicer debug messages, not improve actual handling of errors.

Also, in principle, we could have added such context via boost::exceptions.

veprbl: Regarding narrowing by type, it's probably JException for collection not found.

wdconinc: But we control the input layer through JEventSourcePodio. Does that help here? (On my phone so I'm not able to check in detail.)

veprbl: I don't what you have in mind here, we either have a type match or not in this clause. Generally, control over input layer doesn't help unless we can sort the dependency graph.

wdconinc: Because JANA uses their "factory" pattern, dependencies turn into function call stacks. Any exception thrown in the deepest part of the call stack (frame 0 in a backtrace) is passed up and (hopefully*) rethrown without changing its type. It is up to the factory that allows for optional inputs (which is further up in the call stack) to decide what to do with the exception when it gets there. If we can throw a `jana::absent_in_event_source` then that exception is different from a `std::out_of_range` thrown somewhere.

* I think this is likely the part where it goes wrong. JANA takes a `std::exception` and throws a JException with only the `what()` string.

simonge: Is there some advice with whether to do something different with the exception? This is something I haven't much experience with.

ruse-traveler (commented):
Apologies, but I'm catching up and I think I lost the thread here. Let me ask a few conceptual questions for my own understanding:
1. The goal here is to create reconstructed particles _without_ using any calorimeter inputs, i.e. to be able to run `MatchClusters` without providing a cluster collection, right?
2. So how does the `CollectionCollector` fit into this? Is this a test case for making sure optional inputs work now?

If the answer is yes to (1), this seems to me to be the right path towards accomplishing this, though...

simonge:
> Apologies, but I'm catching up and I think I lost the thread here. Let me ask a few conceptual questions for my own understanding:
> 
> 1. The goal here is to create reconstructed particles _without_ using any calorimeter inputs, i.e. to be able to run `MatchClusters` without providing a cluster collection, right?
> 2. So how does the `CollectionCollector` fit into this? Is this a test case for making sure optional inputs work now?
> 
> If the answer is yes to (1), this seems to me to be the right path towards accomplishing this, though...

The original aim was as you suggested. My initial approach was to just copy what we've done for the tracking and test if the input hit collection exists in the geometry. This was never so allowing properly optional input collections has been implemented instead.

An alternative/future modification would be to have the cluster input into `MatchClusters` optional. I don't know if this would provide any minor performance improvements.

A bit more context. I am testing some downstream software developed by a colleague which just uses the `ReconstructedParticles`, `MCParticles `and Association branches to carry out analysis on various reactions. Testing it on a detector configuration which doesn't include calorimeters didn't produce the `ReconstructedParticles` even when none of the tracks passed could be matched.

simonge:
> But right now this doesn't just test that, but also whether a collection that is added has failed to be constructed for any other reason. My concern in [#1843 (comment)](https://github.com/eic/EICrecon/pull/1843#discussion_r2061559561) is that this would let these other reasons go unnoticed...

I don't see that's any different than the current behaviour? The fact that there are no merged clusters doesn't stop ReconstructedParticles being created, it just silently stops it being populated. 

Other than just checking and understanding the output would we require upstream modifications the JANA2 exceptions or a different approach altogether to handle this properly?

src/extensions/jana/JOmniFactory.h
veprbl: I think I'm starting to get what you want here. I was also thinking about extending types of exceptions. But, in the end, it all requires that JANA throws them - there is a way to throw it from JEventSourcePODIO, I think.

Which of the following statuses are useful:
1. "not being present in input file"
2. "not being present in input file, but is an input that could have come from simulation"
3. "not being present in input file, but can not be produced by any factory available"

It looks like we'd like (2.), but (3.) is probably easier to derive.

simonge: Should we move this further development to a future PR? Or is it likely if we don't tackle it now it will get ignored again for a long time?

wdconinc: From my point of view this can get dealt with in a future PR, but I'd like to ask we print out at least a warning when a collection is ignored (until we can be sure it is ignored for the right reasons). And we should create an issue to keep track of this need.

simonge: That would probably result in a lot of log messages for incomplete configurations, where every event could flag up every missing collection.

Looking at the capybara report again I am now confused as to how tracking_only already has entries in the ReconstructedParticles collection. Something about the behaviour must be different to what I've been assuming.

ruse-traveler:
> The original aim was as you suggested. My initial approach was to just copy what we've done for the tracking and test if the input hit collection exists in the geometry. This was never so allowing properly optional input collections has been implemented instead.
> 
> An alternative/future modification would be to have the cluster input into `MatchClusters` optional. I don't know if this would provide any minor performance improvements.
> 
> A bit more context. I am testing some downstream software developed by a colleague which just uses the `ReconstructedParticles`, `MCParticles `and Association branches to carry out analysis on various reactions. Testing it on a detector configuration which doesn't include calorimeters didn't produce the `ReconstructedParticles` even when none of the tracks passed could be matched.

Gotcha! Thanks for the clarification! That helps a lot! And in light of that, I'd second deferring additional work on std::exception vs. JException to a follow-up PR. I'm pretty happy with this as-is!

src/extensions/jana/JOmniFactory.h
veprbl: Can we agree that exceptions arising from "missing" collections are limited to JException?
```suggestion
      } catch (const JException& e) {
```
This will leave more general class `std::exception`s to the status quo.

veprbl: > Looking at the capybara report again I am now confused as to how tracking_only already has entries in the ReconstructedParticles collection. Something about the behaviour must be different to what I've been assuming.

Do you know of a specific piece that should fail when calorimeters are missing?

simonge: > > Looking at the capybara report again I am now confused as to how tracking_only already has entries in the ReconstructedParticles collection. Something about the behaviour must be different to what I've been assuming.
> 
> Do you know of a specific piece that should fail when calorimeters are missing?

I have done a couple of extra tests on the main branch which show I can recover the behaviour of the tracking_only configuration with the ip6 configuration.

The difference appears to be if I request `EcalClusters `or all three of `EcalEndcapPClusters,EcalBarrelScFiClusters,EcalEndcapNClusters` in my output_collections list, it forces the creation of the empty collection(s) even if the inputs don't exist, these empty collections then exist enough to propagate through the factories to fill the ReconstructedParticles collection.

veprbl (approved):
No objections from me. I think, we should also ping @nathanwbrei on this.

PR #1845, opened 2025-04-28
fix(ci): remove CC, only use CXX

wdconinc:
Since we specify in CMakeLists.txt that this project uses CXX only, we don't need to provide CC. This simplifies the yml file a bit.

- [x] Bug fix (issue: CC unused)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
Nice

PR #1853, opened 2025-05-07
fix: move implementation out of headers for compilation speed

wdconinc:
This PR takes the largest headers, and moves implementations into a dedicate cc file. This may speed up compilation, though the only header that is included in a significant number of other cc files is the CalorimeterIslandCluster.h.

No code changes beyond moving implementation to cc.

Needs:
- [x] #1863 

- [x] Bug fix (issue: definition and implementation should be separate)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


ruse-traveler (approved):
LGTM!

PR #1861, opened 2025-05-09
Add configuration option to output clones from subdivided and collected collections

simonge:
Adds the `output_copies` configuration to tell the CollectionCollector and SubDivideCollection algorithms to copy the data into the output rather than produce a subset collection.

This will allow, for instance, the merging of particles from the central detector and far detectors into a final collection, with the final collection being directly readable in a root tree. Or wanting to examine the contents of the collection in a root tree for debugging.

e.g. `-Pbeam:BeamParticles:outputCopies=true`

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No, but perhaps an example collection should be chosen

### DIFF
--- a/src/algorithms/meta/CollectionCollector.h
+++ b/src/algorithms/meta/CollectionCollector.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2024 Simon Gardner
+// Copyright (C) 2024-2025 Simon Gardner
 
 #include <spdlog/spdlog.h>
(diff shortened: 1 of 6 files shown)


simonge:
I suppose this could be dangerous/cause confusion if used on some intermediate collections and mess up all of the associations.

wdconinc:
I don't know if I like this. There are suitable interfaces to podio for all analysis languages (or, if not, they can be developed). This duplicates storage needs for convenience reasons only.

PR #1864, opened 2025-05-10
feat: avoid explicit loop in CalorimeterIslandCluster

wdconinc:
This PR adds some micro-optimization, primarily as proof of concept :-) and demo of good practice. Godbolt confirms equal or better speed for n from 10 to 100000 for compilers currently in use and for all execution policies. This is just stylistic, in a sense.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: prefer modern c++ with use of generic functions where possible)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
image

PR #1867, opened 2025-05-13
Use time info from TOF and MPGD for CKF

ShujieL:
Try to use time info as an additional dimension for CKF. Only provide this with TOF and MPGD hits. 
CKF should be able to handle a combination of 2D and 3D measurements.

Work in progress. 

- [ ] Bug fix (issue #__)
- [ x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -162,13 +162,16 @@ CKFTracking::process(const edm4eic::TrackParametersCollection& init_trk_params,
     Acts::ActsVector<2> loc = Acts::Vector2::Zero();
     loc[Acts::eBoundLoc0]   = meas2D.getLoc().a;
     loc[Acts::eBoundLoc1]   = meas2D.getLoc().b;
+   // add time info from TOF or MPGD. ACTS can handle a mix of space points, AND space points with time info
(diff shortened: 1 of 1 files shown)


ShujieL:
The calibrator needs to be updated if we want to pass time to CKF 
https://github.com/eic/EICrecon/blob/b44aecda3de2412c4e2f7138245f3f1750ffdec1/src/algorithms/tracking/CKFTracking.cc#L115

@wdconinc suggested "do this based on a flag in the CKFTrackingConfig so it can be turned on/off selectively. Default could remain off."

We should consider this later when there is a clear use case.

simonge:
I am a bit confused why this isn't a priority for tracking considering the importance given to the background studies.

ShujieL:
> I am a bit confused why this isn't a priority for tracking considering the importance given to the background studies.

Sorry the "use case" sentence was just my thoughts.

The current background studies show decent purity when the fast detectors (TOF and MPGD) hits are included as confirmation hits. So I personally didn't see a strong motivation to add the time info to CKF right now until we know clearly what we try to improve.

The tracking priority needs a separate discussion. We can follow up on Thursday tracking meeting.

PR #1869, opened 2025-05-15
fix: Propagate tracks from outermost surface

bschmookler:
This updates the track propagation (projection) to use the track parameters determined on the outermost tracking layer.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, it will change the track propagation results.

### DIFF
--- a/src/algorithms/tracking/TrackPropagation.cc
+++ b/src/algorithms/tracking/TrackPropagation.cc
@@ -234,11 +234,14 @@ TrackPropagation::propagate(const edm4eic::Track& /* track */,
   m_log->trace("  Num measurement in trajectory: {}", m_nMeasurements);
   m_log->trace("  Num states in trajectory     : {}", m_nStates);
 
-  //=================================================
-  //Track projection
-  //Reference sPHENIX code: https://github.com/sPHENIX-Collaboration/coresoftware/blob/335e6da4ccacc8374cada993485fe81d82e74a4f/offline/packages/trackreco/PHActsTrackProjection.h
-  //=================================================
-  const auto& initial_bound_parameters = acts_trajectory->trackParameters(trackTip);
+  // Get track state at last measurement surface
+  // For last measurement surface, filtered and smoothed results are equivalent
+  auto trackState = mj.getTrackState(trackTip);
+  auto initSurface = trackState.referenceSurface().getSharedPtr();
+  const auto& initParams  = trackState.filtered();
+  const auto& initCov     = trackState.filteredCovariance();
+
+  Acts::BoundTrackParameters initBoundParams(initSurface,initParams,initCov,Acts::ParticleHypothesis::pion());
 
   m_log->trace("    TrackPropagation. Propagating to surface # {}",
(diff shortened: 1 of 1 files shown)


mposik1983 (approved):

wdconinc:
Can you compare with https://github.com/eic/epic-capybara/pull/23 to drill down into the changes for small covariances (they should get smaller)?

And can you comment on whether the pathlength variable changes need to be considered negatively? Is there a way to add to the pathlenght the value that we get from the final track state which is now used as starting point? Right now the pathlength would not mean anything really.
image

mposik1983:
I re-ran my propagation to the dirc surface with this updated code. The first 2x2 set of plots shows the theta (left) and phi (right) residual distributions. The bottom row shows the theta error (left) and phi error (right) taken from the cov matrix. 
Screen Shot 2025-05-15 at 10 49 38 AM

The next two plots show the pull distributions (e.g. residual / cov error) for theta (left) and phi (right). These pull distributions seem to show consistency between the two methods. Some caveats: one, this is for pi- at fixed p = 6GeV, and eta ~ -0.05. I have not looked at other regions yet. Two, I had not realized pid detector materials were in the material map, so there is a discrepancy between the material seen by my reference (truth) hit and the propagated trajectory.  

Screen Shot 2025-05-15 at 10 50 08 AM

bschmookler:
I just updated it so that it add the pathlength of the state (i.e. last tracker measurement) and the propagation pathlength.

bschmookler:
I'm not sure how to trigger a capybara branch other than the main.

wdconinc:
> I'm not sure how to trigger a capybara branch other than the main.

Not possible in CI. Only local install. I verified, no changes to the position and momentum errors. Not sure if that's good or bad...

bschmookler:
Ok. I'm surprised since Matt sees some change for the DIRC (theta and phi
errors at least).  The pull distribution sigmas are closer to 1 now.

wdconinc:
Here's the report (before your pathlength fix): [capybara-reports.zip](https://github.com/user-attachments/files/20234372/capybara-reports.zip)

Look at with
```console
unzip capybara-reports.zip
cd capybara-reports
python3 -m http.server 8080
```

wdconinc:
Path length now looks unchanged.

Rebased to get eic_container to run.

wdconinc:
> Ok. I'm surprised since Matt sees some change for the DIRC (theta and phi errors at least). The pull distribution sigmas are closer to 1 now.

I guess the reason for the absence of visible changes in the momentum and direction errors in https://eicrecon.epic-eic.org/pr/1869/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#_CalorimeterTrackProjections_points is that there is probably structure inside the most populated bin, and that's not reflected in the interface. The KS confidence levels indicate zero % likelihood, so that may be what's going on.

wdconinc (approved):
Looks good to me.

PR #1870, opened 2025-05-16
Make tracker measurement from hits more optimal

simonge:
This reduces the time taken in the TrackerMeasurementFromHits algorithm. According to JANA this is reduced in my sample from

`61.02 s (100.0%) edm4eic::Measurement2D:CentralTrackerMeasurements`
to
`1.00 ms ( 33.3%) edm4eic::Measurement2D:CentralTrackerMeasurements`

It appears defining the variable `surfaceMap` may have been cloning the map for every hit.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Hopefully speeds everything up.


src/algorithms/tracking/TrackerMeasurementFromHits.cc
```diff
@@ -60,15 +60,13 @@ TrackerMeasurementFromHits::produce(const edm4eic::TrackerHitCollection& trk_hit
     const auto* vol_ctx = m_converter->findContext(hit.getCellID());
     auto vol_id         = vol_ctx->identifier;
 
-    auto surfaceMap = m_acts_context->surfaceMap();
```
veprbl, line 63: Can we instead make the `surfaceMap` into a reference `& surfaceMap` here?

simonge: Sure, done.

I haven't checked running with logging so that might still be unnecessarily inefficient, but less important.

veprbl (approved):

PR #1876, opened 2025-05-21
fix: abs -> std::abs

wdconinc:
This PR ensures consistent use of `std::abs` (in C++ header `<cmath>`) over `abs` (in C header `<math.h>`). Since the behavior of `abs` when included via `<math.h>` may depend on whether or not `<cmath>` is also included (possibly via another header), it is better to be explicit and required `std::abs`, which operates within the same domain for both int and float types (see https://en.cppreference.com/w/cpp/numeric/math/fabs (yes, that's for `std::abs` too)).

- [x] Bug fix (issue: https://github.com/eic/EICrecon/actions/runs/15169971236/job/42657315496#step:7:646)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/reco/FarForwardLambdaReconstruction.cc
+++ b/src/algorithms/reco/FarForwardLambdaReconstruction.cc
@@ -7,7 +7,7 @@
 #include <edm4hep/Vector3f.h>
 #include <edm4hep/utils/vector_utils.h>
 #include <fmt/core.h>
-#include <math.h>
+#include <cmath>
 #include <cstddef>
 #include <gsl/pointers>
 #include <vector>
@@ -94,7 +94,7 @@ void FarForwardLambdaReconstruction::process(
         }
 
         double mass_rec = lambda.M();
-        if (abs(mass_rec - m_lambda) > m_cfg.lambdaMaxMassDev)
+        if (std::abs(mass_rec - m_lambda) > m_cfg.lambdaMaxMassDev)
           continue;
 
         // rotate everything back to the lab coordinates.
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -47,7 +47,8 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
     // require close to interaction vertex
     auto v = mcparticle.getVertex();
     if (std::abs(v.x) * dd4hep::mm > m_cfg.maxVertexX ||
-        std::abs(v.y) * dd4hep::mm > m_cfg.maxVertexY || abs(v.z) * dd4hep::mm > m_cfg.maxVertexZ) {
+        std::abs(v.y) * dd4hep::mm > m_cfg.maxVertexY ||
+        std::abs(v.z) * dd4hep::mm > m_cfg.maxVertexZ) {
       m_log->trace("ignoring particle with vs = {} [mm]", v);
       continue;
     }
@@ -75,7 +76,7 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
     const auto pdg      = mcparticle.getPDG();
     const auto particle = m_particleSvc.particle(pdg);
     double charge       = std::copysign(1.0, particle.charge);
-    if (abs(charge) < std::numeric_limits<double>::epsilon()) {
+    if (std::abs(charge) < std::numeric_limits<double>::epsilon()) {
       m_log->trace("ignoring neutral particle");
       continue;
     }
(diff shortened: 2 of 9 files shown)


src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -94,7 +94,7 @@ void FarForwardLambdaReconstruction::process(
         }
 
         double mass_rec = lambda.M();
-        if (abs(mass_rec - m_lambda) > m_cfg.lambdaMaxMassDev)
+        if (std::abs(mass_rec - m_lambda) > m_cfg.lambdaMaxMassDev)
```
simonge, line 97:
```suggestion
        if (std::fabs(mass_rec - m_lambda) > m_cfg.lambdaMaxMassDev)
```

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -47,7 +46,8 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
     // require close to interaction vertex
     auto v = mcparticle.getVertex();
     if (std::abs(v.x) * dd4hep::mm > m_cfg.maxVertexX ||
-        std::abs(v.y) * dd4hep::mm > m_cfg.maxVertexY || abs(v.z) * dd4hep::mm > m_cfg.maxVertexZ) {
+        std::abs(v.y) * dd4hep::mm > m_cfg.maxVertexY ||
+        std::abs(v.z) * dd4hep::mm > m_cfg.maxVertexZ) {
```
simonge, line 50:
```suggestion
        std::fabs(v.y) * dd4hep::mm > m_cfg.maxVertexY ||
        std::fabs(v.z) * dd4hep::mm > m_cfg.maxVertexZ) {
```

src/services/geometry/richgeo/IrtGeoPFRICH.cc
```diff
@@ -141,7 +141,7 @@ void richgeo::IrtGeoPFRICH::DD4hep_to_IRT() {
       auto testOrtho  = normXdir.Dot(normYdir); // should be zero, if normX and normY are orthogonal
       auto testRadial = sensorNorm.Cross(normZdir)
                             .Mag2(); // should be zero, if sensor surface normal is as expected
-      if (abs(testOrtho) > 1e-6 || abs(testRadial) > 1e-6) {
+      if (std::abs(testOrtho) > 1e-6 || std::abs(testRadial) > 1e-6) {
```
simonge, line 144:
```suggestion
      if (std::fabs(testOrtho) > 1e-6 || std::fabs(testRadial) > 1e-6) {
```

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -75,7 +75,7 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
     const auto pdg      = mcparticle.getPDG();
     const auto particle = m_particleSvc.particle(pdg);
     double charge       = std::copysign(1.0, particle.charge);
-    if (abs(charge) < std::numeric_limits<double>::epsilon()) {
+    if (std::abs(charge) < std::numeric_limits<double>::epsilon()) {
```
simonge, line 78:
```suggestion
    if (std::fabs(charge) < std::numeric_limits<double>::epsilon()) {
```
Although I don't think this will actually do what it claims.

simonge (changes requested):
Looks good. All of the `std::abs` should probably be `std::fabs`

wdconinc:
> Looks good. All of the `std::abs` should probably be `std::fabs`

There's no difference. The single overloaded function name `std::abs` can be used on all types. The f-prefixed versions are a holdover from C.

simonge:
> There's no difference. The single overloaded function name `std::abs` can be used on all types. The f-prefixed versions are a holdover from C.

Ok, thanks. Can we swap them all to `std::abs` to clean it up, or does it not matter enough.

src/algorithms/tracking/TrackParamTruthInit.cc
simonge, line 78: Can I confirm that this is a bug which should be addressed in a different PR.

simonge (approved):
Looks good again.

src/algorithms/tracking/TrackParamTruthInit.cc
wdconinc:
```suggestion
    if (std::abs(particle.charge) < std::numeric_limits<double>::epsilon()) {
```

wdconinc: But I'll apply this in a new PR to be able to assess impact independently. Good catch!

wdconinc: https://github.com/eic/EICrecon/pull/1878

wdconinc: https://github.com/eic/EICrecon/pull/1878

PR #1878, opened 2025-05-22
fix: properly ignore neutrals in TrackParamTruthInit.cc

wdconinc:
This PR fixes an issue ([pointed out](https://github.com/eic/EICrecon/pull/1876#discussion_r2102977558) by @simonge) where neutrals are not skipped as intended in TrackParamTruthInit

- [x] Bug fix (issue: neutrals not skipped)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -75,7 +75,7 @@ eicrecon::TrackParamTruthInit::produce(const edm4hep::MCParticleCollection* mcpa
     const auto pdg      = mcparticle.getPDG();
     const auto particle = m_particleSvc.particle(pdg);
     double charge       = std::copysign(1.0, particle.charge);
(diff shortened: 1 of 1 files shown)


simonge:
Is the `copysign` needed here in the first place rather than just casting the charge to a double?

wdconinc:
> Is the `copysign` needed here in the first place rather than just casting the charge to a double?

I've wondered about this too. In the past I think @sly2j convinced me there was a good reason for this.

wdconinc:
This removes about 50% of the truth seeds:
image

Unfiltered tracks down by about 50% too:
image

After filtering, tracks mostly unchanged and statistically consistent with before:
image

Tagging @bschmookler @ShujieL @mposik1983 @veprbl for their reference.

wdconinc:
Rebased after #1876.

simonge (approved):
Changes to the output look sensible. Where does this change make its way into small changes in ReconstructedParticles?

PR #1880, opened 2025-05-23
fix: JApplication.h -> JApplicationFwd.h

### LINKED ISSUE #1879: Use JApplicationFwd.h header where possible
The slowest included header in EICrecon (cumulatively) is the JApplication.h which is used all over.

There is a forwarding header since JANA2 v2.2.0, which should be faster and may be sufficient for a majority of cases.

wdconinc:
This PR replaces includes of `JApplication.h` with `JApplicationFwd.h` (except for the `Log_service`, where this leads to compilation failures). The iwyu rules were updated to allow the forwarding header to be used. JANA2 v2.2.0 includes the required forwarding header and is already the minimum version required. Saves about 10% of compilation time.

- [x] Bug fix (issue #1879)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/iwyu.imp
+++ b/.github/iwyu.imp
@@ -26,9 +26,6 @@
   { include: ['@<(edm4(eic|hep))/(Mutable)?([^/]*?)(Collection|Data|Obj)?\.h>', public, '<$1/$4Collection.h>', public] },
   { include: ['<edm4hep/utils/vector_utils_legacy.h>', private, '<edm4hep/utils/vector_utils.h>', public] },
 
-  # JANA2: for now, preserve compatibility with older versions
-  { include: ['<JANA/JApplicationFwd.h>', private, '<JANA/JApplication.h>', public] },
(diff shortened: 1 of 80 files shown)


veprbl (approved):
I don't see compilation speedup on my system (in single-threaded mode its 3 seconds slower for ApplicationFwd with ~5 min compilation overall). Change is generally fine.

wdconinc:
> I don't see compilation speedup on my system (in single-threaded mode its 3 seconds slower for ApplicationFwd with ~5 min compilation overall). Change is generally fine.

Mine went down from 6:30 minutes to just under 6 minutes for a compile from clean without ccache. Consistent in magnitude with the 150 seconds in https://github.com/eic/EICrecon/actions/runs/15199556415/job/42751058355#step:8:237, even though that does use ccache.

PR #1881, opened 2025-05-25
feat: ROOT GIL reproducer (don't merge)

wdconinc:
This is a reproducer for the data race observed in #975.

```
gdb -ex "b exit" -ex "b _exit" -ex "catch syscall exit_group" -ex run --args bin/gil 1 2 3 4 5
```


wdconinc:
Moved to self-contained https://github.com/wdconinc/root-data-race-issue18863-reproducer

PR #1895, opened 2025-06-03
Make 10x100 CI run use epic_craterlake_10x100.xml

simonge:
Changes the CI run of 10x100 to use the `epic_craterlake_10x100.xml` rather than just `epic_craterlake.xml` which is actually 5x41.

I thought historically this 10x100 was correct as the default, but it isn't clear looking at the epic history it was ever anything else.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
CI run and report for 10x100 run will be corrected.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -443,7 +443,7 @@ jobs:
           - beam: 5x41
             detector_config: craterlake_5x41
           - beam: 10x100
-            detector_config: craterlake
+            detector_config: craterlake_10x100
           - beam: 18x275
             detector_config: craterlake_18x275
     steps:
@@ -851,7 +851,7 @@ jobs:
         - CXX: clang++
           beam: 10x100
           minq2: 0
-          detector_config: craterlake
+          detector_config: craterlake_10x100
         - CXX: clang++
           beam: 18x275
           minq2: 0


wdconinc (commented):
What about the craterlake_tracking_only? That's also apparently assuming 10x100 but not explicit about it in the configuration.

simonge:
> What about the craterlake_tracking_only? That's also apparently assuming 10x100 but not explicit about it in the configuration.

Good point, I haven't been checking tracking_only. That's a little trickier to address requiring a decision to be made with most options needing changes to epic.

Options:
1. Create a tracking_only configuration for all beam settings. Why stop there, all configurations using the FF/FB detectors should have all the options available.
2. Explicitly change craterlake_tracking_only to 10x100. This is different to the default 5x41 configuration in craterlake  so potentially confusing.
3. Change the craterlake_tracking_only CI to use 18x275 events. Currently defaults to 18x275 when no beam energies are specified which may also be confusing, this is the smallest possible change which would just need to be done in EICrecon but will effect the test itself if anyone is looking historically.
4. Change the jinja template to use 10x100 when no values are specified. Swaps one default value for another neither of which match the original.

In any case the default beam energy configuration should be expected to be the same everywhere while in actuality it appears to be 3 different values.

wdconinc (approved):

PR #1901, opened 2025-06-04
fix(ci): cancel eicweb pipeline if workflow canceled or failed

wdconinc:
This PR cancels the triggered eicweb pipeline when a GitHub workflow later fails or is canceled (for example by a new commit). This reduces the load on eicweb when multiple commits are submitted in relatively quick succession.

Note that canceling the container pipeline on eicweb does not cancel any benchmark pipelines it may have triggered. That's a problem for eicweb to solve, not here (arguably).

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No .

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -968,7 +968,7 @@ jobs:
     runs-on: ubuntu-24.04
     if: ${{ github.event_name != 'merge_group' && github.event_name != 'schedule' && github.actor != 'dependabot[bot]' }}
     needs:
-    - eicrecon-gun
+    - build
     outputs:
       json: ${{ steps.trigger.outputs.json }}
     steps:


.github/workflows/linux-eic-shell.yml
```diff
@@ -1233,3 +1235,16 @@ jobs:
           issue-number: ${{ github.event.pull_request.number }}
           body-file: capybara_${{ github.event.pull_request.number }}.md
           edit-mode: replace
+
+  cancel-container:
+    if: cancelled() || failure()
```
veprbl, line 1240:
```suggestion
    if: cancelled()
```
All relevant failures will not trigger container in the first place. All irrelevant ones do not matter.

wdconinc: `trigger-container` only depends on `eicrecon-gun`. If e.g. `eicrecon-dis` or so fails, then we want the container pipeline to be canceled as well.

.github/workflows/linux-eic-shell.yml
```diff
@@ -1233,3 +1235,16 @@ jobs:
...
+    needs:
+    - trigger-container
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Cancel container
+        run:
+          echo \
+          curl -X POST \
+             -F "token=${{ secrets.EICWEB_CONTAINER_TRIGGER }}" \
+             https://eicweb.phy.anl.gov/api/v4/projects/290/pipelines/${{ needs.trigger-container.outputs.json.id }}/cancel
```
wdconinc, line 1250: Unsure if json field evaluation this way (`.id`) works...

.github/workflows/linux-eic-shell.yml
```diff
@@ -1233,3 +1235,16 @@ jobs:
...
+    if: cancelled() || failure()
+    needs:
+    - trigger-container
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Cancel container
+        run:
+          echo \
+          curl -X POST \
+             -F "token=${{ secrets.EICWEB_CONTAINER_TRIGGER }}" \
```
wdconinc, line 1249: Probably not the correct token. This probably only allows triggering, not canceling.

.github/workflows/linux-eic-shell.yml
```diff
@@ -1233,3 +1235,16 @@ jobs:
           issue-number: ${{ github.event.pull_request.number }}
           body-file: capybara_${{ github.event.pull_request.number }}.md
           edit-mode: replace
+
+  cancel-container:
+    if: cancelled() || failure()
+    needs:
+    - trigger-container
```
wdconinc, line 1242: This likely `needs: deploy-docs` too to make sure it doesn't get skipped too early.

.github/workflows/linux-eic-shell.yml
wdconinc: Doesn't seem to work.

veprbl: It should work with `fromJSON()` https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#fromjson

.github/workflows/linux-eic-shell.yml
wdconinc:
```suggestion
             -F "token=${{ secrets.EICWEB_CONTAINER_CANCEL }}" \
```

wdconinc:
Triggered and canceled at https://eicweb.phy.anl.gov/containers/eic_container/-/pipelines/120790

.github/workflows/linux-eic-shell.yml
veprbl: No contradiction here. If it was not in dependencies, it didn't matter enough to be added.

veprbl (dismissed):
No strong feelings about applying my suggestion. If this works, this is good.

veprbl (approved):
This way is fine too.

PR #1903, opened 2025-06-05
MatchClusters: do not assign PDG code to trackless EM clusters

veprbl:
There were reports of ReconstructedParticle containing PDG=130 (K0L), which we don't identify. Convention is that we use PID information in Reconstructed* branches. That is already the case for tracks (ChargedParticles), but was never addressed for when we add in clusters in `MatchClusters`. For now the only clusters with PID are from EEEMCal, and only electron/pion separation is implemented which requires a track and does not belong in this branch, so we can't take that. So solution, for now, is to keep zero values.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes, users may observe missing PDG for neutral particles reconstructed.

Yes

### DIFF
--- a/src/algorithms/reco/MatchClusters.cc
+++ b/src/algorithms/reco/MatchClusters.cc
@@ -89,8 +89,8 @@ void MatchClusters::process(const MatchClusters::Input& input,
 
     // get mass/PDG from mcparticles, 0 (unidentified) in case the matched particle is charged.
     const auto mc     = (*mcparticles)[mcID];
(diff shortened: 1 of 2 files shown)


ruse-traveler:
If HCAL clusters could be provided to the algorithm, then we might be able to hack together a quick check on EM vs. hadronic energy to assign to assign a rough PID. But I think this is also a good stopgap...

Issue #1885, opened 2025-06-02
Make calorimeters-to-match-to configurable in `TrackClusterMatch`


ruse-traveler:
Currently, in the `TrackClusterMatch` algorithm, the system IDs of the calorimeters  to match to are hard-coded. To improve the utility of this algorithm, this will need to be changed.

**Describe the solution you'd like**

As a next step, it would be straightforward to convert these numbers to parameters and also include the surface indices as parameters. This algorithm then be run for each calorimeter we want matches to and the output consolidated via a call to `CollectionCollector` at the end.

Note that currently matching is done geometrically in eta-phi space. While this isn't ideal for forward/backward detectors, extending the range of matching-options could be deferred to a subsequent PR.

**Describe alternatives you've considered**

An alternative could be to provide the list of system IDs + surfaces to match as vector parameters, but this will make specifying how to match for each calorimeter tricky in the follow-up PR.

Another alternative could be to hard-code all calorimeter IDs. This is not ideal since it's (1) tough to maintain as our suite of calorimeters evolve, and (2) breaks our usual distinction of data vs. algorithms.

PR #1852, opened 2025-05-06
Removing insert

akioogawa:
### Briefly, what does this PR introduce?

Remove EcalEndcapPInsert since it will not be built
See : https://github.com/eic/epic/pull/855

I'm not sure if I should remove it from
https://github.com/eic/EICrecon/blob/c9ea43d6e21feaafe78bc9f992f3ca70abce51ef/src/services/io/podio/JEventProcessorPODIO.cc#L255
https://github.com/eic/EICrecon/blob/c9ea43d6e21feaafe78bc9f992f3ca70abce51ef/.github/janadot.groups#L7

How can I remove it from auto checks?

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?

No more EcalEndcapPInsert 

### Does this PR change default behavior?

No more EcalEndcapPInsert

PR #1906, opened 2025-06-09
Run track-cluster matching separately for each calorimeter

tlprotzman:
A suggested in #1885, the track-cluster matching should be done separately for each calorimeter.  Later algorithms can then manage the merging and splitting of clusters and integration across calorimeters.  This also makes it simpler to tune the matching criteria separately for each subsystem. 

- [ ] Bug fix (issue #__)
- [X] New feature (issue #1885 )
- [ ] Documentation update
- [ ] Other: __

There is no longer a single TrackClusterMatches collection output, instead is is broken into separate collections for each calorimeter
```
    EcalEndcapPTrackClusterMatches
    EcalEndcapPInsertTrackClusterMatches
    LFHCALTrackClusterMatches
    HcalEndcapPInsertClusterMatches
    EcalBarrelTrackClusterMatches
    HcalBarrelTrackClusterMatches
    EcalEndcapNTrackClusterMatches
    HcalEndcapNTrackClusterMatches
```

See above

### DIFF
--- a/src/algorithms/reco/TrackClusterMatch.cc
+++ b/src/algorithms/reco/TrackClusterMatch.cc
@@ -29,8 +29,7 @@ void TrackClusterMatch::process(const TrackClusterMatch::Input& input,
 
   // Validate the configuration
   if (m_cfg.matching_distance <= 0) {
-    throw std::runtime_error(
-        fmt::format("Invalid matching distance: {}", m_cfg.matching_distance));
+    throw std::runtime_error(fmt::format("Invalid matching distance: {}", m_cfg.matching_distance));
   }
   if (m_cfg.calo_id.empty()) {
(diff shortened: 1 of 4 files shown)


ruse-traveler (dismissed):
Very nice! This is a much appreciated improvement! The diffs for the relevant collections look like I'd expect them to, but I am seeing some diffs for a few collections I'd expect to be unaffected by this PR...

This seems to be similar to what we saw in #1919 ... So maybe the artifacts capybara are using out of date?

Screenshot (28)
Screenshot (29)

tlprotzman:
I just noticed that the EcalEndcapPInsert has been removed between starting this and now (#1852), so I removed its track cluster match factory

veprbl:
@ruse-traveler There are somehow irreproducibilities in B0 calorimetry, and they are not yet understood.

PR #1911, opened 2025-06-12
Fix deprecation warnings about JEventProcessor constructor

nathanwbrei:

JANA version 2.4.2 deprecates the JEventProcessor constructor that requires the JApplication pointer as an argument. The replacement is the default constructor. This eliminates boilerplate with no change in functionality.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No


wdconinc:
> > We aim to maintain some level of backwards compatibility. How does this change affect running with older JANA2 versions?
> 
> If it complies with older container on CI, it's enough?

It must mean some change to the minimum JANA2 version, though, right? We should at least document that.

wdconinc (approved):
Excellent, minimum version satisfied :-)

PR #1915, opened 2025-06-13
Secvertex amvf to be merged into main EICrecon repo

bishoyDHD:
This adds secondary vertex finder to the EICrecon.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No need to make changes to existing code. Full compliance with main EICrecon repo has been maintained throughout

No change to default behaviour


wdconinc:
Fails to build. We typically expect checks to complete before we do a review. Initial comments: I don't think we will want to add TProfile dependencies.

veprbl (commented):
(Not a review)

src/factories/tracking/SecondaryVertexFinder_factory.h
```diff
@@ -0,0 +1,60 @@
+// Created by Joe Osborn
+// Subject to the terms in the LICENSE file found in the top-level directory.
+//
```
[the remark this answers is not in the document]
bishoyDHD: For completeness, I have made the aforementioned suggestions. Still fails for checks beyond my control

src/algorithms/tracking/SecondaryVertexFinder.h
```diff
@@ -0,0 +1,338 @@
...
+                       vtx.fullCovariance()(1, 3), vtx.fullCovariance()(2, 3));
+    auto eicvertex = prmVertices->create();
+    eicvertex.setType(1); // boolean flag if vertex is primary vertex of event
+    eicvertex.setChi2((float)vtx.fitQuality().first); // chi2
+    eicvertex.setNdf((float)vtx.fitQuality().second); // ndf
+    eicvertex.setPosition({
+        (float)vtx.position().x(),
+        (float)vtx.position().y(),
+        (float)vtx.position().z(),
+        (float)vtx.time(),
```
wdconinc, line 174: Avoid C-style casts.

src/algorithms/tracking/SecondaryVertexFinder.h
```diff
@@ -0,0 +1,338 @@
...
+      float loc_a = par.localPosition().x();
+      float loc_b = par.localPosition().y();
+
+      for (const auto& part : *reconParticles) {
+        const auto& tracks = part.getTracks();
+        for (const auto& trk : tracks) {
+          const auto& traj    = trk.getTrajectory();
+          const auto& trkPars = traj.getTrackParameters();
+          for (const auto& par : trkPars) {
+            const double EPSILON = 1.0e-4; // mm
```
wdconinc, line 192: Use proper `std::numeric_limits::epsilon`.

src/algorithms/tracking/SecondaryVertexFinder.h
```diff
@@ -0,0 +1,338 @@
...
+                           secvertex.fullCovariance()(1, 3), secvertex.fullCovariance()(2, 3));
+        auto eicvertex = secVertices->create();
+        eicvertex.setType(0); // boolean flag if vertex is primary vertex of event
+        eicvertex.setChi2((float)secvertex.fitQuality().first); // chi2
+        eicvertex.setNdf((float)secvertex.fitQuality().second); // ndf
+        eicvertex.setPosition({
+            (float)secvertex.position().x(),
+            (float)secvertex.position().y(),
+            (float)secvertex.position().z(),
+            (float)secvertex.time(),
```
wdconinc, line 289: Avoid C-style casts.

src/algorithms/tracking/SecondaryVertexFinder.h
```diff
@@ -0,0 +1,338 @@
...
+          float loc_a = par.localPosition().x();
+          float loc_b = par.localPosition().y();
+
+          for (const auto& part : *reconParticles) {
+            const auto& tracks = part.getTracks();
+            for (const auto& trk : tracks) {
+              const auto& traj    = trk.getTrajectory();
+              const auto& trkPars = traj.getTrackParameters();
+              for (const auto& par : trkPars) {
+                const double EPSILON = 1.0e-4; // mm
```
wdconinc, line 311: Use `std::numeric_limits::epsilon`.

src/algorithms/tracking/SecondaryVertexFinder.h
```diff
@@ -0,0 +1,338 @@
...
+                if (fabs((par.getLoc().a / edm4eic::unit::mm) - (loc_a / Acts::UnitConstants::mm)) <
+                        EPSILON &&
+                    fabs((par.getLoc().b / edm4eic::unit::mm) - (loc_b / Acts::UnitConstants::mm)) <
+                        EPSILON) {
```
wdconinc, line 315: Avoid `fabs`, use `std::abs`. (also earlier)

src/factories/tracking/SecondaryVertexFinder_factory.h
```diff
@@ -0,0 +1,59 @@
...
+    m_algo = std::make_unique<AlgoT>();
+    m_algo->applyConfig(config());
+    m_algo->init(m_ACTSGeoSvc().actsGeoProvider(), logger());
+  }
+
+  void ChangeRun(int32_t /*run_number*/) {}
+
+  void Process(int32_t /*run_number*/, uint64_t /*event_number*/) {
+    std::tie(prm_vertices_output(), sec_vertices_output()) =
+        m_algo->produce(m_reco_input(), m_acts_trajectories_input());
```
wdconinc, line 52: Process will be `void`, so needs two tuples by reference, input and output.

src/factories/tracking/SecondaryVertexFinder_factory.h
```diff
@@ -0,0 +1,59 @@
...
+    /*
+    sec_vertices_output() = m_algo->produce(m_reco_input(),m_acts_trajectories_input());
+    */
```
wdconinc, line 55: Remove.

src/global/tracking/tracking.cc
```diff
@@ -214,6 +215,14 @@ void InitPlugin(JApplication* app) {
   app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::Track, true>>(
       "CombinedTracks", {"CentralCKFTracks", "TaggerTrackerTracks"}, {"CombinedTracks"}, app));
 
+  app->Add(new JOmniFactoryGeneratorT<SecondaryVertexFinder_factory>(
+      "AMVSecondaryTrackVertices", {"ReconstructedParticles", "CentralCKFActsTrajectories"},
+      {
+          "AMVPrimaryVertices",
+          "AMVSecondaryTrackVertices",
```
wdconinc, line 222: Don't prefix with name of algorithm. If anything, suffix. Be consistent and don't call something `PrimaryVertices` but the other thing `SecondaryTrackVertices`. Use Track or not, but consistently.

wdconinc (commented):
Main comment for discussion: name of output collection. I'd suggest `SecondaryVertices`.

Rest is style and maintainability.

src/factories/tracking/SecondaryVertexFinder_factory.h
bishoyDHD: In this case `std::tie` already takes cares of this, and is able to output `std::tuple` which is then written to the output `ROOT file`. Otherwise this would fail at `runtime`.

wdconinc: No, the function signature will be different when you use the algorithms interface as requested elsewhere.

src/algorithms/tracking/SecondaryVertexFinder.h
```diff
@@ -0,0 +1,338 @@
...
+#include "Acts/Vertexing/AdaptiveMultiVertexFinder.hpp"
+#include "ActsExamples/EventData/Trajectories.hpp"
+#include "ActsGeometryProvider.h"
+#include "DD4hepBField.h"
+#include "SecondaryVertexFinderConfig.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+class SecondaryVertexFinder
+    : public eicrecon::WithPodConfig<eicrecon::SecondaryVertexFinderConfig> {
```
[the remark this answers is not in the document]
bishoyDHD: After pushing `git precomp running` produces the following error at bottom of this message(non my work machine and also BNL):

For the new algorithms, the `init(){}` is a default function. However, the older algorithm interface, `init(m_ACTSGeoSvc().actsGeoProvider(), logger())` takes two inputs.  I need and use both inputs, can I revert to the older interface or is something that I am missing with newer interface?

ERROR:
======

`/home/runner/work/EICrecon/EICrecon/src/algorithms/tracking/SecondaryVertexFinder.h:60:8: error: 'eicrecon::SecondaryVertexFinder::init' hides overloaded virtual function [-Werror,-Woverloaded-virtual]
  void init(std::shared_ptr<const ActsGeometryProvider> geo_svc,
       ^
/opt/local/include/algorithms/algorithm.h:87:16: note: hidden overloaded virtual function 'algorithms::Algorithm<algorithms::Input<edm4eic::ReconstructedParticleCollection, std::vector<const ActsExamples::Trajectories *>>, algorithms::Output<edm4eic::VertexCollection, edm4eic::VertexCollection>>::init' declared here: different number of parameters (0 vs 2)
  virtual void init() {}`

bishoyDHD:
Alright folks. Please help me out here. This online precompile check keeps asking for me to `iwyu` by:
`#include <format>`

When I do, the precompiler cannot even find it. I had to upgrade my container in order to be able compile while having the `#include <format>`. I mean, the code is really ready to be reviewed and merged.

wdconinc:
> I had to upgrade my container in order to be able compile while having the `#include <format>`.

Yes, you should run with an up to date container.

> I mean, the code is really ready to be reviewed and merged.

Last time I looked, today afternoon, there was still no algorithms interface as requested.

bishoyDHD:
> > I had to upgrade my container in order to be able compile while having the `#include <format>`.
> 
> Yes, you should run with an up to date container.
> 
> > I mean, the code is really ready to be reviewed and merged.
> 
> Last time I looked, today afternoon, there was still no algorithms interface as requested.

Regarding the algorithms interface:
I have tested out the Algorithm interface last week. The problem, it adds a level of complexity that is not efficient for the secondary vertex setup. For example, I have functions that need to independently calculate the Primary and Secondary Vertex using the AMVF. Using the Algorithm::Input & Algorithm::Output means that I would have to unwrap those types and then pass them to the respective functions. That is just not efficient.

veprbl:
@wdconinc I know that @bishoyDHD did a good faith attempt at adapting his code to the algorithms interface. As you know, there are objective issues with interfacing `ACTS_LOCAL_LOGGER` with algorithms that are the same for existing CKFTracking code. The issues with passing code to functions I can't testify, but we probably can relatively easily address in further refactoring. My proposal is to review PR without asking for algorithms interface. Likely, it will be to either me or you to take on this technical debt.

wdconinc:
> @wdconinc I know that @bishoyDHD did a good faith attempt at adapting his code to the algorithms interface. As you know, there are objective issues with interfacing `ACTS_LOCAL_LOGGER` with algorithms that are the same for existing CKFTracking code. The issues with passing code to functions I can't testify, but we probably can relatively easily address in further refactoring. My proposal is to review PR without asking for algorithms interface. Likely, it will be to either me or you to take on this technical debt.

That's not the objection that's been raised. I'm ok with keeping an extra helper function that passes a spdlog logger from the factory to the algorithm, but there is no technical reason that the rest can't be using the algorithms interface.

src/algorithms/tracking/SecondaryVertexFinder.cc
```diff
@@ -0,0 +1,419 @@
...
+  m_log            = log;
+  auto& serviceSvc = algorithms::ServiceSvc::instance();
+  m_geoSvc         = serviceSvc.service<algorithms::ActsSvc>("ActsSvc")->acts_geometry_provider();
+  m_BField =
+      std::dynamic_pointer_cast<const eicrecon::BField::DD4hepBField>(m_geoSvc->getFieldProvider());
+  m_fieldctx = eicrecon::BField::BFieldVariant(m_BField);
+}
+
+std::tuple<std::unique_ptr<edm4eic::VertexCollection>, std::unique_ptr<edm4eic::VertexCollection>>
+eicrecon::SecondaryVertexFinder::produce(
```
wdconinc, line 55: Please use the function signature `void process(const Input&, const Output&) const`. In particular, this function must be void, just be const, and should not create its own collections.

veprbl:
> > @wdconinc I know that @bishoyDHD did a good faith attempt at adapting his code to the algorithms interface. As you know, there are objective issues with interfacing `ACTS_LOCAL_LOGGER` with algorithms that are the same for existing CKFTracking code. The issues with passing code to functions I can't testify, but we probably can relatively easily address in further refactoring. My proposal is to review PR without asking for algorithms interface. Likely, it will be to either me or you to take on this technical debt.
> 
> That's not the objection that's been raised. I'm ok with keeping an extra helper function that passes a spdlog logger from the factory to the algorithm, but there is no technical reason that the rest can't be using the algorithms interface.

I wrote "passing code" instead of "passing pointers [to other methods of the algorithm]". My assumption is that it's a trivial change, but there is some C++ gotcha that tripped up @bishoyDHD in his attempt. I'd rather work with code that is final and effective before doing the refactoring we are talking about.

veprbl:
Looks like we have two comments unaddressed from the first round of review. One on using spdlog, which we might punt on. And another about units of initialVariances and moving that to config.

bishoyDHD:
> Looks like we have two comments unaddressed from the first round of review. One on using spdlog, which we might punt on. And another about units of initialVariances and moving that to config.

Moved to Config

src/algorithms/tracking/SecondaryVertexFinder.cc
```diff
@@ -0,0 +1,414 @@
...
+                    EPSILON) {
+              m_log->trace(
+                  "From ReconParticles, track local position [Loc a, Loc b] = {} mm, {} mm",
+                  par.getLoc().a / edm4eic::unit::mm, par.getLoc().b / edm4eic::unit::mm);
+              eicvertex.addToAssociatedParticles(part);
+            } // endif
+          } // end for par
+        } // end for trk
+      } // end for part
+    } // end for t
```
wdconinc, line 232: This whole matching is not ideal, and I'd like to see it removed instead of getting into our code base and left for others to refactor later.

The input to seed finding is (only) a set of track parameters. You have these track parameters as input in the form of ReconstructedParticles. But you ALSO pass and use the ActsExamples::Trajectories (which we are actively trying to get away from...). Then after seed finding you have to run this nested set of loops for matching with ReconstructedParticles. Why not start from ReconstructedParticles to fill your input tracks? Then when you get the vertices they will point to the tracks that are filled parallel to the ReconstructedParticles vector. There will be no matching loops necessary.

wdconinc (commented):
Two major remaining comments from my side at this point:
- We should move the creation and initialization of algorithmic structures that are event-independent into the init function, not recreate them for each event in the process function.
- We should not base this algorithm on the input from ActsExamples::Trajectories, but on the input from ReconstructedParticles. That will avoid the combinatorics of the matching loop. (Even if you don't do this, the matching loop should use an early break.)

src/algorithms/tracking/SecondaryVertexFinder.cc
veprbl: We told @bishoyDHD that copy-pasting the https://github.com/eic/EICrecon/blob/246dae58831f605639f8086abab1a0b450d093cb/src/algorithms/tracking/IterativeVertexFinder.cc#L155-L182 is an acceptable workaround to the issue of associating to PODIO. I don't think writing a routine to convert ReconstructedParticles to Trajectories is a useful exercise, if we plan to move to PODIO for Acts EDM anyway. Even once we have that, there will be some matching loop to have.

wdconinc: > I don't think writing a routine to convert ReconstructedParticles to Trajectories is a useful exercise.

Not what I'm suggesting. Look at the code. Instead of getTrackParameters from Trajectories, you can getTrackParameters from ReconstructedParticles, instead of doing exactly that in the loop later.

veprbl: Or the 3D position? But then we need to convert using bound surface.

wdconinc: It's in the loop here in the segment that the comment is about:
```
   for (const auto& part : *reconParticles) { 
     const auto& tracks = part.getTracks(); 
     for (const auto& trk : tracks) { 
       const auto& traj    = trk.getTrajectory(); 
       const auto& trkPars = traj.getTrackParameters();
```
i.e. [edm4eic::ReconstructedParticle::getTracks](https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L257), [edm4eic::Track::getTrajectory](https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L468), [edm4eic::Trajectory::getTrackParameters](https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml#L432)

wdconinc: The idea is this: https://github.com/eic/EICrecon/pull/1992 (not even tested compilation, and has some remaining FIXMEs).

wdconinc: > non-existent conversion routine

In the sense that CKFTracking does this already (minus surface lookup since it's a perigee that is re-created).

src/algorithms/tracking/SecondaryVertexFinder.cc
```diff
@@ -0,0 +1,414 @@
...
+
+  if (m_cfg.useTime) {
+    // When using time, we have an extra contribution to the chi2 by the time
+    // coordinate.
+    vertexfinderConfigSec.tracksMaxSignificance      = m_cfg.tracksMaxSignificance;
+    vertexfinderConfigSec.maxMergeVertexSignificance = m_cfg.maxMergeVertexSignificance;
+  }
+
+  vertexfinderConfigSec.extractParameters.connect<&Acts::InputTrack::extractParameters>();
+  vertexFitterConfigSec.trackLinearizer.connect<&LinearizerSec::linearizeTrack>(&linearizerSec);
```
veprbl: This is a use-after-move?

src/algorithms/tracking/SecondaryVertexFinderConfig.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <Acts/Definitions/Units.hpp>
+
+namespace eicrecon {
+
+struct SecondaryVertexFinderConfig {
+  int maxVertices        = 20;
+  float maxIterations    = 500;
+  float maxSecIterations = 1000;
```
veprbl:
```suggestion
  unsigned int maxVertices        = 20;
  unsigned int maxIterations    = 500;
  unsigned int maxSecIterations = 1000;
```

src/algorithms/tracking/SecondaryVertexFinder.cc
```diff
@@ -0,0 +1,414 @@
...
+
+  if (m_cfg.useTime) {
+    // When using time, we have an extra contribution to the chi2 by the time
+    // coordinate.
+    vertexfinderConfigSec.tracksMaxSignificance      = m_cfg.tracksMaxSignificance;
+    vertexfinderConfigSec.maxMergeVertexSignificance = m_cfg.maxMergeVertexSignificance;
+  }
+
+  vertexfinderConfigSec.extractParameters.connect<&Acts::InputTrack::extractParameters>();
+  vertexfinderConfigSec.trackLinearizer.connect<&LinearizerSec::linearizeTrack>(&linearizerSec);
```
veprbl: Turns out this was not a thing

I don't see this in Acts example
https://github.com/acts-project/acts/blob/a8e9f5cb396c27a999b5bc052635581bf46063fc/Examples/Algorithms/Vertexing/src/AdaptiveMultiVertexFinderAlgorithm.cpp#L187-L188

```suggestion
```

veprbl:
We'll need to update this following
 - [x] #2224
 - [x] #2311
 - [x] #2317

src/algorithms/tracking/SecondaryVertexFinderConfig.h
```diff
@@ -0,0 +1,39 @@
...
+  double maxVertexChi2 = 18.42;
+  bool doFullSplitting = false;
+  // 5 corresponds to a p-value of ~0.92 using `chi2(x=5,ndf=2)`
+  float tracksMaxSignificance      = 6.7;
+  float maxMergeVertexSignificance = 3;
+  float minWeight                  = 1e-04;
+  float maxDistToLinPoint          = 5.5 * Acts::UnitConstants::mm;
+  // Bin extent in z-direction
+  float spatialBinExtent         = 25 * Acts::UnitConstants::um;
+  Acts::Vector4 initialVariances = Acts::Vector4{1e+2, 1e+2, 1e+2, 1e+8};
```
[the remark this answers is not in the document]
veprbl: Done.

src/algorithms/tracking/SecondaryVertexFinder.h
```diff
@@ -0,0 +1,91 @@
...
+                          algorithms::Output<edm4eic::VertexCollection, edm4eic::VertexCollection>>;
+
+class SecondaryVertexFinder : public SecondaryVertexFinderAlgorithm,
+                              public WithPodConfig<eicrecon::SecondaryVertexFinderConfig> {
+public:
+  SecondaryVertexFinder(std::string_view name)
+      : SecondaryVertexFinderAlgorithm{name,
+                                       {"inputReconstructedParticles", "inputActsTrajectories"},
+                                       {"outputPrimaryVertices", "outputSecondaryVertices"},
+                                       ""} {}
```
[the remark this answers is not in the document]
veprbl: Not anymore. Thanks.

ruse-traveler (approved):
Following discussion with @marrbnl and @Simple-Shyam, this is ready to go! There are 2 items to follow-up on for the next maintainer:

- Deduplication of code between vertexing modes (not critical, but preferable) and integration of the Secondary Vertex EDM when ready (critical)
- Validation of algorithm performance, e.g. checking the kaon invariant mass distribution

Issue #1919, opened 2025-06-18
Consistency of energy with momentum and mass/PDG code for ReconstructedChargedParticles


bschmookler:
As we discussed at the Inclusive meeting yesterday, several people have noticed issues with the kinematic reconstruction with hadronic final-state variables. One comment was that this could be related to the Lorentz Rotations applied before calculating the kinematic quantities.

I noticed one issue, which could be minor; but we may still want to correct it. For the ReconstructedChargedParticles, the energy is set equal to the magnitude of the track momentum prior to adding the PID hypothesis:

https://github.com/eic/EICrecon/blob/5954c13dedabef38a339e48b0e72fb560fcc4a44/src/algorithms/tracking/TracksToParticles.cc#L43

If a non-zero PDG ID is set, the energy is not updated. It remains equal to the magnitude of the track momentum, rather than being equal to sqrt(p^2 + m^2).

https://github.com/eic/EICrecon/blob/5954c13dedabef38a339e48b0e72fb560fcc4a44/src/algorithms/pid_lut/PIDLookup.cc#L142-L145

Since for charged particles the ReconstructedParticles collection just copies the ReconstructedChargedParticles information, this energy inconsistency remains in that final collection.

It is quite easy to resolve this. I just did it in this branch:

https://github.com/eic/EICrecon/tree/ReconstructedChargedParticles_energy

But it depends a bit what we want to set the energy equal to for charged particles in the ReconstructedParticles collection. Maybe we want to set it equal to the associated cluster energy, with the momentum vector for the track, and the PDG/mass from the PID hypothesis. But until we decide, leaving the energy equal to magnitude of track momentum is not the best choic.

@tylerkutz @smaple1 @ruse-traveler

bschmookler:
Okay, I will make the pull request. I agree that this is an interim solution for the charged particles.

PR #1928, opened 2025-06-23
Fix: Energy consistency with momentum and mass/PDG

bschmookler:
This resolves the issue detailed in #1919. As discussed there, for the ReconstructedChargedParticles, this sets the energy to be consistent with the momentum and mass/PDG. 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

It will change the reconstructed energy in the ReconstructedChargedParticles collection and the reconstructed energy for charged particles in the ReconstructedParticles Collection.


bschmookler:
> Looks good! The diffs all look like I'd expect them to, except, oddly, for the B0 ECal Cluster Associations for the MT samples... Would we expect these to change for any reason?

I don't think this PR should have any effect, since it just updates the energy when a ReconstructedChargedParticle has a non-zero PDG code.

ruse-traveler:
> I don't think this PR should have any effect, since it just updates the energy when a ReconstructedChargedParticle has a non-zero PDG code.

Same... And it looks like its `ObjectID`s shifting around (namely the different collection ID)... So maybe it's out-of-date like you mentioned offline or it's picking up changes from somewhere else? 🤷

PR #1934, opened 2025-06-27
feat: ingest EventHeader into digi, generate UID to reseed RNG

wdconinc:
One challenge of multi-threaded/concurrent running is that the random number generators are called in no predictable order; this affects both events which can come out of order, or algorithms within the same event which can be called in different orders for different branches. That results in non-reproducibility.

This PR ingests the event header into digitization algorithms (which is our main use for random sampling). The event header is combined with the algorithm name into a random seed for that event/algorithm combination. With this seed, the random number generator is reseeded.

Outside the scope of this PR is reverting back to a single random service instead of the various algorithm specific approaches.

Still TODO in this algorithm:
- [x] expose the top-level seed used to generate event-level seeds as a property for setting through JANA.

- [x] Bug fix (issue: lack of reproducibility for multi-threaded running)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes. It results in a different random number sequence, so all distributions will be shuffled.


src/algorithms/interfaces/UniqueIDGenSvc.h
```diff
@@ -0,0 +1,101 @@
...
+    return getUniqueID(evt_header.getEventNumber(), evt_header.getRunNumber(), name);
+  }
+  // calculate a unique id from name and the first element of collection
+  size_t getUniqueID(const edm4hep::EventHeaderCollection& evt_headers,
+                     const std::string_view& name) const {
+    return getUniqueID(evt_headers.at(0), name);
+  }
+  // calculate a unique id from name and the event and run numbers
+  size_t getUniqueID(const event_num_t evt_num, const run_num_t run_num,
+                     const std::string_view& name) const {
```
wdconinc, line 44: Not strictly necessary to ingest name here, since can pass that on `init`, not `process`, but this is an attempt at sticking close to the k4FWCore approach. Only difference is that we take a `string_view` instead of a `string` (arguably they will want to move to `string_view` anyway).

wdconinc:
Interesting... The new plots for MT and non-MT should now be identical... But aren't.
- https://eicrecon.epic-eic.org/pr/1934/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275_MT/index.html#B0ECalClusters
- https://eicrecon.epic-eic.org/pr/1934/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#B0ECalClusters

Maybe I'm missing some reseeding somewhere.

wdconinc:
> Interesting... The new plots for MT and non-MT should now be identical... But aren't.
> 
> * https://eicrecon.epic-eic.org/pr/1934/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275_MT/index.html#B0ECalClusters
> * https://eicrecon.epic-eic.org/pr/1934/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#B0ECalClusters
> 
> Maybe I'm missing some reseeding somewhere.

Or there's a data race, https://github.com/eic/EICrecon/actions/runs/15937325592/job/44960208488?pr=1934#step:7:928 ...

src/services/unique_id/UniqueIDGen_service.h
wdconinc, line 1: This is entirely generalizable to an `eicrecon::Algorithm_service<algorithms::UniqueIDGenSvc>`. That then allows services in algorithms to be exposed to JANA2, just as we can already expose JANA2 services through the algorithms interface to algorithms.

src/algorithms/interfaces/UniqueIDGenSvc.h
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024 Wouter Deconinck
```
veprbl, line 2:
```suggestion
// Copyright (C) 2025 Wouter Deconinck
```
Other authors?

src/algorithms/interfaces/UniqueIDGenSvc.h
```diff
@@ -0,0 +1,101 @@
...
+#include <algorithms/service.h>
+#include <bitset>
+#include <map>
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+namespace algorithms {
+
+class UniqueIDGenSvc : public LoggedService<UniqueIDGenSvc> {
```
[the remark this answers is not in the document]
wdconinc: I don't think there is a need for an algorithm-specific seed, but it could be added as additional salt. That would then allow only one algorithm to change its salt independently, but that salt on its own would not be sufficient for reproducibility. It would have to be specified alongside the main seed.

src/algorithms/interfaces/UniqueIDGenSvc.h
wdconinc: This comment was for information only. Resolving.

src/algorithms/digi/PhotoMultiplierHitDigi.cc
```diff
@@ -102,10 +88,10 @@ void PhotoMultiplierHitDigi::process(const PhotoMultiplierHitDigi::Input& input,
     trace(" -> hit accepted");
     trace(" -> MC hit id={}", sim_hit.getObjectID().index);
     auto time  = sim_hit.getTime();
-    double amp = m_cfg.speMean + m_rngNorm() * m_cfg.speError;
+    double amp = m_cfg.speMean + gaussian(generator) * m_cfg.speError;
```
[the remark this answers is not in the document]
veprbl: @wdconinc I don't like its suggestions. Do you think we could disable?

.github/workflows/linux-eic-shell.yml
```diff
@@ -929,6 +929,7 @@ jobs:
           export JANA_PLUGIN_PATH=$PWD/install/lib/EICrecon/plugins:/usr/local/plugins
           prmon --json-summary rec_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.prmon.json -- \
           $PWD/install/bin/eicrecon ${{env.JANA_OPTIONS}} ${{matrix.nthreads > 1 && format('-Pnthreads={0}', matrix.nthreads) || ''}} -Ppodio:output_file=rec_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4eic.root sim_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root -Pacts:WriteObj=true -Pacts:WritePly=true -Pplugins=janadot,janatop $(<${{ github.workspace }}/.github/janadot.groups)
+      continue-on-error: ${{matrix.nthreads > 1}}
```
veprbl, line 932: Looks like it will ignore runtime errors for mt running. Why do we want that?

src/algorithms/digi/PhotoMultiplierHitDigi.cc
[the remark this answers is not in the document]
wdconinc: I don't feel strongly. It is true that most people here know relative precedence of `*` and `+` at an above average level compared to typical programmers, but I think they know the relative precedence of bitwise and logical operators at a level below typical programmers.

I don't think the parentheses make the code worse (about the worst from the auto-fixes are `(v.x * p.x + v.y * p.y) / (p.x * p.x + p.y * p.y)` turned into `((v.x * p.x) + (v.y * p.y)) / ((p.x * p.x) + (p.y * p.y))`, which I think is still acceptable).

I wish we could disable this for `*` and `+` in one line arithmetic, but we can't. We don't have many (or any) bitwise and logical operator expressions that are flagged by this.

So, I guess I'm ok with disabling until we can add more fine-grained treatment.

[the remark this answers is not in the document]
wdconinc: `Mxz * (Mxz * Myy - Myz * Mxy) + Myz * (Myz * Mxx - Mxz * Mxy) - Var_z * Cov_xy` is another one that looks a bit more annoying with extra parentheses.

.github/workflows/linux-eic-shell.yml
wdconinc: This was intended for the initial phases of roll-out. I would prefer we first get some more gentle notifications of breakage rather than ending up with broken CI on main due to some algorithm that is thread-unsafe but that hasn't shown up yet in this branch.

src/services/unique_id/UniqueIDGen_service.cc
```diff
@@ -0,0 +1,19 @@
...
+UniqueIDGen_service::UniqueIDGen_service(JApplication* app)
+    : m_app(app), m_uniqueGenIDSvc(algorithms::UniqueIDGenSvc::instance()) {
+  for (const auto& [key, prop] : m_uniqueGenIDSvc.getProperties()) {
+    std::visit(
+        [this, key = key](auto&& val) {
+          m_app->SetDefaultParameter(std::string(key), val); // FIXME add description
+          this->m_uniqueGenIDSvc.setProperty(key, val);
+        },
+        prop.get());
+  }
```
veprbl, line 18: The `UniqueIDGen_service` implements a JANA2 service that is able to
1. Initialize algorithms-style service (`UniqueIDGenSvc`)
2. Capture a `JApplication*` and use it to initialize parameters

There is already such a JANA2 service in EICrecon that does 1. it's `AlgorithmsInit_service` itself. We can move responsibility 2. there and get rid of this extra indirection. What do you think?

veprbl: Oh, that's a fair point. And I agree, we better stick with the current ongoing approach of shoving everything in `AlgorithmsInit_service`, and refactor uniformly later.

.github/workflows/linux-eic-shell.yml
veprbl: If you error and continue in eicrecon, it will not produce an artifact, so `compare-single-multi-threaded` will not be able to download it. That was my thinking.

src/services/unique_id/UniqueIDGen_service.cc
wdconinc: Done in 00b7f2d5e.

Issue #1944, opened 2025-07-02
Not all TOFBarrelHits are converted to TOFBarrelSharedHits by SiliconChargeSharing class


ssedd1123:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released):
- Which version (or `HEAD` for the most recent on git):
- Any specific OS or system where the issue occurs?
- Any special versions of ROOT or Geant4?

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Create simulation files with either dd4hep or npsim. For my own tests, I simulated 500 pi+ events with 
    `npsim --compactFile $DETECTOR_PATH/epic.xml --enableGun --gun.distribution uniform --gun.particle pi+ --numberOfEvents 500 --outputFile output.edm4hep.root`
2. Reconstruct with charge sharing magnitude set to zero,
    `eicrecon -Ppodio:output_collections=TOFBarrelHits,TOFBarrelSharedHits -PBTOF:TOFBarrelSharedHits:sigmaSharingX=0 -PBTOF:TOFBarrelSharedHits:sigmaSharingY=0 output.edm4hep.root`
3. Open the `podio_output.root` and count the number of hits in `TOFBarrelHits `and `TOFBarrelSharedHits`,
    `root podio-output.root`
     `root [1] events->Draw("TOFBarrelHits.position.z", "", "") // the .position.z part doesn't matter. We just want to see the number of entries in the histogram`
     `root [2] events->Draw("TOFBarrelSharedHits.position.z", "", "")`

### Expected Result: (what do you expect when you execute the steps above)

Since `BTOF:TOFBarrelSharedHits:sigmaSharingX=0` and `BTOF:TOFBarrelSharedHits:sigmaSharingY=0`, we expect the charge sharing class to not share any charge with neighbors and just pass all `TOFBarrelHits` along. The number of entries in `TOFBarrelHits.position.z` should be equal to that in `TOFBarrelSharedHits.position.z`.

### Actual Result: (what do you get when you execute the steps above)

There is always less hit in `TOFBarrelSharedHits.position.z` than `TOFBarrelHits.position.z`. SiliconChargeSharing class ignore some hits by mistakes. In my simulation, I get 830 entries in the former and 850 in the latter. The exact number differ for each simulation, but the disagreement persists for all simulations.

PR #1945, opened 2025-07-02
Modified SiliconChargeSharing so it won't miss any hits

ssedd1123:

The `SiliconChargeSharing` terminates it's iteration on `findAllNeighborsInSensor` when `edep < m_cfg.min_edep` for a test cell. The problem is that when the iteration ends, it will also terminate its search for neighbors, and just because `edep < m_cfg.min_edep` for a particular cell, that doesn't mean `edep < m_cfg.min_edep` is true for their neighbors. 

To fix this, I let `SiliconChargeSharing` to keep searching for neighbors even for cells with  `edep < m_cfg.min_edep`. `edep < m_cfg.min_edep` is now only used to determine if a shared hits to be created. 

- [x] Bug fix (issue #1944 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/digi/SiliconChargeSharing.cc
+++ b/src/algorithms/digi/SiliconChargeSharing.cc
@@ -179,10 +179,16 @@ dd4hep::Position SiliconChargeSharing::global2Local(const dd4hep::Position& glob
 float SiliconChargeSharing::energyAtCell(const double xDimension, const double yDimension,
                                          const dd4hep::Position localPos,
                                          const dd4hep::Position hitPos, const float edep) const {
+  auto sigma_sharingx = m_cfg.sigma_sharingx;
+  auto sigma_sharingy = m_cfg.sigma_sharingy;
+  if (m_cfg.sigma_mode == SiliconChargeSharingConfig::ESigmaMode::rel) {
+    sigma_sharingx *= xDimension;
+    sigma_sharingy *= yDimension;
+  }
   float energy = edep *
-                 integralGaus(hitPos.x(), m_cfg.sigma_sharingx, localPos.x() - 0.5 * xDimension,
+                 integralGaus(hitPos.x(), sigma_sharingx, localPos.x() - 0.5 * xDimension,
                               localPos.x() + 0.5 * xDimension) *
-                 integralGaus(hitPos.y(), m_cfg.sigma_sharingy, localPos.y() - 0.5 * yDimension,
+                 integralGaus(hitPos.y(), sigma_sharingy, localPos.y() - 0.5 * yDimension,
                               localPos.y() + 0.5 * yDimension);
   return energy;
 }
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -16,6 +16,7 @@ Checks: '
   -readability-function-cognitive-complexity,
   -readability-identifier-length,
   -readability-magic-numbers,
+  -readability-math-missing-parentheses,
   -cppcoreguidelines-avoid-c-arrays,
   -cppcoreguidelines-avoid-magic-numbers,
   -cppcoreguidelines-narrowing-conversions,
--- a/src/algorithms/digi/SiliconChargeSharingConfig.h
+++ b/src/algorithms/digi/SiliconChargeSharingConfig.h
@@ -7,10 +7,42 @@ namespace eicrecon {
 
 struct SiliconChargeSharingConfig {
   // Parameters of Silicon signal generation
+  // determines the meaning of sigma_sharingx and y.
+  // rel means relative, so charge sharing range = sigma_sharingx * cell_width_x, etc for y
+  // abs means absolute, charge sharing range = sigma_shargeingx directly
+  enum ESigmaMode { abs = 0, rel = 1 } sigma_mode = abs;
   float sigma_sharingx;
   float sigma_sharingy;
   float min_edep;
   std::string readout;
 };
+
+std::istream& operator>>(std::istream& in, SiliconChargeSharingConfig::ESigmaMode& sigmaMode) {
+  std::string s;
+  in >> s;
+  // stringifying the enums causes them to be converted to integers before conversion to strings
+  if (s == "abs" or s == "0") {
+    sigmaMode = SiliconChargeSharingConfig::ESigmaMode::abs;
+  } else if (s == "rel" or s == "1") {
+    sigmaMode = SiliconChargeSharingConfig::ESigmaMode::rel;
+  } else {
+    in.setstate(std::ios::failbit); // Set the fail bit if the input is not valid
+  }
+
+  return in;
+}
+std::ostream& operator<<(std::ostream& out, SiliconChargeSharingConfig::ESigmaMode& sigmaMode) {
+  switch (sigmaMode) {
+  case SiliconChargeSharingConfig::ESigmaMode::abs:
+    out << "abs";
+    break;
+  case SiliconChargeSharingConfig::ESigmaMode::rel:
+    out << "rel";
+    break;
+  default:
+    out.setstate(std::ios::failbit);
+  }
+  return out;
+}
 
 } // namespace eicrecon
--- a/src/detectors/BTOF/BTOF.cc
+++ b/src/detectors/BTOF/BTOF.cc
@@ -7,11 +7,14 @@
 
 #include <Evaluator/DD4hepUnits.h>
 #include <JANA/JApplicationFwd.h>
+#include <JANA/Utils/JTypeInfo.h>
 #include <TMath.h>
 #include <edm4eic/unit_system.h>
-#include <memory>
+#include <cmath>
+#include <string>
+#include <vector>
 
-#include "algorithms/interfaces/WithPodConfig.h"
+#include "algorithms/digi/SiliconChargeSharingConfig.h"
 #include "extensions/jana/JOmniFactoryGeneratorT.h"
 #include "factories/digi/EICROCDigitization_factory.h"
 #include "factories/digi/PulseCombiner_factory.h"
@@ -48,8 +51,9 @@ void InitPlugin(JApplication* app) {
   app->Add(new JOmniFactoryGeneratorT<SiliconChargeSharing_factory>(
       "TOFBarrelSharedHits", {"TOFBarrelHits"}, {"TOFBarrelSharedHits"},
       {
-          .sigma_sharingx = 0.1 * dd4hep::mm,
-          .sigma_sharingy = 0.5 * dd4hep::cm,
+          .sigma_mode     = SiliconChargeSharingConfig::ESigmaMode::rel,
+          .sigma_sharingx = 1,
+          .sigma_sharingy = 0.5,
           .min_edep       = 0.0 * edm4eic::unit::GeV,
           .readout        = "TOFBarrelHits",
       },
--- a/src/factories/digi/SiliconChargeSharing_factory.h
+++ b/src/factories/digi/SiliconChargeSharing_factory.h
@@ -26,6 +26,8 @@ private:
   ParameterRef<float> m_sigma_sharingy{this, "sigmaSharingY", config().sigma_sharingy};
   ParameterRef<float> m_min_edep{this, "minEDep", config().min_edep};
   ParameterRef<std::string> m_readout{this, "readout", config().readout};
+  ParameterRef<eicrecon::SiliconChargeSharingConfig::ESigmaMode> m_sigma_mode{this, "sigmaMode",
+                                                                              config().sigma_mode};
 
 public:
   void Configure() {


simonge:
I'm not sure I follow how a neighbouring cell could have a higher charge deposited than the cell being checked. Particularly for the demo case in your issue where the only cell which should be checked is the only cell with a hit in to start with. I think it's more likely to be a coordinate transformation and cell matching issue.

I don't think we should remove the termination like this as for detectors with large number of pixels it will have to check them all.

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -108,19 +108,18 @@ void SiliconChargeSharing::findAllNeighborsInSensor(
   auto yDimension = segmentation->gridSizeY();
   // Calculate deposited energy in cell
   float edepCell = energyAtCell(xDimension, yDimension, localPos, hitPos, edep);
-  if (edepCell <= m_cfg.min_edep) {
-    return;
```
veprbl, line 112: If you are making this change, then, please, update documentation on line 81.

ssedd1123:
> I'm not sure I follow how a neighbouring cell could have a higher charge deposited than the cell being checked. Particularly for the demo case in your issue where the only cell which should be checked is the only cell with a hit in to start with. I think it's more likely to be a coordinate transformation and cell matching issue.
> 
> I don't think we should remove the termination like this as for detectors with large number of pixels it will have to check them all.

I reverted back to terminating the iteration when `edep < m_cfg.min_edep`. But I now implemented a breath first search algorithm to look for the correct cellID that corresponds to `globalHitPos` before finding the neighbors. 

If the correct cell is close to `hit.getCellID()`, then BFS should be faster than DFS like what was done in `findAllNeighborsInSensors`, although the time complexity should be the same.

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -74,6 +76,13 @@ void SiliconChargeSharing::process(const SiliconChargeSharing::Input& input,
                                       globalHitPos.z * dd4hep::mm),
                      transformIt->second);
 
+    // globalHitPos can sometimes lie outside of cell with cellID given by hit.getCellID()
```
wdconinc, line 79: When does this occur?

ssedd1123: I can't say when exactly, but I confirm that this bug comes directly from `TOFBarrelHits` and not from the charge-sharing class. To see that, run simulation of pions like I suggested in the bug report, then compare `cellID `from the hit themselves with `cellID `estimated from hit position.

```
for(const auto& hit : *tofHits) { // tofHits is the collection for TOFBarrelHits
  auto cellID = hit.getCellID(); 
  auto globalHitPos = hit.getPosition();
  auto posID = m_converter -> cellID(dd4hep::Position(globalHitPos.x * dd4hep::mm, globalHitPos.y * dd4hep::mm, globalHitPos.z * dd4hep::mm));
  // Compare cellID and posID. You will see that sometimes they are not the same
}
```

If I must guess, perhaps `cellID` was generated at the beginning of a track segment while position is the center of the track? Unfortunately I don't know how `TOFBarrelHits` were generated.

wdconinc: I'd be hesitant about introducing a (potentially poorly scaling) search algorithm until we understand if there is not a way to change how TOFBarrelHits are created to make sure this is done correctly.

ssedd1123: I agree that the best strategy is to fix `TOFBarrelHits` such that `hit.getCellID()` and `hit.getPosition()` always agree with each other. Unfortunately I may need external help with that as I never handle the part of the code that generates `TOFBarrelHits`.

The complexity of breath first search (BFS) is O(E+V), where E is the number of edges and V is the number of vertex. In our case, vertexes are the cells, and E is total number of neighbors = 4*number of cells. Therefore, time complexity is just O(n) where n is total number of cells in a sensor. But that's slightly misleading, because that's only true in general where `hit.getCellID()` and `hit.getPosition()` are located completely randomly and uniformly in a sensor. If they are close (which I really hope they are), the search would return early. BFS is guarantee to find the shortest path between 2 points, so I assumed BFS would return earlier than other algorithms such as DFS employed in `findAllNeighborsInSensor`.

There is even a more simple solution. Simply replace `auto cellID = hit.getCellID()` with `auto cellID = m_converter -> cellID(dd4hep::Position(globalHitPos.x * dd4hep::mm, globalHitPos.y * dd4hep::mm, globalHitPos.z * dd4hep::mm))` and this issue will resolve itself. The only problem is, if I recall correctly, `CellIDPositionConverter::cellID` is not thread safe. If there is a scalable, thread safe version of `CellIDPositionConverter::cellID`, please let me know and the entire fix can be done by changing that line that fetch `cellID`.

ssedd1123: I tried replacing `auto cellID = hit.getCellID(); ` with`auto cellID = m_converter->cellID(dd4hep::Position(globalHitPos.x * dd4hep::mm, globalHitPos.y * dd4hep::mm, globalHitPos.z * dd4hep::mm)); ` Surprisingly, even this seemingly straightforward change doesn't fully resolve the issue.

If you make this replacement and then follow the steps outlined in my original bug report, you'll see that the number of entries in `TOFBarrelSharedHits.position.z` becomes very close to that in `TOFBarrelHits.position.z`, but they're still not exactly the same. The two histograms only match perfectly when using BFS.

This is quite strange. There may be issues with both the `cellID` and the coordinate transformation. That could be problematic.

ssedd1123:
Okay I think I have a rough idea on why all this is happening.

When I convert global hit position to local position, most of the hits have local z ~ 0. Since the local z-axis is defined as the axis perpendicular to the sensor surface, having z = 0 is expected.

However, there are times when local z-value is not zero. Below is a distribution of local-z values for all hits in all BTOF sensors. Note that y-axis is a log scale, so the peak is very strong.

image

If I focus only on hits where |local-z| < 1e-4 cm, then `hit.getPosition()` always lie within the cell given by `hit.getCellID()`. All disagreements come from hits with |local-z| > 1e-4 cm. What I guess happen is that when a track enters the sensor at an angle, and if the hit is registered "deep" inside the sensor, the end point will "migrate" to the nearby cells, away from the cell that it first enter the detector. 

With that in mind, I tried to extrapolate the position back to local-z = 0 using the momentum vector, but so far I haven't been able to get that working. If any of you have any ideas better than a searching for the cell center that corresponds to  `hit.getPosition()`, please let me know. I would greatly appreciate your help.

ssedd1123:
> A couple of further checks which could help us understand this better. Were there as many hits put in the wrong cell when the cell size was larger? 

The percentage of missed hits do decrease with increasing cell width. The current cell width is 0.01 cm on BTOF. Here's a table of percentage of missed hits and cell width.

image

With that being said, the efficiency is never 100%. 

>If you run a simulation, dropping the MultiSegmentation and only use a single segmentation do you still see the same thing?

Dropping MultiSegmentation actually lowers the efficiency further. If we keep a cell width of 0.01 cm, dropping MultiSegmentation cases 7.4% of the total hits to be lost, as opposed to 6.4% lost in the original. However, I I think that's just because the half sensors are now unusable and has nothing to do with MultiSegmentation.

veprbl (commented):
It looks like this degrades performance by a lot:
This
```
100 calls, 205.66 s ( 40.6%) JEventProcessorPODIO -> edm4eic::RawTrackerHit:TOFBarrelADCTDC
```
https://github.com/eic/EICrecon/actions/runs/16271502860/job/45943794900?pr=1945

Versus `main`:
```
 100 calls, 27.44 s (  8.5%) JEventProcessorPODIO -> edm4eic::RawTrackerHit:TOFBarrelADCTDC
```
https://github.com/eic/EICrecon/actions/runs/16426599520/job/46419873420

Before going forward with this, we might want to consider some alternatives here. Is volume-based segmentation an option for this? (might cause extra memory consumption) Or we could work with DD4hep upstream to implement hit creation that would be more favorable to our type of analysis. Or any chance truth particle information could somehow be used by your algorithm to improve performance?

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -85,6 +93,52 @@ void SiliconChargeSharing::process(const SiliconChargeSharing::Input& input,
   } // for simhits
 } // SiliconChargeSharing:process
 
+dd4hep::rec::CellID SiliconChargeSharing::findHitCellID(
+    const dd4hep::rec::CellID testCellID, const dd4hep::Position& hitPos,
+    const dd4hep::DDSegmentation::CartesianGridXY* segmentation) const {
```
veprbl, line 98: Is this not more or less just dd4hep::DDSegmentation::Segmentation::cellID ? VolumeID parameter for probably can be initialized from existing cellID of the hit.

veprbl: I've learned from past implementations by you and @simonge that segmentations in DD4hep have a lower-level API than the full CellIDPositionConverter. From what I can tell, it has virtual dispatch, but implementations are much simpler, like: https://dd4hep.web.cern.ch/dd4hep/reference/segmentations_2CartesianGridXY_8cpp_source.html

ssedd1123: I am sorry I misread. I thought you were talking about `CellIDPositionConverter::cellID`.  `dd4hep::DDSegmentation::Segmentation::cellID` is being called internally by  `CellIDPositionConverter::cellID` on line 124 and it looks like it's not straight forward to get volume ID that is required for `dd4hep::DDSegmentation::Segmentation::cellID`. 

I can reimplement `CellIDPositionConverter::cellID` by simply copying their code and get rid of the `TGeoHMatrix*  m = geoManager->GetCurrentMatrix() ;`, but are we sure that is better than a breath first search, i.e. what I am doing right now? It doesn't seem to be any easier to implement than just a direct breath first search for neighbors. I cannot comment on the speed either because I don't know the complexity of the API calls of dd4hep.

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -74,6 +74,13 @@ void SiliconChargeSharing::process(const SiliconChargeSharing::Input& input,
                                       globalHitPos.z * dd4hep::mm),
                      transformIt->second);
 
+    // globalHitPos can sometimes lie outside of cell with cellID given by hit.getCellID()
+    // therefore, we search neighbors in a sensor to find the cell ID that correspond to globalHitPos
+    // Precise reason unknown, but we suspect it's cause by steps in Geant4
+    // Perhaps position is the average of all steps in volume while cellID is just the first cell the track hits
+    // They disagree when there are multiple step and scattering inside the volume
+    cellID = segmentationIt->second->cellID(hitPos, hitPos, cellID);
```
veprbl, line 82:
```suggestion
    cellID = segmentationIt->second->cellID(hitPos, globalHitPos, cellID);
```
This one may be unused, but, I think, it would be a good practice to provide it correctly?

src/algorithms/digi/SiliconChargeSharing.cc
```diff
@@ -74,6 +75,13 @@ void SiliconChargeSharing::process(const SiliconChargeSharing::Input& input,
                                       globalHitPos.z * dd4hep::mm),
                      transformIt->second);
 
+    // globalHitPos can sometimes lie outside of cell with cellID given by hit.getCellID()
+    // therefore, we search neighbors in a sensor to find the cell ID that correspond to globalHitPos
```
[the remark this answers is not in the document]
veprbl: Thank you. This is a nice summary. I could only propose some clarification wrt new implementation
```suggestion
    // globalHitPos can sometimes lie outside of cell with cellID given by hit.getCellID()
    // therefore, we search neighbors within the segmentation of the same volume
    // to find the cell ID that correspond to globalHitPos.
```

ssedd1123:
> > I added an option in `SiliconChargeSharing` such that `sharing_sigma` can be defined relative to the cell size in the latest update, as per our agreement in the reconstruction meeting. BTOF now defaults to using this relative definition of charge sharing magnitude.
> 
> Change looks good. If possible, please, split it into a separate PR.

It's now splitted to https://github.com/eic/EICrecon/pull/1989

src/algorithms/digi/SiliconChargeSharing.cc
ssedd1123: What I meant by "not worth it" was "I am too lazy to type out a line of code for the conversion" :P, not any hit in the performance. 

Your way is better. The code is updated.

src/algorithms/digi/SiliconChargeSharing.cc
veprbl: This is mostly addressed now.

veprbl (approved):

PR #1959, opened 2025-07-08
feat: inherit jana::components::JOmniFactory and transition to using upstream JOmniFactory

wdconinc:
This PR transitions us to the upstream `jana::components::JOmniFactory` (and generator) instead of our local version. To help the transition in an adiabatic way, we first inherit `jana::components::JOmniFactory` into our local JOmniFactory, and slowly transition the plugins over with `using` switches.

This now requires JANA2 v2.3.2, which is present as of 24.10.1. Therefore, this requires #1705.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: use upstream JOmniFactory)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/detectors/B0ECAL/B0ECAL.cc
+++ b/src/detectors/B0ECAL/B0ECAL.cc
@@ -5,13 +5,13 @@
 #include <JANA/Components/JOmniFactoryGeneratorT.h>
 #include <JANA/JApplication.h>
 #include <JANA/JApplicationFwd.h>
+#include <JANA/Utils/JEventLevel.h>
 #include <JANA/Utils/JTypeInfo.h>
 #include <cmath>
 #include <string>
 #include <variant>
 #include <vector>
 
-#include "algorithms/calorimetry/CalorimeterClusterRecoCoGConfig.h"
 #include "factories/calorimetry/CalorimeterClusterRecoCoG_factory.h"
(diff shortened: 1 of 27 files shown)


wdconinc:
No diffs after 6c4519a.

wdconinc:
No diff after a09333c.

wdconinc:
No diff after b162ddb.

src/factories/tracking/ActsTrajectoriesMerger_factory.h
```diff
@@ -26,6 +26,10 @@ class ActsTrajectoriesMerger_factory : public JOmniFactory<ActsTrajectoriesMerge
   void ChangeRun(int32_t /* run_number */) {}
 
   void Process(int32_t /* run_number */, uint64_t /* event_number */) {
+
+    // FIXME clear output since it may not have been initialized or reset
+    m_acts_trajectories_output().clear();
```
wdconinc, line 31:
```suggestion
    // FIXME clear output since it may not have been initialized or reset
    // See https://github.com/eic/EICrecon/issues/1961
    m_acts_trajectories_output().clear();
```

wdconinc:
No diff after 5f11ce4.

PR #1924, opened 2025-06-20
Building pulses from SimCalorimeterHitProcessor

mhkim-anl:
### Briefly, what does this PR introduce?
This algorithm, `CalorimeterPulseGeneration`, builds pulses for each hit implemented by `SimCalorimeterHitProcessor`. Sub-pulses were created for each contribution and combined to build pulse of a hit. Regarding the pulse shape, template pulse implemented by `SiliconPulseGeneration` was used. Before the sub-pulses are created, the energy deposit can be converted to number of photoelectrons if necessary. 

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #1902)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators

PR #1973, opened 2025-07-14
Speed up calculation by reducing BTOF charge sharing magnitude

ssedd1123:

TOF pulse generation currently takes 30% of the workload (see https://github.com/eic/EICrecon/pull/1924). This can be significantly improved by reducing charge sharing sigma. Realistically speaking, charge is being shared to 1-3 neighboring cells as the prototype indicates, but the current code sharing everything to, on average, 9 cells. That's because while the grid size is reduced to an unrealistic width (see https://github.com/eic/epic/pull/883), the charge sharing sigma is not reduced along with it.

This pull request reduces sigmaX of BTOF from 0.1 cm to 0.1 mm. Here's the speed of reconstructing 1000 pi+ in the main branch on my laptop,

image

Here's the speed in this pull request,

image

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

There would be less `TOFBarrelADCTDC` entries due to reduced charge sharing.

### DIFF
--- a/src/detectors/BTOF/BTOF.cc
+++ b/src/detectors/BTOF/BTOF.cc
@@ -61,17 +61,17 @@ void InitPlugin(JApplication* app) {
   const double x_when_landau_min = -0.22278;
   const double landau_min        = TMath::Landau(x_when_landau_min, 0, 1, true);
   const double sigma_analog      = 0.293951 * edm4eic::unit::ns;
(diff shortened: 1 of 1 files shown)


veprbl (approved):
Explanation in the reco meeting was that this is now consistent with a cell pitch size.

PR #1974, opened 2025-07-14
Switched to using constant fraction discriminator for BTOF sensors

ssedd1123:

Switched to using constant fraction discriminator (`CFDROCDigitization`) for BTOF sensors. The main branch uses constant threshold (`EICROCDigitization`) for BTOF sensors. I was informed that EICROC was only used on ETOF and not BTOF. As it turns out, we are using constant fraction on BTOF after all. 

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


veprbl (approved):
Looks good. I have a few comments.

veprbl (approved):

Issue #1412, opened 2024-05-04
Have only one service type


veprbl:
We need to either port all services to algorithms interface or to JANA interface.

Also, whatever system we choose, we need to add support for user-settable verbosity.

nathanwbrei:
JANA2 already provides support for user-settable verbosity on a per-component basis with no additional code needed. We aren't using it because we are using the spdlog from our log service instead of the JANA's built in logger. It would be an easy change to extract the JANA log levels and use them to configure the spdlog logger. I'm also interested in redirecting all JANA log messages through spdlog [1], or even replacing the JANA logger with spdlog outright.

[1] https://github.com/JeffersonLab/JANA2/issues/250

PR #1975, opened 2025-07-14
fix: do not pass ActsExamples::Trajectories between algorithms

wdconinc:
This PR removes the `ActsExamples::Trajectories` exchange between tracking algorithms, in favor of passing `Acts::TrackContainer`s. Only tracks are representable in the Acts storage backends, while `ActsExamples::Trajectories` are purely example objects that have no podio equivalent. A helper is introduced to aid in accessing `Acts::TrackContainer`s through the `ActsExamples::Trajectories` interface.

- [x] Bug fix (issue: remove non-podio-representable types)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/benchmarks/reconstruction/TRACKINGcheck/TRACKINGcheckProcessor.cc
+++ b/src/benchmarks/reconstruction/TRACKINGcheck/TRACKINGcheckProcessor.cc
@@ -4,8 +4,10 @@
 //
 
 #include <Acts/Geometry/GeometryContext.hpp>
+#include <ActsExamples/EventData/Track.hpp>
 #include <ActsExamples/EventData/Trajectories.hpp>
 #include <JANA/JApplication.h>
+#include <JANA/JApplicationFwd.h>
 #include <TDirectory.h>
 #include <vector>
 
--- a/src/benchmarks/reconstruction/tracking_efficiency/TrackingEfficiency_processor.cc
+++ b/src/benchmarks/reconstruction/tracking_efficiency/TrackingEfficiency_processor.cc
@@ -2,6 +2,7 @@
 
 #include <Acts/Definitions/TrackParametrization.hpp>
 #include <Acts/EventData/MultiTrajectoryHelpers.hpp>
+#include <ActsExamples/EventData/Track.hpp>
 #include <ActsExamples/EventData/Trajectories.hpp>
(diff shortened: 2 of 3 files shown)


src/algorithms/tracking/ActsTracksToTrajectoriesHelper.h
```diff
@@ -0,0 +1,59 @@
...
+#pragma once
+
+#include <Acts/EventData/ProxyAccessor.hpp>
+#include <ActsExamples/EventData/Track.hpp>
+#include <ActsExamples/EventData/Trajectories.hpp>
+
+namespace eicrecon {
+
+inline ActsExamples::TrajectoriesContainer
+CreateTrajectories(const ActsExamples::ConstTrackContainer& tracks) {
```
veprbl, line 13: Should we move this to the ActsSvc?

veprbl:
The trade-off there is that we have to create trajectories from scratch in every factory that uses those?

wdconinc:
> The trade-off there is that we have to create trajectories from scratch in every factory that uses those?

Yes, but it's a cheap operation, mostly involving resharing of shared_ptrs. This is also what the ActsExamples framework does, see e.g. https://github.com/acts-project/acts/blob/main/Examples/Algorithms/Utilities/src/TracksToTrajectories.cpp. Ultimately it will probably make sense for the downstream algorithms to use the tracks directory

src/algorithms/tracking/ActsTracksToTrajectoriesHelper.h
veprbl: Right now, we don't have a standard for "helper" libraries in eicrecon. It looks like things are moving towards sprinkling header-only utility collections, so I'm looking for a better way to organize things. Also, it looks like your function can't be completely stateless, it begs for proper logging.

veprbl: Thanks for attempting this. We can move items to a follow-up, of course.

I think getting rid of the two "benchmark" is good. We also should resolve #1412, perhaps it would be best to move to JANA services only (would require porting some services from algorithms).

wdconinc:
Closing in favor of #2317.

PR #1977, opened 2025-07-16
fix: use Navigator with MaterialInteractor action for track propagation

wdconinc:
This PR adds an explicit Navigator with MaterialInteractor action for track propagation. This may impact how the material map is taken into account during propagation.

- [x] Bug fix (issue: material map no explicitly included in track propagation)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/TrackPropagation.cc
+++ b/src/algorithms/tracking/TrackPropagation.cc
@@ -283,8 +283,7 @@ TrackPropagation::propagate(const edm4eic::Track& /* track */,
 #if Acts_VERSION_MAJOR >= 36
   using Propagator = Acts::Propagator<Acts::EigenStepper<>, Acts::Navigator>;
 #if Acts_VERSION_MAJOR >= 37
-  using PropagatorOptions = Propagator::template Options<
-      Acts::ActorList<Acts::MaterialInteractor>>;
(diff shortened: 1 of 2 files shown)


wdconinc:
This loses all HcalBarrel (111) ad HcalEndcapN (113) calorimeter track projections (and thus the track/cluster matches).
image

It would be useful to see if this affects your projection residuals and pulls to the DIRC, @mposik1983. It may rather increase the covariance instead of making the residual width narrower (i.e. doesn't change projection centroid but increases the uncertainty in it).

mposik1983:
This also seems to break my propagation to the dirc reference surface

wdconinc:
> This also seems to break my propagation to the dirc reference surface

If you can, try without the end of world aborter. I'm wondering if that's it.

wdconinc:
I'll remove the end of world aborter here, but we should add a proper end of world definition that lives outside of the relevant target surfaces.

ShujieL:
Shall we merge this to fix the pull distribution, and work on the end of world option with a separate PR?

ShujieL (approved):

veprbl (approved):
Approving based on previous approval by Shujie. Diff LGTM

PR #1978, opened 2025-07-17
Add association for TaggerTrackerReconstructedParticles, use new network with additional inputs

simonge:
Adds associations to the particles reconstructed directly from the onnx workflow in preparation to swapping over to merging these at a later stage of the reconstruction workflow.

Swaps to using a newer onnx model which takes in the TaggerTrackerLocalTracks and does the projection step internally.

~~Also makes the associations and beam collection more optional.~~
Some changes have been made to make the collections optional but not fully tested or understood if this is the approach we will make tree wide for MC associations. A proper review will come later

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Adds missing associations which will be passed along


veprbl (approved):
LGTM

PR #1982, opened 2025-07-18
fix: avoid unused variable warnings with Acts 39

wdconinc:
### Briefly, what does this PR introduce?
This PR ensures that CI can run successfully with `-Wall`.

### What kind of change does this PR introduce?
- [x] Bug fix (issue: https://github.com/eic/EICrecon/actions/runs/16375138003/job/46273101051)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #1983, opened 2025-07-18
linux-eic-shell: update llvm-cov from 15 to 18

veprbl:
eic-shell now ships with llvm/clang 18, so we started running into:
```
warning: src/tests/algorithms_test/algorithms_test.profraw: unsupported instrumentation profile format version
error: no profile can be merged
```


wdconinc (dismissed):
Looks good to me. See also #1982.

wdconinc (approved):

PR #1988, opened 2025-07-22
Random Noise injection factory with its usage in tracker classes

minjungkim12:
- [ ] Bug fix (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, it should be transparent.


src/algorithms/digi/RandomNoise.cc
```diff
@@ -0,0 +1,562 @@
...
+void RandomNoise::process(const Input& in, const Output& out) const {
+  const auto [in_hits] = in;
+  auto [out_hits]      = out;
+
+  //----------------------------------------------------------------
+  // 1.1  Resolve the DD4hep read-out that belongs to the collection
+  //----------------------------------------------------------------
+  std::string readoutname = m_cfg.readout_name;
+
+  dd4hep::Readout rd = m_dd4hepGeo->readout(readoutname);
```
simonge, line 189: All of the readout and geometry scanning could be moved to `init()` so it is only calculated once.

src/algorithms/digi/RandomNoise.cc
```diff
@@ -0,0 +1,562 @@
...
+
+    // 3.2  randomise only missing fields
+    for (auto& kv : fieldDists) {
+      if (base.find(kv.first) != base.end())
+        continue;
+      decoder->set(cid, kv.first, kv.second.uni(rng));
+    }
+
+    // 3.3  duplicate check
+    if (hitMap.find(cid) != hitMap.end())
```
simonge, line 525: I don't think this should be needed, as soon as you distribute hits across time there could be multiple noise hits in the same element. It's not quite that simple as it will be digitization dependent.

simonge (commented):
Nice contribution. A few cursory comments to get things going, I haven't dived into the details.

I think this would sit better as a separate set of hits e.g. there is no input collection just the requested readout. Then the `CollectionCollector` would combine them later. This would stop you needing to clone the hits and help maintain the associations. There might be other opinions on this.

The N noise hits per system might be ok for a first pass but should later be replaced by a frequency per channel or frequency per surface area so when a MC timeframe period can be used it can be directly configured.

wdconinc:
Note the recently merged #1934 which imposes the use of the EventHeader info for any algorithms that use randomness, to ensure reproducibility in concurrent running.

src/algorithms/digi/RandomNoise.cc
minjungkim12: not sure what to do at this moment - I will bring this point to tracking or reco. meeting!

src/algorithms/digi/RandomNoise.cc
```diff
@@ -0,0 +1,562 @@
...
+  dd4hep::Segmentation segH = leaf->sd.readout().segmentation();
+  auto seg                  = segH; // handle
+  const auto* decoder       = seg.decoder();
+
+  /* -------- 2) probe volume ------------------------------------ */
+  std::set<dd4hep::CellID> unique;
+
+  auto* ptr = seg.ptr();
+  if (auto* cart = dynamic_cast<const dd4hep::DDSegmentation::CartesianGrid*>(ptr)) {
+    scanCartesian(*cart, box, unique);
```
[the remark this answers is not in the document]
minjungkim12: Exactly. I moved the scanning part into init as suggested.

ShujieL:
This work is redesigned and continued with another PR

PR #1994, opened 2025-08-01
feat: add CITATION.cff

wdconinc:
This adds a CITATION.cff file to the repository, linking to the zenodo entry, https://zenodo.org/records/15880663. Names autogenerated from zenodo. Unresolved names (i.e. just github username) removed.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,4 +1,4 @@
-abstract: "EICrecon is the EIC online/offline reconstuction software."
+abstract: "EICrecon is the EIC online/offline reconstruction software."
 authors: # in the order of https://zenodo.org/records/15880663/export/cff
 - given-names: The ePIC Collaboration
 - family-names: Wouter Deconinck


CITATION.cff
```diff
@@ -0,0 +1,42 @@
...
+- family-names: Shyam Kumar
+- family-names: Tommy Chun Yuen Tsang
+- family-names: Zhongling Ji
+- family-names: Matt Posik
+- family-names: Peter Steinberg
+- family-names: Minho Kim
+- family-names: Daniel Brandenburg
+- family-names: Dhevan Gangadharan
+cff-version: 1.2.0
+date-released: '<replace with the actual release date>'
```
veprbl, line 31: Omit these?

CITATION.cff
```diff
@@ -0,0 +1,42 @@
...
+identifiers:
+- type: swh
+  value: swh:1:dir:75196c46ce549dded398985e500dafcf6a1e5f7c;origin=https://doi.org/10.5281/zenodo.7942531;visit=swh:1:snp:eeac3d121922af83467c6081ca862b631acd15d7;anchor=swh:1:rel:166eaec6e708067946596cfa344c95189ce99608;path=eic-EICrecon-7fad7b7
+license:
+- lgpl-3.0-only
+message: If you use this software, please cite it using the metadata from this file.
+repository-code: https://github.com/eic/EICrecon
+title: 'eic/EICrecon: <replace with the actual version>'
+type: software
+version: '<replace with the actual version>'
```
veprbl, line 42: And these?

CITATION.cff
```diff
@@ -0,0 +1,42 @@
...
+- family-names: Sakib Rahman
+- family-names: Joseph (Joe) Osborn
+- family-names: Shyam Kumar
+- family-names: Tommy Chun Yuen Tsang
+- family-names: Zhongling Ji
+- family-names: Matt Posik
+- family-names: Peter Steinberg
+- family-names: Minho Kim
+- family-names: Daniel Brandenburg
+- family-names: Dhevan Gangadharan
```
veprbl, line 29: How do we keep this up to date?

wdconinc: I was wondering about that too. It's related to the problem with the versions. If we could automate our version release workflow and include a download of a new CITATION.cff, then we can get away from this. Otherwise, I am just tempted to keep it as is, and replace it maybe on demand (we can add a comment line to the file) or on schedule. We could probably even do a scheduled job that downloads the file, checks for diffs, and opens a PR with the necessary changes. But that's all work...

CITATION.cff
veprbl: I assume if we omit, Zenodo will fill this in?

Looking at https://github.com/scikit-hep/uproot5/blob/main/CITATION.cff
they put date-released for the original release date and version is not specified.

Each version on Zeonodo is already provided with an unique DOI, we don't need to make a CITATION.cff to address that.

wdconinc: My concern is that people won't (only) go to zenodo for the citation, but copy it directly from GitHub, see https://github.com/eic/epic/ and the "Cite this repository" menu on the right sidebar. That's based purely on what's inside the cff file.

veprbl: So we should be removing the cff file from eic/epic and replacing it with a Zenodo link in the README?

wdconinc: That's what we have here... and it isn't working. I think having the "Cite this repository" is a better solution than expecting people to know they can cite zenodo. They (especially now) think of zenodo as a place for internal documents.

CITATION.cff
veprbl: That's until Zenodo will start feeding you your own cff?

PR #1999, opened 2025-08-05
fix: Add ImagingTopoCluster layer mode within same layers

AkshayaVijay:
Added Layer Modes to the helper function to group hits within the same layer for ImagingTopoCluster

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/calorimetry/ImagingTopoCluster.cc
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.cc
@@ -180,8 +180,6 @@ void ImagingTopoCluster::process(const Input& input, const Output& output) const
   //   debug("group {}: {} hits", i, groups[i].size());
   // }
   for (std::size_t i = 0; i < groups.size(); ++i) {
-    debug("");
-    debug("");
     debug("group {}: {} hits", i, groups[i].size());
     for (auto idx : groups[i]) {
       const auto& hit = (*hits)[idx];
--- a/src/algorithms/calorimetry/ImagingTopoCluster.cc
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.cc
@@ -63,14 +63,14 @@ void ImagingTopoCluster::init() {
   sameLayerDistXY[1]     = std::visit(_toDouble, m_cfg.sameLayerDistXY[1]) / dd4hep::mm;
   diffLayerDistXY[0]     = std::visit(_toDouble, m_cfg.diffLayerDistXY[0]) / dd4hep::mm;
   diffLayerDistXY[1]     = std::visit(_toDouble, m_cfg.diffLayerDistXY[1]) / dd4hep::mm;
-  sameLayerDistEtaPhi[0] = std::visit(_toDouble, m_cfg.sameLayerDistEtaPhi[0]);
-  sameLayerDistEtaPhi[1] = std::visit(_toDouble, m_cfg.sameLayerDistEtaPhi[1]) / dd4hep::rad;
-  diffLayerDistEtaPhi[0] = std::visit(_toDouble, m_cfg.diffLayerDistEtaPhi[0]);
-  diffLayerDistEtaPhi[1] = std::visit(_toDouble, m_cfg.diffLayerDistEtaPhi[1]) / dd4hep::rad;
-  sameLayerDistTZ[0]     = std::visit(_toDouble, m_cfg.sameLayerDistTZ[0]) / dd4hep::mm;
-  sameLayerDistTZ[1]     = std::visit(_toDouble, m_cfg.sameLayerDistTZ[1]) / dd4hep::mm;
-  diffLayerDistTZ[0]     = std::visit(_toDouble, m_cfg.diffLayerDistTZ[0]) / dd4hep::mm;
-  diffLayerDistTZ[1]     = std::visit(_toDouble, m_cfg.diffLayerDistTZ[1]) / dd4hep::mm;
+  sameLayerDistEtaPhi[0] = m_cfg.sameLayerDistEtaPhi[0];
+  sameLayerDistEtaPhi[1] = m_cfg.sameLayerDistEtaPhi[1] / dd4hep::rad;
+  diffLayerDistEtaPhi[0] = m_cfg.diffLayerDistEtaPhi[0];
+  diffLayerDistEtaPhi[1] = m_cfg.diffLayerDistEtaPhi[1] / dd4hep::rad;
+  sameLayerDistTZ[0]     = m_cfg.sameLayerDistTZ[0] / dd4hep::mm;
+  sameLayerDistTZ[1]     = m_cfg.sameLayerDistTZ[1] / dd4hep::mm;
+  diffLayerDistTZ[0]     = m_cfg.diffLayerDistTZ[0] / dd4hep::mm;
+  diffLayerDistTZ[1]     = m_cfg.diffLayerDistTZ[1] / dd4hep::mm;
 
   sectorDist           = m_cfg.sectorDist / dd4hep::mm;
   minClusterHitEdep    = m_cfg.minClusterHitEdep / dd4hep::GeV;
--- a/src/algorithms/calorimetry/ImagingTopoCluster.cc
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.cc
@@ -63,10 +63,10 @@ void ImagingTopoCluster::init() {
   sameLayerDistXY[1]     = std::visit(_toDouble, m_cfg.sameLayerDistXY[1]) / dd4hep::mm;
   diffLayerDistXY[0]     = std::visit(_toDouble, m_cfg.diffLayerDistXY[0]) / dd4hep::mm;
   diffLayerDistXY[1]     = std::visit(_toDouble, m_cfg.diffLayerDistXY[1]) / dd4hep::mm;
-  sameLayerDistEtaPhi[0] = m_cfg.sameLayerDistEtaPhi[0];
-  sameLayerDistEtaPhi[1] = m_cfg.sameLayerDistEtaPhi[1] / dd4hep::rad;
-  diffLayerDistEtaPhi[0] = m_cfg.diffLayerDistEtaPhi[0];
-  diffLayerDistEtaPhi[1] = m_cfg.diffLayerDistEtaPhi[1] / dd4hep::rad;
+  sameLayerDistEtaPhi[0] = std::visit(_toDouble, m_cfg.sameLayerDistEtaPhi[0]);
+  sameLayerDistEtaPhi[1] = std::visit(_toDouble, m_cfg.sameLayerDistEtaPhi[1]) / dd4hep::rad;
+  diffLayerDistEtaPhi[0] = std::visit(_toDouble, m_cfg.diffLayerDistEtaPhi[0]);
+  diffLayerDistEtaPhi[1] = std::visit(_toDouble, m_cfg.diffLayerDistEtaPhi[1]) / dd4hep::rad;
   sameLayerDistTZ[0]     = std::visit(_toDouble, m_cfg.sameLayerDistTZ[0]) / dd4hep::mm;
   sameLayerDistTZ[1]     = std::visit(_toDouble, m_cfg.sameLayerDistTZ[1]) / dd4hep::mm;
   diffLayerDistTZ[0]     = std::visit(_toDouble, m_cfg.diffLayerDistTZ[0]) / dd4hep::mm;
--- a/src/algorithms/calorimetry/ImagingTopoCluster.cc
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.cc
@@ -158,9 +158,9 @@ void ImagingTopoCluster::process(const Input& input, const Output& output) const
        indices.empty() ? idx = indices.end() : idx) {
 
     trace("hit {:d}: local position = ({}, {}, {}), global position = ({}, {}, {}), energy = {}",
-          *idx, (*hits)[*idx].getLocal().x, (*hits)[*idx].getLocal().y,
-          (*hits)[*idx].getLocal().z, (*hits)[*idx].getPosition().x,
-          (*hits)[*idx].getPosition().y, (*hits)[*idx].getPosition().z, (*hits)[*idx].getEnergy());
+          *idx, (*hits)[*idx].getLocal().x, (*hits)[*idx].getLocal().y, (*hits)[*idx].getLocal().z,
+          (*hits)[*idx].getPosition().x, (*hits)[*idx].getPosition().y,
+          (*hits)[*idx].getPosition().z, (*hits)[*idx].getEnergy());
 
     // not energetic enough for cluster center, but could still be cluster hit
     if ((*hits)[*idx].getEnergy() < minClusterCenterEdep) {
--- a/src/algorithms/calorimetry/ImagingTopoCluster.cc
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.cc
@@ -29,6 +29,7 @@
 #include <cmath>
 #include <cstdlib>
 #include <gsl/pointers>
+#include <stdexcept>
 #include <utility>
 #include <variant>
 #include <vector>
--- a/src/algorithms/calorimetry/ImagingTopoCluster.cc
+++ b/src/algorithms/calorimetry/ImagingTopoCluster.cc
@@ -54,8 +54,28 @@ void ImagingTopoCluster::init() {
     error("Expected 2 values (x_dist, y_dist) for localDistXY");
     return;
   }
-  if (m_cfg.layerDistEtaPhi.size() != 2) {
-    error("Expected 2 values (eta_dist, phi_dist) for layerDistEtaPhi");
+  if (m_cfg.sameLayerDistXY.size() != 2) {
+    error("Expected 2 values (x_dist, y_dist) for sameLayerDistXY");
+    return;
+  }
+  if (m_cfg.diffLayerDistXY.size() != 2) {
+    error("Expected 2 values (x_dist, y_dist) for diffLayerDistXY");
+    return;
+  }
+  if (m_cfg.sameLayerDistEtaPhi.size() != 2) {
+    error("Expected 2 values (eta_dist, phi_dist) for sameLayerDistEtaPhi");
+    return;
+  }
+  if (m_cfg.diffLayerDistEtaPhi.size() != 2) {
+    error("Expected 2 values (eta_dist, phi_dist) for diffLayerDistEtaPhi");
+    return;
+  }
+  if (m_cfg.sameLayerDistPhiZ.size() != 2) {
+    error("Expected 2 values (phi_dist, z_dist) for sameLayerDistPhiZ");
+    return;
+  }
+  if (m_cfg.diffLayerDistPhiZ.size() != 2) {
+    error("Expected 2 values (phi_dist, z_dist) for diffLayerDistPhiZ");
     return;
   }
   if (m_cfg.minClusterCenterEdep < m_cfg.minClusterHitEdep) {
@@ -66,10 +86,19 @@ void ImagingTopoCluster::init() {
   // using juggler internal units (GeV, dd4hep::mm, dd4hep::ns, dd4hep::rad)
   localDistXY[0]       = std::visit(_toDouble, m_cfg.localDistXY[0]) / dd4hep::mm;
   localDistXY[1]       = std::visit(_toDouble, m_cfg.localDistXY[1]) / dd4hep::mm;
-  layerDistXY[0]       = std::visit(_toDouble, m_cfg.layerDistXY[0]) / dd4hep::mm;
-  layerDistXY[1]       = std::visit(_toDouble, m_cfg.layerDistXY[1]) / dd4hep::mm;
-  layerDistEtaPhi[0]   = m_cfg.layerDistEtaPhi[0];
-  layerDistEtaPhi[1]   = m_cfg.layerDistEtaPhi[1] / dd4hep::rad;
+  sameLayerDistXY[0]       = std::visit(_toDouble, m_cfg.sameLayerDistXY[0]) / dd4hep::mm;
+  sameLayerDistXY[1]       = std::visit(_toDouble, m_cfg.sameLayerDistXY[1]) / dd4hep::mm;
+  diffLayerDistXY[0]       = std::visit(_toDouble, m_cfg.diffLayerDistXY[0]) / dd4hep::mm;
+  diffLayerDistXY[1]       = std::visit(_toDouble, m_cfg.diffLayerDistXY[1]) / dd4hep::mm;
+  sameLayerDistEtaPhi[0]   = m_cfg.sameLayerDistEtaPhi[0];
+  sameLayerDistEtaPhi[1]   = m_cfg.sameLayerDistEtaPhi[1]/ dd4hep::rad;
+  diffLayerDistEtaPhi[0]   = m_cfg.diffLayerDistEtaPhi[0];
+  diffLayerDistEtaPhi[1]   = m_cfg.diffLayerDistEtaPhi[1] / dd4hep::rad;
+  sameLayerDistPhiZ[0]   = std::visit(_toDouble,m_cfg.sameLayerDistPhiZ[0])/dd4hep::mm;
+  sameLayerDistPhiZ[1]   = std::visit(_toDouble,m_cfg.sameLayerDistPhiZ[1])/dd4hep::mm;
+  diffLayerDistPhiZ[0]   = std::visit(_toDouble,m_cfg.diffLayerDistPhiZ[0])/dd4hep::mm;
+  diffLayerDistPhiZ[1]   = std::visit(_toDouble,m_cfg.diffLayerDistPhiZ[1])/dd4hep::mm;
+  
   sectorDist           = m_cfg.sectorDist / dd4hep::mm;
   minClusterHitEdep    = m_cfg.minClusterHitEdep / dd4hep::GeV;
   minClusterCenterEdep = m_cfg.minClusterCenterEdep / dd4hep::GeV;
@@ -85,12 +114,12 @@ void ImagingTopoCluster::init() {
   case ImagingTopoClusterConfig::ELayerMode::etaphi:
     info("Same-layer clustering (same sector and same layer): "
          "Global [eta, phi] distance between hits <= [{:.4f}, {:.4f} rad].",
-         layerDistEtaPhi[0], layerDistEtaPhi[1]);
+         sameLayerDistEtaPhi[0], sameLayerDistEtaPhi[1]);
     break;
   case ImagingTopoClusterConfig::ELayerMode::phiz:
     info("Same-layer clustering (same sector and same layer): "
          "Global [phi, z] distance between hits <= [{:.4f} rad, {:.4f} mm].",
-         layerDistEtaPhi[1], layerDistXY[1]);
+         sameLayerDistPhiZ[0], sameLayerDistPhiZ[1]);
     break;
   default:
     error("Unknown same-layer mode.");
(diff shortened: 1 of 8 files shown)


src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -153,8 +176,21 @@ void ImagingTopoCluster::process(const Input& input, const Output& output) const
     idx = indices.erase(idx); // takes role of idx++
   }
   debug("found {} potential clusters (groups of hits)", groups.size());
+  // for (std::size_t i = 0; i < groups.size(); ++i) {
+  //   debug("group {}: {} hits", i, groups[i].size());
+  // }
```
wdconinc, line 181: General comment: remove all commented code. It's still there in the git history and when it's commented it's not tested for correctness so it's not to be relied upon.

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -153,8 +176,21 @@ void ImagingTopoCluster::process(const Input& input, const Output& output) const
...
   for (std::size_t i = 0; i < groups.size(); ++i) {
+    debug("");
+    debug("");
```
wdconinc, line 184:
```suggestion
```

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -153,8 +176,21 @@ void ImagingTopoCluster::process(const Input& input, const Output& output) const
...
     debug("group {}: {} hits", i, groups[i].size());
+    for (auto idx : groups[i]) {
+      const auto& hit = (*hits)[idx];
+      debug("  hit {} -> energy = {:.6f}, layer = {}, sector = {}, local = ({:.2f}, {:.2f}, "
```
wdconinc, line 188: There's a level that is even more noisy than `debug`, and that's `trace`. That's intended for output at the individual hit level in clustering algorithms like this.
```suggestion
      trace("  hit {} -> energy = {:.6f}, layer = {}, sector = {}, local = ({:.2f}, {:.2f}, "
```

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -190,21 +226,67 @@ bool ImagingTopoCluster::is_neighbour(const edm4eic::CalorimeterHit& h1,
...
+    switch (m_cfg.diffLayerMode) {
     case eicrecon::ImagingTopoClusterConfig::ELayerMode::etaphi:
       return (std::abs(edm4hep::utils::eta(h1.getPosition()) -
                        edm4hep::utils::eta(h2.getPosition())) <= layerDistEtaPhi[0]) &&
              (std::abs(edm4hep::utils::angleAzimuthal(h1.getPosition()) -
                        edm4hep::utils::angleAzimuthal(h2.getPosition())) <= layerDistEtaPhi[1]);
+
     case eicrecon::ImagingTopoClusterConfig::ELayerMode::xy:
       return (std::abs(h1.getPosition().x - h2.getPosition().x) <= layerDistXY[0]) &&
              (std::abs(h1.getPosition().y - h2.getPosition().y) <= layerDistXY[1]);
```
wdconinc, line 271: Who uses this? This is a dangerous mode since it is global XY positions instead of local XY positions (as the same-layer mode of the same name), and it will therefore only work for endcap detectors.

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -190,21 +226,67 @@ bool ImagingTopoCluster::is_neighbour(const edm4eic::CalorimeterHit& h1,
...
+
+    case eicrecon::ImagingTopoClusterConfig::ELayerMode::phiz: {
```
wdconinc, line 273: It's a bit risky to call this `phiz` since the `phi` might seem to indicate that the `0` coordinate is an angle. But it is a distance.

I don't have any better suggestions though.

src/algorithms/calorimetry/ImagingTopoClusterConfig.h
```diff
@@ -22,7 +22,11 @@ struct ImagingTopoClusterConfig {
   // maximum distance of global (x, y) to be considered as neighbors at different layers (if layerMode==xy)
   std::vector<std::variant<std::string, double>> layerDistXY = {1.0 * dd4hep::mm, 1.0 * dd4hep::mm};
   // determines how neighbors are determined for hits in different layers (using either eta and phi, or x and y)
-  enum ELayerMode { etaphi = 0, xy = 1 } layerMode = etaphi;
+  // enum ELayerMode { etaphi = 0, xy = 1 } layerMode = etaphi;
+
+  enum ELayerMode { etaphi = 0, xy = 1, phiz = 2 };
```
wdconinc, line 27: Probably should make this a [scoped enum](https://en.cppreference.com/w/cpp/language/enum.html) since otherwise it means nobody can use the variable `xy` anywhere (or they have to be careful about shadowing).
```suggestion
  enum class ELayerMode { etaphi = 0, xy = 1, phiz = 2 };
```

src/factories/calorimetry/ImagingTopoCluster_factory.h
```diff
@@ -24,8 +24,13 @@ class ImagingTopoCluster_factory
...
   ParameterRef<std::vector<double>> m_ldep{this, "layerDistEtaPhi", config().layerDistEtaPhi};
   // ParameterRef<std::vector<double>> m_ldxy_adjacent{this, "layerDistXY", config().layerDistXY};
-  ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_laymode{this, "layerMode",
-                                                                         config().layerMode};
+  // ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_laymode{this, "layerMode",
+  //                                                                        config().layerMode};
+  ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_sameLayMode{
+      this, "sameLayerMode", config().sameLayerMode};
+  ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_diffLayMode{
+      this, "diffLayerMode", config().diffLayerMode};
```
wdconinc, line 32: This is invisible but it's just better to give this the same name.
```suggestion
  ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_sameLayerMode{
      this, "sameLayerMode", config().sameLayerMode};
  ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_diffLayerMode{
      this, "diffLayerMode", config().diffLayerMode};
```

src/algorithms/calorimetry/ImagingTopoCluster.cc
AkshayaVijay:
```suggestion
```

src/algorithms/calorimetry/ImagingTopoCluster.cc
AkshayaVijay:
```suggestion
    case eicrecon::ImagingTopoClusterConfig::ELayerMode::phiz: {
    // Layer mode 'phiz' uses the average phi of the hits to define a rotated direction. The coordinate is a distance, not an angle.


```

AkshayaVijay:
In this PR, a new phiz layermode is introduced in the Imaging Topological Clustering Algorithm to group hits both within the same layer and across different layers.

- Currently, hits in the same layer are grouped based on their local x and y positions. The new phiz mode converts the azimuthal (phi) separation between hits into a tangential distance (in millimeters) by projecting hit positions relative to the average phi of the two hits.
- In same-layer phiz mode, the algorithm checks whether two hits are sufficiently close in the phi direction (converted to distance) and in z, instead of using the local xand y positions. In different-layer phiz mode, the same logic is applied across layers within the neighbourLayersRange.

The impact of these changes is studied using a 2-photon simulation, and the corresponding plots are attached.
[Layermode_xy_phiz_comparison.pdf](https://github.com/user-attachments/files/21800338/Layermode_xy_phiz_comparison.pdf)

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -75,23 +75,46 @@ void ImagingTopoCluster::init() {
...
   case ImagingTopoClusterConfig::ELayerMode::etaphi:
-    info("Neighbour layers clustering (same sector and layer id within +- {:d}: "
+    info("Same-layer clustering (same sector and same layer): "
+         "Global [eta, phi] distance between hits <= [{:.4f}, {:.4f} rad].",
+         layerDistEtaPhi[0], layerDistEtaPhi[1]);
+    break;
+  case ImagingTopoClusterConfig::ELayerMode::phiz:
+    info("Same-layer clustering (same sector and same layer): "
+         "Global [phi, z] distance between hits <= [{:.4f} rad, {:.4f} mm].",
+         layerDistEtaPhi[1], layerDistXY[1]);
```
veprbl, line 93: Indices look suspicious here. Only `layerDistXY` is used below. Reusing XY and EtaPhi for PhiZ is a bit confusing. Can we tidy this up somehow? One option is to add layerDistPhiZ. Other is to throw away unit handling and have a single array for every mode.

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -75,23 +75,46 @@ void ImagingTopoCluster::init() {
...
-    info("Neighbour layers clustering (same sector and layer id within +- {:d}: "
-         "Local [x, y] distance between hits <= [{:.4f} mm, {:.4f} mm].",
+    info("Neighbour layers clustering (same sector and layer id within +- {:d}): "
+         "Global [x, y] distance between hits <= [{:.4f} mm, {:.4f} mm].",
          m_cfg.neighbourLayersRange, layerDistXY[0], layerDistXY[1]);
     break;
+  case ImagingTopoClusterConfig::ELayerMode::phiz:
+    info("Neighbour layers clustering (same sector and layer id within +- {:d}): "
+         "Global [phi, z] distance between hits <= [{:.4f} rad, {:.4f} mm].",
+         m_cfg.neighbourLayersRange, layerDistEtaPhi[1], layerDistXY[1]);
```
veprbl, line 114: Similar concern for layerDist.

src/detectors/BEMC/BEMC.cc
```diff
@@ -204,6 +205,9 @@ void InitPlugin(JApplication* app) {
           .neighbourLayersRange = 2, //  # id diff for adjacent layer
           .localDistXY          = {2.0 * dd4hep::mm, 2 * dd4hep::mm},     //  # same layer
           .layerDistEtaPhi      = {10 * dd4hep::mrad, 10 * dd4hep::mrad}, //  # adjacent layer
+          .sameLayerMode        = eicrecon::ImagingTopoClusterConfig::ELayerMode::
```
veprbl, line 208: So, this configures phi distance cut to localDistXY[0] = 2 mm?

src/algorithms/calorimetry/ImagingTopoClusterConfig.h
```diff
@@ -22,7 +22,9 @@ struct ImagingTopoClusterConfig {
   // maximum distance of global (x, y) to be considered as neighbors at different layers (if layerMode==xy)
   std::vector<std::variant<std::string, double>> layerDistXY = {1.0 * dd4hep::mm, 1.0 * dd4hep::mm};
   // determines how neighbors are determined for hits in different layers (using either eta and phi, or x and y)
-  enum ELayerMode { etaphi = 0, xy = 1 } layerMode = etaphi;
+  enum class ELayerMode { etaphi = 0, xy = 1, phiz = 2 };
+  ELayerMode sameLayerMode = ELayerMode::xy; // for ldiff =0
+  ELayerMode diffLayerMode = ELayerMode::xy; // for ldiff <= neighbourLayersRange
```
veprbl, line 27: Would it make sense to have a naming consistent with parameters:
```suggestion
  ELayerMode localMode = ELayerMode::xy; // for ldiff =0
  ELayerMode layerMode = ELayerMode::xy; // for ldiff <= neighbourLayersRange
```

src/factories/calorimetry/ImagingTopoCluster_factory.h
```diff
@@ -24,8 +24,11 @@ class ImagingTopoCluster_factory
   // ParameterRef<std::vector<double>> m_ldxy{this, "localDistXY", config().localDistXY};
   ParameterRef<std::vector<double>> m_ldep{this, "layerDistEtaPhi", config().layerDistEtaPhi};
   // ParameterRef<std::vector<double>> m_ldxy_adjacent{this, "layerDistXY", config().layerDistXY};
-  ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_laymode{this, "layerMode",
-                                                                         config().layerMode};
+  // ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_sameLayerMode{
+  //     this, "sameLayerMode", config().sameLayerMode};
+  // ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_diffLayerMode{
+  //     this, "diffLayerMode", config().diffLayerMode};
```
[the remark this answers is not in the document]
wdconinc: This is a problem that needs resolving in JANA2. When using strongly typed enums, the `JParameterManager` handling fails to compile even when a proper operator<< and operator>> are defined (which are used by `Stringify` for weakly typed enums).

See https://github.com/eic/EICrecon/pull/2015 for demo. (FYI @nathanwbrei)

src/algorithms/calorimetry/ImagingTopoClusterConfig.h
veprbl: Yes, this is much better.

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -191,20 +246,65 @@ bool ImagingTopoCluster::is_neighbour(const edm4eic::CalorimeterHit& h1,
...
+    case eicrecon::ImagingTopoClusterConfig::ELayerMode::phiz: {
+      auto phi  = 0.5 * (edm4hep::utils::angleAzimuthal(h1.getPosition()) +
+                        edm4hep::utils::angleAzimuthal(h2.getPosition()));
+      auto h1_t = (h1.getPosition().x * sin(phi)) - (h1.getPosition().y * cos(phi));
+      auto h2_t = (h2.getPosition().x * sin(phi)) - (h2.getPosition().y * cos(phi));
+      auto h1_z = h1.getPosition().z;
+      auto h2_z = h2.getPosition().z;
+
+      return (std::abs(h1_t - h2_t) <= diffLayerDistPhiZ[0]) &&
+             (std::abs(h1_z - h2_z) <= diffLayerDistPhiZ[1]);
```
veprbl, line 299: I would rename `PhiZ` to `TZ` (and `phiz` to `tz`). You'll be able to remove some explanatory comments too.

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -50,12 +50,28 @@ void ImagingTopoCluster::init() {
...
+    return;
+  }
+  if (m_cfg.sameLayerDistPhiZ.size() != 2) {
+    error("Expected 2 values (phi_dist, z_dist) for sameLayerDistPhiZ");
+    return;
+  }
+  if (m_cfg.diffLayerDistPhiZ.size() != 2) {
+    error("Expected 2 values (phi_dist, z_dist) for diffLayerDistPhiZ");
     return;
   }
```
veprbl, line 76: We could move these into the the switch statements below to not require all values to be specified. Also consider replacing `return` with `throw std::runtime_error("<Same message as in error()>")`.

veprbl (approved):
Diff looks good to me. Capybara report shows no changes to ZDC, as expected. https://eicrecon.epic-eic.org/pr/1999/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#HcalFarForwardZDCClusters

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -50,48 +51,103 @@ void ImagingTopoCluster::init() {
...
-  layerDistEtaPhi[0]   = m_cfg.layerDistEtaPhi[0];
-  layerDistEtaPhi[1]   = m_cfg.layerDistEtaPhi[1] / dd4hep::rad;
+  sameLayerDistXY[0]     = std::visit(_toDouble, m_cfg.sameLayerDistXY[0]) / dd4hep::mm;
+  sameLayerDistXY[1]     = std::visit(_toDouble, m_cfg.sameLayerDistXY[1]) / dd4hep::mm;
+  diffLayerDistXY[0]     = std::visit(_toDouble, m_cfg.diffLayerDistXY[0]) / dd4hep::mm;
+  diffLayerDistXY[1]     = std::visit(_toDouble, m_cfg.diffLayerDistXY[1]) / dd4hep::mm;
+  sameLayerDistEtaPhi[0] = m_cfg.sameLayerDistEtaPhi[0];
+  sameLayerDistEtaPhi[1] = m_cfg.sameLayerDistEtaPhi[1] / dd4hep::rad;
+  diffLayerDistEtaPhi[0] = m_cfg.diffLayerDistEtaPhi[0];
+  diffLayerDistEtaPhi[1] = m_cfg.diffLayerDistEtaPhi[1] / dd4hep::rad;
```
wdconinc, line 69:
```suggestion
  sameLayerDistEtaPhi[0] = std::visit(_toDouble, m_cfg.sameLayerDistEtaPhi[0]);
  sameLayerDistEtaPhi[1] = std::visit(_toDouble, m_cfg.sameLayerDistEtaPhi[1]) / dd4hep::rad;
  diffLayerDistEtaPhi[0] = std::visit(_toDouble, m_cfg.diffLayerDistEtaPhi[0]);
  diffLayerDistEtaPhi[1] = std::visit(_toDouble, m_cfg.diffLayerDistEtaPhi[1]) / dd4hep::rad;
```

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -50,48 +51,103 @@ void ImagingTopoCluster::init() {
...
+      throw std::runtime_error(msg);
+    }
+    info(
+        "Neighbour layers clustering (same sector and layer id within +- {:d}): "
+        "Global [phi, z] distance between hits <= [{:.4f} mm, {:.4f} mm].", // The coordinate Phi is the projected Phi and thus is a distance, not an angle.
+        m_cfg.neighbourLayersRange, diffLayerDistTZ[0], diffLayerDistTZ[1]);
     break;
   default:
-    error("Unknown layer mode.");
+    throw std::runtime_error("Unknown different-layer mode.");
```
wdconinc, line 150: Keep the explicit error logging, since not all runtime error exceptions lead to a printout.
```suggestion
    error("Unknown different-layer mode.");
    throw std::runtime_error("Unknown different-layer mode.");
```

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -50,48 +51,103 @@ void ImagingTopoCluster::init() {
...
+  case ImagingTopoClusterConfig::ELayerMode::tz:
+    if (m_cfg.sameLayerDistTZ.size() != 2) {
+      const std::string msg = "Expected 2 values (phi_dist, z_dist) for sameLayerDistTZ";
+      error(msg);
+      throw std::runtime_error(msg);
+    }
+    info("Same-layer clustering (same sector and same layer): "
+         "Global [phi, z] distance between hits <= [{:.4f} rad, {:.4f} mm].",
+         sameLayerDistTZ[0], sameLayerDistTZ[1]);
+    break;
```
wdconinc, line 111:
```suggestion
  case ImagingTopoClusterConfig::ELayerMode::tz:
    if (m_cfg.sameLayerDistTZ.size() != 2) {
      const std::string msg = "Expected 2 values (t_dist, z_dist) for sameLayerDistTZ";
      error(msg);
      throw std::runtime_error(msg);
    }
    info("Same-layer clustering (same sector and same layer): "
         "Global [t, z] distance between hits <= [{:.4f} mm, {:.4f} mm].",
         sameLayerDistTZ[0], sameLayerDistTZ[1]);
    break;
```

src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -50,48 +51,103 @@ void ImagingTopoCluster::init() {
...
+    if (m_cfg.diffLayerDistTZ.size() != 2) {
+      const std::string msg = "Expected 2 values (phi_dist, z_dist) for diffLayerDistTZ";
+      error(msg);
+      throw std::runtime_error(msg);
+    }
+    info(
+        "Neighbour layers clustering (same sector and layer id within +- {:d}): "
+        "Global [phi, z] distance between hits <= [{:.4f} mm, {:.4f} mm].", // The coordinate Phi is the projected Phi and thus is a distance, not an angle.
+        m_cfg.neighbourLayersRange, diffLayerDistTZ[0], diffLayerDistTZ[1]);
     break;
```
wdconinc, line 148:
```suggestion
  case ImagingTopoClusterConfig::ELayerMode::tz:
    if (m_cfg.diffLayerDistTZ.size() != 2) {
      const std::string msg = "Expected 2 values (t_dist, z_dist) for diffLayerDistTZ";
      error(msg);
      throw std::runtime_error(msg);
    }
    info(
        "Neighbour layers clustering (same sector and layer id within +- {:d}): "
        "Global [t, z] distance between hits <= [{:.4f} mm, {:.4f} mm].",
        m_cfg.neighbourLayersRange, diffLayerDistTZ[0], diffLayerDistTZ[1]);
break;
```

src/algorithms/calorimetry/ImagingTopoClusterConfig.h
```diff
@@ -15,14 +15,28 @@ struct ImagingTopoClusterConfig {
...
-  // maximum distance of local (x, y) to be considered as neighbors at the same layer
-  std::vector<std::variant<std::string, double>> localDistXY = {1.0 * dd4hep::mm, 1.0 * dd4hep::mm};
-  // maximum distance of global (eta, phi) to be considered as neighbors at different layers (if layerMode==etaphi)
-  std::vector<double> layerDistEtaPhi = {0.01, 0.01};
+  // maximum distance of global (x, y) to be considered as neighbors at same layers (if layerMode==xy)
+  std::vector<std::variant<std::string, double>> sameLayerDistXY = {1.0 * dd4hep::mm,
+                                                                    1.0 * dd4hep::mm};
+  // maximum distance of global (eta, phi) to be considered as neighbors at same layers (if layerMode==etaphi)
+  std::vector<double> sameLayerDistEtaPhi = {0.01, 0.01};
+  // maximum distance of global (x, y) to be considered as neighbors at same layers (if layerMode==tz)
```
wdconinc, line 23:
```suggestion
  // maximum distance of global (t, z) to be considered as neighbors at same layers (if layerMode==tz)
```

src/algorithms/calorimetry/ImagingTopoClusterConfig.h
```diff
@@ -15,14 +15,28 @@ struct ImagingTopoClusterConfig {
...
+  std::vector<std::variant<std::string, double>> sameLayerDistTZ = {1.0 * dd4hep::mm,
+                                                                    1.0 * dd4hep::mm};
   // maximum distance of global (x, y) to be considered as neighbors at different layers (if layerMode==xy)
-  std::vector<std::variant<std::string, double>> layerDistXY = {1.0 * dd4hep::mm, 1.0 * dd4hep::mm};
+  std::vector<std::variant<std::string, double>> diffLayerDistXY = {1.0 * dd4hep::mm,
+                                                                    1.0 * dd4hep::mm};
+  // maximum distance of global (eta, phi) to be considered as neighbors at different layers (if layerMode==etaphi)
+  std::vector<double> diffLayerDistEtaPhi = {0.01, 0.01};
+  // maximum distance of global (x, y) to be considered as neighbors at different layers (if layerMode==tz)
```
wdconinc, line 31:
```suggestion
  // maximum distance of global (t, z) to be considered as neighbors at different layers (if layerMode==tz)
```

src/algorithms/calorimetry/ImagingTopoClusterConfig.h
```diff
@@ -15,14 +15,28 @@ struct ImagingTopoClusterConfig {
...
   int neighbourLayersRange = 1;
-  // maximum distance of local (x, y) to be considered as neighbors at the same layer
-  std::vector<std::variant<std::string, double>> localDistXY = {1.0 * dd4hep::mm, 1.0 * dd4hep::mm};
-  // maximum distance of global (eta, phi) to be considered as neighbors at different layers (if layerMode==etaphi)
-  std::vector<double> layerDistEtaPhi = {0.01, 0.01};
+  // maximum distance of global (x, y) to be considered as neighbors at same layers (if layerMode==xy)
+  std::vector<std::variant<std::string, double>> sameLayerDistXY = {1.0 * dd4hep::mm,
+                                                                    1.0 * dd4hep::mm};
+  // maximum distance of global (eta, phi) to be considered as neighbors at same layers (if layerMode==etaphi)
+  std::vector<double> sameLayerDistEtaPhi = {0.01, 0.01};
```
wdconinc, line 22:
```suggestion
  std::vector<std::variant<std::string, double>> sameLayerDistEtaPhi = {0.01, 0.01};
```

src/algorithms/calorimetry/ImagingTopoClusterConfig.h
```diff
@@ -15,14 +15,28 @@ struct ImagingTopoClusterConfig {
...
+  // maximum distance of global (t, z) to be considered as neighbors at same layers (if layerMode==tz)
+  std::vector<std::variant<std::string, double>> sameLayerDistTZ = {1.0 * dd4hep::mm,
+                                                                    1.0 * dd4hep::mm};
   // maximum distance of global (x, y) to be considered as neighbors at different layers (if layerMode==xy)
-  std::vector<std::variant<std::string, double>> layerDistXY = {1.0 * dd4hep::mm, 1.0 * dd4hep::mm};
+  std::vector<std::variant<std::string, double>> diffLayerDistXY = {1.0 * dd4hep::mm,
+                                                                    1.0 * dd4hep::mm};
+  // maximum distance of global (eta, phi) to be considered as neighbors at different layers (if layerMode==etaphi)
+  std::vector<double> diffLayerDistEtaPhi = {0.01, 0.01};
```
wdconinc, line 30:
```suggestion
  std::vector<std::variant<std::string, double>> diffLayerDistEtaPhi = {0.01, 0.01};
```

src/factories/calorimetry/ImagingTopoCluster_factory.h
```diff
@@ -22,10 +22,19 @@ class ImagingTopoCluster_factory
...
-  ParameterRef<eicrecon::ImagingTopoClusterConfig::ELayerMode> m_laymode{this, "layerMode",
-                                                                         config().layerMode};
+  ParameterRef<std::vector<double>> m_ldxy_same{this, "sameLayerDistXY", config().sameLayerDistXY};
+  ParameterRef<std::vector<double>> m_ldxy_diff{this, "diffLayerDistXY", config().diffLayerDistXY};
+  ParameterRef<std::vector<double>> m_ldep_same{this, "sameLayerDistEtaPhi",
+                                                config().sameLayerDistEtaPhi};
+  ParameterRef<std::vector<double>> m_ldep_diff{this, "diffLayerDistEtaPhi",
+                                                config().diffLayerDistEtaPhi};
+  ParameterRef<std::vector<double>> m_ldxy_same{this, "sameLayerDistTZ", config().sameLayerDistTZ};
+  ParameterRef<std::vector<double>> m_ldxy_diff{this, "diffLayerDistTZ", config().diffLayerDistTZ};
```
wdconinc, line 32:
```suggestion
  ParameterRef<std::vector<double>> m_ldtz_same{this, "sameLayerDistTZ", config().sameLayerDistTZ};
  ParameterRef<std::vector<double>> m_ldtz_diff{this, "diffLayerDistTZ", config().diffLayerDistTZ};
```

wdconinc:
So, we can't support variants of string and double with JANA's ParameterRef, only XY mode actually uses it, and the other modes need to be optimized which will be easier with command line controls. So I reverted the variants back to just doubles in the case of eta-phi and t-z modes. At some point we need to determine whether ParameterRef can be modified to support variants, but not urgent. (FYI @nathanwbrei)

wdconinc (dismissed):

src/factories/calorimetry/ImagingTopoCluster_factory.h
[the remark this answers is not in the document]
wdconinc: The inability to use a scoped enum here turned out to be caused by an incorrect `operator<<` definition. The second argument needs to be `const &`. Without `const &`, the unscoped enum is implicitly converted to an `int`, which as another `operator<<`. The scoped enum does not have an implicit conversion to `int` so it fails the JTypeInfo traits SFINAE. I updated #2015 from demonstrating to fixing this issue.

wdconinc:
Rebased for #2057.

PR #2002, opened 2025-08-07
Support JANA2 v2.4.3

nathanwbrei:
PR #1959 pulled in JANA2 v2.4.2's JOmniFactory and JOmniFactoryGenerator after JANA2 v2.4.3 was already released. The new release includes backwards incompatible changes to JOmniFactory{Generator}. They add support for multiple variadic inputs and outputs with arbitrary arity, which is needed for timeframe splitting.

### DIFF
--- a/src/detectors/DRICH/DRICH.cc
+++ b/src/detectors/DRICH/DRICH.cc
@@ -131,15 +131,14 @@ void InitPlugin(JApplication* app) {
       "DRICHGasTracks", {"CentralCKFTracks", "CentralCKFActsTrajectories", "CentralCKFActsTracks"},
       {"DRICHGasTracks"}, gas_track_cfg));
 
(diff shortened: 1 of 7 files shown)


wdconinc:
Sorry.... https://github.com/eic/EICrecon/pull/2005

nathanwbrei:
#2005 makes my life easier actually, it gives me a chance to avoid some backwards compatibility headaches. Let's discuss what you need on your side in order to un-revert, because I have a couple of modifications as well

wdconinc:
> #2005 makes my life easier actually, it gives me a chance to avoid some backwards compatibility headaches.

It might not avoid the backwards compatibility headaches: in EICrecon we would still expect support from JANA 2.3.2 all the way up to the current version (and all version in between, unless you explicitly block a version from being used in the CMakeLists.txt).

> Let's discuss what you need on your side in order to un-revert, because I have a couple of modifications as well

I am not sure yet what the cause of this issue is. It is (probably) something about how exceptions are treated differently, but I don't see yet where.

Issue #2001, opened 2025-08-07
Pulse generation from time-clustered SimCalorimeterHits using a shared template


mhkim-anl:
This issue is about pulse generation from time-clustered `SimCalorimeterHit`s. To generate a pulse from the contributions that are close in time, a `timeID` is assigned in the `SimCalorimeterHitProcessor`. Since both `SimCalorimeterHit` and `SimTrackerHit` share a common pulse generation mechanism, a shared template algorithm, `PulseGeneration` has been implemented.

PR #2004, opened 2025-08-07
Pulse generation from time-clustered Simcalorimeterhits using a shared template

mhkim-anl:
This PR is for pulse generation from time-clustered `SimCalorimeterHit`s. To generate a pulse from the contributions that are close in time, a `timeID` is assigned in the `SimCalorimeterHitProcessor`. Since both `SimCalorimeterHit` and `SimTrackerHit` share a common pulse generation mechanism, a shared template algorithm, `PulseGeneration`, has been implemented.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #2001)
- [ ] Documentation update
- [ ] Other: __


veprbl:
Looks like this increases number of hit contributions and hits:
image
image

mhkim-anl:
@wdconinc @veprbl 
Thank you for your answers and suggestions. I understood the factor 4 increase is too much to save the `EcalBarrelScFi{P, N}Pulses`. I removed them from the output. 

Then, regarding the next step, I would like to discuss it including the downstream algorithm at the coming simulation meeting. I would appreciate it if you could join the meeting on Aug 26.

By the way, could you let me know up to what factor of data size increase would be acceptable? I think there may still be a possible way to reduce the output size by optimizing both `timestep` and `ignore_thres`.

wdconinc:
> By the way, could you let me know up to what factor of data size increase would be acceptable? I think there may still be a possible way to reduce the output size by optimizing both `timestep` and `ignore_thres`.

As a rule of thumb, we don't want a single detector subsystem to take more than 20% of resources (storage, computing).

veprbl (approved):
This is in good shape now.

veprbl (approved):
Approval still stands

PR #2013, opened 2025-08-14
fix: explicitly convert signed run number to unsigned for bitset (ubsan fix)

wdconinc:
This PR avoids an Undefined Behavior warning due to the default run number of -1 being used in the constructor of a bitset, which only accepts unsigned long long. Negative run numbers are static cast from -1 to 4294967295 instead of just taking abs because that would lead to -1, 0, 1 run sequences having random IDs that are identical in close proximity.
```
SUMMARY: UndefinedBehaviorSanitizer: implicit-integer-sign-change /home/wdconinc/git/EICrecon/src/algorithms/interfaces/UniqueIDGenSvc.h:52:52 
```

See also:
- https://github.com/JeffersonLab/JANA2/blob/a2c5bbeb00c545ed44fb7698bd9e7f1d74feb8b0/src/libraries/JANA/JFactory.h#L184
- https://github.com/key4hep/EDM4hep/commit/5219b52264285910c65f21d7d047fb93fafd85d2

Note: no warning is suppressed for event numbers since those are not ever -1 (and if they are, we want to know why it changed).

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/interfaces/UniqueIDGenSvc.h
+++ b/src/algorithms/interfaces/UniqueIDGenSvc.h
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <edm4hep/EventHeaderCollection.h>
+#include <edm4hep/EDM4hepVersion.h>
 #include <algorithms/logger.h>
 #include <algorithms/service.h>
 #include <bitset>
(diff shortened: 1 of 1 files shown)


src/algorithms/interfaces/UniqueIDGenSvc.h
```diff
@@ -49,7 +50,12 @@ class UniqueIDGenSvc : public LoggedService<UniqueIDGenSvc> {
                      const std::string_view& name) const {
     std::bitset<seed_digits> seed_bits           = m_seed.value();
     std::bitset<event_num_digits> event_num_bits = evt_num;
+#if EDM4HEP_BUILD_VERSION > EDM4HEP_VERSION(0, 99, 2)
```
wdconinc, line 53:
```suggestion
#if EDM4HEP_BUILD_VERSION >= EDM4HEP_VERSION(0, 99, 2)
```

veprbl (approved):
LGTM

PR #2014, opened 2025-08-14
linux-eic-shell.yml: compare MT to previous MT

veprbl:
We should be fine now, with new RNG service. Even if we not, we should have an issue/PR associated with every TODO.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc:
OT, slightly. I've been wanting to make it clearer what the `ST_MT` comparison is. I don't want people to look at that and say "oh, look, this PR doesn't introduce any differences.

PR #2017, opened 2025-08-19
Reduce verbosity of CKFTracking algorithm

simonge:
Change empty tips log message from info to debug.

Certain MC samples and detector configurations result in by default most events printing an info message for both the Central and B0 reconstruction. This seems unnecessary and the information would only be used in debugging anyway.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Reduces the number of log outputs.

### DIFF
--- a/src/algorithms/tracking/CKFTracking.cc
+++ b/src/algorithms/tracking/CKFTracking.cc
@@ -454,7 +454,6 @@ CKFTracking::process(const edm4eic::TrackParametersCollection& init_trk_params,
   // Seed number column accessor
   const Acts::ConstProxyAccessor<unsigned int> constSeedNumber("seed");
 
-
   ActsExamples::Trajectories::IndexedParameters parameters;
   std::vector<Acts::MultiTrajectoryTraits::IndexType> tips;
 


simonge:
> I think this is possibly the wrong fix to an underlying problem. If we attempt CKF, and we consistently find empty trajectories (and I think in this case it is likely _only_ empty trajectories), then there's something wrong with what we ask CKF to do. I've been meaning to look into this too, since I suspect it's because we incorrectly include seeds that have no possibility of ever resulting in a track.

For me this is due to the input collections always being empty. I could move the creation of the output variables to the beginning of process, add a check for the empty collections and return immediately.

wdconinc:
> > I think this is possibly the wrong fix to an underlying problem. If we attempt CKF, and we consistently find empty trajectories (and I think in this case it is likely _only_ empty trajectories), then there's something wrong with what we ask CKF to do. I've been meaning to look into this too, since I suspect it's because we incorrectly include seeds that have no possibility of ever resulting in a track.
> 
> For me this is due to the input collections always being empty. I could move the creation of the output variables to the beginning of process, add a check for the empty collections and return immediately.

Yeah, we shouldn't run CKF if there are no hits :-)

wdconinc (approved):
Looks good to me. Thanks!

PR #2018, opened 2025-08-19
Adopt onnx workflow over TMVA injecting low-q2 reconstructed particles later

simonge:
Swaps out the particles reconstructed via the TMVA network for the onnx reconstruction. The particles are added to the ReconstructedParticles collection workflow later to remove unnecessary combinatorics in the tracking and pid stages.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Several intermediate collections have been removed from the output. Anyone using ReconstructedChargedParticles will no longer have the Low-Q2 electrons included.
The resolution of the particles reconstructed from the low-q2 tagger should be slightly improved over the current method.


simonge:
Comparison benchmark plots for the currently trained TMVA vs. onnx networks available on the pythia6 test sample. (Note there is nothing inherently better about the onnx network for the reconstruction just how it was trained)

TMVA being replaced:
image

New onnx reconstruction:
image

src/global/pid_lut/pid_lut.cc
```diff
@@ -1,62 +1,23 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2022-2025 Christopher Dilks, Simon Gardner
 
-#include <Evaluator/DD4hepUnits.h>
 #include <JANA/JApplicationFwd.h>
 #include <JANA/Utils/JTypeInfo.h>
 #include <cmath>
 #include <string>
```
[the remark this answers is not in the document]
wdconinc: This is a false positive triggered by the polar edge 2.7175 below... probably acceptable to use `// NOLINTNEXTLINE` on that. Or use a generator expression for the equidistant bin ranges to avoid the number and avoid typos in them.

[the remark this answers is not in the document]
simonge: I don't follow why this was triggered or why the 2.7175 is special. It's not related to this PR but could add the `// NOLINTNEXTLINE`, does it have to come at the end of this line or the next one:
https://github.com/eic/EICrecon/pull/2018/files#diff-7e6870a9db7bc36732e352daa8150dc06568f4b4a4927300ee65d70b92e3a758L71

src/global/pid_lut/pid_lut.cc
```diff
@@ -97,8 +58,8 @@ void InitPlugin(JApplication* app) {
       "RICHEndcapNLUTPID",
       {
           "EventHeader",
-          "ReconstructedChargedWithFBPIDParticles",
-          "ReconstructedChargedWithFBPIDParticleAssociations",
+          "ReconstructedChargedWithoutFBPIDParticles",
+          "ReconstructedChargedWithoutFBPIDParticleAssociations",
```
veprbl, line 62:
```suggestion
          "ReconstructedChargedWithoutPIDParticles",
          "ReconstructedChargedWithoutPIDParticleAssociations",
```
I would think, WithoutFBPID is not a thing.

src/global/reco/reco.cc
```diff
@@ -77,12 +77,27 @@ void InitPlugin(JApplication* app) {
        "EcalEndcapPClusterAssociations"},
       {"EcalClusterAssociations"}, app));
 
+  // Combine reconstructed charged particles and their associations from different sources
+  app->Add(
+      new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::ReconstructedParticle, true>>(
+          "CombinedReconstructedChargedParticles",
+          {"ReconstructedChargedParticles", "TaggerTrackerReconstructedParticles"},
```
veprbl, line 84: This changes definition of `ReconstructedChargedParticles` to not include FB particles. We probably don't want that.

src/global/pid_lut/pid_lut.cc
```diff
@@ -1,62 +1,23 @@
...
-          "ReconstructedTruthSeededChargedWithoutPIDParticleAssociations",
-      },
-      {
-          "ReconstructedTruthSeededChargedWithFBPIDParticles",
-          "ReconstructedTruthSeededChargedWithFBPIDParticleAssociations",
-          "FarBackwardTruthSeededPhaseSpacePIDParticleIDs",
-      },
-      phase_space_pid_cfg, app));
-
-  app->Add(new JOmniFactoryGeneratorT<PhaseSpacePID_factory>(
```
veprbl, line 47: Was that the only use of PhaseSpacePID_factory. If yes, let's consider removing it.

src/global/reco/reco.cc
simonge: There are no clean ways of doing this. Assuming we don't want to replace `ReconstructedChargedParticles` with a subset collection we need to pass it though another algorithm first, this would mean merging it before the final `PID` factory, which may be the only reasonable approach but feels unnecessary.

veprbl: If I understand correctly, the issue is with needing association rewriting. Letting pid_lut algorithm do the rewriting for you is one way. In one of the PRs on calorimetry we've envisioned a generic association rewriting algorithm for such use cases, but wasn't implemented so far.

simonge:
@wdconinc It looks like your recent changes to the iwyu CI doesn't cover the option of headers being removed in the PR.

src/global/pid_lut/pid_lut.cc
```diff
@@ -210,6 +177,40 @@ void InitPlugin(JApplication* app) {
...
+          {"ReconstructedChargedWithPFRICHTOFDIRCPIDParticles",
+           "TaggerTrackerReconstructedParticles"},
+          {"ReconstructedWithPFRICHTOFDIRCLOWQ2PIDChargedParticles"}, app));
+
+  app->Add(new JOmniFactoryGeneratorT<
+           CollectionCollector_factory<edm4eic::MCRecoParticleAssociation, true>>(
+      "ReconstructedChargedWithPFRICHTOFDIRCLOWQ2PIDParticleAssociations",
+      {"ReconstructedChargedWithPFRICHTOFDIRCPIDParticleAssociations",
+       "TaggerTrackerReconstructedParticleAssociations"},
+      {"ReconstructedChargedWithPFRICHTOFDIRCLOWQ2PIDParticleAssociations"}, app));
```
veprbl, line 196: I think, this is broken. See https://eicrecon.epic-eic.org/pr/2018/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#_ReconstructedParticles_particleIDs
image
This loses 2 out of 4 PID LUTs.

Looking again at logic in
https://github.com/eic/EICrecon/blob/6e3493bd5a2b1d851660e66a72561af2109d263f/src/algorithms/pid_lut/PIDLookup.cc#L67-L72
I don't see what is wrong yet. It should match entries just fine, if it's just a subset collection, and it should be:
https://github.com/eic/EICrecon/blob/6e3493bd5a2b1d851660e66a72561af2109d263f/src/algorithms/meta/CollectionCollector.h#L39

simonge: Further comments, I'm not sure what changed to cause the changes between your original figure and the one which is in the link now. 

Of the originally missing ParticleIDs. One of them was the `FarBackwardPhaseSpacePIDParticleIDs` which has been removed and the other the `RICHEndcapNParticleIDs`.

For some reason the `RICHEndcapNParticleIDs` appeared to be adding a PID to the LowQ2 particles. The `pfrich.lut` file covers values all the way up to pi based however the binning provided in the pid_lut plugin doesn't appear to. Looking at the implementation the binning in the config should provide the histogram limits and the file should just fill it.

simonge: Updated image after #2083 from: 
 https://eicrecon.epic-eic.org/pr/2018/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#_ReconstructedParticles_particleIDs
image

Shows only the `FarBackwardPhaseSpacePIDParticleIDs` collection has been removed an no changes to the content of the others. I can't comment on why the previous plots appeared so different with totally different counts in even the completely filled ParticleIDs.

PR #2065, opened 2025-09-03
fix the weight calculation in raw to sim hits association

ShujieL:
In Si digi, the weight between raw and sim hit was set to 1 manually regardless of how many sim hits on the same cell. This update should provide a rough but more correct weight.

- [ x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/digi/SiliconTrackerDigi.cc
+++ b/src/algorithms/digi/SiliconTrackerDigi.cc
@@ -88,13 +88,17 @@ void SiliconTrackerDigi::process(const SiliconTrackerDigi::Input& input,
   }
 
   for (auto item : cell_hit_map) {
-    raw_hits->push_back(item.second);
-
+    auto raw_hit=item.second;
+    raw_hits->push_back(raw_hit);
     for (const auto& sim_hit : *sim_hits) {
       if (item.first == sim_hit.getCellID()) {
         // set association
         auto hitassoc = associations->create();
-        hitassoc.setWeight(1.0);
+        double weight = 0.0;
+        if (raw_hit.getCharge() > 0) {
+            weight = std::round((sim_hit.getEDep() * 1e6 / raw_hit.getCharge()) * 10.0) / 10.0;
+        }
+        hitassoc.setWeight(weight);
         hitassoc.setRawHit(item.second);
         hitassoc.setSimHit(sim_hit);
       }


ShujieL:
To add a bit more context of this PR, I looked at the background merged events, and noticed that the photons from Synchrotron radiation almost always left a pair of sim hits on one pixel, see attached an example from the vertex barrel hits. 

In our current digi code, the energy deposit on a given cell is the sum of all sim hits so they almost always pass the edep threshold. But any idea on how DD4hep created two hits here? The non-zero quality code in this case seems to indicate overlay (bit=31). 
image

src/algorithms/digi/SiliconTrackerDigi.cc
```diff
@@ -88,13 +88,17 @@ void SiliconTrackerDigi::process(const SiliconTrackerDigi::Input& input,
...
+    auto raw_hit = item.second;
+    raw_hits->push_back(raw_hit);
     for (const auto& sim_hit : *sim_hits) {
       if (item.first == sim_hit.getCellID()) {
         // set association
         auto hitassoc = associations->create();
-        hitassoc.setWeight(1.0);
+        double weight = 0.0;
+        if (raw_hit.getCharge() > 0) {
+          weight = std::round((sim_hit.getEDep() * 1e6 / raw_hit.getCharge()) * 10.0) / 10.0;
```
wdconinc, line 99: Please remove these magic numbers.

wdconinc: And why the rounding?

ShujieL: by "magic number" I guess you mean the times 10 for rounding? 

In line 72 will fill the charge with (std::int32_t)std::llround(sim_hit.getEDep() * 1e6). one could use exactly this formula. I wanted to use rounding to make sure it will return 1 instead of 0.99999 for example, and I assume we don't care about too much about precision. Would times 100 (two decimal digits) be better?

wdconinc: Don't round.

Other magical number is the 1e6.

ShujieL: 1e6 is the unit conversion. Can you suggest some alternative?

wdconinc: https://github.com/eic/EDM4eic/blob/main/utils/include/edm4eic/unit_system.h,  https://github.com/AIDASoft/DD4hep/blob/master/DDParsers/include/Evaluator/DD4hepUnits.h

ShujieL: Thanks! I changed to use eV for rawhit.charge so it's always integer. It was keV with the magic number of 1e6, which is not the best choice because our detector threshold is often <1keV. 
We can have a separate discussion later on whether it makes sense to declare RawTrackerHit.charge as int32_t.

wdconinc:
I don't know how we've come this far without proper energy deposition digitization in the silicon tracker. What bit depth and dynamic range do we expect there? This digitization algorithm can learn a lot from the calorimeter digitization... It almost doesn't make sense to tack on more hacky hard coded bit depth and dynamic range assumptions with this PR until the underlying logic is fixed.

simonge:
> I don't know how we've come this far without proper energy deposition digitization in the silicon tracker.

I thought we'd decided to leave this in place, rather than spend time patching it, while pursuing proper technology dependant approaches.

> To add a bit more context to this PR, I looked at the background merged events, and noticed that the photons from Synchrotron radiation almost always left a pair of sim hits on one pixel, see attached an example from the vertex barrel hits.
In our current digi code, the energy deposit on a given cell is the sum of all sim hits so they almost always pass the edep threshold. But any idea on how DD4hep created two hits here? The non-zero quality code in this case seems to indicate overlay (bit=31).

Is that not the quality flag for `secondary` rather than overlay. Does the associated MCParticle point to the same particle? Is this just the synchrotron backgrounds or are other sources showing the same thing?

ShujieL:
> I don't know how we've come this far without proper energy deposition digitization in the silicon tracker. What bit depth and dynamic range do we expect there? This digitization algorithm can learn a lot from the calorimeter digitization... It almost doesn't make sense to tack on more hacky hard coded bit depth and dynamic range assumptions with this PR until the underlying logic is fixed.

I will follow up with you and Simon to arrange a dedicated discussion on edep digitization scheme. Any more suggestions on the current change?

PR #2078, opened 2025-09-15
Flag clusters as EMCal or HCal

ruse-traveler:

This PR introduces two new flags to the `Jug::Reco::ClusterType` in `algorithms/calorimetry/ClusterTypes`, `kClusterEMCal` and `kClusterHCal`. These flags are intended to help facilitate generic algorithms downstream which need to distinguish between EM and hadronic energy.

These are new types are used to flag only the types of clusters which will be passed on to downstream, holistic algorithms such as the output of `CalorimeterClusterRecoCoG` (for all calorimeters other than the BIC) or `EnergyPositionClusterMerger` (for the BIC).

**Note:** this is intended to be an interim solution, aimed at facilitating development on downstream PF algorithms and maintaining momentum (see the presentation [here](https://indico.global/event/15249/contributions/133795) for more details on the PF development plan). A longer-term, more flexible approach is discussed in [EDM4eic#122](https://github.com/eic/EDM4eic/issues/122).

- [ ] Bug fix (issue #__)
- [x] New feature (issue #2086)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/calorimetry/TruthEnergyPositionClusterMerger.cc
+++ b/src/algorithms/calorimetry/TruthEnergyPositionClusterMerger.cc
@@ -173,7 +173,7 @@ std::map<int, edm4eic::Cluster> TruthEnergyPositionClusterMerger::indexedCluster
       continue;
     }
 
-    const bool duplicate = matched.count(mcID);
+    const bool duplicate = matched.contains(mcID);
     if (duplicate) {
       trace("   --> WARNING: this is a duplicate mcID, keeping the higher energy cluster");
       if (cluster.getEnergy() < matched[mcID].getEnergy()) {


wdconinc (commented):
I don't think we should be encoding Ecal vs Hcal clusters in the data type since it is accessible from the detector type in DD4hep, e.g. https://github.com/AIDASoft/DD4hep/blob/e369490a18a9857d5a61771f340c1791e8f9f7a3/DDCore/include/DD4hep/DetType.h#L53-L54

wdconinc:
> Ah! I wasn't aware of this `enum`! Thanks! Based on a cursory glance, it seems like many of our calorimeters don't set these...

Yeah, it's only imposed by ACTS for tracking, which is why those have it...

> But then I'm curious about how I would use this downstream. For example, would I extract these from a cell ID? Or would I look them up based on the system ID? In the former case, I suspect I would need to select a hit from the cluster to identify a cell ID; and in the latter, I expect I would need to provide the system IDs to use as a parameter to the algorithm...

Cell ID would give you system ID as the 8 lsb. Presumably there is nothing in our data model that is so disconnected from hits that you can't get there in a hop or two.

If the use is for within EICrecon, then I'd prefer to not just add unspecific bit fields to data model. I don't know what it adds to the end user if they know it's a 2D vs 3D cluster. And within EICrecon at what point is there confusion about this, and does it matter? In any case, having to specify what an algorithm produces in its configuration seems redundant.

wdconinc:
> the EMCal and HCal DD4hep tags shouldn't impact the actual detector simulation, right? They're just there for description?

Should not impact simulation indeed. May impact overall event classification. This is used to indicate in an event bit pattern whether the event had any energy deposition in an Ecal or Hcal, but we don't do anything with that.

ruse-traveler:
After testing it out, I can confirm that the `dd4hep::DetType` flags will work for this purpose! I'll close this PR in favor of [epic#994](https://github.com/eic/epic/pull/994).

PR #2083, opened 2025-09-18
attempt to remove the pid_lut theta overflow bins

simonge:
Removes filling of PID lookup tables outside of their theta bin range.

As noted in #2018 the `pfrich.lut` file contains entries out with the range of the binning defined for it in https://github.com/eic/EICrecon/blob/main/src/global/pid_lut/pid_lut.cc#L72. This has given access to a bin up to `theta<inf` so particles in the far-backward region are provided with a `ParticleID` from the `RichEndcapN`.

~~This PR attempts to remove this overflow bin using boost histograms own methods but if it doesn't work,~~ manual filtering is possibly simpler.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Stops the PID LUTs from assigning values outside of their range.

### DIFF
--- a/src/services/pid_lut/PIDLookupTable.cc
+++ b/src/services/pid_lut/PIDLookupTable.cc
@@ -100,9 +100,9 @@ void PIDLookupTable::load_file(const std::string& filename,
     if ((bool)(iss >> pdg >> charge >> momentum >> eta >> phi) &&
         (binning.missing_electron_prob || (bool)(iss >> prob_electron)) &&
         (bool)(iss >> prob_pion >> prob_kaon >> prob_proton)) {
-          
-      if(eta * angle_fudge < polar_bins.bin(0).lower() ||
-         eta * angle_fudge > polar_bins.bin(polar_bins.size() -1).upper()) {
+
+      if (eta * angle_fudge < polar_bins.bin(0).lower() ||
+          eta * angle_fudge > polar_bins.bin(polar_bins.size() - 1).upper()) {
         debug("Out of bounds: eta");
         continue; // out of bounds
       }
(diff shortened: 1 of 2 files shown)


veprbl:
Just to confirm, this is my interpretation of your finding: The original code relies on histogram binning to impose limits on table acceptance. However, it turns out that the values filled to overflow bins get interpreted.

If that's the case and, if there are multiple entries in overflow bins, then those should fail the check for a single entry, but, unfortunately, `bh::indexed(m_hist)` ignores overflow bins (`coverage = coverage::inner` is default https://www.boost.org/doc/libs/1_80_0/libs/histogram/doc/html/histogram/reference.html#header.boost.histogram.indexed_hpp).

You have tried to pass `bh::axis::option::none_t` to `options` to disable the underflow/overflow as per https://www.boost.org/doc/libs/1_80_0/libs/histogram/doc/html/histogram/reference.html#header.boost.histogram.axis.option_hpp, but that didn't work.

veprbl (approved):
It's fine to merge as hotfix, we can figure out details later.

veprbl:
This has no effect on ParticleIDs though:
https://eicrecon.epic-eic.org/pr/2083/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#_ReconstructedParticles_particleIDs

Which distribution was fixed?

veprbl:
My point is that there might be more than one issue. For posterity, this is what I saw
image
and, while one could argue it's a small reduction, that is not the complete disappearance that we see in #2018. There must be another separate issue, I believe.

simonge:
> while one could argue it's a small reduction, that is not the complete disappearance that we see in #2018. There must be another separate issue, I believe.

The complete disappearance of the `RICHEndcapNParticleIDs` seemed to somehow resolve itself between the image you posted to the comment and some time in the last week if you check the current capybara for #2018 . The other collection of `ParticleIDs` which disappears is from the removed PhaseSpaceID algorithm.

PR #2094, opened 2025-09-29
EcalBarrelScFiProtoClusters: use localDistXY instead of localDistXZ

wdconinc:
The island clustering in the BEMC EcalBarrelScFiProtoClusters should happen in the XY plane, not the XZ plane. This should remove the unreasonable effectiveness of the island clustering for two gammas in the same sector. We only gain Z information when we add timing analysis, and we don't lose the Y information at that point but combine it all in topological clustering.

- [x] Bug fix (issue: EcalBarrelScFiProtoClusters unreasonably effective)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, makes EcalBarrelScFiProtoClusters worse.

### DIFF
--- a/src/detectors/BEMC/BEMC.cc
+++ b/src/detectors/BEMC/BEMC.cc
@@ -162,7 +162,7 @@ void InitPlugin(JApplication* app) {
           .adjacencyMatrix{},
           .peakNeighbourhoodMatrix{},
           .readout{},
-          .sectorDist = 50. * dd4hep::mm,
+          .sectorDist  = 50. * dd4hep::mm,
           .localDistXY = {80 * dd4hep::mm, 80 * dd4hep::mm},
           .localDistXZ{},
           .localDistYZ{},


wdconinc:
> Do I understand correctly, that Y is the radial direction? I'm not entirely sure what is the effect is being fixed or worked around here. Ultimately, we want XZ, but the artificial Z segmentation is not representative of the resolution?

Island clustering for the scfi should have as primary directions X and Y (i.e. the tangential and radial directions at the end of sector readout plane*). The Z information is only secondarily available (since it requires time info, and then resolving which pulses on both ends are matching the same original hit). We therefore don't want Z information in our island clustering (which is 2D only).

Right now we are ignoring radial direction entirely when doing the clustering, and instead we are using Z information which is leading to a too optimistic operation.

So, this will make scfi clustering intentionally worse since we are now applying clustering in a way that is unrealistically efficient for same-sector clusters.

sly2j:
I think there's a bit of a misunderstanding: In local coordinates, the fibers are in the y-direction and the grid cells are in the x-z plane (similar to the tracker disks). See for example here: https://github.com/eic/epic/blob/fc2eb7a5f35264d46c5c5c04e0404e3b816b8e82/src/BarrelCalorimeterScFi_geo.cpp#L279C50-L279C51

In other words, we shouldn't change this as this would break what the algorithm is supposed to be doing.

wdconinc:
> I think there's a bit of a misunderstanding: In local coordinates, the fibers are in the y-direction and the grid cells are in the x-z plane (similar to the tracker disks). See for example here: https://github.com/eic/epic/blob/fc2eb7a5f35264d46c5c5c04e0404e3b816b8e82/src/BarrelCalorimeterScFi_geo.cpp#L279C50-L279C51
> 
> In other words, we shouldn't change this as this would break what the algorithm is supposed to be doing.

You can print out the positions it operates on to see that the local position is not rotated, only shifted.

wdconinc:
Here are the input x and y local coordinates for EcalBarrelScFiRecHits:
image
And here is the z local coordinate:
image
(sorry for splitting, they are on different rows in capybara)

These are not local in the sector coordinate system anymore. As discussed with @sly2j on mm, there is possibly a change in the geometry that caused this, e.g. a change from placing an actual sector volume to placing fibers in an assembly. I can't find anything while blame surfing epic.

We (=I) should get back to adding the capybara output to the epic repository, so we capture these issues, if it is a geometry change.

The other possible cause could be in our global-to-local handling in CalorimeterHitReco which may have started to use a different reference volume. I don't see any relevant changes to that in a long time, so I don't know what could have changed.

PR #2144, opened 2025-10-20
SecondaryVertices factory based on Helix method

starsdong:
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

None

New SecondaryVerticesHelix factory added into reco plugin, SecondaryVerticesHelix object (currently with edm4eic:Vertex structure) saved in PODIO

### DIFF
--- a/src/factories/reco/SecondaryVerticesHelix_factory.h
+++ b/src/factories/reco/SecondaryVerticesHelix_factory.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2024 Xin Dong
+// Copyright (C) 2025 Xin Dong
 
 #pragma once
 
--- a/src/algorithms/reco/SecondaryVerticesHelix.cc
+++ b/src/algorithms/reco/SecondaryVerticesHelix.cc
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2024 Daniel Brandenburg, Xin Dong
+// Copyright (C) 2025 Daniel Brandenburg, Xin Dong
 
 #include <DD4hep/Fields.h>
 #include <Evaluator/DD4hepUnits.h>


veprbl:
> Hi Wouter and Dmitri,
> 
> The two remaining failed checks seem to beyond my knowledge. Will you be able to help identify what went wrong there? Thanks and Regards
> 
> /xin

This looks like an irrelevant issue. Should resolve itself once latest geometry makes it into the container. I'm rerunning container build without cache now, so, hopefully, this will clear away. We can review the code meanwhile.

starsdong:
> > Hi Wouter and Dmitri,
> > The two remaining failed checks seem to beyond my knowledge. Will you be able to help identify what went wrong there? Thanks and Regards
> > /xin
> 
> This looks like an irrelevant issue. Should resolve itself once latest geometry makes it into the container. I'm rerunning container build without cache now, so, hopefully, this will clear away. We can review the code meanwhile.

Hi Dmitri and Wouter,

Do you have any suggestions/comments to this PR? Any adjustment is needed before it gets merged? Thanks

/xin

src/algorithms/reco/Helix.cc
wdconinc: Is this copied from another framework?

starsdong: Yes, Wouter. The core part is copied from the code used in STAR. I made some adjustments to be compatible with edm4eic or edm4hep objects. And I added one constructor to create a Helix object from edm4eic::ReconstructedParticle.

starsdong:
> This is on my radar, but I haven't had chance to look deeper yet. I can share a few early comments. We don't have a convention for utility libraries in EICrecon. For now, do you think we could merge Helix.cc and Helix.h into SecondaryVerticesHelix.cc? I also see you provide Config structure, but don't expose the parameters in the factory. And last is the hardcoded b_field, would it make sense to get the value from the geometry service instead (e.g. sample at the 0,0,0 coordinate)?

I believe it is in principle possible to get merged into SecondaryVerticesHelix.cc. But the Helix.cc is pretty sizable, and I believe many functions can be used in other down stream analysis. The code is largely copied from STAR with adjustments to be adaptable to edm4eic constainers. I think it is probably preferrable to keep them separated so it is easier to maintain. 

I wasn't sure that I understand completely about your comment "I also see you provide Config structure, but don't expose the parameters in the factory." Do you suggest to explicitly put them out in reco.cc when creating this factory? I see there, some factories do this way, some don't. I am fine with either requirement.

Regarding the b_field, Yes, I was thinking about the way you suggested too. Let me try to update this part soon. Thanks

starsdong:
> > I believe it is in principle possible to get merged into SecondaryVerticesHelix.cc. But the Helix.cc is pretty sizable, and I believe many functions can be used in other down stream analysis. The code is largely copied from STAR with adjustments to be adaptable to edm4eic constainers. I think it is probably preferrable to keep them separated so it is easier to maintain.
> 
> Maybe it's for the best to keep it separate, especially if we find that it has questions about authorship/license. But in principle, if left as is, we may have to relocate it at some point.
> 
> > I wasn't sure that I understand completely about your comment "I also see you provide Config structure, but don't expose the parameters in the factory." Do you suggest to explicitly put them out in reco.cc when creating this factory? I see there, some factories do this way, some don't. I am fine with either requirement.
> 
> I did not mean this. See my inline comment above, I hope that will clarify.

Hi Dmitri,

The PR has been updated with new commits addressing the magnetic field input and the parameters declaration in Config file. Thanks

/xin

src/factories/reco/SecondaryVerticesHelix_factory.h
veprbl: IVF lives in `tracking/` should we do secondary keep vertexing there too? (Same for algorithms)

src/algorithms/reco/Helix.h
```diff
@@ -0,0 +1,222 @@
...
+
+}; // end class Helix
+
+//
+//     Non-member functions
+//
+//int operator== (const Helix&, const Helix&);
+//int operator!= (const Helix&, const Helix&);
+//std::ostream& operator<<(std::ostream&, const Helix&);
+
```
veprbl:
```suggestion
```

src/factories/reco/SecondaryVerticesHelix_factory.h
veprbl: Let's ignore this for now.

src/algorithms/reco/Helix.cc
starsdong: Wouter, I talked to STAR about the license issue. Currently there is no explicit license in STAR, though mostly will be MIT or CC BY 4.0. There will be some discussion and a conclusion soon. Shall we wait for the conclusion or we can proceed with merging and update the license later?

wdconinc: Please point them to https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software

If they read that, one would hope they land automatically with the MIT license.

I'd be happy just moving ahead on the assumption of MIT. Please modify the SPDX headers to reflect this, with the copyright holding body of the STAR Collaboration as names. E.g.
```
// SPDX-License-Identifier: MIT
// Copyright (C) 2020-2024 The STAR Collaboration, Xin Dong, Rongrong Ma
```

PR #2153, opened 2025-10-28
Update/Fix MatchClusters.cc

qunib:
Previously, some photons that overlapped with charged tracks were being incorrectly discarded, leading to reduced reconstruction efficiency. This change adjusts the matching logic to retain valid photons without affecting charged particle suppression.

This PR improves cluster-to-MC association logic by choosing the highest-weight match per cluster. 
Default behavior is slightly changed for clusters with multiple MC associations; 
otherwise, results are identical. No API or function signature changes are required for users.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


wdconinc:
@qunib This can likely be closed. Can you copy the more detailed description from here into #2154, update the title there to be more descriptive, merge the iwyu fixes in #2155, and keep an eye on #2159 (i.e. subscribe) so you know when it is merged?

PR #2154, opened 2025-10-28
Fix missing photon reconstruction caused by charged track overlap in MatchClusters.cc

qunib:
Previously, some photons that overlapped with charged tracks were being incorrectly discarded, leading to reduced reconstruction efficiency. This change adjusts the matching logic to retain valid photons without affecting charged particle suppression.

This PR improves the cluster-to-MC association logic in MatchClusters by selecting the best MC particle per cluster based on association weight, instead of taking the first match found. Default behavior is slightly changed for clusters with multiple MC associations; otherwise, results are identical. No API or function signature changes are required for users.

This code change :
1.  Builds a map keyed by cluster object index, storing the best-associated MC id and its weight.
2. When matching charged tracks, it only removes a cluster if that cluster's best association is to that charged MC particle (and chooses the best cluster for that MC by highest association weight if multiple clusters claim the same MC). Remaining clusters are treated as neutrals.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/reco/MatchClusters.cc
+++ b/src/algorithms/reco/MatchClusters.cc
@@ -132,7 +132,7 @@ std::map<int, edm4eic::Cluster> MatchClusters::indexedClusters(
   for (const auto cluster : *clusters) {
 
     int bestMcID     = -1;
-    float bestWeight = -1.f;
+    float bestWeight = -1.F;
 
     // find best associated MC particle for this cluster (largest association weight)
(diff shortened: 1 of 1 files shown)


veprbl (approved):
I can't say that I understand all the effects here, but change looks reasonable. Not considering weights of associations is definitely not what we want to do.

PR #2161, opened 2025-10-29
refactor: use std::map with comparator for deterministic ordering in SimCalorimeterHitProcessor

wdconinc:
This PR aims to remove the non-deterministic behavior between single- and multi-threading observed in e.g. `_EcalBarrelScFiPPulses_calorimeterHits` and brethren. Compared to unordered_map, we get
- slower insertion with map (O(log n) instead of O(1)),
- technically* slower lookup (O(log n) instead of O(1)), and 
- the same iteration complexity (O(n)).

In the map we define an explicit comparator since the default OrderKey in podio is memory-address based.

At this point (and maybe always), deterministic output outweighs performance consideration.

(*Technically since we don't do random lookups.)

- [x] Bug fix (issue: non-deterministic behavior)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, we might have slightly slower performance and better determinism.

### DIFF
--- a/src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
+++ b/src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
@@ -199,8 +199,8 @@ void SimCalorimeterHitProcessor::process(const SimCalorimeterHitProcessor::Input
   auto [out_hits, out_hit_contribs] = output;
 
   // Map for staging output information. We have 2 levels of structure:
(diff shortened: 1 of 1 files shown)


wdconinc:
@veprbl This resolves the irreproducible behavior. I wonder how we can avoid this in the future? It's really probably whenever we rely on podio's OrderKey, so maybe whenever that's in our AST we need to fail?

(This will also likely fix the tracking branch algorithms.)

Issue #2090, opened 2025-09-22
Digitization algorithm to measure ADC, TOA, and TOT


mhkim-anl:
This issue is about a digitization algorithm to measure ADC, TOA, and TOT from a pulse.

PR #2169, opened 2025-10-30
Digitization algorithm that reproduces the CALOROC measurement

mhkim-anl:
This PR reproduces the CALOROC measurement and store the corresponding data types, `RawCALOROCHit` and `CALOROC1{A, B}Sample`, from pulses. When CALOROC1A measures the TOT, it is stored in the same sample where the TOA is recorded, even though the TOT is measured in a later sample than the TOA. This means we should not finalize/output the sample immediately after scanning a time window in the code. Instead, we need to temporarily store the sample information somewhere, such as in a buffer, until the TOT is measured or all amplitudes have been scanned. For this purpose, a struct, `RawEntry`, was used in this algorithm.

Please refer to the [slides](https://indico.bnl.gov/event/30532/contributions/118515/attachments/67625/116206/BIC20260121_CALOROC_workfest.pdf) on the basic structure of this digitization algorithm and the basic cross-checks.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #2090)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/interfaces/CompareObjectID.h
+++ b/src/algorithms/interfaces/CompareObjectID.h
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2026 Derek Anderson
+
+#pragma once
+
+#include <podio/ObjectID.h>
+
+namespace eicrecon {
+
+/*! Comparator struct for podio::ObjectID. Organizes podio objects by
+ *  their ObjectID's in decreasing collection ID first, and second by
+ *  decreasing index second.
+ */
+template <typename T> struct CompareObjectID {
+  bool operator()(const T& lhs, const T& rhs) const {
+    if (lhs.getObjectID().collectionID == rhs.getObjectID().collectionID) {
+      return (lhs.getObjectID().index < rhs.getObjectID().index);
+    } else {
+      return (lhs.getObjectID().collectionID < rhs.getObjectID().collectionID);
+    }
+  }
+};
+
+} // namespace eicrecon


src/algorithms/digi/PulseDigi.h
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2025 Minho Kim
+
+#pragma once
+
+#include <algorithms/algorithm.h>
+#include <edm4eic/SimPulseCollection.h>
+#include <edm4eic/RawHGCROCHitCollection.h>
```
veprbl: You will need to require EDM4eic 8.3.0 to use your code. Basically, see some of older changes removed in https://github.com/eic/EICrecon/pull/2121/files for inspiration.

src/factories/digi/PulseDigi_factory.h
```diff
@@ -0,0 +1,48 @@
...
+
+#pragma once
+
+#include "algorithms/digi/PulseDigi.h"
+#include "services/algorithms_init/AlgorithmsInit_service.h"
+#include "extensions/jana/JOmniFactory.h"
+
+namespace eicrecon {
+
+class PulseDigi_factory : public JOmniFactory<PulseDigi_factory, PulseDigiConfig> {
```
veprbl: Please come up with a more descriptive name than `PulseDigi`.

src/algorithms/digi/PulseDigi.cc
veprbl: This algorithm needs a use somewhere. We don't add dead code. Either make a unit test or include this in the analysis chain.

src/factories/digi/PulseDigi_factory.h
```diff
@@ -0,0 +1,48 @@
...
+  ParameterRef<double> m_time_window{this, "time_window", config().time_window};
+  ParameterRef<double> m_adc_phase{this, "adc_phase", config().adc_phase};
+  ParameterRef<double> m_toa_thres{this, "toa_thres", config().toa_thres};
+  ParameterRef<double> m_tot_thres{this, "tot_thres", config().tot_thres};
+  ParameterRef<unsigned int> m_capADC{this, "capADC", config().capADC};
+  ParameterRef<double> m_dyRangeADC{this, "dyRangeADC", config().dyRangeADC};
+  ParameterRef<unsigned int> m_capTOA{this, "capTOA", config().capTOA};
+  ParameterRef<double> m_dyRangeTOA{this, "dyRangeTOA", config().dyRangeTOA};
+  ParameterRef<unsigned int> m_capTOT{this, "capTOT", config().capTOT};
+  ParameterRef<double> m_dyRangeTOT{this, "dyRangeTOT", config().dyRangeTOT};
```
veprbl:
```suggestion
  ParameterRef<double> m_time_window{this, "timeWindow", config().time_window};
  ParameterRef<double> m_adc_phase{this, "adcPhase", config().adc_phase};
  ParameterRef<double> m_toa_thres{this, "toaThres", config().toa_thres};
  ParameterRef<double> m_tot_thres{this, "totThres", config().tot_thres};
  ParameterRef<unsigned int> m_capADC{this, "capADC", config().capADC};
  ParameterRef<double> m_dyRangeADC{this, "dyRangeADC", config().dyRangeADC};
  ParameterRef<unsigned int> m_capTOA{this, "capTOA", config().capTOA};
  ParameterRef<double> m_dyRangeTOA{this, "dyRangeTOA", config().dyRangeTOA};
  ParameterRef<unsigned int> m_capTOT{this, "capTOT", config().capTOT};
  ParameterRef<double> m_dyRangeTOT{this, "dyRangeTOT", config().dyRangeTOT};
```
camelCase for JANA2  parameter names, and underscores for config names. Bonus points for making variable names consistent with what is used elsewhere.

veprbl (commented):
I have some early comments.

src/factories/digi/PulseDigi_factory.h
mhkim-anl: Thank you for pointing out that the JANA2 parameters were not using camelCase. I've applied your suggested changes in `PulseDigi_factory.h`. 
For the parameter names in `PulseDigiConfig.h`, would it be better to use underscores consistently for all of them? Since `CalorimeterHitDigiConfig` uses camelCase for its parameters (it has `capADC` and `dyRangeADC`), it is a bit hard for me to decided, so I wanted to follow your suggestion.

src/algorithms/digi/PulseDigi.h
```diff
@@ -0,0 +1,43 @@
...
+      : PulseDigiAlgorithm{name,
+                           {"InputPulses"},
+                           {"OutputDigiHits"},
+                           {"ADC, TOA, and TOT are measured referring to the"
+                            "working principle of the HGCROC."}} {}
+  virtual void init() final;
+  void process(const Input&, const Output&) const;
+
+private:
+  edm4eic::HGCROCSample sample;
```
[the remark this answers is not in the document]
mhkim-anl: Thank you for pointing out that there is an unused variable. I've removed it.

src/algorithms/digi/PulseDigi.h
mhkim-anl: Thank you for pointing out the EDM4eic version requirement. I believe I understood the changes in the page you linked, and have updated the `EDM4EIC_VERSION_MIN 8.0` to `8.3.0`. Please let me know if I've misunderstood your comments.

mhkim-anl:
Regarding @veprbl's  and @wdconinc's following comments,
> Please come up with a more descriptive name than `PulseDigi`

> This algorithm needs a use somewhere.

>  introduce sufficient wiring that this actually gets used

Since BIC has come to prefer CALOROC1B chip over HGCROC, I've implemented CALOROC1B data model [(PR)](https://github.com/eic/EDM4eic/pull/134) and I'm now considering whether this algorithm should reproduce the HGCROC measurement, the CALOROC1B measurement, or both. 

I understood that only the HGCROC data model has been implemented from @ruse-traveler's previous [PR](https://github.com/eic/EDM4eic/pull/101) because the CALOROC1B chip is not available yet. However, if we implement the CALOROC1B data model, we can compare its performance with the HGCROC-based one and use this to help decide which chip will ultimately be used for the ePIC calorimeters. This study will also provide useful input for the chip development, including parameter optimization such as thresholds and dynamic ranges. 

Therefore, I would like to discuss the CALOROC1B data model implementation and the direction of this algorithm at the upcoming BIC simulation meeting. I'll invite reviewers to that discussion.

src/algorithms/digi/PulseDigi.cc
```diff
@@ -0,0 +1,137 @@
...
+
+		for (const auto& pulse : *in_pulses) {
+			double pulse_t     = pulse.getTime();
+			double pulse_dt    = pulse.getInterval();
+			std::size_t n_amps = pulse.getAmplitude().size();
+
+			// Calculate the number of samples.
+			std::size_t time_stamp_first = static_cast<std::size_t>(std::ceil((pulse_t - m_cfg.adc_phase) / m_cfg.time_window));
+			std::size_t time_stamp_last = static_cast<std::size_t>(std::ceil((pulse_t + (n_amps - 1) * pulse_dt - m_cfg.adc_phase) / m_cfg.time_window));
+			std::size_t n_samps = time_stamp_last - time_stamp_first + 1;
```
simonge: It sounds like n_samps should be a configuration parameter rather than set dynamically (between 1-7 for the real chip)

mhkim-anl:
Thank you very much for your many helpful comments. (I'm sorry for the delayed response..) Based on your comments, I've updated the algorithm. The following have been updated and I've clicked [Resolve conversation] for the related comments. 
* `PulseDigi` has been renamed to `CALOROCDigitization`.
* Description in `CALOROCDigitization.h` has been updated as it is CALOROC-specific.
* `clang-format` has been applied except for the part where it was suggested to forgo aligning the `=` signs.
* Number of samples is now a configuration parameter.
* BIC now uses this algorithm to generate `EcalBarrelScFi{P, N}CALOROCHits`.

I'll also work on the version issue and try to resolve it soon.

src/algorithms/digi/CALOROCDigitization.cc
```diff
@@ -0,0 +1,150 @@
...
+
+    const auto& entries = raw_sample.getEntries();
+
+    for (const auto& entry : entries) {
+      edm4eic::CALOROC1ASample aSample;
+      auto adc = std::max(std::llround(entry.adc / m_cfg.dyRangeSingleGainADC * m_cfg.capADC), 0LL);
+      aSample.ADC = adc > m_cfg.capADC ? m_cfg.capADC : adc;
+      auto toa    = std::max(std::llround(entry.toa / m_cfg.dyRangeTOA * m_cfg.capTOA), 0LL);
+      aSample.timeOfArrival = toa > m_cfg.capTOA ? m_cfg.capTOA : toa;
+      auto tot = std::max(std::llround(entry.tot / m_cfg.dyRangeTOT * m_cfg.capTOT), 0LL);
```
wdconinc: What's the reason for `std::max(value, 0LL)` here? Why not `std::abs`? Or why not ensure this is already positive earlier?

src/algorithms/digi/PulseDigi.cc
```diff
@@ -0,0 +1,137 @@
...
+
+		void setADC(std::size_t idx, double adc) { rawEntries[idx].adc = adc; }
+		void setTOA(std::size_t idx, double toa) { rawEntries[idx].toa = toa; }
+		void setTOT(std::size_t idx, double tot) { rawEntries[idx].tot = tot; }
+
+		const std::vector<RawEntry>& getEntries() const { return rawEntries; }
+
+	private:
+		std::vector<RawEntry> rawEntries;
+};
```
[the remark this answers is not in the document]
wdconinc: Your buffer forces a two-pass operation with type copies, so it's a slowdown. Podio datatypes avoid this through two-layer implementation with handles and data storage backend types.

src/algorithms/digi/CALOROCDigitization.cc
```diff
@@ -0,0 +1,150 @@
...
+      aSample.timeOverThreshold = tot > m_cfg.capTOT ? m_cfg.capTOT : tot;
+      out_digi_hit.addToASamples(aSample);
+
+      edm4eic::CALOROC1BSample bSample;
```
wdconinc: Place in collection right away:
```suggestion
      edm4eic::CALOROC1BSample& bSample = out_digi_hit->create();
```

CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ endif()
 set(Acts_VERSION_MIN 36.1.0)
 set(algorithms_VERSION_MIN 1.0.0)
 set(DD4hep_VERSION_MIN 1.21)
-set(EDM4EIC_VERSION_MIN 8.0)
+set(EDM4EIC_VERSION_MIN 8.7.0)
```
wdconinc: We can't require the latest release. We need backwards compatibility. This means you have to put preprocessor guards or otherwise in the code to make sure everything runs in older versions too (even if the output won't include the necessary collections).

src/algorithms/digi/CALOROCDigitization.cc
```diff
@@ -0,0 +1,150 @@
...
+
+class CALOROCRawSample {
+public:
+  struct RawEntry {
+    double adc{0};
+    double toa{0};
+    double tot{0};
+  };
+
+  CALOROCRawSample(std::size_t n_samps) { rawEntries.resize(n_samps); }
```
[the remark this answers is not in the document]
mhkim-anl: Thank you for your comment. I think I overthought the buffer implementation for ADC, TOA, and TOT. I'll update the algorithm to use a simple `RawEntry` struct rather than a class.

src/algorithms/digi/CALOROCDigitization.cc
mhkim-anl: I'm sorry if I'm misunderstanding, but I don't see a `create()`-style method in `MutableRawCALOROCHit.h`, so I can't apply this change. I suspect this may be related to why a single-pass fill (avoiding the second loop) is difficult because the TOT is determined after a couple of subsequent samples have been recorded. Please correct me if I've misunderstood something.

src/algorithms/digi/CALOROCDigitization.cc
```diff
@@ -0,0 +1,137 @@
...
+      if (idx_amp < n_amps)
+        raw_samples[i].adc = pulse.getAmplitude()[idx_amp];
+      else
+        break;
+    }
+
+    std::size_t idx_sample = 0;
+    std::size_t idx_toa    = 0;
+    double t_upcross       = 0;
+    bool tot_progress      = false;
```
wdconinc: Can we call this `is_above_threshold` or so, since "progress" is generic?

src/algorithms/digi/CALOROCDigitization.cc
```diff
@@ -0,0 +1,137 @@
...
+
+    // Measure the TOAs and TOTs while scanning the amplitudes.
+    for (std::size_t i = 1; i < n_amps; i++) {
```
wdconinc: Add a note about why you start at `i = 1` and not `i = 0` (use of `i - 1` later).

src/algorithms/digi/CALOROCDigitization.cc
veprbl: tot_progress is the hgcroc-specific terminology?

src/algorithms/digi/CALOROCDigitization.cc
```diff
@@ -0,0 +1,137 @@
...
+      if (idx_sample == m_cfg.n_samples)
+        break;
+
+      // Measure up-crossing time for TOA
+      if (!tot_progress && pulse.getAmplitude()[i] > m_cfg.toa_thres) {
+        idx_toa   = idx_sample;
+        t_upcross = get_crossing_time(m_cfg.toa_thres, pulse_dt, t, pulse.getAmplitude()[i],
+                                      pulse.getAmplitude()[i - 1]);
+        raw_samples[idx_toa].toa =
+            m_cfg.adc_phase + (time_stamp + idx_toa) * m_cfg.time_window - t_upcross;
```
veprbl: Some of this algebra looks a bit tricky, I would recommend adding a test.

veprbl (commented):
This in a very good shape. Well done! I have two thoughts on this.

src/algorithms/digi/CALOROCDigitization.cc
mhkim-anl: Thanks for the suggestion. I've added a note for why it starts from 1.

ruse-traveler (dismissed):
Awesome work on this, Minho! This is looking great from my end, and I think once Wouter & Dima are satisfied with the responses this is ready to ship! 🚀

Issue #2188, opened 2025-11-15
Inconsistent photon count in fEMCal


veprbl:
I'm realizing there might be an issue in be68024de6594cc870cfacf1e88ce9d44f3e141f. It has to do with deposited energy scales being different for homogenious and ScFi variants, which is inconsistent with using the same number for `lightYield`/`photonDetectionEfficiency`.

The homogenious path uses `corrMeanScale = 1`, whereas ScFi uses `corrMeanScale = 1/(sampling fraction)` to correct the energy, but that does not help us. That is because the correction is applied after the distribution parameter is chosen based on yields/efficiencies and photon count is determined from it:
https://github.com/eic/EICrecon/blob/be68024de6594cc870cfacf1e88ce9d44f3e141f/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L246-L247

The fix might be to rescale yield using sampling fraction for one of the variants.

cc @akioogawa

akioogawa:
Yes... I think corrMeanScale=1 for both and lightYield  = EcalEndcapP_nPhotonPerGeV / EcalEndcapP_PhotonCollectionEff / EcalEndcapP_sampFrac for ScFi should fix this.

PR #2190, opened 2025-11-17
FEMCal photon count for ScFI

akioogawa:

Fix photon count inconsistency for ScFI

- [ x] Bug fix (issue #2188)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterHitDigi.cc
+++ b/src/algorithms/calorimetry/CalorimeterHitDigi.cc
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Chao Peng, Wouter Deconinck, Sylvester Joosten, Barak Schmookler, David Lawrence
+// Copyright (C) 2022 - 2025 Chao Peng, Wouter Deconinck, Sylvester Joosten, Barak Schmookler, David Lawrence, Akio Ogawa
 
 // A general digitization for CalorimeterHit from simulation
 // 1. Smear energy deposit with a/sqrt(E/GeV) + b + c/E or a/sqrt(E/GeV) (relative value)


src/detectors/FEMC/FEMC.cc
```diff
@@ -46,12 +46,21 @@ void InitPlugin(JApplication* app) {
   decltype(CalorimeterHitDigiConfig::resolutionTDC) EcalEndcapP_resolutionTDC =
       10 * dd4hep::picosecond;
   const double EcalEndcapP_sampFrac = 0.029043; // updated with ratio to ScFi model
-  decltype(CalorimeterHitDigiConfig::corrMeanScale) EcalEndcapP_corrMeanScale =
-      fmt::format("{}", 1.0 / EcalEndcapP_sampFrac);
+  decltype(CalorimeterHitDigiConfig::corrMeanScale) EcalEndcapP_corrMeanScale = "1.0";
```
veprbl: I think, the old version was correct – it simulated the fact that gain is adjusted between the configurations to have a similar dynamic range.

veprbl: My original point was that the lightYield stuff happens before corrMeanScale is applied in digi.

akioogawa:
Also applying resolution smearing (1.0 + eResRel) factor should be before SiPM saturation. I thought we talked about it? Cannot recall maybe you rejected for a reason. At least for FEMC, edep should be modifyed by (1.0 + eResRel) factor, then go through all photon statistics and SiPM sauration.

src/detectors/FEMC/FEMC.cc
veprbl: I don't think that's the case. The formulas are supposed to divide the sampled number of photons by the expected number of photons again to convert energy back to GeV scale, so that will multiply by sampFrac once to cancel one of them.

src/algorithms/calorimetry/CalorimeterHitDigi.cc
```diff
@@ -232,6 +232,7 @@ void CalorimeterHitDigi::process(const CalorimeterHitDigi::Input& input,
                                             n_max_photons * m_cfg.capADC),
                      0LL);
     } else if (readoutType == kSipmReadout) {
+      edep *= (1.0 + eResRel);
```
veprbl: Put before this `if` block so that `kSimpleReadout` and `kPoissonPhotonReadout` cases are supported.

veprbl: I don't think it makes sense to handle `kPoissonPhotonReadout` differently. It's not used by any calorimeter. and is mostly for testing questions like "what is our kSipmReadout calorimeter didn't have saturation". If you are concerned that other users may need to smear after saturation (e.g. dominated by electronics), you'd need to add a switch to support their case.

src/detectors/FEMC/FEMC.cc
```diff
@@ -101,7 +101,8 @@ void InitPlugin(JApplication* app) {
             .threshold =
                 0.0, // 15MeV threshold for a single tower will be applied on ADC at Reco below
             .readoutType = "sipm",
-            .lightYield  = EcalEndcapP_nPhotonPerGeV / EcalEndcapP_PhotonCollectionEff,
+            .lightYield =
+                EcalEndcapP_nPhotonPerGeV / EcalEndcapP_PhotonCollectionEff / EcalEndcapP_sampFrac,
```
veprbl: I see the benchmarks are failing:
```
JException
  Message:  Timeout in worker thread
  Backtrace:
   0: ???
      /lib/x86_64-linux-gnu/libm.so.6:43c6b
   1: int std::binomial_distribution<int>::_M_waiting<std::linear_congruential_engine<unsigned long, 16807ul, 0ul, 2147483647ul> >(std::linear_congruential_engine<unsigned long, 16807ul, 0ul, 2147483647ul>&, int, double)
      /opt/software/linux-x86_64_v2/eicrecon-git.90bc377fbbcfcbc8704e54948853825b9859138d_main-kfpoooubirrt7kb7ypllshf4fmo7i3pe/lib/libalgorithms_calorimetry.so:644eb
   2: int std::binomial_distribution<int>::operator()<std::linear_congruential_engine<unsigned long, 16807ul, 0ul, 2147483647ul> >(std::linear_congruential_engine<unsigned long, 16807ul, 0ul, 2147483647ul>&, std::binomial_distribution<int>::param_type const&)
```
This means that too much time is spent calculating binomial sampling, which means we are end up with unreasonably high numbers.

The lightYield value should reflect number of photons that reach the sensor. Substituting the numbers into this formula I get 1500 / 0.285 / 0.029043 = 181,219 photons/GeV, which is unphysical. I also don't entirely understand why you divide by `EcalEndcapP_PhotonCollectionEff` here.

PR #2195, opened 2025-11-19
Implement Calo Remnant Combiner (PFA2)

### LINKED ISSUE #2129: Implement PFA2: Remnant Cluster Combiner
The 3rd stage of PFAlpha, the remnant cluster combiner, is necessary to form EMCal + HCal clusters left _after_ removing all energy due to charged particles into neutral particle candidates, i.e. `edm4eic::ReconstructedParticle`s with only select relations filled.

palsgit:
This PR implements the CaloRemnantCombiner algorithm. This algorithm ingests remnant ecal and hcal clusters after track-cluster matching and converts them to neutral particle candidates, i.e. edm4eic::ReconstructedParticles with only associated clusters filled.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #2129 )
- [ ] Documentation update
- [ ] Other: __

No.
PFAlpha neutral candidate collections will now be output by default.


src/algorithms/particle/CaloRemnantCombiner.h
```diff
@@ -0,0 +1,41 @@
...
+#include "CaloRemnantCombinerConfig.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+using CaloRemnantCombinerAlgorithm =
+    algorithms::Algorithm<algorithms::Input<std::vector<edm4eic::ClusterCollection>>,
+                          algorithms::Output<edm4eic::ReconstructedParticleCollection>>;
+
+class CaloRemnantCombiner : public CaloRemnantCombinerAlgorithm,
```
ruse-traveler: I'd suggest adding docstrings in several places. These should get propagated to our doxygen page, and help explain what's happening in the code for anyone who's looking at the source code!
```suggestion
// ==========================================================================
//! Calorimeter Remnant Cluster Combiner
// ==========================================================================
/*! An algorithm which takes multiple calorimeter cluster collections and combines them into
 *  neutral-particle candidates based on distance matching.
 */
class CaloRemnantCombiner : public CaloRemnantCombinerAlgorithm,
```

src/factories/particle/CaloRemnantCombiner_factory.h
```diff
@@ -0,0 +1,65 @@
...
+  VariadicPodioInput<edm4eic::Cluster> m_in_calo_clusters{this};
+
+  // Declare outputs
+  PodioOutput<edm4eic::ReconstructedParticle> m_out_neutral_candidates{this};
+
+  // Declare parameters
+  /*ParameterRef<double> m_min_energy_over_momentum{this, "minEnergyOverMomentum",
+                                                  config().min_energy_over_momentum};
+  ParameterRef<double> m_max_energy_over_momentum{this, "maxEnergyOverMomentum",
+                                                  config().max_energy_over_momentum};*/
```
ruse-traveler: Typo?
```suggestion
  ParameterRef<double> m_deltaRAddEM{this, "deltaRAddEM", config().deltaRAddEM};
  ParameterRef<double> m_deltaRAddH{this, "deltaRAddH", config().deltaRAddH};
```

src/factories/particle/CaloRemnantCombiner_factory.h
```diff
@@ -0,0 +1,65 @@
...
+  void Configure() {
+    // This is called when the factory is instantiated.
+    // Use this callback to make sure the algorithm is configured.
+    // The logger, parameters, and services have all been fetched before this is called
+    m_algo = std::make_unique<CaloRemnantCombiner>(GetPrefix());
+    m_algo->level(static_cast<algorithms::LogLevel>(logger()->level()));
+    // Pass config object to algorithm
+    m_algo->applyConfig(config());
+
+    m_algo->init();
```
ruse-traveler: I think we can safely drop the comments here! Those are best geared towards some template code.
```suggestion
    m_algo = std::make_unique<CaloRemnantCombiner>(GetPrefix());
    m_algo->level(static_cast<algorithms::LogLevel>(logger()->level()));
    m_algo->applyConfig(config());
    m_algo->init();
```

src/global/particle/particle.cc
```diff
@@ -0,0 +1,119 @@
...
+  /* TODO add PFA1(b) FHCal insert here */
+
+  // --------------------------------------------------------------------
+  // PFA (2) arbitration: combine remnants, form neutral candidates
+  // --------------------------------------------------------------------
+
+  // backward -----------------------------------------------------------
+
+  /* TODO add PFA2 EEEMCal here */
+  /* TODO add PFA2 EHCal here */
```
ruse-traveler: No longer to-do 😉 
```suggestion
```

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -394,6 +394,13 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
       "EcalEndcapNTrackClusterMatches",
       "HcalEndcapNTrackClusterMatches",
 
+      // particle flow
+      /* TODO move PFA0 collections here when ready */
+      /* TODO add PFA1 collections here */
+      /* TODO add PFA2 collections here */
+      "ReconstructedNeutralCandidates",
+      /* TODO add PFA3 collections here */
```
ruse-traveler: Propagating the above comments
```suggestion
      "EndcapNNeutralCandidateParticlesAlpha",
```

src/algorithms/particle/CaloRemnantCombiner.h
```diff
@@ -0,0 +1,41 @@
...
+                            public WithPodConfig<CaloRemnantCombinerConfig> {
+
+public:
+  CaloRemnantCombiner(std::string_view name)
+      : CaloRemnantCombinerAlgorithm{name,
+                                     {"CaloClusters"},
+                                     {"NeutralParticleCandidate"},
+                                     "make neutral candidates from remnant clusters"} {}
+
+  void init() final {};
```
ruse-traveler: Since you're not overwriting this method, I'd suggest dropping it from here!
```suggestion
```

src/algorithms/particle/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,154 @@
...
+      }
+      out_neutral_candidates->push_back(neutral_candidate_h);
+
+    } // end of if (seed_rem_hcal_index != -1)
+
+  } // end of while (visits_hcal != visits_hcal_true)
+
+} // end of process
+
+std::size_t CaloRemnantCombiner::findSeedCluster_index(const edm4eic::ClusterCollection& clusters,
```
ruse-traveler: Another good place for a docstring! Here you'll definitely want to note the criteria for a seed cluster and what it's for.
```suggestion
// ----------------------------------------------------------------------------
//! Find seed cluster
// ----------------------------------------------------------------------------
/*! Identifies a seed (highest energy) cluster in a collection 
 *  which sets the center of the cone in which clusters are
 *  combined.
 */
std::size_t CaloRemnantCombiner::find_seed_cluster_index(const edm4eic::ClusterCollection& clusters,
```

ruse-traveler (changes requested):
Thanks for the awesome work on this, @palsgit ! This is looking great so far! Most of my comments are really about documentation (e.g. docstrings) and being stylistically consistent. The biggest things I noted in this round are some missing declarations in the factory, and adding the factory generators for the other eta regions in the plugin.

I think there might a few ways to simplify the algorithm further using `podio::ObjectID`s and operating on the clusters themselves, but I'll think on it and we can follow-up in a subsequent pass if need be!

src/algorithms/particle/CaloRemnantCombinerConfig.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include <Evaluator/DD4hepUnits.h>
+
+namespace eicrecon {
+
+struct CaloRemnantCombinerConfig {
+
+  double deltaRAddEM = 0.03;
+  double deltaRAddH  = 0.15;
+};
```
ruse-traveler: Just for easy reference, can't hurt to add a couple more docstrings!
```suggestion
  ///! radius around seed to add EMCal clusters
  double deltaRAddEM = 0.03;

  ///! radius around see to add HCal clusters
  double deltaRAddH  = 0.15;

};
```

ruse-traveler (dismissed):
Thanks for integrating my feedback! I have a couple of extremely minor final suggestions, but this looks good from my end! I think this is ready to open for an external review!

src/factories/particle/CaloRemnantCombiner_factory.h
```diff
@@ -0,0 +1,45 @@
...
+    m_algo = std::make_unique<CaloRemnantCombiner>(GetPrefix());
+    m_algo->level(static_cast<algorithms::LogLevel>(logger()->level()));
+    m_algo->applyConfig(config());
+    m_algo->init();
+  }
+
+  void Process(int32_t /* run_number */, uint64_t /* event_number */) {
+    auto in_clusters = m_in_calo_clusters();
+    std::vector<gsl::not_null<const edm4eic::ClusterCollection*>> in_algo;
+    std::copy(in_clusters.cbegin(), in_clusters.cend(), std::back_inserter(in_algo));
```
wdconinc: Not canonical. Explain why not or make canonical.

src/factories/particle/CaloRemnantCombiner_factory.h
```diff
@@ -0,0 +1,45 @@
...
+#include "extensions/jana/JOmniFactory.h"
+#include "services/algorithms_init/AlgorithmsInit_service.h"
+
+namespace eicrecon {
+
+class CaloRemnantCombiner_factory
+    : public JOmniFactory<CaloRemnantCombiner_factory, CaloRemnantCombinerConfig> {
+private:
+  // Underlying algorithm
+  std::unique_ptr<eicrecon::CaloRemnantCombiner> m_algo;
```
wdconinc: Define private alias AlgoT. No difference in operation but no need to deviate from practice in other factories.

src/algorithms/particle/CaloRemnantCombinerConfig.h
```diff
@@ -0,0 +1,19 @@
...
+#pragma once
+
+#include <Evaluator/DD4hepUnits.h>
+
+namespace eicrecon {
+
+struct CaloRemnantCombinerConfig {
+
+  ///! radius around seed to add EMCal clusters
+  double deltaRAddEM = 0.03;
```
wdconinc: Avoid abbreviations when they make the variable less clear.

wdconinc (commented):
I think this algorithm would benefit from using set earlier.

src/algorithms/particle/CaloRemnantCombinerConfig.h
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2025 Subhadip Pal
+
+#pragma once
+
+#include <Evaluator/DD4hepUnits.h>
+
```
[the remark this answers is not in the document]
palsgit: It is kept to add flexibility on the distance parameters (i.e., in case of 2D or 3D euclidean distances) for future.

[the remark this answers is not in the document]
wdconinc: You won't be able to get checks to succeed when you have unused headers. We generally don't rely on copilot to flag them; there's a dedicated check for it.

src/algorithms/particle/CaloRemnantCombiner.h
```diff
@@ -0,0 +1,58 @@
...
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+// --------------------------------------------------------------------------
+//! Algorithm input/output
+// --------------------------------------------------------------------------
+/*! Input is a vector of calorimeter cluster collections. For now:
+  *    - 1st entry in the vector should be the EMCal collection, and
+  *    - 2nd entry in the vector should be the HCal collection.
```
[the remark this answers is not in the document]
palsgit: The order does matter here. [VariadicPodioInput()](https://eicrecon.epic-eic.org/doxygen/classJOmniFactory_1_1VariadicPodioInput.html#ae3681ea5405c0b2c4b2573893b30fbdb) uses std::vector< std::string > as its underlying container, and the algorithm accesses the collections positionally: calo_clusters[0] for Ecal and calo_clusters[1] for Hcal, as documented in the comment. I think the 'variadic' aspect here refers to accepting a variable number of collections, not to them being unordered.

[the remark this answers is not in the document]
wdconinc: Yes, but we try hard not to do positional interpretation of collections passed to the algorithms. If it is physically different, it probably should be a separate argument.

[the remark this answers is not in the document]
ruse-traveler: Adding some context here: our original intent behind using `VariadicPodioInput` was to preempt situations where we might want to provide multiple EMCal or HCal cluster collections (e.g. a particle showering into the barrel & forward endcap).

After some discussion, we felt that it would be reasonable to just stick to an input EMCal collection and an input HCal collection for now!

src/algorithms/particle/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,180 @@
...
+    //   - FIXME expand to allow for other distance metrics
+    edm4hep::Vector3f seed_pos    = seed[seed_cluster_index].getPosition();
+    edm4hep::Vector3f cluster_pos = clusters[i].getPosition();
+
+    float eta_seed    = edm4hep::utils::eta(seed_pos);
+    float phi_seed    = edm4hep::utils::angleAzimuthal(seed_pos);
+    float eta_cluster = edm4hep::utils::eta(cluster_pos);
+    float phi_cluster = edm4hep::utils::angleAzimuthal(cluster_pos);
+
+    float dphi     = phi_cluster - phi_seed;
```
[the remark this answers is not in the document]
ruse-traveler: ~~Yeah... How about `std::remainder`?~~
```suggestion
    float dphi     = std::remainder(phi_cluster - phi_seed, 2.*M_PI);
```

**EDIT:** I didn't catch what range of commits I was looking at 😅

src/algorithms/particle_flow/CaloRemnantCombiner.h
```diff
@@ -0,0 +1,116 @@
...
+      float phi_cluster             = edm4hep::utils::angleAzimuthal(cluster_pos);
+
+      float dphi = phi_cluster - phi_seed;
+      if (dphi > M_PI) {
+        dphi -= 2 * M_PI;
+      } else if (dphi < -M_PI) {
+        dphi += 2 * M_PI;
+      }
+      float deta     = eta_cluster - eta_seed;
+      float distance = std::sqrt(deta * deta + dphi * dphi);
```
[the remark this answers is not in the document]
ruse-traveler: `M_PI` is used consistently throughout our codebase, and haven't made a move to adapt `std::numbers::pi_v` yet.

src/algorithms/particle_flow/CaloRemnantCombinerConfig.h
```diff
@@ -0,0 +1,19 @@
...
+// #include <Evaluator/DD4hepUnits.h>
+
+namespace eicrecon {
+
+struct CaloRemnantCombinerConfig {
+
+  ///! DeltaR around seed to add Ecal clusters
+  double ecalDeltaR = 0.03;
+
+  ///! DeltaR around seed to add Hcal clusters
```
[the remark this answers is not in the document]
ruse-traveler: `//!` also would be a reasonable doc marker.

src/factories/particle_flow/CaloRemnantCombiner_factory.h
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2025 Subhadip Pal
+
+#pragma once
+
+#include "algorithms/particle_flow/CaloRemnantCombiner.h"
+#include "extensions/jana/JOmniFactory.h"
+#include "services/algorithms_init/AlgorithmsInit_service.h"
```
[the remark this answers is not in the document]
ruse-traveler: This will be fixed when we switch over to dedicated EMCal, HCal inputs.

src/algorithms/particle/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,180 @@
...
+  const auto [calo_clusters]    = input;
+  auto [out_neutral_candidates] = output;
+
+  std::vector<bool> visits_ecal(calo_clusters[0]->size(), false);
+  std::vector<bool> visits_hcal(calo_clusters[1]->size(), false);
+
+  std::vector<bool> visits_ecal_true(calo_clusters[0]->size(), true);
+  std::vector<bool> visits_hcal_true(calo_clusters[1]->size(), true);
+
+  while (visits_ecal != visits_ecal_true) {
```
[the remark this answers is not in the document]
palsgit: Thanks! I have implemented these suggestions in[ this commit:](https://github.com/eic/EICrecon/pull/2195/changes/7579f55f56996502d75c9367e8ff12bd2ee5b949#diff-4991667c5e6b022f6254b30f8cb68369b1dc9f29378acdb1e3a8cc3e05c28d17R50:~:text=std%3A%3Aset%3Cstd,remaining_hcal(hcal_cmp)%3B)

src/algorithms/particle/CaloRemnantCombinerConfig.h
palsgit: I have changed the variable names to ecalDeltaR and hcalDeltaR.

src/factories/particle/CaloRemnantCombiner_factory.h
palsgit: Thanks! I have implemented these suggestions in[ this commit:](https://github.com/eic/EICrecon/pull/2195/changes/7579f55f56996502d75c9367e8ff12bd2ee5b949#diff-e7a05c018dec1ca8558f4ae7f51f1fb361290958f6cd98281bf4ee9d0bc91cd9:~:text=%2B-,public%3A,std%3A%3Aunique_ptr%3CAlgoT%3E%20m_algo%3B,-17)

src/factories/particle/CaloRemnantCombiner_factory.h
palsgit: Alright. I have made this more canonical in [this commit](https://github.com/eic/EICrecon/pull/2195/changes/7579f55f56996502d75c9367e8ff12bd2ee5b949#diff-e7a05c018dec1ca8558f4ae7f51f1fb361290958f6cd98281bf4ee9d0bc91cd9R44-R45:~:text=auto%20in_algo%20%3D%20std)

src/algorithms/particle_flow/CaloRemnantCombiner.h
```diff
@@ -0,0 +1,116 @@
...
+
```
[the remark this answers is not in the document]
palsgit: cmath has been included in [this commit](https://github.com/eic/EICrecon/pull/2195/changes/a18b4a83a535e856ec972af451fad3f073e9dbcf#diff-1e9f30813644a66cfd56d8b02d478bee78596f393f88565abf15ecfcb0bad56c)

src/algorithms/particle_flow/CaloRemnantCombinerConfig.h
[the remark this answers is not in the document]
palsgit: updated in [this commit](https://github.com/eic/EICrecon/pull/2195/changes/a18b4a83a535e856ec972af451fad3f073e9dbcf#diff-3aae527f407e0f34bf1d3b87ceabb14c0a7191e00b5c149749be24f471f20753:~:text=%2B-,//!%20DeltaR%20around%20seed%20to%20add%20Hcal%20clusters,-16).

src/global/particle_flow/particle_flow.cc
```diff
@@ -0,0 +1,53 @@
...
+
+  // --------------------------------------------------------------------
+  // PFA (2) arbitration: combine remnants, form neutral candidates
+  // --------------------------------------------------------------------
+
+  // backward -----------------------------------------------------------
+
+  app->Add(new JOmniFactoryGeneratorT<CaloRemnantCombiner_factory>(
+      "EndcapNNeutralCandidateParticlesAlpha", {"EcalEndcapNClusters", "HcalEndcapNClusters"},
+      {"EndcapNNeutralCandidateParticlesAlpha"}, {.ecalDeltaR = 0.03, .hcalDeltaR = 0.15}, app));
```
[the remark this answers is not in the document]
ruse-traveler: This will be resolved in a follow-up PR to wire the entire workflow together. Full cluster collections are used only for development/unit testing.

src/factories/particle_flow/CaloRemnantCombiner_factory.h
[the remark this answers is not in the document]
palsgit: The copilot suggestion has been implemented in [this commit](https://github.com/eic/EICrecon/pull/2195/changes/a18b4a83a535e856ec972af451fad3f073e9dbcf#diff-e7a05c018dec1ca8558f4ae7f51f1fb361290958f6cd98281bf4ee9d0bc91cd9R9-R10:~:text=%23-,include%20%3Cgsl/pointers%3E,-9)

src/global/particle_flow/particle_flow.cc
```diff
@@ -234,11 +236,35 @@ void InitPlugin(JApplication* app) {
...
 
+  app->Add(new JOmniFactoryGeneratorT<CaloRemnantCombiner_factory>(
+      "EndcapNNeutralCandidateParticlesAlpha", {"EcalEndcapNClusters", "HcalEndcapNClusters"},
+      {"EndcapNNeutralCandidateParticlesAlpha"}, {.ecalDeltaR = 0.03, .hcalDeltaR = 0.15}, app));
+
+  // central ------------------------------------------------------------
+
+  app->Add(new JOmniFactoryGeneratorT<CaloRemnantCombiner_factory>(
+      "BarrelNeutralCandidateParticlesAlpha", {"EcalBarrelClusters", "HcalBarrelClusters"},
+      {"BarrelNeutralCandidateParticlesAlpha"}, {.ecalDeltaR = 0.03, .hcalDeltaR = 0.15}, app));
```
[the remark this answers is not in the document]
veprbl: Can we figure out if this is right?

src/global/particle_flow/particle_flow.cc
```diff
@@ -234,11 +236,35 @@ void InitPlugin(JApplication* app) {
...
+
+  // forward ------------------------------------------------------------
+
+  app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::Cluster, false>>(
+      "HcalEndcapPClusters", {"LFHCALClusters", "HcalEndcapPInsertClusters"},
+      {"HcalEndcapPClusters"}, app));
```
veprbl: I don't think we are still going through with idea for LFHCAL -> HcalEndcapP rename.
There is `HcalEndcapPInsert`, but not `HcalEndcapP` in the actual code.

src/global/particle_flow/particle_flow.cc
```diff
@@ -234,11 +236,35 @@ void InitPlugin(JApplication* app) {
...
+
+  app->Add(new JOmniFactoryGeneratorT<CaloRemnantCombiner_factory>(
+      "EndcapPNeutralCandidateParticlesAlpha", {"EcalEndcapPClusters", "HcalEndcapPClusters"},
+      {"EndcapPNeutralCandidateParticlesAlpha"}, {.ecalDeltaR = 0.03, .hcalDeltaR = 0.15}, app));
```
veprbl: Any way to indicate that this is a calorimeter collection? `CalEndcapPNeutralCandidateParticlesAlpha`?

src/global/particle_flow/particle_flow.cc
palsgit: ```HcalEndcapPClusters``` is a ```CollectionCollector_factory``` which collects the ```LFHCALClusters``` and ```HcalEndcapPInsertClusters```. Since PFA2 takes one ECal cluster collection and one HCal cluster collection as inputs, I used ```HcalEndcapPClusters``` to combine the ```LFHCALClusters``` and ```HcalEndcapPInsertClusters```.

veprbl: Thanks for your explanation. The issue is that in the DD4hep (and original Athena/ePIC) convention `HcalEndcapP` literally means `LFHCAL`. I would check what are other precedents in our codebase for naming merged cluster (cluster matches?) collections.

palsgit: Okay, I see the confusion here. I had a look at the naming conventions of other ```CollectionCollector_factory```s. For instance, https://github.com/eic/EICrecon/blob/main/src/global/reco/reco.cc#L66. It is very similar to what I have currently. I have a few suggestions: ```HcalEndcapPositiveClusters``` , ```PositiveHcalEndcapClusters```,  ```CombinedHcalEndcapPClusters``` (following https://github.com/eic/EICrecon/blob/fcea66d38d21bf91cd510af3400f76dd8891a8a7/src/global/tracking/tracking.cc#L462)

src/utilities/eicrecon/eicrecon.cc
```diff
@@ -25,6 +25,7 @@ std::vector<std::string> EICRECON_DEFAULT_PLUGINS = {
     "beam",
     "reco",
     "tracking",
+    "particle_flow",
```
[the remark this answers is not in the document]
ruse-traveler: This will add new collections to the EICrecon output and is not a meaningful change to default behavior from the users' perspective.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -575,6 +575,11 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
       "EcalEndcapNTrackClusterMatches",
       "HcalEndcapNTrackClusterMatches",
 
+      // particle flow
+      "EndcapNNeutralCandidateParticlesAlpha",
+      "BarrelNeutralCandidateParticlesAlpha",
+      "EndcapPNeutralCandidateParticlesAlpha",
+
```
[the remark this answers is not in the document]
ruse-traveler: See other comment: these will not be large changes in the output size and are not meaningful changes in default behavior from the users' perspective.

src/global/particle_flow/particle_flow.cc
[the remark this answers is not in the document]
ruse-traveler: This is part of the development strategy: since PFA1 was also in development when we started on this, we opted to use the existing clusters to make sure the algorithm runs. The intent was to come back and change the inputs to the remnants in a follow-up PR.

I suppose now we don't need to wait, though!

src/global/particle_flow/particle_flow.cc
ruse-traveler: Between these I personally prefer `CombinedHcalEndcapPClusters`! I think that stays in line with the patterns we've adopted elsewhere and communicates that we've merged the forward clusters...

src/global/particle_flow/particle_flow.cc
ruse-traveler: Hmmm... Is there a situation where we would ever have a neutral particle made from info other than the calorimeters? I ask because I'm not sure we add much information by noting that these are formed from calorimeter inputs...

The name lines up with the naming convention adapted for [PFA1b](https://github.com/eic/EICrecon/blob/848c29bf70ff66ac6be6116129d26b74d7edcbe3/src/global/particle_flow/particle_flow.cc#L220) (though those, of course, are made of both tracks and clusters). It's also close to the naming scheme for the [`ClustersToParticles` output](https://github.com/eic/EICrecon/blob/848c29bf70ff66ac6be6116129d26b74d7edcbe3/src/global/reco/reco.cc#L156), where this algorithm acts as the PF-parallel to that one.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,114 @@
...
+ *          radius of hcalDeltaR around seed Hcal
+ *          cluster
+ *       c. Combine all identified clusters into a neutral particle
+ *          candidate
+ */
+void CaloRemnantCombiner::process(const CaloRemnantCombiner::Input& input,
+                                  const CaloRemnantCombiner::Output& output) const {
+
+  const auto [calo_clusters]    = input;
+  auto [out_neutral_candidates] = output;
```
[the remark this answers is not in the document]
ruse-traveler: A unit test would definitely be ideal! But I think since this is new functionality (and relatively transparent functionality) and that we're writing out the collections by default it would be okay deferring the unit test to a follow-up PR.

src/global/particle_flow/particle_flow.cc
veprbl: Good point. Then let's not indicate it.

Does this collection (an its name) need to always be public, or we eventually remove it from PODIO output in favor of some more general "Reconstructed" particles?

ruse-traveler: That's my thinking! Once we have the full chain, I don't think we'll need these or the charged candidates to be public.

veprbl: Okay, so we can punt this discussion.

src/global/particle_flow/particle_flow.cc
palsgit: This is updated in the latest commit https://github.com/eic/EICrecon/pull/2195/commits/0a56c2f1fcaa0ed924a5c53922cb43d8b0d4bcc6

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,109 @@
...
+    // Gather hcal clusters within hcalDeltaR of the ecal seed
+    std::vector<std::size_t> hcal_to_merge = get_cluster_indices_for_merging(
+        *calo_clusters[1], remaining_hcal, seed_ecal_index, m_cfg.hcalDeltaR, *calo_clusters[0]);
+
+    for (const auto& idx : hcal_to_merge) {
+      neutral_candidate_eh.addToClusters((*calo_clusters[1])[idx]);
+    }
+
+    out_neutral_candidates->push_back(neutral_candidate_eh);
+
```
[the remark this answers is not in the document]
ruse-traveler: This is a reasonable suggestion, but something I think that should be handled in a follow-up PR.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,151 @@
...
+ */
+void CaloRemnantCombiner::process(const CaloRemnantCombiner::Input& input,
+                                  const CaloRemnantCombiner::Output& output) const {
+
+  const auto [ecal_clusters, hcal_clusters] = input;
+  auto [out_neutral_candidates]             = output;
+
+  // Skip event if both cluster collections are empty
+  if ((ecal_clusters->size() == 0) && (hcal_clusters->size() == 0)) {
+    error("Both ECAL and HCAL inputs are empty; skipping event.");
```
wdconinc: Is this really an error? What if someone is just running a full craterlake geometry reconstruction against a simulation that was obtained with tracking_only?

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,151 @@
...
+    return;
+  }
+
+  // Build ordered sets of remaining cluster indices (highest energy first)
+  ClusterEnergyCompare ecal_cmp{ecal_clusters};
+  ClusterEnergyCompare hcal_cmp{hcal_clusters};
```
wdconinc: This seems weird. We shouldn't need to use a struct that *holds* the clusters to define a comparator object. See the following code from ImagingTopoCluster:
```
  // Sort hit indices (podio collections do not support std::sort)
  auto compare = [&hits](const auto& a, const auto& b) {
    // if !(a < b) and !(b < a), then a and b are equivalent
    // and only one of them will be allowed in a set
    if ((*hits)[a].getLayer() == (*hits)[b].getLayer()) {
      return (*hits)[a].getObjectID().index < (*hits)[b].getObjectID().index;
    }
    return (*hits)[a].getLayer() < (*hits)[b].getLayer();
  };
  // indices contains the remaining hit indices that have not
  // been assigned to a group yet
  std::set<std::size_t, decltype(compare)> indices(compare);
```
You see how it also defines an index set but captures the hits in the lambda.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,151 @@
...
+    remaining_ecal.insert(i);
+  }
+  for (std::size_t i = 0; i < hcal_clusters->size(); ++i) {
+    remaining_hcal.insert(i);
+  }
+
+  // Phase 1: Ecal-seeded candidates
+  while (!remaining_ecal.empty()) {
+
+    edm4eic::MutableReconstructedParticle neutral_candidate_eh;
```
wdconinc: Since there's no way to break here, can we create in place up front?
```suggestion
    auto& neutral_candidate_eh = out_neutral_candidates->create();
```
and avoid the push_back?

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,151 @@
...
+
+    out_neutral_candidates->push_back(neutral_candidate_h);
+
+  } // end of hcal-seeded loop
+} // end of process
+
+/*! Collects indices of clusters within `delta_r_add` of the seed cluster,
+ *  removes them from `remaining`, and returns the collected indices.
+ */
+std::vector<std::size_t> CaloRemnantCombiner::get_cluster_indices_for_merging(
```
wdconinc: To indicate the semantics of what happens to matching indices better, can we rename to
```suggestion
std::vector<std::size_t> CaloRemnantCombiner::move_cluster_indices_for_merging(
```

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    out_neutral_candidates->push_back(neutral_candidate_eh);
+
+  } // end of ecal-seeded loop
+
+  // Phase 2: Hcal-seeded candidates (remaining hcal clusters)
+  while (!remaining_hcal.empty()) {
+
+    edm4eic::MutableReconstructedParticle neutral_candidate_h;
```
wdconinc: Same comment as above on create without push_back.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,151 @@
...
+    std::set<std::size_t, ClusterEnergyCompare>& remaining, std::size_t seed_cluster_index,
+    double delta_r_add, const edm4eic::ClusterCollection& seed) const {
+
+  std::vector<std::size_t> merged_indices;
+
+  // get the position of the seed cluster to calculate distance to other clusters
+  edm4hep::Vector3f seed_pos = seed[seed_cluster_index].getPosition();
+  float eta_seed             = edm4hep::utils::eta(seed_pos);
+  float phi_seed             = edm4hep::utils::angleAzimuthal(seed_pos);
+  double delta_r_add_use     = delta_r_add_use < 0.0 ? 0.0 : delta_r_add;
```
wdconinc:
```suggestion
  double delta_r_add_use     = delta_r_add < 0.0 ? 0.0 : delta_r_add;
```

wdconinc: Also, why introduce this and not modify `delta_r_add`? You have pass-by-value, and you don't use the passed value below.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
```diff
@@ -0,0 +1,151 @@
...
+    edm4hep::Vector3f cluster_pos = clusters[i].getPosition();
+    float eta_cluster             = edm4hep::utils::eta(cluster_pos);
+    float phi_cluster             = edm4hep::utils::angleAzimuthal(cluster_pos);
+
+    float dphi = phi_cluster - phi_seed;
+    if (dphi > M_PI) {
+      dphi -= 2 * M_PI;
+    } else if (dphi < -M_PI) {
+      dphi += 2 * M_PI;
+    }
```
wdconinc: Consider what @dvperepelitsa noticed in our calorimeter clustering when we have phi_cluster = pi - eps and phi_seed = -pi + eps. The dphi will have an unphysical wrap-around.

wdconinc (commented):
Two actual issues. Rest are relatively minor.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
ruse-traveler: Whoops! Good catch! That should be info, not an error...

src/algorithms/particle_flow/CaloRemnantCombiner.cc
ruse-traveler: Sure! Just for my own understanding: why is creating in place preferable to push back? Is it to avoid a copy?

src/algorithms/particle_flow/CaloRemnantCombiner.cc
ruse-traveler: Whoops! Thanks for catching that! And true! TBH the `delta_r_add_use` was more just a personal idiosyncracy: in cases like these I sort of like having an explicit "here's what was passed" and "here's what was used" 🤷

src/algorithms/particle_flow/CaloRemnantCombiner.cc
wdconinc: Even info seems high to me...

src/algorithms/particle_flow/CaloRemnantCombiner.cc
wdconinc: It probably doesn't even really avoid a copy, but it avoids an unowned object without collection ID or index.

ruse-traveler: Ahhhh gotcha!

veprbl: Don't do the `auto&` for objects, always `auto`.

wdconinc: Yeah, sorry, the `create` that returns a reference isn't this one.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
palsgit: I am considering to make it a debug message.

src/algorithms/particle_flow/CaloRemnantCombiner.cc
veprbl: I think logic here is correct, but there is a shorter expression that is validated:
```suggestion
    float dphi = std::remainder(phi_cluster - phi_seed, 2 * M_PI);
```

src/algorithms/particle_flow/CaloRemnantCombiner.cc
palsgit: I made this a debug message in the new commit ce3922f0e98f18e9340bb46d27adee41e1d93322

src/algorithms/particle_flow/CaloRemnantCombiner.cc
palsgit: using two lambdas for sorting ecal and hcal clusters instead of a struct holding the cluster in this commit ce3922f0e98f18e9340bb46d27adee41e1d93322

src/algorithms/particle_flow/CaloRemnantCombiner.cc
palsgit: in-place creation instead of push back is done in ce3922f0e98f18e9340bb46d27adee41e1d93322

src/algorithms/particle_flow/CaloRemnantCombiner.cc
palsgit: this change has been implemented here ce3922f0e98f18e9340bb46d27adee41e1d93322

src/algorithms/particle_flow/CaloRemnantCombiner.cc
palsgit: ```delta_r_add_use``` is no longer used ce3922f0e98f18e9340bb46d27adee41e1d93322

src/algorithms/particle_flow/CaloRemnantCombiner.cc
palsgit: the suggestion is implemented ce3922f0e98f18e9340bb46d27adee41e1d93322

veprbl (approved):
LGTM

ruse-traveler (approved):
🚀Ship it!🚀

PR #2198, opened 2025-11-22
Don't forget unattenuated energy and restore discarded sim variables in SimCalorimeterHitProcessor

sly2j:
This PR does addresses two issues in SimCalorimeterHitProcessor:
1. The unattenuated energy is necessary to benchmark e.g. clustering algorithm performance, or to train AI algorithms. While it was originally foreseen to be available through 2-step simulated hit processor, but was then not added once we combined both steps for performance reasons.
2. Some variables in the SimCalorimeterHit were silently removed in #2076 as they were deemed "not needed". However, this was not documented, and leaves the actual SimCalorimeterHits that we use for our digi/reco in a partially filled state with some information irretrievably lost. 
I addressed the issues as follows
1. Keep track of the unattenuated energy and store it with the contribution members of the SimCalorimeterHit so it can still be accessed. It means that SimCalorimeterHit.energy/sum(contribution.energies) equals the effective attenuation.
2. Restored the discarded PID/position.xyz/... fields

- [X] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
+++ b/src/algorithms/calorimetry/SimCalorimeterHitProcessor.cc
@@ -87,18 +87,22 @@ edm4hep::MCParticle lookup_primary(const edm4hep::CaloHitContribution& contrib) 
 class HitContributionAccumulator {
 private:
   float m_energy{0};
+  float m_att_energy{0};
   float m_avg_time{0};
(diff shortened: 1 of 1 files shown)


wdconinc:
(Failing eicweb trigger is known issue, and being resolved.)

PR #2202, opened 2025-11-24
treewide: add level() or applyConfig() where missing

veprbl:
There are factories that don't call `level` or `applyConfig`, which prevents `LogLevel` and other configuration parameters from being passed to the algorithm. This leads to surprising behaviors when using or debugging EICrecon.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


ruse-traveler (approved):
Nice! Good catch!

PR #2226, opened 2025-12-11
linux-eic-shell.yml: use clang for ASAN builds

veprbl:
The CI has false positive failures with ASAN/Eigen. We probably can overcome those with correct setting of `EIGEN_MALLOC_ALREADY_ALIGNED` once #2223 is ready. This should allow to unblock meanwhile.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -539,7 +539,7 @@ jobs:
     - npsim-gun
     strategy:
       matrix:
-        CXX: [g++]
+        CXX: [clang++]
         particle: [e]
         detector_config: [craterlake]
     steps:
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
> The CI has false positive failures with ASAN/Eigen. 

Not a false positive (since memory behavior is actually different). And this is technically not a solution either, since this still results in the same discrepancy between Eigen functions as compiled in ACTS and Eigen functions as compiled in EICrecon.

So, this may mean that the only reason this works it that address sanitizer in clang doesn't detect it, not that there are no issues there.

But ok, let's ship it.

veprbl:
I agree. We can revert everything to g++ (and move dis/gun jobs to g++ in addition), when the other PR is live.

PR #2255, opened 2025-12-17
Access podio metadata from JEvent parent

simonge:
Alternative to https://github.com/eic/EICrecon/pull/2253 where the "runs" podio frame is read in as a `JEvent` with `JEventLevel::Run` and set as the parent of all physics events read in through the "events" podio frames. The algorithm config is then updated with these values when `BeginRun` is called, although could be loaded elsewhere.

I prefer this approach to the service approach but there are probably advantages in both cases. There is definitely less additional code required.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #https://github.com/eic/EICrecon/issues/1336)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/extensions/jana/JOmniFactory.h
+++ b/src/extensions/jana/JOmniFactory.h
@@ -295,6 +295,40 @@ public:
   template <typename T> class ParameterRef : public ParameterBase {
 
     T* m_data;
+
+  public:
+    // Constructor
+    ParameterRef(JOmniFactory* owner, std::string name, T& slot, std::string description = "") {
+      this->m_name             = name;
+      this->m_description      = description;
+      this->m_podio_param_name = "";
+      this->m_event_level      = JEventLevel::None;
+      m_data                   = &slot;
+
+      owner->RegisterParameter(this);
+    }
+
+    const T& operator()() { return *m_data; }
+
+  private:
+    friend class JOmniFactory;
+
+    void Configure(JParameterManager& parman, const std::string& prefix) override {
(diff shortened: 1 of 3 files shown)


simonge:
Can we update the eic-shell build requirements to be more recent containers?

src/extensions/jana/JOmniFactory.h
```diff
@@ -263,20 +266,101 @@ class JOmniFactory : public JMultifactory {
   struct ParameterBase {
     std::string m_name;
     std::string m_description;
+    std::string m_podio_param_name; // Optional: name in run frame (empty means not synced)
+    JEventLevel m_event_level; // Event level to extract from (ignored if m_podio_param_name empty)
+
     virtual void Configure(JParameterManager& parman, const std::string& prefix) = 0;
     virtual void Configure(std::map<std::string, std::string> fields)            = 0;
+    virtual void UpdateFromPodioParams(const podio::Frame* frame,
```
wdconinc: I don't think we should modify the ParameterBase or ParameterRef. I think it makes more sense to implement a new PodioParameterRef:
```cpp
template <typename T> class PodioParameterRef : public ParameterBase;
```
That would allow us to encapsulate the podio metadata analysis cleanly, and separate it from the sturct ParameterRef. A factory would then *either* get its parameters from metadata *or* from a config setting, not a mixture of both. If the parameters are set from podio run metadata, then that should have primacy.

wdconinc (commented):
Partial review. I think it would be useful to discuss this in a reconstruction meeting.

PR #2257, opened 2025-12-17
Replace 25.01.1-stable with 25.07.0-stable release in compilation CI

simonge:
Updates the containers which the CI checks the compilation works with. Removing the oldest container from January and adding a new November alternative. This brings us closer to the requirement of only supporting compilation with dependencies from the last ~6 months.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -113,7 +113,7 @@ jobs:
             USE_TSAN: ON
           - CXX: clang++
             CMAKE_BUILD_TYPE: Release
-            release: 25.05.0-stable
+            release: 25.07.0-stable
           - CXX: clang++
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -113,10 +113,10 @@ jobs:
             USE_TSAN: ON
           - CXX: clang++
             CMAKE_BUILD_TYPE: Release
-            release: 25.01.1-stable
+            release: 25.05.0-stable
```
veprbl:
```suggestion
            release: 25.07.0-stable
```

veprbl:
This is good, and is the right step. The rest we will track in the issue.

PR #2264, opened 2025-12-18
Multi sensitive mpgd clusters

ybedfer:
Implement Clusters into `MultiSensitiveMPGD`, extending  "PR#https://github.com/eic/EICrecon/pull/2177".

- [ *] New feature

- [* ] Changes have been communicated to collaborators

Breaking changes if and only if MPGD XMLs from `Multi-SensitiveVolume-MPGD`  (see "https://github.com/eic/epic/pull/1014") are loaded or "-PMPGD:SiFactoryPattern" option is specified and set != 3.

No (_w.r.t. PR #2177_).


ybedfer:
This elaborates on PR #2177 (see "https://github.com/eic/EICrecon/pull/2177"), implementing Clusters.
 - Implementing them at the digitization level (_operation I call "clusterization"_).
 - Processing the clusters to produce `TrackerHits`  (_"clustering"_) in a newly added `MPGDHitReconstruction` class.

At present, the clusters are simplistic: 2-hit clusters.

It is enabled by loading dedicated `_2DStrip` MPGD XMLs, with a special `<detector_name>__2DStrip` constant defined, just as PR #2177 (_and also if "-PMPGD:SiFactoryPattern" option is specified and set != 3_). For the clusterization to be meaningful, a new version of the XMLs is needed, which requires new `epic` PR, elaborating on PR #1014.

Checks done:
 - Hit-to-Track association: 
[ClusterTrackAssociation.pdf](https://github.com/user-attachments/files/24244256/ClusterTrackAssociation.pdf)
 - Residuals RecHit-SimHit: 
[cCyMBaLRes_mu_q0.pdf](https://github.com/user-attachments/files/24244268/cCyMBaLRes_mu_q0.pdf)
[cOuterRes_mu_q0.pdf](https://github.com/user-attachments/files/24244270/cOuterRes_mu_q0.pdf)
 (SimHits here being coalesced ones, following a procedure mimicking what's done in `MPGDTrackerDigi`. )

ybedfer:
Two fixes committed in #b71832ad.

New residuals RecHit-SimHit are now more in line w/ expectations, _viz._ Std Dev = 150 µm :
[cCyMBaLRes_mu_Q0.pdf](https://github.com/user-attachments/files/24366972/cCyMBaLRes_mu_Q0.pdf)
[cOuterRes_mu_Q0.pdf](https://github.com/user-attachments/files/24366975/cOuterRes_mu_Q0.pdf)

ybedfer:
[pre-commit.ci] auto fixes (https://github.com/eic/EICrecon/pull/2264/commits/e10c282c8d67e411d9e37ab7901f4dd570a72135) keeps changing the layout of `MPGD.cc` in a way:
  - that is ugly,
  - and that is anyway not consistent.

ybedfer:
> Would you be able to contribute your changes back to the version in git?

There's a problem: in the OuterBarrel case, as I said supra, I transform the RecHit position from Master to Local. I think I cannot avoid to do this. I am not 100% sure of the fact, though. And the parameters of the transform are built-in. Meaning that the macro would cease to work when some of the parameters of the XML (IDDescriptor, position along Z) are changed. A robust modified version would require that some kind of info (you mentioned yourself a check sum) be stored in the podio file allowing to retrieve those parameters.

src/algorithms/digi/MPGDTrackerDigi.cc
```diff
@@ -294,109 +303,172 @@ void MPGDTrackerDigi::process(const MPGDTrackerDigi::Input& input,
...
+      int status = get2HitCluster(refID, locPos, surfPos, pn, generator, cluster);
+      if (status != 0) {
+        CellID vIDs = (std::uint32_t)refID;
+        CellID hIDs = refID >> 32;
+        error(R"(SimHit (= 0x{:08x}, 0x{:08x}, {:.2f},{:.2f} cm) beyond limits of {}Strips.)", hIDs,
+              vIDs, surfPos[0] / cm, surfPos[1] / cm, (pn != 0) ? 'p' : 'n');
+      }
+      // ***** DEBUGGING INFO
+      if (level() >= algorithms::LogLevel::kDebug) {
+        std::string sCellID = (pn != 0) ? "cellIDn" : "cellIDp";
```
veprbl, line 347: Inconsistent with `(pn != 0) ? 'p' : 'n'` just above.

src/detectors/MPGD/MPGD.cc
```diff
@@ -148,30 +179,51 @@ void InitPlugin(JApplication* app) {
...
+        "OuterMPGDBarrelRecHits", {"OuterMPGDBarrelRawHits"}, // Input data collection tags
+        {"OuterMPGDBarrelRecHits"},                           // Output data tag
         {
-            .readout        = "OuterMPGDBarrelHits",
-            .threshold      = 100 * dd4hep::eV,
             .timeResolution = 10,
         },
         app));
+  } else {
+    MPGDHitReconstructionConfig digi_cfg;
```
veprbl:
```suggestion
    MPGDHitReconstructionConfig reco_cfg;
```

veprbl (commented):
Couple thing I spotted so far.

src/tests/algorithms_test/algorithmsInit.cc
```diff
@@ -61,6 +68,99 @@ class algorithmsInitListener : public Catch::EventListenerBase {
     detector->add(id_desc_Silicon);
     detector->add(readoutSilicon);
 
+    // Mock MPGD readout with full geometry registered in the VolumeManager.
```
veprbl, line 77: What you think about this approach, @wdconinc? Or we just load full detector?

wdconinc: At some point maybe we need to add a local test geometry here, in xml with standard DDDetector plugins?

src/algorithms/digi/MPGDTrackerDigi.cc
ybedfer: Fixed.

src/detectors/MPGD/MPGD.cc
ybedfer: Done.

src/tests/algorithms_test/algorithmsInit.cc
veprbl: My original idea for these was to implement mocks instead of using proper DD4hep. If we commit to real geometry further, it might make sense to just trivial setups to epic. We could also advertise them for beam tests.

veprbl (approved):
I have no more comments on this.

Tests were vibe coded. I did not push it to look for edge cases, but hopefully it will at least help with maintainability.

ybedfer:
Hello,
Where are we, in term of PR being accepted? I"m lost...

veprbl:
Sorry for confusion. This is approved. As usual, let's wait a day in case any more comments, if someone else takes a look, and then merge.

veprbl:
(Also during review I checked capybara report, it looks reasonable)

PR #2278, opened 2025-12-31
fix: iwyu: map fmt/base.h to fmt/format.h

wdconinc:
This PR fixes the mapping to ensure that both fmt-10 and fmt-11 headers are supported simultaneously.


veprbl (approved):
Makes sense now. This is good.

PR #2282, opened 2026-01-02
refactor: use std::map with comparator for deterministic ordering in ActsToTracks

wdconinc:
This PR is similar to #2161 but for ActsToTracks.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/ActsToTracks.cc
+++ b/src/algorithms/tracking/ActsToTracks.cc
@@ -30,7 +30,7 @@
 
 namespace eicrecon {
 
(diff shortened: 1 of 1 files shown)


wdconinc:
Does not seem to fix tracking collection irreproducible behavior.

wdconinc:
> Does not seem to fix tracking collection irreproducible behavior.

But still probably a useful thing to add. It may cause irreproducible behavior in some cases since we iterate over the map.

veprbl (approved):

PR #2314, opened 2026-01-11
fix: initialize variables in LGADHitClustering

wdconinc:
This PR initializes variables which g++ flags with maybe-uninitialized warnings (which we suppress from -Werror).

- [x] Bug fix (issue: uninitialized and later relied upon in a court of law)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/LGADHitClustering.cc
+++ b/src/algorithms/tracking/LGADHitClustering.cc
@@ -21,7 +21,7 @@
 #include <edm4eic/Cov3f.h>
 #include <edm4eic/CovDiag3f.h>
 #include <edm4hep/Vector2f.h>
-#include <stddef.h>
+#include <cstddef>
 #include <Eigen/Core>
 #include <cmath>
 #include <gsl/pointers>


src/algorithms/tracking/LGADHitClustering.cc
```diff
@@ -68,14 +67,14 @@ void LGADHitClustering::_calcCluster(const Output& output,
...
-  float max_charge_sigma2_x;
-  float max_charge_sigma2_y;
+  dd4hep::rec::CellID cellID = 0;
+  auto max_charge            = std::numeric_limits<float>::min();
+  auto earliest_time         = std::numeric_limits<float>::max();
+  float time_err{0};
+  float max_charge_x{0};
+  float max_charge_y{0};
+  float max_charge_sigma2_x{0};
+  float max_charge_sigma2_y{0};
```
veprbl:
```suggestion
  float time_err{NAN};
  float max_charge_x{NAN};
  float max_charge_y{NAN};
  float max_charge_sigma2_x{NAN};
  float max_charge_sigma2_y{NAN};
```
Any reason to not declare those (currently) unreachable variables as undefined?

wdconinc: They are not unreachable. There is no (typed) reason why `getEdep()` is positive, and the presence of an `error` for negative `Edep` seems to indicate that it can happen. If there are only negative `Edep` hits, then these values persist into the output for the configuration where `m_cfg.useAve == false`.

PR #2325, opened 2026-01-12
JEventProcessorPODIO: Add Forward{RomanPot,OffMTracker}Hits to the default output

### LINKED ISSUE #2322: Question/problem with getting MC particles from ForwardRomanPotRawHitAssociations
Hi experts,

I am trying to use "ForwardRomanPotRawHitAssociations" to find the MC particles that hit the RP, but it does not seem to store the MC information.

When I do:

```
auto& rawAssoc = mEvent->get<edm4eic::MCRecoTrackerHitAssociationCollection>("ForwardRomanPotRawHitAssociations");
for (const auto& assoc : rawAssoc) 
{
        std::cout << "rec: " << assoc.getRawHit() << std::endl;
        std::cout << "sim: " << assoc.getSimHit() << std::endl;
}

```

I get:
> rec:  id: 6d8311ae|8
> cellID : 18439707237955568283
> charge : 86
> timeStamp : 103735
>
> sim: [not available]

The the simHit seems to be always not available. 

The simulation sample I am looking at is the Oct. campaign BeAGLE eHe3 sample. If there is another way to check what particles hit the RP and the OMD, I am happy to try that too.

veprbl:
Resolves: #2322

We don't expect size of the output to increase significantly from this, but it would enable some studies.

### DIFF
--- a/src/services/io/podio/JEventProcessorPODIO.cc
+++ b/src/services/io/podio/JEventProcessorPODIO.cc
@@ -178,8 +178,10 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
       "ForwardRomanPotStaticRecParticles",
       "ForwardOffMRecParticles",
 
+      "ForwardRomanPotHits",
       "ForwardRomanPotRawHits",
       "ForwardRomanPotRawHitAssociations",
+      "ForwardOffMTrackerHits",
       "ForwardOffMTrackerRawHits",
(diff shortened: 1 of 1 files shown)


ruse-traveler (approved):
LGTM!

Meatbally2 (approved):
Thank you so much! I can see the MC particle information from RawHitAssociations.getSimHits now!

PR #2380, opened 2026-01-30
feat(ci): rm truth in two-stage digi+reco running; add to capybara

wdconinc:
This PR aims to make the split digi + reco two-stage running in CI functional and testable, to the point where we can identify what still has hard requirements on truth collections and where they need to be made optional.

This PR expands two-stage digitization from just `EcalBarrel.*RawHits` to all `.*RawHits` and `.*PulsesWithNoise` (note: we need a better nomenclature for the pulses that are going to be coming from DAQ, e.g. `.*RawPulses` or something).

This PR modifies two-stage reconstruction such that any limitation on which output collections to produce is removed, so *default* EICrecon will need to be able to handle whatever we throw at it.

This two-stage running is only running single particle e- simulations, but at some point this could be changed to DIS to have better coverage for other detector systems.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: allow running EICrecon on digitization-only files)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/reco/Beam.h
+++ b/src/algorithms/reco/Beam.h
@@ -6,6 +6,7 @@
 #include <Math/Vector4D.h>
 #include <edm4hep/MCParticleCollection.h>
 #include <edm4eic/ReconstructedParticleCollection.h>
+#include <algorithm>
 #include <set>
 
 using ROOT::Math::PxPyPzEVector;
@@ -15,11 +16,10 @@ namespace eicrecon {
 template <class T> auto find_first_with_pdg(const T* parts, const std::set<int32_t>& pdg) {
   T c;
   c.setSubsetCollection();
-  for (const auto& p : *parts) {
-    if (pdg.count(p.getPDG()) > 0) {
-      c.push_back(p);
-      break;
-    }
+  const auto it = std::find_if(parts->begin(), parts->end(),
+                               [&pdg](const auto& p) { return pdg.count(p.getPDG()) > 0; });
+  if (it != parts->end()) {
+    c.push_back(*it);
   }
   return c;
 }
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -608,7 +608,7 @@ jobs:
           export DETECTOR_CONFIG=${DETECTOR}_${{ matrix.detector_config }}
           export LD_LIBRARY_PATH=$PWD/install/lib:$LD_LIBRARY_PATH
           export JANA_PLUGIN_PATH=$PWD/install/lib/EICrecon/plugins${JANA_PLUGIN_PATH:+:${JANA_PLUGIN_PATH}}
-          $PWD/install/bin/eicrecon ${{env.JANA_OPTIONS}} ${{env.JANA_OPTIONS_GUN}} -Ppodio:output_collections=EventHeader,MCParticles,EcalBarrelScFiRawHits,EcalBarrelImagingRawHits -Ppodio:output_file=two_stage_raw_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root -Pplugins=dump_flags,janadot -Pdump_flags:json=${{ matrix.particle }}_${{ matrix.detector_config }}_flags.json
+          $PWD/install/bin/eicrecon ${{env.JANA_OPTIONS}} ${{env.JANA_OPTIONS_GUN}} -Ppodio:output_collections="EventHeader,.*RawHits,.*PulsesWithNoise" -Ppodio:output_file=two_stage_raw_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4eic.root sim_${{ matrix.particle }}_1GeV_20GeV_${{ matrix.detector_config }}.edm4hep.root -Pplugins=dump_flags,janadot -Pdump_flags:json=${{ matrix.particle }}_${{ matrix.detector_config }}_flags.json
```
veprbl: I don't think anything will work to start from RawCalorimeterHits without MCParticles. Should we pick something from tracking instead?

PR #2383, opened 2026-01-31
fix: use find_if instead of iterator loops with break

wdconinc:
Needs:
- [x] podio-1.3 (https://github.com/AIDASoft/podio/pull/626), in eic-shell as of 2025.09

In a few places, due to historical limitations of podio collections, we were still using iterator-based loops with tests and early break to do things that are more nicely done with `find_if` nowadays. This PR introduces the `find_if` based approach in a few cases, and removes a comment block complaints about all that used to not work...

- [x] Bug fix (issue: tech debt)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


simonge (approved):
Looks good. Capybara differences appear to be from other simultaneous PRs.

Issue #2258, opened 2025-12-17
Retire support for compiling in old eic containers (December 2025)


veprbl:
|           Tag|    acts|  dd4hep| edm4eic| edm4hep|   jana2|   podio|
|-|-|-|-|-|-|-|
|25.02.0-stable|  35.2.0|    1.30|   8.0.0|  0.10.5|   2.4.0|     1.1|
|25.03.1-stable|  35.2.0|    1.30|   8.0.0|  0.10.5|   2.4.0|     1.1|
|25.04.1-stable|  35.2.0|    1.30|   8.0.0|  0.99.1|   2.4.0|     1.1|
|25.05.0-stable|  35.2.0|    1.31|   8.0.1|  0.99.1|   2.4.0|     1.2|
|25.06.1-stable|  36.3.2|    1.32|   8.0.1|  0.99.1|   2.4.2|     1.2|
|**25.07.0-stable**|  36.3.2|  1.32.1|   8.0.1|  0.99.1|   2.4.2|     1.2|
|**25.08.0-stable**|  39.2.0|  1.32.1|   8.0.1|  0.99.1|   2.4.2|     1.2|
|**25.09.0-stable**|  39.2.0|  1.32.1|   8.0.1|  0.99.1|   2.4.2|     1.3|
|**25.10.4-stable**|  39.2.0|  1.32.1|   8.0.1|  0.99.1|   2.4.2|     1.3|
|**25.11.2-stable**|  39.2.0|  1.32.1|   8.4.0|  0.99.1|   2.4.2|     1.3|
|**25.12.0-stable**|  39.2.0|    1.33|   8.5.0|  0.99.1|   2.4.3|     1.6|

The 6-month support period suggests that we are retiring:
- #2259
- ~~DD4hep 1.30, 1.31~~
- #2260
- #2387
- ~~PODIO 1.1~~

#### Code
<details>
  <summary>versions.sh</summary>
<pre>
#!/usr/bin/env bash
tags=(
25.02.0-stable
25.03.1-stable
25.04.1-stable
25.05.0-stable
25.06.1-stable
25.07.0-stable
25.08.0-stable
25.09.0-stable
25.10.4-stable
25.11.2-stable
25.12.0-stable
)
packages=(
acts
dd4hep
edm4eic
edm4hep
jana2
podio
)

printf "|%14s|" "Tag"
for package in ${packages[@]}; do
    printf "%8s|" "$package"
done
printf "\n"
printf "|-|"
[...]

PR #2387, opened 2026-01-31
drop support for JANA 2.3.x and older

veprbl:
This is overdue for https://github.com/eic/EICrecon/issues/2258 and ones before.


wdconinc (commented):
Don't we need an updated min version in CMakeLists.txt too?

wdconinc (approved):

PR #2388, opened 2026-01-31
drop support for PODIO 1.1 and older

veprbl:
This is for https://github.com/eic/EICrecon/issues/2258


veprbl:
Ah, misread the condition. We need this for 1.2 still. Revisit this in several months.

PR #2397, opened 2026-02-02
fix: avoid nan in InclusiveKinematicsJB when y >= 1 and/or x >= 1

### LINKED ISSUE #2391: InclusiveKinematicsJB produces nan output
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? CI 
- Any special versions of ROOT or Geant4? CI

1. https://github.com/eic/EICrecon/actions/runs/21556164858/job/62113469757?pr=2289#step:7:24

No nan.

Nan in W.

wdconinc:
This PR fixes #2391 by avoiding any inclusive kinematics with y > 1 or x > 1. These are possible in reconstruction (since stuff may be reconstructed in ways that violate energy conservation), and may lead to nan in W due to sqrt of negative numbers.

Note: In the case of ion targets where x > 1 is allowable this will need modifications. But I assume that will require changes to the assumptions early in this algorithm too (e.g. mass of proton assumption).

- [x] Bug fix (issue #2391)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/reco/InclusiveKinematicsJB.cc
+++ b/src/algorithms/reco/InclusiveKinematicsJB.cc
@@ -62,14 +62,14 @@ void InclusiveKinematicsJB::process(const InclusiveKinematicsJB::Input& input,
   // Calculate kinematic variables
   static const auto m_proton = m_particleSvc.particle(2212).mass;
   const auto y_jb            = sigma_h / (2. * ei.energy());
-  if (y_jb > 1) {
+  if (y_jb >= 1) {
     // y > 0 is mathematically guaranteed by sigma_h > 0, but y < 1 is not
     debug("InclusiveKinematicsJB: event with y > 1 skipped");
     return;
   }
   const auto Q2_jb = ptsum * ptsum / (1. - y_jb);
   const auto x_jb  = Q2_jb / (4. * ei.energy() * pi.energy() * y_jb);
-  if (x_jb > 1) {
+  if (x_jb >= 1) {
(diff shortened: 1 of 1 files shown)


src/algorithms/reco/InclusiveKinematicsJB.cc
```diff
@@ -62,11 +62,21 @@ void InclusiveKinematicsJB::process(const InclusiveKinematicsJB::Input& input,
...
-  const auto W_jb            = sqrt(m_proton * m_proton + 2 * m_proton * nu_jb - Q2_jb);
-  auto kin                   = out_kinematics->create(x_jb, Q2_jb, W_jb, y_jb, nu_jb);
+  if (y_jb >= 1) {
+    // y > 0 is mathematically guaranteed by sigma_h > 0, but y < 1 is not
+    debug("InclusiveKinematicsJB: event with y >= 1 skipped");
+    return;
+  }
+  const auto Q2_jb = ptsum * ptsum / (1. - y_jb);
+  const auto x_jb  = Q2_jb / (4. * ei.energy() * pi.energy() * y_jb);
+  if (x_jb >= 1) {
```
smaple1: I understand that `y_jb == 1` must be disallowed to prevent division by 0 - is there a reason to disallow `x_jb == 1`?

smaple1 (approved):
LGTM

PR #2399, opened 2026-02-02
Add particleconverter pfa3

### LINKED ISSUE #2130: Implement PFA3: Candidate Particle Regressor
The final stage of PFAlpha, the candidate particle regressor/converter, is necessary to turn candidate particles into full-blown reconstructed particles.

emolinac:
The ParticleConverter factory. This factory takes information from tracking and calorimeters systems in order to create a final reconstructed particle.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #2130)
- [ ] Documentation update
- [ ] Other: __

No
No


src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+
+#include "ParticleConverter.h"
+
+namespace eicrecon {
+void ParticleConverter::process(const Input& input, const Output& output) const {
+  const auto [in_particles] = input;
+  auto [out_particles]      = output;
+
+  trace("----------------------------------------------------------------");
+  trace("                                                                ");
```
ruse-traveler: EICrecon is already pretty verbose, so I'd suggest keeping debugging output to the essentials.
```suggestion
```

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+  trace("We have {} particles as input", in_particles->size());
```
ruse-traveler: I'd suggest adding a guard against no particles in input. And for debugging messages outside of internal loops, we can output them with `debug` (1 level above `trace` in verbosity)!
```suggestion
  if (in_particles->size() == 0) {
    debug("No particle candidates in input");
    return;
  } else {
    debug("We have {} particles as input", in_particles->size());
  }
```

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+          if (det_element_type.find(ecal_string) != std::string::npos) {
+            isECal = true;
+
+            ecal_energy += calo_hit.getEnergy();
+          }
+          if (det_element_type.find(hcal_string) != std::string::npos) {
+            isHCal = true;
+
+            hcal_energy += calo_hit.getEnergy();
+          }
```
ruse-traveler: I'd suggest using the DD4hep type enum here: it should be a fairly robust against upstream changes and lines up with their intended usage!
```suggestion
          if (type.is(ddh4ep::DetType::CALORIMETER) && type.is(dd4hep::DetType::ELECTROMAGNETIC)) {
            isECal = true;
            ecal_energy += calo_hit.getEnergy();
          }
          if (type.is(ddh4ep::DetType::CALORIMETER) && type.is(dd4hep::DetType::HADRONIC)) {
            isHCal = true;
            hcal_energy += calo_hit.getEnergy();
          }
```

src/algorithms/particle/ParticleConverterConfig.h
```diff
@@ -0,0 +1,16 @@
...
+
+#pragma once
+
+namespace eicrecon {
+struct ParticleConverterConfig {
+  double tracking_resolution = 1;
+
+  double ecal_resolution  = 1;
+  double hcal_resolution  = 1;
+  double calo_energy_norm = 1;
```
ruse-traveler: We'll also need a factor to account for the EM-hadronic calorimeter scale!
```suggestion
  double calo_hadron_scale = 1;
  double calo_energy_norm = 1;
```

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+    if (isECal && !isHCal && !isTrack)
+      prelim_pid = 22; //photon
+    if (!isECal && isHCal && !isTrack)
+      prelim_pid = 2112; // neutron
+
+    // Step 2 : Calculate calo energy (PRELIMINARY IMPLEMENTATION)
+    if (ecal_energy > 0)
+      calo_energy += ecal_energy;
+    if (hcal_energy > 0)
+      calo_energy += hcal_energy;
```
ruse-traveler: Lining up with my suggestion on the config: we'll also want to scale the hadronic energy relative to the electromagnetic!
```suggestion
      calo_energy += m_cfg.calo_hadron_scale * hcal_energy;
```

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+    trace(" Total energy of the particle is E = {} GeV, calo.Energy = {}, track.Energy = {}",
+          avge_energy, calo_energy, track_energy);
+
+    // Step 4 : Store information on a mutable collection
+    double mass_calculated = std::sqrt(std::pow(avge_energy, 2) - std::pow(track_momentum_mag, 2));
+
+    edm4eic::MutableReconstructedParticle out_reco_particle;
+
+    out_reco_particle.setType(
+        0); // Placeholder value until I know what this variable really stands for
```
ruse-traveler: For reconstructed particles, we don't really use the type for anything right now (so 0 is fine). But we do use it to store the generator status for MC particles!

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+
+    out_reco_particle.setStartVertex(particle.getStartVertex());
+    out_reco_particle.setParticleIDUsed(particle.getParticleIDUsed());
+
+    // NOTE: Not necessary?
+    // out_reco_particle.setClusters(particle.getClusters());
+    // out_reco_particle.setTracks(particle.getTracks());
+    // out_reco_particle.setParticles(particle.getParticles());
+    // out_reco_particle.setParticleIDs(particle.getParticleIDs());
+
```
ruse-traveler: And if you clone the particle, this block will be unnecessary! FYI if you don't clone the particle, though, then you will need to fill in the relations again.
```suggestion
```

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+    out_particles->push_back(out_reco_particle);
+
+    trace("                                                                ");
+  }
+
+  trace("                                                                ");
+  trace("----------------------------------------------------------------");
```
ruse-traveler: Ditto.
```suggestion
```

src/factories/particle/ParticleConverter_factory.h
```diff
@@ -0,0 +1,54 @@
+// author : Esteban Molina, Derek Anderson
+// date   : January 2026
```
ruse-traveler:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2026 Esteban Molina, Derek Anderson
```

src/factories/particle/ParticleConverter_factory.h
```diff
@@ -0,0 +1,54 @@
...
+  // - Reconstructed particles
+  PodioOutput<edm4eic::ReconstructedParticle> m_recoparticles_output{this};
+
+  // Parameters
+  // ParameterRef<double> m_tracking_resolution {this, "tracking_resolution", config().tracking_resolution};
+  // ParameterRef<double> m_ecal_resolution     {this, "ecal_resolution", config().ecal_resolution};
+  // ParameterRef<double> m_hcal_resolution     {this, "hcal_resolution", config().hcal_resolution};
+  // ParameterRef<double> m_calo_energy_norm    {this, "calo_energy_norm", config().calo_energy_norm};
+
+  // ParameterRef<bool> m_use_resolution_in_ecalc {this, "use_resolution_in_ecalc", config().use_resolution_in_ecalc};
```
ruse-traveler: Definitely will want these uncommented! You'll need these to make sure that any configs provided by a plugin or user will get applied!
```suggestion
  ParameterRef<double> m_tracking_resolution {this, "tracking_resolution", config().tracking_resolution};
  ParameterRef<double> m_ecal_resolution     {this, "ecal_resolution", config().ecal_resolution};
  ParameterRef<double> m_hcal_resolution     {this, "hcal_resolution", config().hcal_resolution};
  ParameterRef<double> m_calo_hadronic_scale {this, "calo_hadronic_scale", config().calo_hadronic_scale);
  ParameterRef<double> m_calo_energy_norm    {this, "calo_energy_norm", config().calo_energy_norm};
```

src/global/particle/particle.cc
```diff
@@ -0,0 +1,124 @@
...
+  /* TODO add PFA3 FEMC here */
+  /* TODO add PFA3 LFHCAL here */
+  /* TODO add PFA3 FHCal insert here */
+
+  /* TODO collect reconstructed particles here */
+
+  // ASK ABOUT THIS SECOND ARGUMENT -> FALSE
+  app->Add(new JOmniFactoryGeneratorT<ParticleConverter_factory>(
+      "FinalReconstructedParticles", {"ReconstructedParticles"}, {"FinalReconstructedParticles"},
+      {}, app));
```
ruse-traveler: Then, even though it'll be dummy values for now, I'd suggest adding in the config values here! We'll have to modify them for each eta region!
```suggestion
  app->Add(new JOmniFactoryGeneratorT<ParticleConverter_factory>(
      "FinalReconstructedParticles", {"ReconstructedParticles"}, {"FinalReconstructedParticles"},
      {
        .ecal_resolution  = 1.0,
        .hcal_resolution  = 1.0,
        .calo_hadronic_scale = 1.0,
        .calo_energy_norm = 1.0
      }, app));
```

ruse-traveler (commented):
Thanks for the great work on this, @emolinac ! This is looking great so far! Most of comments are either building on what we discussed earlier (e.g. adding a hadronic scale factor to the config) or on cleaning up stuff!

I'll likely have another round of comments on smaller things (camel/snake case in the config, docstrings in a couple more places, etc.). But those can wait for the next round!

src/algorithms/particle/ParticleConverter.cc
emolinac: Unnecessary prints were removed. I was using them while doing the debugging and forgot to remove them oops

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,155 @@
...
+      track_momentum_mag =
+          std::sqrt(std::pow(track_momentum_vector.x, 2) + std::pow(track_momentum_vector.y, 2) +
+                    std::pow(track_momentum_vector.z, 2));
+
+      track_energy = std::sqrt(std::pow(track_momentum_mag, 2) + std::pow(track_mass, 2));
+    }
+
+    // Looking for clusters
+    for (auto cluster : particle.getClusters()) {
+      for (auto calo_hit : cluster.getHits()) {
```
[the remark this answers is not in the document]
emolinac: This is related to what you were saying to me the other day about no hcal in the reco process at the moment?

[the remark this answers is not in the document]
ruse-traveler: We're also facing bottlenecks right now from sheer combinatorics, so anywhere we can avoid loops is welcome 😅

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,159 @@
...
+    if (!isECal && isHCal && !isTrack)
+      prelim_pid = 2112; // neutron
+
+    // Step 2 : Calculate calo energy (PRELIMINARY IMPLEMENTATION)
+    if (ecal_energy > 0)
+      calo_energy += ecal_energy;
+    if (hcal_energy > 0)
+      calo_energy += m_cfg.calo_hadron_scale * hcal_energy;
+
+    calo_energy *= m_cfg.calo_energy_norm;
```
ruse-traveler: There might be cases where you have only ECal clusters (e.g. a photon), and in these you'll want to use the ECal energy directly! We'll only need the hadron scale + normalization when the HCal is also involved.

The other edge case is when you have only the HCal directly. I suspect we'll need a separate parameter for these cases, but for now I think it's okay to use the HCal energy directly and flag it as a TODO.

So I'd suggest the following change:
```suggestion
    // TODO need to consider how to handle cases
    // with only HCal clusters. Will use HCal energy
    // directly in interim.
    if (hasECal) {
      calo_energy = ecal_energy;
      if (hasHCal) {
        calo_energy += m_cfg.calo_hadron_scale * hcal_energy;
        calo_energy *= m_cfg.calo_energy_norm;
      }
    } else if (hasHCal) {
      calo_energy = hcal_energy;
    }
```

src/algorithms/particle/ParticleConverter.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <string>
+#include <string_view>
+
+#include "algorithms/interfaces/WithPodConfig.h"
+#include "algorithms/particle/ParticleConverterConfig.h"
+
+// Class definition
+namespace eicrecon {
+// Define an "alias" for the templated algorithms constructor
```
ruse-traveler:
```suggestion
```

src/algorithms/particle/ParticleConverter.h
```diff
@@ -0,0 +1,47 @@
...
+using ParticleConverterAlgorithm =
+    algorithms::Algorithm<algorithms::Input<edm4eic::ReconstructedParticleCollection>,
+                          algorithms::Output<edm4eic::ReconstructedParticleCollection>>;
+
+// Define the class of particle converter which uses as a base the algorithms class
+class ParticleConverter : public ParticleConverterAlgorithm,
+                          public WithPodConfig<ParticleConverterConfig> {
+public:
+  // Constructor of ParticleConverter inherits from the constructor of ParticleConverterAlgorithm
```
ruse-traveler:
```suggestion
```

src/algorithms/particle/ParticleConverterConfig.h
```diff
@@ -0,0 +1,17 @@
...
+namespace eicrecon {
+struct ParticleConverterConfig {
+  double tracking_resolution = 1;
+
+  double ecal_resolution   = 1;
+  double hcal_resolution   = 1;
+  double calo_hadron_scale = 1;
+  double calo_energy_norm  = 1;
+
+  bool use_resolution_in_ecalc = false;
```
ruse-traveler: There are 2 bigger suggestions I have here:
1. Dropping the individual ECal/HCal resolution since we'll only use the combined resolution
2. Adding docstrings for doxygen and internal documentation

And then, this isn't something we've really enforced, but we tend to use camel case for algorithm parameters in most places. So, for consistency, I'd suggest considering camel case! 
```suggestion
  ///! Momentum resolution of tracker in this eta region
  double tracking_resolution = 1;

  ///! Energy resolution of combined EMCal + HCal system
  double caloResolution   = 1;

  ///! Controls relative contribution of HCal energy to total
  double caloHadronScale = 1;

  ///! Normalizes total EMCal + HCal energy
  double caloEnergyNorm  = 1;

  ///! Turn on/off taking resolution-weighted average of track
  ///! and calo energy
  bool useResolutionInEnergyCalc = false;
```

src/algorithms/particle/ParticleConverter.h
```diff
@@ -0,0 +1,47 @@
...
+  ParticleConverter(std::string_view name)
+      : ParticleConverterAlgorithm(name, {"inputRecoParticles"}, {"outputRecoParticles"},
+                                   "Particles as such (?)") {};
+
+  void init() final {}; // what is the use of making it final?
```
ruse-traveler: Ah! I'm pretty sure it's for devirtualization: this should let our compilers use these as direct functions instead of going through the vtable...
```suggestion
  void init() final {};
```

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,159 @@
...
+    out_reco_particle.setReferencePoint(particle.getReferencePoint());
+
+    out_reco_particle.setCharge(particle.getCharge());
+    out_reco_particle.setMass(mass_calculated);
+    out_reco_particle.setGoodnessOfPID(particle.getGoodnessOfPID());
+    out_reco_particle.setCovMatrix(particle.getCovMatrix());
+    out_reco_particle.setPDG(prelim_pid);
+
+    out_reco_particle.setStartVertex(particle.getStartVertex());
+    out_reco_particle.setParticleIDUsed(particle.getParticleIDUsed());
```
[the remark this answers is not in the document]
ruse-traveler: The `.getCharge()` field will be empty for candidate particles, so you'll need to calculate it here. For now, I think it's okay to just grab the charge from the track, e.g.
```c++
charge_used = 0.0;
if (track.has_value()) {
  /*... other stuff ...*/
  charge_used = track.get_value().getCharge();
}
```

ruse-traveler (changes requested):
Thanks for incorporating my feedback! This is looking great! I've left a 2nd round of comments: there were some edge-cases I didn't catch last time and some areas where I think you can really simplify things. Once these are addressed, I think we're ready to open for review!

One thing I didn't note in a dedicated comment is that for cases where you don't have a track, you'll have to calculate the 3 vector from the clusters! My suggestion would be to do something like the following:

(1) We'll need to know the primary vertex, so we'll have to add an `edm4eic::VertexCollection` as input.
- The name of the primary vertex collection is "PrimaryVertices"
- If it's empty, then we can use `(0, 0, 0)`
- And otherwise we can just use the highest-ranked vertex, which is defined to be the 1st entry in the collection:
```c++
edm4hep::Vector3f primary_vertex(0.0, 0.0, 0.0);
if (in_vertices.size() > 0) {
  primary_vertex = (*in_vertices)[0].getPosition();
} else {
  /* emit a debugging message */
}
```

(2) Calculate an energy weighted average of the ECal and HCal cluster positions
```c++
edm4hep::Vector3f ecal_position(0.0, 0.0, 0.0);
edm4hep::Vector3f hcal_position(0.0, 0.0, 0.0);
for (const auto& cluster : clusters) {
  /*... other stuff ...*/
  if (isECal) {
    ecal_energy += cluster.getEnergy();
    ecal_position = ecal_position + (cluster.getEnergy() * cluster.getPosition());
  }
  if (isHCal) {
    hcal_energy += cluster.getEnergy();
    hcal_position = hcal_position + (cluster.getEnergy() * cluster.getPosition());
  }
}

// normalize average position appropriately
if (ecal_energy > 0.0) ecal_position = ecal_position / ecal_energy;
if (hcal_energy > 0.0) hcal_position = hcal_position / hcal_energy;
```

(3) Then, after we've determined what energy and mass to use, we'll have to calculate the 3-momentum. There are basically 2 cases:
```c++
// 1st get magnitude of 3-momentum based on updated energy
const double magnitude_used = std::sqrt((energy_used * energy_used) - (mass_calculated * mass_calculated));

// Case 1: we have a track, so we should default to that for the 3-momentum
//   --> Get unit vector from track, and then update with new magnitude
const double track_magnitude = edm4hep::utils::magnitude(track_momentum);
const edm4hep::Vector3f track_direction = edm4hep::utils::normalizeVector(track_momentum, track_magnitude);
const edm4hep::Vector3f momentum_used = magnitude_used * track_direction;

// Case 2: we only have calorimeters
//   --> Get direction going from vertex to calorimeter cell, and use magnitude
const edm4hep::Vector3f calo_displacement = calo_position - primary_vertex;
const double calo_distance = edm4hep::utils::magnitude(calo_direction);
const edm4hep::Vector3f calo_direction - edm4hep::utils::normalizeVector(calo_displacement, calo_distance);
const edm4hep::Vector3f momentum_used = magnitude_used * calo_direction;
```

If you have both ECal and HCal clusters, I'd suggest just defaulting to the ECal position!

src/algorithms/particle/ParticleConverter.cc
```diff
@@ -0,0 +1,282 @@
...
+      prelimPID     = track.getPdg();
+      trackMass     = m_particleSvc.particle(prelimPID).mass;
+
+      trackMomentumMag = std::sqrt(std::pow(trackMomentum.x, 2) + std::pow(trackMomentum.y, 2) +
+                                   std::pow(trackMomentum.z, 2));
+
+      trackEnergy = std::sqrt(std::pow(trackMomentumMag, 2) + std::pow(trackMass, 2));
+
+      reconstructedMomentum = trackMomentum;
+    }
```
ruse-traveler: I'd suggest including this in the `if (ntracks > 0)` block to make sure the track exists!
```suggestion

      if (track.getChi2() > 0) {
        hasTrack = true;

        trackMomentum = track.getMomentum();
        prelimPID     = track.getPdg();
        trackMass     = m_particleSvc.particle(prelimPID).mass;

        trackMomentumMag = std::sqrt(std::pow(trackMomentum.x, 2) + std::pow(trackMomentum.y, 2) +
                                     std::pow(trackMomentum.z, 2));

        trackEnergy = std::sqrt(std::pow(trackMomentumMag, 2) + std::pow(trackMass, 2));

        reconstructedMomentum = trackMomentum;
      }
    }
```

ruse-traveler (commented):
Nice! Thanks for integrating my feedback! I think this ready to open up for public review!

Issue #2324, opened 2026-01-12
Tracker hits associations without Tracker hits


DraTeots:
Bump. So far I see problems in calorimeters: 

### Environment: (where does this bug occur, have you tried other environments)
- the latest simulation campaign background data

e.g. 

```
root://dtn-eic.jlab.org//volatile/eic/EPIC/RECO/25.10.4/epic_craterlake/Bkg_Exactly1SignalPer2usFrame/SIDIS/D0_ABCONV/pythia8.306-1.1/10x100/q2_1/hiDiv/pythia8.306-1.1_D0_10x100_hiDiv_q2_1_run10.2110.eicrecon.edm4eic.root ./
```

### Expected results

TrackerHits -> RawHits are one side relationship. So in order to connect "particle" to Tracker hits, one of the strategy is to pull all associations and then find a trackerHit that corresponds to RawHits from the association. It looks like there is no TrackerHits -> RawHits connection set for subsystems: 

- TOFEndcapRawHitAssociations
- SiEndcapTrackerRawHitAssociations

DraTeots:
Actually TrackerHits -> RawHit is working, I checked it again. What tricked me in the first place, is that I didn't know, that if one puts a wrong collection name into `frame.get` (not corresponding to the type) it silently returns an empty collection. 

So only DRICHRawHitsAssociations is left if it is considered a bug

PR #2410, opened 2026-02-04
fix: link to objects in collections, not temporaries

wdconinc:
This PR fixes a bug where we set the raw hit in associations to the object in a temporary map that is not part of a collection. This does not result in a valid association since podio cannot determine index and collectionID.

In this PR we push_back (copy) the object into the collection and get the rbegin iterator (since podio collections don't have a back method to get a reference to the last entry).

- [x] Bug fix (issue: https://github.com/eic/EICrecon/pull/2372#discussion_r2756906284)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/digi/SiliconTrackerDigi.cc
+++ b/src/algorithms/digi/SiliconTrackerDigi.cc
@@ -7,7 +7,6 @@
 #include <edm4hep/MCParticleCollection.h>
 #include <edm4hep/Vector3d.h>
 #include <edm4hep/Vector3f.h>
-#include <fmt/core.h>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


veprbl:
Did we not find that it was all okay in #2324? The temporaries in question have been converted to MaybeSharedPtr https://github.com/AIDASoft/podio/blob/8bb4eba33872186724c2c4f06fe53119d087fe20/include/podio/utilities/MaybeSharedPtr.h#L29 and insertions are supposed to update ObjectID for each instance. That's my recollection, at least.

Issue #2000, opened 2025-08-06
Segfault running eic_xl:nightly


faustus123:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (often `main` for latest released):

This was the /cvmfs/singularity.opensciencegrid.org/eic/eic_xl:nightly on 2025-08-06

- Which version (or `HEAD` for the most recent on git):

I assume it was HEAD

- Any specific OS or system where the issue occurs?

The host system was almalinux 9.6 and the container was being run using apptainer.

- Any special versions of ROOT or Geant4?

No

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Grab an input file of simulated DVCS events:

xrdcp root://dtn-eic.jlab.org//work/eic3/users/davidl/2025.08.05.S1_demo/simulation/DVCS_5x41_02/all/output4998.edm4hep.root .

2.  apptainer exec /cvmfs/singularity.opensciencegrid.org/eic/eic_xl:nightly eic-shell

3. source /opt/detector/epic-main/bin/thisepic.sh

4. cores=$(nproc)

5. eicrecon -Pnthreads=$cores output4998.edm4hep.root 

### Expected Result: (what do you expect when you execute the steps above)

After startup, it should process events at a few 10s of Hz (depending on cores available) then exit leaving a podio_output.root file.

### Actual Result: (what do you get when you execute the steps above)

If goes through what appears to be all of the initial startup but right when it is ready to process events, segfaults. Here is a log file.

[log.txt](https://github.com/user-attachments/files/21634218/log.txt)

wdconinc:
There's a number of things here.
- B0 geometry is currently not being translated well into Acts geometry (since upgrade to acts-39),
- regression on JOmniFactory when using malformed input that causes exceptions.

PR #2458, opened 2026-02-11
fix: try-catch on cellID position lookup in TrackerHitReco/LGAD

wdconinc:
This PR (cherry-picked from #2450) adds a try-catch around cellID position lookup in the TrackerHitReconstruction.cc and LGADHitCalibration.cc. This helps build resiliency when running EICrecon against a simulation file created with a different geometry than that used in reconstruction (see e.g. #2000 which was ultimately caused by this, along with stricter treatment of exceptions in JANA2 factories used there). This shifts geometry mismatch from exceptional error to logic error (but likely a noisy one).

One could argue that geometry mismatches should be treated as exceptional, but that is more true if we can do this on `init` instead of during `process`.

- [x] Bug fix (issue: don't treat cellID lookup failures as exceptional)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/reco/LGADHitCalibration.cc
+++ b/src/algorithms/reco/LGADHitCalibration.cc
@@ -34,7 +34,13 @@ void LGADHitCalibration::process(const LGADHitCalibration::Input& input,
     auto id = TDCADC_hit.getCellID();
 
     // Get position and dimension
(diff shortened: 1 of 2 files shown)


veprbl (approved):
LGTM

PR #2460, opened 2026-02-12
chore!: require podio-1.3; test as of 25.09.0

wdconinc:
This PR bumps the podio requirement to 1.3, available as of 25.09.0 which is now our 6-month compatibility window.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: upgrade podio-1.3)
- [ ] Documentation update
- [ ] Other: __

Yes, requires podio 1.3 now.

No.

### DIFF
--- a/src/services/io/podio/datamodel_select_compat.h
+++ b/src/services/io/podio/datamodel_select_compat.h
@@ -3,16 +3,12 @@
 
 #pragma once
 
-#include <podio/podioVersion.h>
-
 // Note that the JANA version components are defined by the build system,
 // since they are not available for use in preprocessor directives when included
 // from JANA/JVersion.h.
 
-// Use modern implementation for podio >= 1.3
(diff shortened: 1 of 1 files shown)


veprbl (dismissed):

wdconinc:
Podio 1.3 supports the type enumeration used in the code that avoids the data model glue, but it also requires JANA 2.4.3 or later so we can't quite remove the data model glue yet :-(

src/services/io/podio/datamodel_select_compat.h
```diff
@@ -11,10 +11,8 @@
 
 // Use modern implementation for podio >= 1.3
 #if defined(podio_VERSION_MAJOR) && defined(podio_VERSION_MINOR)
```
wdconinc:
```suggestion
```

PR #2464, opened 2026-02-13
fix: centralize Acts contexts for consistency

wdconinc:
This PR centralizes all default-constructed Acts contexts to be consistent with each other and handed out (by const reference) by the ActsGeometryProvider. This ensures that any future run-dependent contexts can be provided without major changes. It also avoids the many deprecation warnings that are associated with default constructed contexts (when not gotten from a new static factory) in Acts v45 (https://github.com/acts-project/acts/pull/4957).

- [x] Bug fix (issue: too many contexts)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
LGTM

PR #2496, opened 2026-02-18
add justfile

veprbl:
An example of a justfile (just is added to container in https://github.com/eic/containers/pull/171), which is my proposal for addressing request for CI workflows runnable locally and another request for having a convenience wrappers around common build workflows.


wdconinc:
> I think users want to do some easy to type commands.

Until, inevitably, they want to do something beyond what's provided in this intermediate layer. Then we end up duplicating more and more functionality in that intermediate layer because the users have not learned the basics of CMake, for example... That's sort of what I think we were trying to avoid by using industry-standard tools that are well-documented online (not saying `just` isn't that; at least we're not writing something from scratch). So, in some sense, I'd rather the intermediate layer is building capacity (by being clear about what's actually being done) in the underlying technology instead of shielding people from it.

wdconinc (approved):

PR #2502, opened 2026-02-19
feat: support free surfaces in Acts v45+

wdconinc:
In Acts v45 surfaces can exist without connection to a detector element (e.g. effective surfaces inside calorimeters). This PR aims to support these surfaces. This functionality is currently unused.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: support free surfaces)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/algorithms/tracking/ActsGeometryProvider.cc
+++ b/src/algorithms/tracking/ActsGeometryProvider.cc
@@ -178,7 +178,13 @@ void ActsGeometryProvider::initialize(const dd4hep::Detector* dd4hep_geo, std::s
 #endif
 
       if (det_element == nullptr) {
-        m_init_log->error("invalid det_element!!! det_element == nullptr ");
+        if (surface->geometryId().value() != 0) {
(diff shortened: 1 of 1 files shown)


wdconinc:
> Can we use this to provide surfaces for MPGD?

I haven't looked into it in too much detail. I was thinking this may indeed be useful for MPGD, for projection to DIRC center plane, and for definition of effective calorimeter surfaces. I just wanted to make sure it's supported here in case someone wants to experiment with it. I don't know if ODD has a demonstrator for this, for example.

veprbl:
I'd love to see a demonstrator!

wdconinc:
> I'd love to see a demonstrator!

None is referenced in the PRs I have looked through. I do think maybe my expectations should be calibrated a bit though: this will now allow free surfaces to be sensitive surface with measurements, so yeah it may work for MPGD, but the other stuff (DRICH, DIRC, calorimeters) is already possible since those are 'regular' surfaces without measurements on them. At least, that's what I understand...

PR #2506, opened 2026-02-20
Expose clang-tidy and clang-format as CMake targets

nathanwbrei:
This PR introduces some new CMake targets for running clang-tidy and clang-format locally. This allows users to validate that their PR passes the style checks proactively and interactively, without having to wait for CI.
 
```bash

# Run clang-tidy over the entire codebase and apply, but don't commit, fixes
cmake --build build --target tidy

# Run clang-tidy over the entire codebase and return an error if changes are still needed
cmake --build build --target check-tidy     

# Run clang-format over the entire codebase and apply, but don't commit, fixes
cmake --build build --target format

# Run clang-format over the entire codebase and return an error if changes are still needed
cmake --build build --target check-format   

# "One-shot" style fixes: Runs `format` followed by `tidy` followed by `check-tidy`
cmake --build build --target style
```

Introduces no breaking changes.

Changes no default behavior.

### DIFF
--- a/cmake/CodeStyleTargets.cmake
+++ b/cmake/CodeStyleTargets.cmake
@@ -1,4 +1,3 @@
-
 find_program(RUN_CLANG_TIDY run-clang-tidy-19)
 find_program(CLANG_APPLY_REPLACEMENTS clang-apply-replacements-19)
 find_program(CLANG_FORMAT clang-format-19)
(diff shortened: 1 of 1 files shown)


nathanwbrei:
Curiously, the CMake-ized clang-tidy emits a bunch of warnings on the unmodified codebase, even though it is using the provided config file. A lot of the warnings are from ACTS headers, but there's 22 files in the EICrecon source tree that got reasonable-looking fixes. So I'm curious what CI is doing differently.

wdconinc:
> So I'm curious what CI is doing differently.

Maybe you already figured this out, but CI only runs on changed files. People don't want to get issues flagged in files they didn't change. Over time, since we introduce this with clang-16 or so, this means that some newer clang-20 checks are not satisfied by code that hasn't been touched recently.

> A lot of the warnings are from ACTS headers

In CI we suppress issues in headers that are not under our version control.

PR #2229, opened 2025-12-12
Integrating IRT2 into EICRecon

chchatte92:
### Briefly, what does this PR introduce?
This PR introduces the changes related to the IRT2 based PID of pfRICH and dRICH. 

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [X ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ X] Tests for the changes have been added
      - The preliminary tests  has been made locally, multi-threading is not ensured.    
- [ ] Documentation has been added / updated
- [X] Changes have been communicated to collaborators
      This is a draft PR at this moment, @veprbl @alexander-kiselev @kumardeepraman are aware of the changes.
      The suggested changes of this PR will require some discussions.

PR #2515, opened 2026-02-25
Initial IRT2 integration

alexander-kiselev:
Makes IRT2 codes consistent with irt-2.1e branches in EDM4eic/irt/epic repositories and 26.02.0-stable image

TODO:
 - [ ] Address remaining comments from https://github.com/eic/EICrecon/pull/2229
   - [x] RNG seeds (example: https://github.com/eic/EICrecon/blob/af0e8fbbbe5c0af9d3ac0fd4ffe8e8c5ec5a659c/src/algorithms/calorimetry/CalorimeterHitDigi.h#L81 https://github.com/eic/EICrecon/blob/af0e8fbbbe5c0af9d3ac0fd4ffe8e8c5ec5a659c/src/algorithms/calorimetry/CalorimeterHitDigi.cc#L147-L150)
 - [x] Pick up config files from geometry
 - [x] Destructor of m_ReconstructionFactory crashes the reconstruction
 - [ ] Move logic from RICH IRT into algorithms init as much as possible
 - [x] Get rid of G4DataInterpolation - the whole implementation is copied over and (incorrectly) used to intialize m_RefractiveIndex in IRT. Potential Geant4 licensing issues.
 - [x] IrtConfig must be a POD type, it's value must not be mutated by the algorithm. All pointers need to move to the algorithm.
 - [x] Adhoc tree/histogram outputs - this is generally not allowed. The functionality must be translated to PODIO macros. Alternatively we can setup a JEventProcessor that would consume m_EventBranch values, custom file output can be safely done there. TApplication/TCanvas must never be instantiated from an algorithm (this produces multiple instances).
   - [x] validation benchmarks for IRT2 using PODIO outputs
   - [ ] (optional) calibrations scripts using PODIO outputs
 - [x] License headers to reflect LGPL in spdx format


src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2024, 2025, 2026, Alexander Kiselev
+
+#ifdef WITH_IRT2_SUPPORT
```
wdconinc: Do we need this preprocessor flag? IMHO we should require IRT2 support.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+#include <random>
+#include <set>
+#include <tuple>
+#include <utility>
+#include <vector>
+
+#include "algorithms/pid/IrtInterfaceConfig.h"
+#include "algorithms/pid/Tools.h"
+
+using namespace IRT2;
```
wdconinc, line 46: Can we narrow the using namespace to individual using aliases where it really reduces the code duplication? It can be annoying for external projects that may define common names for functions without thinking about how it may conflict with other included header or local symbols.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+
+#include "IrtInterface.h"
```
wdconinc, line 48:
```suggestion
#include "algorithms/pid/IrtInterface.h"
```

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+
+// -------------------------------------------------------------------------------------
+
+namespace eicrecon {
+IrtInterface::~IrtInterface() {
+  //printf("@Q@ IrtInterface::~IrtInterface() ... %s\n", m_cfg.m_irt_detector->GetName());
```
wdconinc: Remove all commented code throughout.

```suggestion
```

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+  } //if
+
+  if (m_irt_detector) {
+    for (auto [name, rad] : m_irt_detector->Radiators()) {
+      if (rad && rad->m_RefractiveIndex) {
+        delete rad->m_RefractiveIndex;
+        rad->m_RefractiveIndex = nullptr;
+      }
+    }
+  }
```
wdconinc: Can we use smart pointers here and rely on RAII for deletion? Likely makes destructor sufficient as default.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+void IrtInterface::init() {
+  //printf("@Q@ IrtInterface::init() ... %s\n", m_cfg.m_irt_detector->GetName());
+
+  // Cannot fail (see RICH-IRT.cc);
+  m_irt_geometry = IRT2::CherenkovDetectorCollection::Instance();
+  m_irt_detector = m_irt_geometry->GetDetector(m_cfg.m_detector_name.c_str());
+
+  m_Event = new IRT2::CherenkovEvent();
+
+  m_ReconstructionFactory = new IRT2::ReconstructionFactory(m_irt_geometry, m_irt_detector, m_Event,
```
wdconinc: Smart pointers would change this here too.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+                                                            m_cfg.m_json_config_file_name.c_str());
+  // JANA2 prints out event progress; the rest is kind of irrelevant;
+  m_ReconstructionFactory->SetQuietMode();
+
+  {
```
wdconinc: Unneeded scope block when unlocked.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+        std::unique_ptr<double[]> ri(new double[dim]);
+
+        for (unsigned row = 0; row < rindex_matrix->GetRows(); row++) {
+          e[row]  = rindex_matrix->Get(row, 0) / dd4hep::eV;
+          ri[row] = rindex_matrix->Get(row, 1);
+        } //for row
+
+        auto ptr = rad->m_RefractiveIndex = new DataInterpolation(e.get(), ri.get(), dim);
+        // FIXME: 100 hardcoded;
+        ptr->CreateLookupTable(100);
```
wdconinc: FIXME :-) Add to config as variable?

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+// -------------------------------------------------------------------------------------
+
+void IrtInterface::init() {
+  //printf("@Q@ IrtInterface::init() ... %s\n", m_cfg.m_irt_detector->GetName());
+
+  // Cannot fail (see RICH-IRT.cc);
+  m_irt_geometry = IRT2::CherenkovDetectorCollection::Instance();
+  m_irt_detector = m_irt_geometry->GetDetector(m_cfg.m_detector_name.c_str());
+
+  m_Event = new IRT2::CherenkovEvent();
```
wdconinc: What is the cost of `IRT2::CherenkovEvent`? Can it be an object local to process instead of a persisting pointer?

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+} // IrtInterface::init()
+
+// -------------------------------------------------------------------------------------
+
+void IrtInterface::process(const IrtInterface::Input& input,
+                           const IrtInterface::Output& output) const {
+  //printf("@Q@ IrtInterface::process() ...\n");
+
+  // Reset the event structure;
+  m_Event->Reset();
```
wdconinc, line 123: Can we create a new `IRT2::CherenkovEvent` here?

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+  // Intermediate variables, for less typing;
+  const auto [headers, in_mc_particles, in_tracks, in_track_associations, in_track_projections,
+              in_sim_hits]                        = input;
+  auto [out_irt_radiator_info, out_irt_particles] = output;
+
+  //Random number generator
+  auto seed = m_uid.getUniqueID(*headers, name());
+  // safe access to header
+  if (headers->empty()) {
+    error("No EventHeader available");
```
wdconinc: I'd make this a hard assumption to have a header. I think we assume this elsewhere too.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+  } else {
+    const auto& header = headers->at(0);
+    auto event_num     = header.getEventNumber();
+    auto run_num       = header.getRunNumber();
+    debug("algorithm: {}  run: {}  event: {}  seed: {}", name(), run_num, event_num, seed);
```
wdconinc: Not sure why we need `name()`, `run_num`, `event_num` as they are in the logging prefix. Remove and that avoids the need  for anything other than `seed`.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+  }
+  std::default_random_engine generator(seed);
+  std::uniform_real_distribution<double> uniform(0.0, 1.0);
+
+  // First build MC->reco lookup table;
+  std::map<unsigned, std::vector<unsigned>> MCParticle_to_Tracks_lut;
```
wdconinc: We should consider always using `podio::LinkNavigator` here instead of a homebrew LUT, and in any case put it in with preprocessor directives when it is available.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+    if (rctracks.size() > 1)
+      continue;
+    unsigned rctrack = rctracks[0];
+
+    // Do not want to deal with particles outside of the nominal acceptance; FIXME: do it better later;
+    {
+      double eta = Tools::PodioVector3_to_TVector3(mcparticle.getMomentum()).Eta();
+      if (eta < m_cfg.m_eta_min || eta > m_cfg.m_eta_max)
+        continue;
+    }
```
wdconinc: Unneeded scoping.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+    if (MCParticle_to_Tracks_lut.find(mcid) == MCParticle_to_Tracks_lut.end())
+      continue;
+    auto& rctracks = MCParticle_to_Tracks_lut[mcid];
+    if (rctracks.size() > 1)
+      continue;
+    unsigned rctrack = rctracks[0];
+
+    // Do not want to deal with particles outside of the nominal acceptance; FIXME: do it better later;
+    {
+      double eta = Tools::PodioVector3_to_TVector3(mcparticle.getMomentum()).Eta();
```
wdconinc: Can we hoist `Tools::PodioVector3_to_TVector3` to edm4hep's vector_utils treewide if there is nothing templated there already yet that we can reuse?

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+    MCParticle_to_ChargedParticle[mcid] = particle;
+
+    // Create history records for all known radiators; FIXME: may want to optimize a bit;
+    for (auto [name, rad] : m_irt_detector->Radiators()) {
+      auto history = new RadiatorHistory();
+      particle->StartRadiatorHistory(std::make_pair(rad, history));
+    } //for radiator
+
+    // Record track projections; FIXME: do it only for radiators used for imaging?;
+    {
```
wdconinc: Unneeded scoping.

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,377 @@
...
+    // Create an optical photon class instance and populate it; units: [mm], [ns], [eV];
+    auto photon = new IRT2::OpticalPhoton();
+
+    // Information provided by the hit itself: detection position and time;
+    photon->SetDetectionPosition(Tools::PodioVector3_to_TVector3(mchit.getPosition()));
+    photon->SetDetectionTime(mchit.getTime());
+
+    // Information inherited from photon MCParticle; FIXME: units?;
+    photon->SetVertexPosition(Tools::PodioVector3_to_TVector3(mcparticle.getVertex()));
+    photon->SetVertexMomentum(1E9 * Tools::PodioVector3_to_TVector3(mcparticle.getMomentum()));
```
wdconinc: Use proper units conversion over magic number.

src/algorithms/pid/IrtInterface.h
```diff
@@ -0,0 +1,72 @@
...
+
+public:
+  IrtInterface(std::string_view name)
+      : IrtInterfaceAlgorithm{name,
+                              {"eventHeaderCollection", "inputMCParticles", "inputTracks",
+                               "inputTrackAssotiations", "inputTrackSegments", "inputSimHits"},
+                              {"outputIrtRadiatorInfo", "outputIrtParticles"},
+                              "Performs PID evaluation based on IRT2 algorithm"}
+      , m_Event(0)
+      , m_ReconstructionFactory(0) {};
```
wdconinc: Smart pointers preferred here too.

src/algorithms/pid/IrtInterface.h
```diff
@@ -0,0 +1,72 @@
...
+  IRT2::CherenkovDetectorCollection* m_irt_geometry;
+  IRT2::CherenkovDetector* m_irt_detector;
+
+  IRT2::CherenkovEvent* m_Event;
+
+  const algorithms::UniqueIDGenSvc& m_uid = algorithms::UniqueIDGenSvc::instance();
+
+  IRT2::ReconstructionFactory* m_ReconstructionFactory;
+
+  const algorithms::GeoSvc& m_geo = algorithms::GeoSvc::instance();
```
wdconinc, line 67: It might be useful to create an IRT2GeoSvc and get that on construction, or open an issue if not in scope here.

Also, `m_detector = m_geo.detector()` can already be filled on initialization here too.

src/algorithms/pid/IrtInterface.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "IrtInterfaceConfig.h"
+#include "algorithms/interfaces/UniqueIDGenSvc.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+using IrtInterfaceAlgorithm = algorithms::Algorithm<
+    algorithms::Input<edm4hep::EventHeaderCollection, edm4hep::MCParticleCollection,
+                      edm4eic::TrackCollection, edm4eic::MCRecoTrackParticleAssociationCollection,
+                      edm4eic::TrackSegmentCollection, edm4hep::SimTrackerHitCollection>,
```
wdconinc, line 36: Can we already forward compatible declare MC collections are `std::optional` and put handling in there (i.e. warning when MC based reconstruction is used)?

src/algorithms/pid/IrtInterfaceConfig.h
```diff
@@ -0,0 +1,26 @@
...
+// of either the optics file or a JSON configuration file twice;
+struct IrtConfig {
+  IrtConfig() : m_eta_min(0.0), m_eta_max(0.0) {};
+
+  std::string m_detector_name;
+  std::string m_json_config_file_name;
+
+  // FIXME: perhaps do it better later; but in general see no reason in parsing
+  // the same fields in a JSON file twice;
+  double m_eta_min, m_eta_max;
```
wdconinc:
```suggestion
  double m_eta_min{0}, m_eta_max{0};
```
and avoid explicit constructor definition.

src/algorithms/CMakeLists.txt
```diff
@@ -8,3 +8,4 @@ add_subdirectory(digi)
 add_subdirectory(reco)
 add_subdirectory(fardetectors)
 add_subdirectory(onnx)
+add_subdirectory(meta)
```
wdconinc: Unrelated change?

src/global/pid/IrtInterface_factory.h
```diff
@@ -0,0 +1,55 @@
...
+#include <edm4eic/RawTrackerHitCollection.h>
+
+namespace eicrecon {
+
+class IrtInterface_factory : public JOmniFactory<IrtInterface_factory, IrtConfig> {
+
+private:
+  using AlgoT = eicrecon::IrtInterface;
+  std::unique_ptr<AlgoT> m_algo;
+  PodioInput<edm4hep::EventHeader> m_event_header_input{this}; // Added
```
wdconinc:
```suggestion
  PodioInput<edm4hep::EventHeader> m_event_header_input{this};
```

src/utilities/eicrecon/eicrecon.cc
```diff
@@ -38,9 +38,11 @@ std::vector<std::string> EICRECON_DEFAULT_PLUGINS = {
     "ZDC",
     "BTRK",
     "BVTX",
-    "PFRICH",
     "DIRC",
     "DRICH",
+#ifdef WITH_IRT2_SUPPORT
+    "RICH-IRT",
+#endif
```
wdconinc:
```suggestion
    "RICH-IRT",
```

src/utilities/eicrecon/CMakeLists.txt
```diff
@@ -20,5 +20,10 @@ target_compile_definitions(
           JANA_VERSION_MINOR=${JANA_VERSION_MINOR}
           JANA_VERSION_PATCH=${JANA_VERSION_PATCH})
 
+find_package(IRT2 ${IRT2_VERSION_MIN})
+if(IRT2_FOUND)
+  target_compile_definitions(eicrecon PRIVATE WITH_IRT2_SUPPORT)
+endif()
```
wdconinc, line 30: Unneeded when not using `WITH_IRT2_SUPPORT`.
```suggestion
```

src/detectors/RICH-IRT/RICH-IRT.cc
```diff
@@ -0,0 +1,272 @@
...
+static const char* RICHes[] = {"DRICH", "PFRICH"};
+
+using namespace eicrecon;
+
+extern "C" {
+void InitPlugin(JApplication* app) {
+  InitJANAPlugin(app);
+
+  auto dd4hep_service = app->GetService<DD4hep_service>();
+  auto det            = dd4hep_service->detector();
```
wdconinc, line 68: Anti-pattern to parse geometry in plugin. In this case we don't use the geometry beyond checking that the specific RICH is there, so we can skip this here and rely on the algorithm not doing anything if the RICH is not in the geometry.

src/detectors/RICH-IRT/RICH-IRT.cc
```diff
@@ -0,0 +1,272 @@
...
+
+using json = nlohmann::json;
+
+// Have to start hardcoding known RICH detectors in some way, before trying
+// to pull their configuration data out; for the time being the list is:
+//
+// PFRICH: ePIC backward proximity focusing RICH
+//  DRICH: ePIC forward dual radiator RICH
+//
+static const char* RICHes[] = {"DRICH", "PFRICH"};
```
wdconinc:
```suggestion
static std::vector<std::string> RICHes = {"DRICH", "PFRICH"};
```

src/algorithms/pid/IrtInterface.cc
veprbl: This is the way to comply with requirement for the code to be compilable/runnable with older containers.

src/algorithms/pid/IrtInterface.cc
veprbl: It's normal to import whole namespace inside one translation unit.

src/algorithms/pid/IrtInterface.cc
veprbl: Not sure what does this do for us other than making files harder to relocate. If this is the standard elsewhere, we can apply the change, of course.

src/utilities/eicrecon/eicrecon.cc
veprbl: We need WITH_IRT2_SUPPORT

src/utilities/eicrecon/CMakeLists.txt
veprbl: We need WITH_IRT2_SUPPORT

src/algorithms/pid/IrtInterface.cc
wdconinc: It's mosty relevant for headers (which are installed and don't resolve when using a local directory), not as much in cc. Having it consistent in cc and h files is helpful. A bit more common to use full path to algorithms.
```
11:21:28 wdconinc@menelaos ~/git/EICrecon (avoid-get_evt_srces $%=) $ grep -r include src/algorithms/ | grep '"[^/]*\.h"' | wc -l
127
11:21:38 wdconinc@menelaos ~/git/EICrecon (avoid-get_evt_srces $%=) $ grep -r include src/algorithms/ | grep '"algorithms/.*\.h"' | wc -l
197
```

veprbl: I'm not sure about these stats. Cross-algorithm includes can be global. This is a cc file including its own header.
This is the norm:
https://github.com/eic/EICrecon/blob/3b4eb635fb9e5ab4090f48af1aa1bb794582283b/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.cc#L34

src/algorithms/pid/IrtInterface.cc
veprbl: My thought too. I suspect IRT2 won't use smart pointers inside due to ROOT streamer issues. I will treat `m_Event`/`m_ReconstructionFactory` that we have full control over.

src/algorithms/pid/IrtInterface.cc
veprbl: I think it should be cheap, but it should be cheaper to just let it clear its vectors/sets.
https://github.com/eic/irt/blob/main/include/CherenkovEvent.h

src/algorithms/pid/IrtInterface.cc
veprbl: Removed

src/algorithms/pid/IrtInterface.cc
veprbl: Removed

src/algorithms/pid/IrtInterface.cc
veprbl: Removed

src/algorithms/pid/IrtInterface.cc
veprbl: Removed

src/algorithms/pid/IrtInterface.cc
veprbl: Can be done, but we need https://github.com/eic/EICrecon/pull/2760 first

src/algorithms/pid/IrtInterface.cc
veprbl: Treewide probably not, but in this case to calculate eta there is a function.

src/algorithms/pid/IrtInterface.cc
veprbl: I'm not aware of what is proper unit conversion. I can only guess we are doing GeV to eV conversion.

src/algorithms/CMakeLists.txt
veprbl: https://github.com/eic/EICrecon/pull/2763

src/algorithms/pid/IrtInterfaceConfig.h
veprbl: Done

src/detectors/RICH-IRT/RICH-IRT.cc
veprbl: Done

src/algorithms/pid/IrtInterface.cc
veprbl: Removed

src/algorithms/pid/IrtInterface.cc
veprbl: Answered above

src/algorithms/pid/IrtInterface.cc
veprbl: Not clear if this even needed.

src/algorithms/pid/IrtInterface.h
veprbl: This is currently not optional. We are discussing moving digitization out of IRT2 or making it more modular.

src/detectors/RICH-IRT/RICH-IRT.cc
veprbl: I'm not sure. I think this is a valid question and the FIXME states as much. I suspect, there might be a crash whenever IRT2 tries to access a detector that is not in the configuration, and this is a guard against that. I don't think it's in scope to fix that now - we'll need to revisit this.

src/algorithms/pid/IrtInterface.h
veprbl: Probably doesn't make sense to make this into a service. We only plan to use it from this algorithm and `IRT2::CherenkovDetectorCollection` is basically a service.

src/algorithms/pid/IrtInterface.cc
veprbl: They did

src/algorithms/pid/IrtInterface.cc
veprbl: Now done.

src/algorithms/pid/IrtInterface.cc
wdconinc: If `SetVertexMomentum` expects in eV, then we could do:
```
    photon->SetVertexMomentum(Tools::PodioVector3_to_TVector3(mcparticle.getMomentum()) / edm4hep::eV);
```

veprbl: I was already doing edm4eic::unit, but this is more correct, I think. Now updated.

veprbl:
@wdconinc I'm happy with the current state of this PR. I think I've addressed your comments to the extent that we can do now.

wdconinc (dismissed):

veprbl:
I've tested that I didn't break/change anything in PODIO outputs since 5753debe3803ab68a6235030a28ee8d7ae7bc6cc
Also run with explicit TSAN, didn't see any new issue.

veprbl (dismissed):

src/algorithms/pid/IrtInterface.cc
```diff
@@ -0,0 +1,340 @@
...
+      const unsigned dim = rindex_matrix->GetRows();
+      std::unique_ptr<double[]> e(new double[dim]);
+      std::unique_ptr<double[]> ri(new double[dim]);
+
+      for (unsigned row = 0; row < rindex_matrix->GetRows(); row++) {
+        e[row]  = rindex_matrix->Get(row, 0) / dd4hep::eV;
+        ri[row] = rindex_matrix->Get(row, 1);
+      } //for row
+
+      auto ptr = rad->m_RefractiveIndex = new DataInterpolation(e.get(), ri.get(), dim);
```
wdconinc, line 111: This races, https://github.com/eic/EICrecon/actions/runs/30104707397/job/89520309493#step:7:629.

alexander-kiselev:
Sure, it does. I recall we discussed this with Dmitry just recently, and the conclusion was that these init() calls would never be run concurrently, then why bother. Adding a mutex lock which was there in one of the branches, though for a different reason, should fix this problem.

veprbl:
> Sure, it does. I recall we discussed this with Dmitry just recently, and the conclusion was that these init() calls would never be run concurrently, then why bother.

No, that would be pleasant to hear, but unfortunately was not my message. The TSAN failure reported is not hypothetical, it's a real race that was detected between concurrent inits. We discussed how in JANA2 each concurrent thread will use its own copy of `IrtInterface`.

> Adding a mutex lock which was there in one of the branches, though for a different reason, should fix this problem.

Serializing init/deinit is fine, but doing the same for processing would be bad. I think IRT2 is not that far from being proper thread safe. In order to achieve that we need to ensure that things are either not shared (e.g. m_Event and m_ReconstructionFactory are not shared) or immutable. Here is an example of how ReconstructionFactory mutates radiator object:
https://github.com/eic/irt/blob/9dbd5c5d0885bbd1108cbfcf2e53f9f6dbc49878/source/ReconstructionFactory.cc#L487
Can we move this refractive index to e.g. RadiatorHistory that is local to each process call? Ideally ReconstructionFactory does not have access to radiator without the C++ const qualifier.

src/algorithms/pid/IrtInterface.cc
veprbl: @wdconinc I've pushed a fix for the race that we see. Let me know how much you are against it.

alexander-kiselev:
Right, we discussed the init() serialization only. What you are pointing out now is indeed a race condition which cannot be protected by a mutex lock since these variable appear to be shared. Unfortunately, making them thread_local (would be an easy fix) won't work either as far as I can tell.

src/algorithms/pid/IrtInterface.cc
wdconinc: Ok with me. I'd have put a regular (single) mutex lock in place, but call_once works too of course (and is probably better, semantically and performance-wise). I'd probably have insisted on making the IRT2 side of things a more thread safe by design (as you're already discussing), but that would just postpone this PR by another few weeks.

veprbl: I think call_once is semantically more correct than mutex, we are truly not looking to multiply initialization or destruction of those tables. It also made me think that, if we to keep dealing with global state, this justifies having an IRT service, like you suggested earlier.

veprbl:
> Right, we discussed the init() serialization only. What you are pointing out now is indeed a race condition which cannot be protected by a mutex lock since these variable appear to be shared. Unfortunately, making them thread_local (would be an easy fix) won't work either as far as I can tell.

You can make a whole `std::map<const CherenkovRadiator*, double>`, which could be scoped in ProcessHits (might need to pass it by reference as parameter to `Solve()`).

wdconinc (dismissed):

wdconinc:
The good:
image

The absent:
- no nhits, no npe, no links to radiators
  https://eicrecon.epic-eic.org/pr/2515/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#DRICHIrtParticles
- no track length (or error on it)
  https://eicrecon.epic-eic.org/pr/2515/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#DRICHTracks
- no radiator info (with angle, nhits or npe)
  https://eicrecon.epic-eic.org/pr/2515/capybara/rec_dis_18x275_minQ2=1000_craterlake_18x275/index.html#DRICHIrtRadiatorInfo

Certainly still good for merging (and fixing up in follow-up), but I don't understand how we get PDG without the angle being stored.

wdconinc (dismissed):
From the compare output logs this now seems good.

veprbl:
This is not good. You are right, this does not work in normal epic jobs. I'm investigating.

veprbl:
There were several issues.
 - the `std::call_once` approach I introduced was incorrect - only one of the two RICHes was initialized
 - Two bugs in IRT are problematic for intialization and processing: https://github.com/eic/irt/pull/61
 - Hard dependency on optical photons present in `MCParticles`. The irt-sandbox relies on `--part.userParticleHandler=` to get those, but for production we need something like https://github.com/eic/npsim/pull/82 (costs disk space)

veprbl:
Forgot to mention another one:
- https://github.com/eic/epic-data/pull/37 https://github.com/eic/epic/pull/1149 The calibration file paths are not correct and TApplication stuff is enabled, so nothing works, still.

chchatte92:
Dear colleagues, I noticed that the truth PDG and the reco PDG had 100 percent matches in kinematic regions where it should not have. Upon investigation, I noticed that the reco PDG inside IrtInterface.cc was stored by calling GetPDG(). The  irtParticle inherits this method from Transientparticle of IRT2, which by defaults sets the PDG of the MC information. However, this GetRecoPdgCode stores the reconstructed PDG information as podio output. The reconstructed PDG information is *correctly* stored inside the LaunchRingFinder method of IRT2 ReconstructionFactory (https://github.com/eic/irt/blob/05826b2545312f63939c0ed9ee3f8ed182b024bc/source/ReconstructionFactory.cc#L830).
After this modification, locally everything compiles and  I get meaningful results. 
**As this commit does not change any core code of IRT2 or EICrecon. I am committing this change.**  
Alexander and Dmitry has been informed about this change. 
See the following image after the modification. 
image

Issue #2405, opened 2026-02-03
Primary Vertex Reconstruction Under Machine Background Conditions


Simple-Shyam:
**Is your feature request related to a problem? Please describe.**

This study concerns the simulation of machine-induced backgrounds. As is well known, a large number of ghost 3-hit tracks are produced due to synchrotron radiation. These tracks are clearly fake, but their multiplicity is very high (~1000) compared to tracks originating from real DIS events. Therefore, for reliable primary vertex reconstruction, it is necessary to reject these three-hit ghost tracks. Otherwise, the large background contribution biases the determination of the primary vertex position.
Accurate primary vertex reconstruction is particularly important for heavy-flavor (HF) analyses, which are highly sensitive to both primary and secondary vertex resolutions.
**We will also gain in time to vertex finding, in this case you will reject ~1000 fake tracks used for primary vertex finding.**   

The secondary vertex reconstruction is not affected at this stage, since it is performed at the analysis level rather than within the EIC reconstruction framework.

**Describe the solution you'd like**
It would be good to introduce a flag requiring nHits > 3, which must be activated only for background simulations. For general physics simulations, this cut is not required.

**Describe alternatives you've considered**
There is no alternative except to perform the primary vertex reconstruction again with nHits > 3.
**Additional context**
If this can be fixed in the upcoming campaign, it will help us better understand the impact of background on heavy-flavor (HF) measurements.

veprbl:
Please don't assign people unless they are known to be actively working on the issue.

wdconinc:
https://github.com/eic/EICrecon/blob/main/src/algorithms/tracking/CKFTrackingConfig.h#L15

That'd be where we can make a change, for all running conditions, by requiring 4 measurements instead of 3.

But the impact of that needs to be studied. In particular, it didn't seem like the ghost track rate after ambiguity resolution was that high that it would dominate the vertex finder. Can you bring some numbers and plots for your claims?

veprbl:
I would not redo the whole tracking twice, the requirement would need to go to IVF.

PR #2527, opened 2026-03-01
feat: add minTrackHits option to IterativeVertexFinder

veprbl:
This addresses #2405 by adding an option to set the hit requirement and set of vertex collections that require a higher hit requirement.
I don't think there are any final confirmations on whether people are still looking to study vertices that use tracks with 3 hits.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, only additional collections are being added.

No


veprbl:
I've checked, this doesn't seem to increase file size of rec_dis_18x275_minQ2=0_craterlake_18x275.edm4eic.root by much.
image
image
^ this is from CI with only 100 events.

PR #2579, opened 2026-03-19
Conversion from energy deposit to the number of photoelectrons

### LINKED ISSUE #2467: Conversion from energy deposit sum to number of photoelectrons
This issue is to convert the energy deposit sum to the number of photoelectrons (Npe) before the pulse generation.

mhkim-anl:
This PR adds a conversion from energy deposit to the number of photoelectrons in the `PulseGeneration` algorithm. If the conversion factor is a constant and doesn't depend on any field, `edep_to_npe` is used for conversion. It the conversion factor depends on certain fields, the algorithm gets the field-dependent conversion factors from a lookup table using `m_edep_to_npe_fields` and `m_edep_to_npe_filename`. Poisson smearing is also applied after the conversion.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #2467)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/src/algorithms/digi/PulseGeneration.cc
+++ b/src/algorithms/digi/PulseGeneration.cc
@@ -5,6 +5,7 @@
 // Convert energy deposition into ADC pulses
 // ADC pulses are assumed to follow the shape of landau function
 
+#include <DDSegmentation/BitFieldCoder.h>
 #include <RtypesCore.h>
 #include <TMath.h>
 #include <algorithms/service.h>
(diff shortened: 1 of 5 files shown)


mhkim-anl:
I think we could first discuss whether the Edep-to-Npe conversion should remain in the `PulseGeneration` algorithm or be separated into a new algorithm.

If we keep it in `PulseGeneration`, this algorithm may end up doing too many things. If I understand correctly, `SimTackerHit` doesn't need this conversion as well. If we instead implement a separate algorithm, the output would be `SimCalorimeterHit`, but the `energy` member would store the number of photoelectrons rather than energy. If that is acceptable, I'll implement a new algorithm for it.

I would appreciate hearing your thoughts. Thanks!

ruse-traveler:
> I think we could first discuss whether the Edep-to-Npe conversion should remain in the `PulseGeneration` algorithm or be separated into a new algorithm.
> 
> If we keep it in `PulseGeneration`, this algorithm may end up doing too many things. If I understand correctly, `SimTackerHit` doesn't need this conversion as well. If we instead implement a separate algorithm, the output would be `SimCalorimeterHit`, but the `energy` member would store the number of photoelectrons rather than energy. If that is acceptable, I'll implement a new algorithm for it.
> 
> I would appreciate hearing your thoughts. Thanks!

I think we discussed a little offline, but I wanted to write my thoughts down here as well! After mulling it over, I think it makes sense to place the Edep-to-Npe conversion in a separate algorithm; mainly since the pulse generation algorithm operates on both calorimeter & tracker hits, but also since I feel like having the conversion functionality more readily available could be useful.

mhkim-anl:
@ruse-traveler 
I'm sorry for the very delayed response. Thank you sharing your thoughts. I agree with your comments, and other people also seem to prefer having a separate algorithm rather than implementing the Edep-to-Npe conversion into the `PulseGeneration` algorithm. I'll close this PR and submit a new one for the separate algorithm. Thanks!

PR #2589, opened 2026-04-01
Implement H4, H3 and S2 staggering in HEXPLIT

sebouh137:
Update the HEXPLIT algorithm to be able to handle square cells, which are used in the latest version of the ZDC geometry.  Also update the parameters in the ZDC reconstruction to match the geometry (which also has thicker cells, so larger sampling fraction and larger MIP value.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Users should use the square-cell version of the ZDC geometry in the epic repository

Yes.  Presumes that the ZDC has square cells, and runs the square-cell version of HEXPLIT.


src/algorithms/calorimetry/ImagingTopoCluster.cc
```diff
@@ -58,10 +58,10 @@ void ImagingTopoCluster::init() {
   }
 
   // using juggler internal units (GeV, dd4hep::mm, dd4hep::ns, dd4hep::rad)
-  sameLayerDistXY[0]     = std::visit(_toDouble, m_cfg.sameLayerDistXY[0]) / dd4hep::mm;
```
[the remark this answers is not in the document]
sebouh137: I presume the same can be said for the std::variant<std::string, double> thing in ImagingTopoClusterConfig.h?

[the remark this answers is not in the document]
wdconinc: Yes, the will be set but not be available as command line override options.

sebouh137:
@wdconinc It appears that the pull request in epic ( https://github.com/eic/epic/pull/1070 , which changes the ZDC geometry to match the pre-TDR) and this one (updating the reconstruction to match said geometry) would need to be approved simultaneously to allow the zdc_pi0 benchmark from detector_benchmarks to pass.  This benchmark works on my local machine when I run it the two repositories on these two branches.  What would the procedure be to get these two pull requests approved simultaneously?  
Thanks,
Sebouh

wdconinc:
> @wdconinc It appears that the pull request in epic ( [eic/epic#1070](https://github.com/eic/epic/pull/1070) , which changes the ZDC geometry to match the pre-TDR) and this one (updating the reconstruction to match said geometry) would need to be approved simultaneously to allow the zdc_pi0 benchmark from detector_benchmarks to pass. This benchmark works on my local machine when I run it the two repositories on these two branches. What would the procedure be to get these two pull requests approved simultaneously? Thanks, Sebouh

Generally the approach is to design this with backwards and forward compatibility. By saying this needs to be merged simultaneously, you are also saying that users will be forced to upgrade both epic and EICrecon at the same time. We can't reasonably enforce that.

sebouh137:
@wdconinc There are two benchmarks that currently fail.  The first is the bench:campaign, which appears to be unrelated to this change.  I keep gettting this error on there:

[e:97] [FOFFMTRK:ForwardOffMRecParticles] [error] No matrix found with matching beam momentum
[e:97] [JEventProcessorPODIO] [error] Omitting PODIO collection 'ForwardOffMRecParticles' due to exception: No matrix found with matching beam momentum.

The other is the bench:zdc_pi0, which is related to this pull request.  I'll have to look more into what is going on.  I'm suspecting that it's more difficult to separate the two clusters with square tiles (S2 staggering) than hexagonal tiles (H4 staggering).

wdconinc:
> @wdconinc There are two benchmarks that currently fail.  The first is the bench:campaign, which appears to be unrelated to this change.  I keep gettting this error on there:
> 
> [e:97] [FOFFMTRK:ForwardOffMRecParticles] [error] No matrix found with matching beam momentum
> [e:97] [JEventProcessorPODIO] [error] Omitting PODIO collection 'ForwardOffMRecParticles' due to exception: No matrix found with matching beam momentum.
> 
> The other is the bench:zdc_pi0, which is related to this pull request.  I'll have to look more into what is going on.  I'm suspecting that it's more difficult to separate the two clusters with square tiles (S2 staggering) than hexagonal tiles (H4 staggering).  
> 

@veprbl?

veprbl:
The campaign benchmark was fixed in https://github.com/eic/simulation_campaign_hepmc3/pull/88.

For this PR, since this is not a critical subsystem, we may not need to require a perfect backward compatibility (which is is always welcome in any case), but the minimal goals would be to at least:
1. to not fatally crash due to incompatibilities (i.e. not break stuff for other users)
2. have some understanding on what's the impact of incompatibilities (e.g. is it that no clusters are found, or clusters are found with unreasonable energy/position/etc)
3. test the existing benchmark with the new geometry/EICrecon combination and ensure that the results will be sensible

sebouh137:
> The campaign benchmark was fixed in [eic/simulation_campaign_hepmc3#88](https://github.com/eic/simulation_campaign_hepmc3/pull/88).
> 
> For this PR, since this is not a critical subsystem, we may not need to require a perfect backward compatibility (which is is always welcome in any case), but the minimal goals would be to at least:
> 
> 1. to not fatally crash due to incompatibilities (i.e. not break stuff for other users)
> 2. have some understanding on what's the impact of incompatibilities (e.g. is it that no clusters are found, or clusters are found with unreasonable energy/position/etc)
> 3. test the existing benchmark with the new geometry/EICrecon combination and ensure that the results will be sensible

@veprbl Hi Dmitry, thanks for the info.  This doesn't cause crashes when running the new EICrecon with the previous version of epic, however this does affect the results from the EICrecon, which can cause the analysis scripts in the benchmarks to fail.  Essentially, clusters can still be found, but since the subcell splitting of the HEXPLIT algorithm requires the geometry to match the one used in the simulation, it will distribute the energy in a given cell evenly (since it doesn't know how the cells in different layers are supposed to overlap).  This causes any analysis that requires multiple clusters per event to fail because the clusters are getting merged together, causing the analysis step of the benchmarks to crash (in the case of the zdc_pi0 benchmark), or produce garbage results.  I have locally checked that the zdc_pi0 benchmark works on my laptop with both repositories on the correct branches.  So, all three of these criteria appear to be fulfilled.  There will be followup work that will need to be done, due to new geometry, but that can be done in future pull requests.  
Are there any other steps I will need to do in order to get this and the other pull request approved?  Best regards, Sebouh

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
```diff
@@ -32,7 +32,7 @@ class CalorimeterClusterRecoCoG_factory
   PodioOutput<edm4eic::MCRecoClusterParticleAssociation> m_assoc_output{this};
 
   ParameterRef<std::string> m_energyWeight{this, "energyWeight", config().energyWeight};
-  ParameterRef<double> m_samplingFraction{this, "samplingFraction", config().sampFrac};
+  //ParameterRef<double> m_samplingFraction{this, "samplingFraction", config().sampFrac};
```
veprbl: Removing the sampling fraction option doesn't seem like a good idea. It's a useful option.
Would it work to grab the detector `auto detector   = app->GetService<DD4hep_service>()->detector();` and then call `_toDouble` just in ZDC.cc?

src/detectors/ZDC/ZDC.cc
```diff
@@ -194,36 +195,36 @@ void InitPlugin(JApplication* app) {
...
-                                                       ));
+  app->Add(new JOmniFactoryGeneratorT<HEXPLIT_factory>(
+      "HcalFarForwardZDCSubcellHits", {"HcalFarForwardZDCRecHits"},
+      {"HcalFarForwardZDCSubcellHits"},
+      {
+          .MIP           = 630. * dd4hep::keV,
+          .Emin_in_MIPs  = 0.5,
+          .delta_in_MIPs = 0.01,
+          .tmax          = 269 * dd4hep::ns,
+          .stag_type     = HEXPLITConfig::StaggerType::S2,
```
veprbl, line 207: If you want, you could infer staggering type from segmentation, e.g.:
https://github.com/eic/EICrecon/blob/94a9995fea443e2834d15304ed2fb8be89bd3064/src/algorithms/tracking/LGADHitClustering.cc#L45
Would that make it also backwards compatible?

veprbl (commented):
Thank you for checking the benchmarks. I have couple of comments.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
sebouh137: I'll fix this, thanks.  I was trying to have it be able to scale the sampling fraction based on the thickness of the tiles for backwards compatibility, but what you suggested is probably a better way to do this.

src/detectors/ZDC/ZDC.cc
sebouh137: I'll try that and let you know if it works

sebouh137: @veprbl After looking into this, it appears that implementing this would require changing all instances where HEXPLIT is used to include a "readout" field in it, and then making the HEXPLIT algorithm check to see what staggering is used.  Someone could potentially do that as a followup pull request but I don't think this is necessary at this point.

src/factories/calorimetry/CalorimeterClusterRecoCoG_factory.h
sebouh137: It looks like this works now; thanks.

sebouh137:
@veprbl So I've resolved one of the two conversations, and have determined that the second comment would be better addressed by a future followup pull request.  Do you approve now?  Thanks and best regards, Sebouh

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -15,8 +15,9 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdlib>
-#include <gsl/pointers> // for not_null
+#include <iostream>
```
veprbl:
```suggestion
```

src/algorithms/calorimetry/HEXPLIT.cc
```diff
@@ -79,7 +80,23 @@ const std::vector<double> HEXPLIT::subcell_offsets_y = []() {
...
+const std::vector<double> HEXPLIT::subcell_offsets_x_S2 = {0.5, -0.5, -0.5, 0.5};
+const std::vector<double> HEXPLIT::subcell_offsets_y_S2 = {0.5, 0.5, -0.5, -0.5};
+
+void HEXPLIT::init() {
+  if (m_cfg.stag_type == HEXPLITConfig::StaggerType::H4) {
+    stag = stag_H4;
+  } else if (m_cfg.stag_type == HEXPLITConfig::StaggerType::S2) {
+    stag = stag_S2;
+  } else if (m_cfg.stag_type == HEXPLITConfig::StaggerType::H3) {
+    std::cout << "error   H3 staggering not implemented yet in EICrecon" << std::endl;
```
veprbl:
```suggestion
    error("H3 staggering not implemented yet in EICrecon");
```

veprbl (commented):
One more.

src/algorithms/calorimetry/HEXPLIT.h
```diff
@@ -38,23 +38,72 @@ class HEXPLIT : public HEXPLITAlgorithm, public WithPodConfig<HEXPLITConfig> {
...
+      .subcell_offsets_x  = subcell_offsets_x_S2,
+      .subcell_offsets_y  = subcell_offsets_y_S2,
+  };
+
+  stagger_pattern stag = stag_H4;
 
 private:
   const dd4hep::Detector* m_detector{algorithms::GeoSvc::instance().detector()};
-};
+}; // namespace eicrecon
```
veprbl:
```suggestion
};
```

veprbl (approved):
LGTM

wdconinc:
This PR causes breakage in pipelines beyond what was discussed here, see e.g. 27 failed jobs in https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/7573720 and its parent pipeline.

I propose we revert this until this feature can be made more robust. It's blocking everyone from having pipelines succeed now.

veprbl:
This was reverted. @sebouh137, could you please resubmit. We need to make sure this does not crash with the "old" geometry. `_toDouble` with undefined constants leads to a fatal exception.

veprbl:
It may be that the issue is manifested specifically with `epic_inner_detector` configuration, as revealed in the failing benchmark.

sebouh137:
@wdconinc @veprbl I am just reading now your remarks from a few days ago.  I will resubmit the pull request and fix the things that were causing the issues.

sebouh137:
> Update the HEXPLIT algorithm to be able to handle square cells, which are used in the latest version of the ZDC geometry. Also update the parameters in the ZDC reconstruction to match the geometry (which also has thicker cells, so larger sampling fraction and larger MIP value.

@wdconinc @veprbl I just resubmitted.

PR #2590, opened 2026-04-01
Increase requirement on minimum number of track measurements to 4

bschmookler:
This increases the minimum number of measurements required in the CKF for a given track from 3 to 4. This follows extensive studies by the tracking/vertexing group with beam-induced backgrounds, as well as studies by the PWGs.

The minimum number of hits per track in the Ambiguity Solver and Iterative Vertex Finder/Fitter is also increased to 4. This is done so that we always have 

```numMeasurementsMin_CKF <= numMeasurementsMin_AmbiguitySolver <= numMeasurementsMin_VertexFinder```

The extra vertexing factories that were created for the tests done in March are now redundant and have been removed.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: __

- [x] Changes have been communicated to collaborators @ShujieL @Simple-Shyam @rachel-m @sfazio

No

Yes. It impacts what tracks (and therefore ReconstructedParticles) will be saved

### DIFF
--- a/src/algorithms/tracking/AmbiguitySolverConfig.h
+++ b/src/algorithms/tracking/AmbiguitySolverConfig.h
@@ -12,6 +12,6 @@ struct AmbiguitySolverConfig {
   /// Maximum number of iterations
   std::uint32_t maximum_iterations = 100000;
   /// Minimum number of measurement to form a track.
-  std::size_t n_measurements_min = 3;
+  std::size_t n_measurements_min = 4;
 };
 } // namespace eicrecon
--- a/src/algorithms/tracking/CKFTrackingConfig.h
+++ b/src/algorithms/tracking/CKFTrackingConfig.h
@@ -12,6 +12,6 @@ struct CKFTrackingConfig {
   std::vector<double> chi2CutOff                 = {15.};
   std::vector<std::size_t> numMeasurementsCutOff = {10};
 
-  std::size_t numMeasurementsMin = 3;
+  std::size_t numMeasurementsMin = 4;
 };
 } // namespace eicrecon
--- a/src/algorithms/tracking/IterativeVertexFinderConfig.h
+++ b/src/algorithms/tracking/IterativeVertexFinderConfig.h
@@ -5,7 +5,7 @@ namespace eicrecon {
 struct IterativeVertexFinderConfig {
   int maxVertices                  = 10;
   bool reassignTracksAfterFirstFit = true;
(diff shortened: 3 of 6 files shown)


ShujieL:
The change itself looks fine. @veprbl I saw you introduced the 4HitVertice last month. Do you know any analysis group is actively using that branch? Are PWGs aware of this change?

bschmookler:
> > > The change itself looks fine. @veprbl I saw you introduced the 4HitVertice last month. Do you know any analysis group is actively using that branch? Are PWGs aware of this change?
> > 
> > 
> > I think it should be understood that it was a temporary solution for March. I assume, the only user is @Simple-Shyam .
> 
> Yes, as I presented a first look in the tracking meeting that primary vertexing using 4 hits works well in many cases. I am thinking to quantify them. You can keep the primary vertexing using 4 hits tracks.
> 
> There was also a comment by Barbara, she was mentioning to apply it the level of tracking basically store the tracks which have at least 4 hits. As I mentioned, some PWGs may not agree for this (keeping tracks with at least 4 hits).

Hi Shyam. I spoke to Sal and Rachel, and they are okay with the new 4-measurement track cut that is being applied here at the CKF level. So, the 4-measurement track cut will be automatically applied in the Iterative Vertex Finder without the extra factory.

Simple-Shyam:
> > > > The change itself looks fine. @veprbl I saw you introduced the 4HitVertice last month. Do you know any analysis group is actively using that branch? Are PWGs aware of this change?
> > > 
> > > 
> > > I think it should be understood that it was a temporary solution for March. I assume, the only user is @Simple-Shyam .
> > 
> > 
> > Yes, as I presented a first look in the tracking meeting that primary vertexing using 4 hits works well in many cases. I am thinking to quantify them. You can keep the primary vertexing using 4 hits tracks.
> > There was also a comment by Barbara, she was mentioning to apply it the level of tracking basically store the tracks which have at least 4 hits. As I mentioned, some PWGs may not agree for this (keeping tracks with at least 4 hits).
> 
> Hi Shyam. I spoke to Sal and Rachel, and they are okay with the new 4-measurement track cut that is being applied here at the CKF level. So, the 4-measurement track cut will be automatically applied in the Iterative Vertex Finder without the extra factory.

Hi Barak,

Thanks for the information. I was thinking of writing to them, but since you have already confirmed, we can move ahead with that.

ShujieL (approved):

PR #2604, opened 2026-04-10
Introduce managed processing with control over ZMQ

veprbl:
This implements a mode of operation needed for fast processing payload.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


CMakeLists.txt
```diff
@@ -296,6 +297,11 @@ endif()
 # ONNX Runtime
 find_package(onnxruntime ${onnxruntime_MIN_VERSION} CONFIG)
 
+# ZeroMQ
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(ZeroMQ REQUIRED libzmq>=${ZeroMQ_VERSION_MIN})
+find_package(cppzmq REQUIRED)
```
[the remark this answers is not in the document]
veprbl: I think, this is fine. Our user environments contain those dependencies.

cmake/jana_plugin.cmake
```diff
@@ -451,3 +451,33 @@ macro(plugin_add_onnxruntime _name)
   plugin_link_libraries(${PLUGIN_NAME} onnxruntime::onnxruntime)
 
 endmacro()
+
+# Adds ZeroMQ for a plugin
+macro(plugin_add_zeromq _name)
+
+  if(NOT ZeroMQ_FOUND)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(ZeroMQ REQUIRED libzmq>=${ZeroMQ_VERSION_MIN})
```
wdconinc: libzmq also installs cmake config so I don't think we need to deal with pkgconfig (just because this will break container build until we add pkgconfig as a build dependency to EICrecon)

veprbl: I didn't see it:
```
$ ls $(spack location --install-dir libzmq)/lib
==> Warning: Missing dependency not in database: gcc-runtime/54ch43q needs glibc-uvk4zqr
libzmq.a  libzmq.so  libzmq.so.5  libzmq.so.5.2.5  pkgconfig
```

veprbl:
```
$ ls $(spack location --install-dir libzmq)/share
==> Warning: Missing dependency not in database: gcc-runtime/54ch43q needs glibc-uvk4zqr
ls: cannot access '/opt/software/linux-x86_64_v2/libzmq-4.3.5-5m5gf4fryl4kn3lyq6to5c4rb3uzg2dt/share': No such file or directory
```

wdconinc: I was looking at https://github.com/zeromq/libzmq/blob/b946c18f676760387276cd095bbdd8c0e18c09bf/CMakeLists.txt#L1782 ... seems like maybe we somehow don't install that?

wdconinc: Ah, we use autotools to install it, not cmake.

src/services/io/podio/JEventProcessorManagedPODIO.cc
```diff
@@ -0,0 +1,370 @@
...
+      std::lock_guard<std::mutex> lock(m_file_mutex);
+      m_current_input_file     = input_file;
+      m_current_output_file    = output_file;
+      m_file_processing_active = true;
+      m_events_processed       = 0;
+      m_pending_response       = true; // Mark that we have a pending response to send
+    }
+
+    // Open output file before notifying the source
+    OpenOutputFile(output_file);
```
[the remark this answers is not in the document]
ruse-traveler: Ahhh... This is an interesting case! How _would_ we want EICrecon to behave if the collection sets differ between inputs? For example, maybe one input doesn't have backward endcap hits: would we want to run the reconstruction as-is for the rest of the detector?

That being said, I find it kinda hard to imagine a situation right now where that might happen (esp. while we're only running simulations). So think this edge case might be out of scope for this PR...

src/services/io/podio/JEventSourcePODIO.h
```diff
@@ -34,7 +34,9 @@ class JEventSourcePODIO : public JEventSource {
   void PrintCollectionTypeTable(void);
 
   std::vector<std::string_view> getAvailableCategories() const;
-  std::size_t getEntries(const std::string& category) const;
+  std::size_t getEntries(const std::string& category) const {
+    return m_reader->getEntries(category);
+  }
```
ruse-traveler: For my own understanding: why make this inline here?

src/services/io/podio/JEventProcessorManagedPODIO.cc
```diff
@@ -0,0 +1,389 @@
...
+void JEventProcessorManagedPODIO::CloseOutputFile() {
+  if (m_writer) {
+    try {
+      // Propagate non-event frames (same as parent class)
+      auto* app          = GetApplication();
+      auto event_sources = app->GetService<JComponentManager>()->get_evt_srces();
+      for (auto* source : event_sources) {
+        auto* podio_source = dynamic_cast<JEventSourcePODIO*>(source);
+        if (podio_source == nullptr)
+          continue;
```
ruse-traveler: Minor suggestion:
```suggestion
        if (podio_source == nullptr) {
          continue;
        }
```

src/services/io/podio/JEventProcessorManagedPODIO.cc
```diff
@@ -0,0 +1,389 @@
...
+        for (const auto& _category : podio_source->getAvailableCategories()) {
+          std::string category{_category};
+          if (category == "events")
+            continue;
```
ruse-traveler: Minor suggestion:
```suggestion
          if (category == "events") {
            continue;
          }
```

ruse-traveler (commented):
Thanks, Dima! This a huge advancement! Just to check my understanding: if a new file is received while one's being processed, then the new one won't start processing until the current one's done, right?

If so, I feel like a lot of Copilot's suggestions are little over-defensive and assume that possibly files will be processed concurrently... Most of the suggestions don't seem _outrageous_ if we were processing files concurrently, but I don't think that's the case :/

src/services/io/podio/JEventSourcePODIO.h
veprbl: I actually don't think this needs to be in the header.

veprbl:
> Thanks, Dima! This a huge advancement! Just to check my understanding: if a new file is received while one's being processed, then the new one won't start processing until the current one's done, right?

It is not currently possible to queue files. The ZMQ REQ/REP sockets do not support this in their state machine. This was a conscious choice, the queuing logic is meant to live on the side of the testbed.

> If so, I feel like a lot of Copilot's suggestions are little over-defensive and assume that possibly files will be processed concurrently... Most of the suggestions don't seem outrageous if we were processing files concurrently, but I don't think that's the case :/

I agree.

> Ahhh... This is an interesting case! How would we want EICrecon to behave if the collection sets differ between inputs? For example, maybe one input doesn't have backward endcap hits: would we want to run the reconstruction as-is for the rest of the detector?

This is a good question. When writing a singular PODIO file you are explicitly not allowed to change set of collections that you write from event to event. There is a comment on that in JEventProcessorPODIO.cc (although, I don't think the logic actually implemented even works). I don't think anything on JANA2 side prevents us from processing different sets of collections for different files. We probably don't need to do that in the short term, but it would require doing some additional resets to the state of the JEventProcessorPODIO when JEventProcessorManagedPODIO needs to advance to the next file. I would do it now, if not for potential bugs in JEventProcessorPODIO that may need a separate PR.

cmake/jana_plugin.cmake
veprbl: Actually pkgtools here do nothing for us since cppzmq already does its own discovery. I did have to implement a hack to combat other hacks related to autotools.

I don't know if we want to switch libzmq from autotools and cmake, since most package managers use autotools for it, we could just hit new class of issues (and I prefer hacking cmake hacks over autotools bugs).

What you think?

src/scripts/eicrecon-client.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python3
+"""
+Simple client for EICrecon managed PODIO processor.
+Submits a file processing request and listens for all responses.
+"""
```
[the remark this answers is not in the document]
veprbl:
```suggestion
"""
Simple client for EICrecon managed PODIO processor.
Submits a file processing request and listens a response.
"""
```

src/services/io/podio/JEventProcessorManagedPODIO.cc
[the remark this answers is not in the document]
veprbl: Actually, this seems correct.

src/services/io/podio/JEventProcessorManagedPODIO.cc
```diff
@@ -0,0 +1,391 @@
...
+#include "services/io/podio/JEventSourceManagedPODIO.h"
+#include "services/io/podio/JEventSourcePODIO.h"
+#include "services/log/Log_service.h"
+
+JEventProcessorManagedPODIO::JEventProcessorManagedPODIO() : JEventProcessorPODIO() {
+  SetTypeName(NAME_OF_THIS);
+
+  // Register the managed-mode socket path parameter (used to enable managed mode from the CLI)
+  japp->SetDefaultParameter("podio:managed_socket_path", m_socket_path,
+                            "UNIX socket path for managed PODIO processing");
```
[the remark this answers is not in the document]
veprbl:
```suggestion
  japp->SetDefaultParameter("podio:managed_socket_path", m_socket_path,
                            "UNIX socket path for managed PODIO processing");
```

veprbl:
Okay, I think the problems reported are getting more obscure. Perhaps, this is enough.

cmake/jana_plugin.cmake
```diff
@@ -447,3 +447,39 @@ macro(plugin_add_onnxruntime _name)
...
 endmacro()
+
+# Adds ZeroMQ for a plugin
+macro(plugin_add_zeromq _name)
+
+  if(NOT TARGET libzmq)
+    # This and set_target_properties below addresses a discrepancy between
+    # https://github.com/zeromq/cppzmq/blob/master/libzmq-pkg-config/FindZeroMQ.cmake
+    # and
+    # https://github.com/JeffersonLab/JANA2/blob/master/cmake/FindZeroMQ.cmake
```
wdconinc, line 458: Why are we picking up the JANA2 FindZeroMQ? It is a build infrastructure component and should not get installed or found by EICrecon's build.

src/services/io/podio/podio.cc
```diff
@@ -5,24 +5,26 @@
...
-  // input file as an argument.
-  // Only add a EICRootWriter if the user has specified a configuration parameter relevant to writing
-  //    if( app->GetJParameterManager()->Exists("podio:output_file")
-  //        ||  app->GetJParameterManager()->Exists("podio:output_file_copy_dir")
-  //        ||  app->GetJParameterManager()->Exists("podio:output_include_collections")
-  //        ||  app->GetJParameterManager()->Exists("podio:output_exclude_collections")        ){
-  app->Add(new JEventProcessorPODIO());
-  //    }
+  // Check if managed mode is requested
+  if (app->GetJParameterManager()->Exists("podio:managed_socket_path")) {
```
wdconinc, line 22: If a managed podio instance inherits the regular podio functionality, why do we need to branch on the parameters here? Can we not just use managed podio always?

ruse-traveler:
> It is not currently possible to queue files. The ZMQ REQ/REP sockets do not support this in their state machine. This was a conscious choice, the queuing logic is meant to live on the side of the testbed.

Ahhhh I see! Thanks for the clarification!
 
> This is a good question. When writing a singular PODIO file you are explicitly not allowed to change set of collections that you write from event to event. There is a comment on that in JEventProcessorPODIO.cc (although, I don't think the logic actually implemented even works). I don't think anything on JANA2 side prevents us from processing different sets of collections for different files. We probably don't need to do that in the short term, but it would require doing some additional resets to the state of the JEventProcessorPODIO when JEventProcessorManagedPODIO needs to advance to the next file. I would do it now, if not for potential bugs in JEventProcessorPODIO that may need a separate PR.

Agreed! I don't think we need this near term, and it can definitely be handled in a later PR! But it's good food-for-thought...

ruse-traveler (approved):
LGTM!

cmake/jana_plugin.cmake
veprbl: This is how transient target dependencies work in CMake
https://github.com/zeromq/cppzmq/blob/7f0530688804c2b5b6b0d985773405593fd25ca8/cppzmqConfig.cmake.in#L20-L26
https://github.com/JeffersonLab/JANA2/blob/28d1ea76b4e9c48c3008e08b359d1cbe8a3bf08b/cmake/JANAConfig.cmake.in#L20-L23
it's needed for things like header includes, static linking, or imposing version constraints.

src/services/io/podio/podio.cc
veprbl: The managed source does not accept file names from command line arguments at all. The managed processor will not begin processing anything until someone connects to the socket and submits a file.

cmake/jana_plugin.cmake
wdconinc: I know *why* we need the find_dependency, but I there's a zeromq CMake config then JANA2 needs to use that and not its own bundled Find module which is incompatible with the zeromq CMake config.

veprbl: How do you want to achieve that?

wdconinc: We should fix to JANA2. cppzmq has distributed a proper [cppzmqConfig.cmake.in](https://github.com/zeromq/cppzmq/commits/master/cppzmqConfig.cmake.in) for 9 years now, so maybe it's time to start using it instead of bundling a custom search module that then conflicts with other projects that find the upstream config. FYI @nathanwbrei

wdconinc: (but that said, I don't want to hold this up, so don't see it as a request to pass review)

veprbl: Your suggestion does not work as exactly stated since JANA2 does not depend on cppzmq. It would work if JANA2 copied exact same FindZMQ implementation, though.

wdconinc: Now I'm completely lost. What is JANA2's FindZeroMQ.cmake then for if there's no dependency, and why do we even pick it up if there's then no find_dependency?

veprbl: I filed https://github.com/JeffersonLab/JANA2/pull/507

PR #2609, opened 2026-04-10
Add FF Lambda multi-calorimeter reconstruction

baptistefraisse:
## Summary
This PR adds far-forward Lambda reconstruction using multi-calorimeter inputs (improving the previous ZDC-only approach).

## Main changes
- improve far-forward Lambda reconstruction algorithms (from ZDC-only to multi-calorimeters)
- add far-forward neutrals reconstruction for EcalEndCapP, LFHCAL, B0-Ecal.
- integrate new components into the reco chain
- update related factories and configuration

## Motivation
This extends the far-forward Lambda reconstruction beyond the previous setup and makes it available from `main`.

## Notes
- rebased on the latest `main`
- resolved conflict in `JEventProcessorPODIO.cc` by keeping upstream logic

### DIFF
--- a/src/algorithms/reco/FarForwardLambdaReconstruction.cc
+++ b/src/algorithms/reco/FarForwardLambdaReconstruction.cc
@@ -1,182 +1,376 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2025 Sebouh Paul
+// Update/modification 2026 by Baptiste Fraisse
 
 #include <Evaluator/DD4hepUnits.h>
 #include <TVector3.h>
 #include <edm4eic/ReconstructedParticleCollection.h>
 #include <edm4hep/Vector3f.h>
-#include <edm4hep/utils/vector_utils.h>
 #include <fmt/core.h>
 #include <cmath>
 #include <cstddef>
 #include <gsl/pointers>
 #include <stdexcept>
 #include <vector>
+#include <type_traits>
 
 #include "FarForwardLambdaReconstruction.h"
 #include "TLorentzVector.h"
-/**
-Creates Lambda candidates from a neutron and two photons from a pi0 decay
- */
 
 namespace eicrecon {
 
-void FarForwardLambdaReconstruction::init() {
+// helpers 
 
+static inline void toTVector3(TVector3& v1, const edm4hep::Vector3f& v2) {
+  v1.SetXYZ(v2.x, v2.y, v2.z);
+}
+
+void FarForwardLambdaReconstruction::init() {
   try {
     m_zMax = m_detector->constant<double>(m_cfg.offsetPositionName);
   } catch (std::runtime_error&) {
-    m_zMax = 35800; // default value
-    trace("Failed to get {} from the detector, using default value of {}", m_cfg.offsetPositionName,
-          m_zMax);
+    m_zMax = 35800;
+    trace("Failed to get {} from the detector, using default value of {}", m_cfg.offsetPositionName, m_zMax);
   }
 }
 
-/* converts one type of vector format to another */
-void toTVector3(TVector3& v1, const edm4hep::Vector3f& v2) { v1.SetXYZ(v2.x, v2.y, v2.z); }
+// reconstruction machinery from n+g+g
 
-void FarForwardLambdaReconstruction::process(
-    const FarForwardLambdaReconstruction::Input& input,
-    const FarForwardLambdaReconstruction::Output& output) const {
-  const auto [neutrals]                                = input;
-  auto [out_lambdas, out_decay_products]               = output;
-  std::vector<edm4eic::ReconstructedParticle> neutrons = {};
-  std::vector<edm4eic::ReconstructedParticle> gammas   = {};
-  for (auto part : *neutrals) {
-    if (part.getPDG() == 2112) {
-      neutrons.push_back(part);
-    }
-    if (part.getPDG() == 22) {
-      gammas.push_back(part);
-    }
-  }
-
-  if (neutrons.empty() || gammas.size() < 2) {
-    return;
-  }
-
+bool FarForwardLambdaReconstruction::reconstruct_from_triplet(
+    const edm4eic::ReconstructedParticle& n_in,
+    const edm4eic::ReconstructedParticle& g1_in,
+    const edm4eic::ReconstructedParticle& g2_in,
+    edm4eic::ReconstructedParticleCollection* out_lambdas,
+    edm4eic::ReconstructedParticleCollection* out_decay_products) const
+{
   static const double m_neutron = m_particleSvc.particle(2112).mass;
   static const double m_pi0     = m_particleSvc.particle(111).mass;
   static const double m_lambda  = m_particleSvc.particle(3122).mass;
 
-  for (std::size_t i_n = 0; i_n < neutrons.size(); i_n++) {
-    for (std::size_t i_1 = 0; i_1 < gammas.size() - 1; i_1++) {
-      for (std::size_t i_2 = i_1 + 1; i_2 < gammas.size(); i_2++) {
+  const double En = n_in.getEnergy();
+  const double E1 = g1_in.getEnergy();
+  const double E2 = g2_in.getEnergy();
 
-        double En = neutrons[i_n].getEnergy();
-        double pn = sqrt(En * En - m_neutron * m_neutron);
-        double E1 = gammas[i_1].getEnergy();
-        double E2 = gammas[i_2].getEnergy();
-        TVector3 xn;
-        TVector3 x1;
-        TVector3 x2;
+  if (En <= m_neutron) return false;
+  const double pn = std::sqrt(std::max(0.0, En*En - m_neutron*m_neutron));
 
-        toTVector3(xn, neutrons[0].getReferencePoint() * dd4hep::mm);
-        toTVector3(x1, gammas[0].getReferencePoint() * dd4hep::mm);
-        toTVector3(x2, gammas[1].getReferencePoint() * dd4hep::mm);
+  TVector3 xn, x1, x2;
 
-        xn.RotateY(-m_cfg.globalToProtonRotation);
-        x1.RotateY(-m_cfg.globalToProtonRotation);
-        x2.RotateY(-m_cfg.globalToProtonRotation);
+  const auto& rn  = n_in.getReferencePoint();
+  const auto& rg1 = g1_in.getReferencePoint();
+  const auto& rg2 = g2_in.getReferencePoint();
 
-        debug("nx recon = {}, g1x recon = {}, g2x recon = {}", xn.X(), x1.X(), x2.X());
-        debug("nz recon = {}, g1z recon = {}, g2z recon = {}, z face = {}", xn.Z(), x1.Z(), x2.Z(),
-              m_zMax);
+  xn.SetXYZ(rn.x * dd4hep::mm,
+            rn.y * dd4hep::mm,
+            rn.z * dd4hep::mm);
 
-        TVector3 vtx(0, 0, 0);
-        double f                   = 0;
-        double df                  = 0.5;
-        double theta_open_expected = 2 * asin(m_pi0 / (2 * sqrt(E1 * E2)));
-        TLorentzVector n;
-        TLorentzVector g1;
-        TLorentzVector g2;
-        TLorentzVector lambda;
-        for (int i = 0; i < m_cfg.iterations; i++) {
-          n                 = {pn * (xn - vtx).Unit(), En};
-          g1                = {E1 * (x1 - vtx).Unit(), E1};
-          g2                = {E2 * (x2 - vtx).Unit(), E2};
-          double theta_open = g1.Angle(g2.Vect());
-          lambda            = n + g1 + g2;
-          if (theta_open > theta_open_expected) {
-            f -= df;
-          } else if (theta_open < theta_open_expected) {
-            f += df;
-          }
+  x1.SetXYZ(rg1.x * dd4hep::mm,
+            rg1.y * dd4hep::mm,
+            rg1.z * dd4hep::mm);
 
-          vtx = lambda.Vect() * (f * m_zMax / lambda.Z());
-          df /= 2;
-        }
+  x2.SetXYZ(rg2.x * dd4hep::mm,
+            rg2.y * dd4hep::mm,
+            rg2.z * dd4hep::mm);
 
-        double mass_rec = lambda.M();
-        if (std::abs(mass_rec - m_lambda) > m_cfg.lambdaMaxMassDev) {
-          continue;
-        }
+  xn.RotateY(-m_cfg.globalToProtonRotation);
+  x1.RotateY(-m_cfg.globalToProtonRotation);
+  x2.RotateY(-m_cfg.globalToProtonRotation);
 
-        // rotate everything back to the lab coordinates.
-        vtx.RotateY(m_cfg.globalToProtonRotation);
-        lambda.RotateY(m_cfg.globalToProtonRotation);
-        n.RotateY(m_cfg.globalToProtonRotation);
-        g1.RotateY(m_cfg.globalToProtonRotation);
-        g2.RotateY(m_cfg.globalToProtonRotation);
+  TVector3 vtx(0, 0, 0);
+  double f  = 0.0;
+  double df = 0.5;
 
-        auto b = -lambda.BoostVector();
-        n.Boost(b);
-        g1.Boost(b);
-        g2.Boost(b);
+  const double theta_open_expected = 2 * std::asin(m_pi0 / (2 * std::sqrt(E1 * E2)));
 
-        //convert vertex to mm:
-        vtx = vtx * (1 / dd4hep::mm);
+  TLorentzVector n, g1, g2, lambda;
 
-        auto rec_lambda = out_lambdas->create();
-        rec_lambda.setPDG(3122);
+  for (int i = 0; i < m_cfg.iterations; i++) {
+    n      = { pn * (xn - vtx).Unit(), En };
+    g1     = { E1 * (x1 - vtx).Unit(), E1 };
+    g2     = { E2 * (x2 - vtx).Unit(), E2 };
+    lambda = n + g1 + g2;
 
-        rec_lambda.setEnergy(lambda.E());
-        rec_lambda.setMomentum({static_cast<float>(lambda.X()), static_cast<float>(lambda.Y()),
-                                static_cast<float>(lambda.Z())});
-        rec_lambda.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()),
-                                      static_cast<float>(vtx.Z())});
-        rec_lambda.setCharge(0);
-        rec_lambda.setMass(mass_rec);
+    const double theta_open = g1.Angle(g2.Vect());
+    if (theta_open > theta_open_expected)      f -= df;
+    else if (theta_open < theta_open_expected) f += df;
 
-        auto neutron_cm = out_decay_products->create();
-        neutron_cm.setPDG(2112);
-        neutron_cm.setEnergy(n.E());
-        neutron_cm.setMomentum(
-            {static_cast<float>(n.X()), static_cast<float>(n.Y()), static_cast<float>(n.Z())});
-        neutron_cm.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()),
-                                      static_cast<float>(vtx.Z())});
-        neutron_cm.setCharge(0);
-        neutron_cm.setMass(m_neutron);
-        //link the reconstructed lambda to the input neutron,
-        // and the cm neutron to the input neutron
-        rec_lambda.addToParticles(neutrons[i_n]);
-        neutron_cm.addToParticles(neutrons[i_n]);
+    vtx = lambda.Vect() * (f * m_zMax / lambda.Z());
+    df /= 2.0;
+  }
 
-        auto gamma1_cm = out_decay_products->create();
-        gamma1_cm.setPDG(22);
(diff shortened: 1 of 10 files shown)


src/factories/reco/FarForwardLambdaReconstruction_factory.h
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2025 Sebouh Paul
+// Update/modification 2026 by Baptiste Fraisse
```
veprbl:
```suggestion
// Copyright (C) 2025 - 2026 Sebouh Paul, Baptiste Fraisse
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2025 Sebouh Paul
+// Update/modification 2026 by Baptiste Fraisse
```
ruse-traveler, line 3: So far we've usually just added contributors to the copyright!
```suggestion
// Copyright (C) 2026 Sebouh Paul, Baptiste Fraisse
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2025 Sebouh Paul
+// Update/modification 2026 by Baptiste Fraisse
+
```
ruse-traveler, line 4: Ditto.
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2025 Sebouh Paul, Baptiste Fraisse

```

src/factories/reco/FarForwardLambdaReconstruction_factory.h
ruse-traveler, line 3: Ditto.
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2025 Sebouh Paul, Baptiste Fraisse
```

src/factories/reco/FarForwardNeutralsReconstruction_factory.h
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2025 Sebouh Paul
+// Update/modification 2026 by Baptiste Fraisse
```
ruse-traveler, line 3: Ditto.
```suggestion
// SPDX-License-Identifier: LGPL-3.0-or-later
// Copyright (C) 2025 Sebouh Paul, Baptiste Fraisse
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -10,139 +12,302 @@
...
 
-/**
- Creates photon candidate Reconstructed Particles, using clusters which fulfill cuts on the position of their CoG position, length (sqrt of largest eigenvector of their moment matrix), and width (sqrt of second largest eigenvector of their moment matrix).  Its energy is the energy of the cluster times a correction factor.
- Also creates a "neutron candidate" Reconstructed Particle consisting of all remaining clusters in the collection. Its energy is the sum of the energies of the constituent clusters
- times a correction factor, and its direction is the direction from the origin to the position
- of the most energetic cluster.  The correction factors in both cases are given by 1/(1+c[0]+c[1]/sqrt(E)+c[2]/E),
- where c is the coefficients and E is the uncorrected energy in GeV.  Different correction coefficients are used for photons vs for neutrons.  This form was chosen
- empirically based on the discrepancies in single-neutron and single-photon MC simulations between the uncorrected
- reconstructed energies and the truth energies of the neutrons.  The parameter values should be co-tuned with those of the clustering algorithm being used.
- */
```
ruse-traveler, line 24: I think this documentation is useful, so I'd prefer us to keep it! Granted, where it's currently placed it won't get picked up by Doxygen (c.f. the link below), so my suggestion would be to instead move it over to the class definition.

https://eicrecon.epic-eic.org/doxygen/classeicrecon_1_1FarForwardLambdaReconstruction.html

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
   } catch (std::runtime_error&) {
-    m_zMax = 35800; // default value
+    m_zMax = 35800;
     trace("Failed to get {} from the detector, using default value of {}", m_cfg.offsetPositionName,
           m_zMax);
   }
 }
 
-/* converts one type of vector format to another */
-void toTVector3(TVector3& v1, const edm4hep::Vector3f& v2) { v1.SetXYZ(v2.x, v2.y, v2.z); }
```
ruse-traveler, line 36: Out of scope for this PR, but leaving it as a note: this would be a handy snippet to add to `edm4eic::utils`!

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
+
+  const auto& rn  = n_in.getReferencePoint();
+  const auto& rg1 = g1_in.getReferencePoint();
+  const auto& rg2 = g2_in.getReferencePoint();
+
+  xn.SetXYZ(rn.x * dd4hep::mm, rn.y * dd4hep::mm, rn.z * dd4hep::mm);
+
+  x1.SetXYZ(rg1.x * dd4hep::mm, rg1.y * dd4hep::mm, rg1.z * dd4hep::mm);
+
+  x2.SetXYZ(rg2.x * dd4hep::mm, rg2.y * dd4hep::mm, rg2.z * dd4hep::mm);
```
ruse-traveler, line 61: I'm not sure I understand why the formatting is different here than in the surrounding blocks... I would suggest making it consistent and remove the extra space!
```suggestion
  xn.SetXYZ(rn.x * dd4hep::mm, rn.y * dd4hep::mm, rn.z * dd4hep::mm);
  x1.SetXYZ(rg1.x * dd4hep::mm, rg1.y * dd4hep::mm, rg1.z * dd4hep::mm);
  x2.SetXYZ(rg2.x * dd4hep::mm, rg2.y * dd4hep::mm, rg2.z * dd4hep::mm);
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
+
+  if (!neutrons_zdc.empty())
+    try_neutrons(neutrons_zdc, 0);
+  if (!neutrons_other.empty())
+    try_neutrons(neutrons_other, 1);
+
+  if (cands.empty())
+    return;
+
+  // (3) best candidate selection (your policy)
```
ruse-traveler, line 354:
```suggestion
  // (3) best candidate selection
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
+    int n_idx;
+    int n_cat; // 0: ZDC neutron, 1: other neutron
+    int g_cat; // 0:2ZDC, 1:1ZDC, 2:0ZDC
+    double chi2;
+    double E;
+    double pz;
+  };
+
+  std::vector<LambdaCand> cands;
+  cands.reserve(64);
```
ruse-traveler, line 307: I assume the 64 here is because we have 4 input collections and we expect at most 3 particles from each?

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
-        gamma2_cm.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()),
-                                     static_cast<float>(vtx.Z())});
-        gamma2_cm.setCharge(0);
-        gamma2_cm.setMass(0);
-        rec_lambda.addToParticles(gammas[i_2]);
-        gamma2_cm.addToParticles(gammas[i_2]);
+
+        const double dL = (mL - m_lambda);
+
+        // same score as your new code
```
ruse-traveler, line 327:
```suggestion
```

src/algorithms/reco/FarForwardNeutralsReconstruction.h
```diff
@@ -21,18 +21,33 @@
...
+      : FarForwardNeutralsReconstructionAlgorithm{
+            name,
+
+            {"clustersHcal", "clustersB0", "clustersEcalEndCapP", "clustersLFHCAL"},
+
+            {"outputNeutralsHcal", "outputNeutralsB0", "outputNeutralsEcalEndCapP",
+             "outputNeutralsLFHCAL"},
+
+            "Merges all HCAL clusters in a collection into a neutron candidate and photon "
+            "candidates "} {}
```
ruse-traveler, line 50: Since the algorithm now handles both EMCal and HCal clusters, I'd suggest updating the algorithm description!
```suggestion
            {"clustersHcal", "clustersB0", "clustersEcalEndCapP", "clustersLFHCAL"},
            {"outputNeutralsHcal", "outputNeutralsB0", "outputNeutralsEcalEndCapP",
             "outputNeutralsLFHCAL"},
            "Convert EMCal and HCal clusters into neutron or photon candidates"
            "candidates "} {}
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
-    }
+  const auto [neutralsHcal, neutralsB0, neutralsEcalEndcapP, neutralsLFHCAL] = input;
+  auto [out_lambdas, out_decay_products]                                     = output;
+
+  const double m_pi0    = m_particleSvc.particle(111).mass;
+  const double m_lambda = m_particleSvc.particle(3122).mass;
+
+  // (A) collect by detector category
+  std::vector<edm4eic::ReconstructedParticle> neutrons_zdc, neutrons_other;
+  std::vector<edm4eic::ReconstructedParticle> gammas_zdc, gammas_other;
```
ruse-traveler, line 179: For my own understanding: why do we need to separate the neutrals into 2 categories?

My initial thought was that this is effectively coarse-graining the calculation and reducing combinatorics, but that doesn't seem to be the case... Rather, it seems that the ZDC neutrals are preferred over the others?

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
-  std::vector<edm4eic::ReconstructedParticle> neutrons = {};
-  std::vector<edm4eic::ReconstructedParticle> gammas   = {};
-  for (auto part : *neutrals) {
-    if (part.getPDG() == 2112) {
-      neutrons.push_back(part);
-    }
-    if (part.getPDG() == 22) {
-      gammas.push_back(part);
-    }
+  const auto [neutralsHcal, neutralsB0, neutralsEcalEndcapP, neutralsLFHCAL] = input;
```
ruse-traveler, line 171: Hardcoding particular collections as inputs is something we generally try to avoid with our algorithms (though aren't always successful). So while these particular calorimeters will likely be the only consumers for now, it would be preferable to keep this flexible! For example, what if we wanted to incorporate the forward insert?

One possible approach to generalizing this could be to instead have it accept a vector of neutral collections. Aside from the preference given to ZDC neutrals, identifying and ranking the lambda candidates seems like it could handle an arbitrary number of inputs.

The [MergeTracks](https://github.com/eic/EICrecon/blob/main/src/algorithms/pid/MergeTracks.h) algorithm is one good reference of using a vector of inputs!

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,383 @@
...
+    const double Eb = b.getEnergy();
+    const double Ec = c.getEnergy();
+    const double px = pa.x + pb.x + pc.x;
+    const double py = pa.y + pb.y + pc.y;
+    const double pz = pa.z + pb.z + pc.z;
+    const double E  = Ea + Eb + Ec;
+    return E * E - (px * px + py * py + pz * pz);
+  };
+
+  // (1) pi0 candidates
```
ruse-traveler, line 261: These (1), (2), ... comments seem to be demarcating stages of the calculation. Something that would be very useful would be documentation that summarizes the algorithm in broad strokes (e.g. "1. Form list of pi0 candidates by grouping all photons into pairs"). My suggestion would be to write this out as a doxygen comment on `process`!

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -10,139 +12,302 @@
...
+
 double FarForwardNeutralsReconstruction::calc_corr(double Etot, const std::vector<double>& coeffs) {
   return coeffs[0] + coeffs[1] / sqrt(Etot) + coeffs[2] / Etot;
 }
 
-/**
-     check that the cluster position is within the correct range,
-     and that the sqrt(largest eigenvalue) is less than gamma_max_length,
-     and that the sqrt(second largest eigenvalue) is less than gamma_max_width
-  */
```
ruse-traveler, line 63: Similar to the above suggestion on removing comments: even though these won't get picked up by doxgen for now, I'd prefer we keep this documentation. Having non-code explanations of what these functions are doing (or trying to do) can go a long way in helping others understand the code. Lacking this documentation this something has already frustrated users several times...

My suggestion would be to either keep these or make them more descriptive!

src/factories/reco/FarForwardLambdaReconstruction_factory.h
```diff
@@ -19,14 +20,20 @@ class FarForwardLambdaReconstruction_factory
...
+
   PodioOutput<edm4eic::ReconstructedParticle> m_lambda_output{this};
   PodioOutput<edm4eic::ReconstructedParticle> m_lambda_decay_products_cm_output{this};
 
   ParameterRef<std::string> m_offset_position_name{this, "offsetPositionName",
                                                    config().offsetPositionName};
   ParameterRef<double> m_rot_y{this, "globalToProtonRotation", config().globalToProtonRotation};
-  ParameterRef<double> m_lambda_max_mass_dev{this, "lambdaMaxMassDev", config().lambdaMaxMassDev};
+  ParameterRef<double> m_lambda_max_mass_dev{this, "lambdaMassWindow", config().lambdaMassWindow};
+  ParameterRef<double> m_pi0_max_mass_dev{this, "pi0Window", config().pi0Window};
```
ruse-traveler, line 36: Small point here, I'd suggest making the variable names consistent!
```suggestion
  ParameterRef<double> m_lambda_max_mass_window{this, "lambdaMassWindow", config().lambdaMassWindow};
  ParameterRef<double> m_pi0_max_mass_window{this, "pi0Window", config().pi0Window};
```

src/factories/reco/FarForwardLambdaReconstruction_factory.h
```diff
@@ -40,8 +47,12 @@ class FarForwardLambdaReconstruction_factory
   }
 
   void Process(int32_t /* run_number */, uint64_t /* event_number */) {
-    m_algo->process({m_neutrals_input()},
-                    {m_lambda_output().get(), m_lambda_decay_products_cm_output().get()});
+    m_algo->process(
+
+        {m_hcal_neutrals_input(), m_b0_neutrals_input(), m_ecalendcapp_neutrals_input(),
+         m_lfhcal_neutrals_input()},
+
```
ruse-traveler, line 54:
```suggestion
        {m_hcal_neutrals_input(), m_b0_neutrals_input(), m_ecalendcapp_neutrals_input(),
         m_lfhcal_neutrals_input()},
```

ruse-traveler (commented):
Thanks for the significant amount work on this, Baptiste! This is a big PR, so I'll likely have another round of comments.

Overall, I like the direction this is heading! After my 1st pass, something I think you should consider is if it's possible to avoid hardcoding the input collections? While this might require a some rethinking how the lambda candidates are formed, this will help keep the algorithm generic and make it easier to extend/maintain in the future!

veprbl:
The remaining tests will not pass because this is submitted from a fork repository.
To fix that, please push your branch to the shared repository and resubmit your PR.

baptistefraisse:
Thanks, I will resubmit this PR from a branch in the shared repository so the workflows can run properly.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Thanks, fixed.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
baptistefraisse: Thanks, fixed.

src/factories/reco/FarForwardLambdaReconstruction_factory.h
baptistefraisse: Thanks, fixed.

src/factories/reco/FarForwardNeutralsReconstruction_factory.h
baptistefraisse: Thanks, fixed.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
baptistefraisse: Thanks, it will be included in the next PR.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Thanks, fixed.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Thanks, fixed.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Yes exactly, I will include a comment to make it clearer for users.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Thanks, fixed.

src/algorithms/reco/FarForwardNeutralsReconstruction.h
baptistefraisse: You are right, I update the title for the next PR.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Thank you for the advise. I will let those comments in the code for future users, but also include a comment into on the process to be caught by doxygen, as you suggest.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
baptistefraisse: Thank you. I include those comments for the next PR.

src/factories/reco/FarForwardLambdaReconstruction_factory.h
baptistefraisse: Thanks, fixed.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Thanks for the question, good point!
The separation into two categories is not meant to reduce the combinatorics. All combinations are considered when building the candidate pool. It is used for detector preference. In our implementation, candidates involving ZDC neutrals are 'prioritized' over those from other calorimeters. This reflects the fact that the ZDC is more forward and typically provides a better handle and efficiency for the kinematics of forward neutrons and photons relevant for the Lambda reconstruction. The other calorimeters are used to build additional candidates, but are treated as complementary inputs rather than primary ones if ZDC did not catch the triplet (n+g1+g2).

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Thanks, that's a very good point.
I agree that hardcoding the input collections is not ideal and that a more flexible approach based on a vector of inputs would improve extensibility.
For this PR, what about keeping the explicit inputs to avoid introducing a larger refactor of the factory and wiring? However, I will clean up the implementation by centralizing the detector-specific handling into a local input-description table to reduce duplication and make the logic clearer. I suggest we open a follow-up issue to properly generalize the interface and clean this up in a dedicated PR later. Do you agree with this strategy?

src/factories/reco/FarForwardLambdaReconstruction_factory.h
baptistefraisse: Thanks, fixed.

baptistefraisse:
This PR has been resubmitted from a branch in the shared repository.
Please refer to the new PR for further review.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
ruse-traveler: Sounds good to me!

src/algorithms/reco/FarForwardLambdaReconstruction.cc
ruse-traveler: Gotcha! Thanks!

src/algorithms/reco/FarForwardLambdaReconstruction.cc
ruse-traveler: I see! Makes sense! Thanks for the clarification!

PR #2618, opened 2026-04-15
Hexplit s2 --resubmit

sebouh137:
resubmision of the https://github.com/eic/EICrecon/pull/2589

Update the HEXPLIT algorithm to be able to handle square cells, which are used in the latest version of the ZDC geometry. Also update the parameters in the ZDC reconstruction to match the geometry (which also has thicker cells, so larger sampling fraction and larger MIP value.

- [X] High
- [ ] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [X] Updated parameters, constants (issue #__)
- [ ] Updated documentation
- [ ] other: __

- [X] This PR requires changes to geometry (epic PR: https://github.com/eic/epic/pull/1070 )
- [X] This PR changes default behavior. Please describe changes below.

Changed parameters used in ZDC reconstruction.  Also the HEXPLIT algorithm is set to assume square (S2) staggering in the ZDC.


sebouh137:
@veprbl @wdconinc I have removed the _toDouble statements that were causing the code to break.  I was attempting to obtain the thickness of the scintillator tiles in order to have the sampling fraction scale with the scintillator-tile thickness, but if there isn't a way to reliably do this, then it is better to hardcode the value than to have it rely on the presence of a variable that may or may not exist.

veprbl:
Thanks.
I think the real issue might be that some geometries don't have your detector. You just need some sort of graceful handling of the exceptions that naturally arise in that situation.

This time let's wait on benchmark completion before merging.

sebouh137:
> Thanks. I think the real issue might be that some geometries don't have your detector. You just need some sort of graceful handling of the exceptions that naturally arise in that situation.
> 
> This time let's wait on benchmark completion before merging.

Thanks @veprbl .  The only jobs failing are the clean_docker:pipeline-new and one of the sim:backwards_ecal jobs (which are marked as "allowed to fail") the ZDC pi0 (which fails due to geometry incompatibility), and the campaign, which we have already discussed as being unrelated.  

Let me know if there's anything else you need on my end.

veprbl (approved):

veprbl:
I reviewed the benchmark failures, they look good. Let's look at the epic geometry change next.

sebouh137:
> I reviewed the benchmark failures, they look good. Let's look at the epic geometry change next.

Thanks!   I'll rerun the checks for epic now.  https://github.com/eic/epic/pull/1070

PR #2622, opened 2026-04-17
Add multi-calorimeter reconstruction for Lambdas

### LINKED ISSUE #2398: Enhanced reconstruction of Lambdas
### Motivation

The Meson Structure Function WG proposes the integration of an updated Lambda reconstruction algorithm into EICrecon, particularly to enable realistic projections of future kaon structure measurements at the EIC using the Sullivan process: e + p → Λ + K.

This new version builds upon the existing implementation previously merged by Sebouh Paul, which searches for the neutral decay signature of the Lambda as one neutron and two gamma rays detected in the ZDC.

Using realistic Lambdas from event generators, the acceptance obtained with this first implementation is found to be extremely small (close to 0%). This is expected due to the combined effects of the energy and angular distributions of Lambdas produced via the Sullivan process: the probability that all three neutral decay products (1n + 2γ) are simultaneously detected in the ZDC is nearly zero.

These acceptance effects are illustrated in the Meson Structure Function Working Group studies : https://jeffersonlab.github.io/meson-structure/

### Description of the proposed changes

The goal of this modification is to improve the acceptance of Lambdas by including other forward calorimeters, in addition to ZDC, so as to intercept more neutral decay products. The overall reconstruction method remains similar, but neutrons and gamma-rays are also searched in other sub-systems. 

The main improvements are:
- Inclusion of forward calorimeters beyond the ZDC:
    - EcalEndcap
    - LFHCAL
    - B0Ecal
- Significantly improved reconstruction efficiency for Lambdas, especially from realistic event generators.

### Technical scope

The changes are intentionally limited in scope:
- `src/algorithms/reco/FarForwardLambdaReconstruction.cc`
- `src/algorithms/reco/FarForwardNeutralsReconstruction.cc`
All other parts of the codebase remain unchanged.

### Proposed workflow

1. Fork the repository
2. Submit a Pull Request implementing the new algorithm
3. Create a dedicated branch
4. Merge to the main branch

### Tagging

@DraTeots @ruse-traveler

baptistefraisse:
## Summary

This PR extends the far-forward Λ reconstruction from the previous ZDC-only approach to a multi-calorimeter setup. 
This PR is a resubmission of PR #2609  from a branch in the shared repository so the workflows can run properly.

### Main changes
- Use additional calorimeters (B0 ECal, EndcapP ECal, LFHCAL) alongside the ZDC
- Add corresponding neutrals reconstruction
- Integrate everything into the reconstruction chain
- Cleanup following review comments of the closed PR #2609 have been addressed. 

### Motivation

The goal is to improve acceptance and make the far-forward Λ reconstruction usable in a more realistic multi-detector setup.

### Notes

- Rebased on the latest main

### DIFF
--- a/src/algorithms/reco/FarForwardLambdaReconstruction.cc
+++ b/src/algorithms/reco/FarForwardLambdaReconstruction.cc
@@ -1,182 +1,376 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2025 Sebouh Paul
+// Update/modification 2026 by Baptiste Fraisse
 
 #include <Evaluator/DD4hepUnits.h>
 #include <TVector3.h>
 #include <edm4eic/ReconstructedParticleCollection.h>
 #include <edm4hep/Vector3f.h>
-#include <edm4hep/utils/vector_utils.h>
 #include <fmt/core.h>
 #include <cmath>
 #include <cstddef>
 #include <gsl/pointers>
 #include <stdexcept>
 #include <vector>
+#include <type_traits>
 
 #include "FarForwardLambdaReconstruction.h"
 #include "TLorentzVector.h"
-/**
-Creates Lambda candidates from a neutron and two photons from a pi0 decay
- */
 
 namespace eicrecon {
 
-void FarForwardLambdaReconstruction::init() {
+// helpers 
 
+static inline void toTVector3(TVector3& v1, const edm4hep::Vector3f& v2) {
+  v1.SetXYZ(v2.x, v2.y, v2.z);
+}
+
+void FarForwardLambdaReconstruction::init() {
   try {
     m_zMax = m_detector->constant<double>(m_cfg.offsetPositionName);
   } catch (std::runtime_error&) {
-    m_zMax = 35800; // default value
-    trace("Failed to get {} from the detector, using default value of {}", m_cfg.offsetPositionName,
-          m_zMax);
+    m_zMax = 35800;
+    trace("Failed to get {} from the detector, using default value of {}", m_cfg.offsetPositionName, m_zMax);
   }
 }
 
-/* converts one type of vector format to another */
-void toTVector3(TVector3& v1, const edm4hep::Vector3f& v2) { v1.SetXYZ(v2.x, v2.y, v2.z); }
+// reconstruction machinery from n+g+g
 
-void FarForwardLambdaReconstruction::process(
-    const FarForwardLambdaReconstruction::Input& input,
-    const FarForwardLambdaReconstruction::Output& output) const {
-  const auto [neutrals]                                = input;
-  auto [out_lambdas, out_decay_products]               = output;
-  std::vector<edm4eic::ReconstructedParticle> neutrons = {};
-  std::vector<edm4eic::ReconstructedParticle> gammas   = {};
-  for (auto part : *neutrals) {
-    if (part.getPDG() == 2112) {
-      neutrons.push_back(part);
-    }
-    if (part.getPDG() == 22) {
-      gammas.push_back(part);
-    }
-  }
-
-  if (neutrons.empty() || gammas.size() < 2) {
-    return;
-  }
-
+bool FarForwardLambdaReconstruction::reconstruct_from_triplet(
+    const edm4eic::ReconstructedParticle& n_in,
+    const edm4eic::ReconstructedParticle& g1_in,
+    const edm4eic::ReconstructedParticle& g2_in,
+    edm4eic::ReconstructedParticleCollection* out_lambdas,
+    edm4eic::ReconstructedParticleCollection* out_decay_products) const
+{
   static const double m_neutron = m_particleSvc.particle(2112).mass;
   static const double m_pi0     = m_particleSvc.particle(111).mass;
   static const double m_lambda  = m_particleSvc.particle(3122).mass;
 
-  for (std::size_t i_n = 0; i_n < neutrons.size(); i_n++) {
-    for (std::size_t i_1 = 0; i_1 < gammas.size() - 1; i_1++) {
-      for (std::size_t i_2 = i_1 + 1; i_2 < gammas.size(); i_2++) {
+  const double En = n_in.getEnergy();
+  const double E1 = g1_in.getEnergy();
+  const double E2 = g2_in.getEnergy();
 
-        double En = neutrons[i_n].getEnergy();
-        double pn = sqrt(En * En - m_neutron * m_neutron);
-        double E1 = gammas[i_1].getEnergy();
-        double E2 = gammas[i_2].getEnergy();
-        TVector3 xn;
-        TVector3 x1;
-        TVector3 x2;
+  if (En <= m_neutron) return false;
+  const double pn = std::sqrt(std::max(0.0, En*En - m_neutron*m_neutron));
 
-        toTVector3(xn, neutrons[0].getReferencePoint() * dd4hep::mm);
-        toTVector3(x1, gammas[0].getReferencePoint() * dd4hep::mm);
-        toTVector3(x2, gammas[1].getReferencePoint() * dd4hep::mm);
+  TVector3 xn, x1, x2;
 
-        xn.RotateY(-m_cfg.globalToProtonRotation);
-        x1.RotateY(-m_cfg.globalToProtonRotation);
-        x2.RotateY(-m_cfg.globalToProtonRotation);
+  const auto& rn  = n_in.getReferencePoint();
+  const auto& rg1 = g1_in.getReferencePoint();
+  const auto& rg2 = g2_in.getReferencePoint();
 
-        debug("nx recon = {}, g1x recon = {}, g2x recon = {}", xn.X(), x1.X(), x2.X());
-        debug("nz recon = {}, g1z recon = {}, g2z recon = {}, z face = {}", xn.Z(), x1.Z(), x2.Z(),
-              m_zMax);
+  xn.SetXYZ(rn.x * dd4hep::mm,
+            rn.y * dd4hep::mm,
+            rn.z * dd4hep::mm);
 
-        TVector3 vtx(0, 0, 0);
-        double f                   = 0;
-        double df                  = 0.5;
-        double theta_open_expected = 2 * asin(m_pi0 / (2 * sqrt(E1 * E2)));
-        TLorentzVector n;
-        TLorentzVector g1;
-        TLorentzVector g2;
-        TLorentzVector lambda;
-        for (int i = 0; i < m_cfg.iterations; i++) {
-          n                 = {pn * (xn - vtx).Unit(), En};
-          g1                = {E1 * (x1 - vtx).Unit(), E1};
-          g2                = {E2 * (x2 - vtx).Unit(), E2};
-          double theta_open = g1.Angle(g2.Vect());
-          lambda            = n + g1 + g2;
-          if (theta_open > theta_open_expected) {
-            f -= df;
-          } else if (theta_open < theta_open_expected) {
-            f += df;
-          }
+  x1.SetXYZ(rg1.x * dd4hep::mm,
+            rg1.y * dd4hep::mm,
+            rg1.z * dd4hep::mm);
 
-          vtx = lambda.Vect() * (f * m_zMax / lambda.Z());
-          df /= 2;
-        }
+  x2.SetXYZ(rg2.x * dd4hep::mm,
+            rg2.y * dd4hep::mm,
+            rg2.z * dd4hep::mm);
 
-        double mass_rec = lambda.M();
-        if (std::abs(mass_rec - m_lambda) > m_cfg.lambdaMaxMassDev) {
-          continue;
-        }
+  xn.RotateY(-m_cfg.globalToProtonRotation);
+  x1.RotateY(-m_cfg.globalToProtonRotation);
+  x2.RotateY(-m_cfg.globalToProtonRotation);
 
-        // rotate everything back to the lab coordinates.
-        vtx.RotateY(m_cfg.globalToProtonRotation);
-        lambda.RotateY(m_cfg.globalToProtonRotation);
-        n.RotateY(m_cfg.globalToProtonRotation);
-        g1.RotateY(m_cfg.globalToProtonRotation);
-        g2.RotateY(m_cfg.globalToProtonRotation);
+  TVector3 vtx(0, 0, 0);
+  double f  = 0.0;
+  double df = 0.5;
 
-        auto b = -lambda.BoostVector();
-        n.Boost(b);
-        g1.Boost(b);
-        g2.Boost(b);
+  const double theta_open_expected = 2 * std::asin(m_pi0 / (2 * std::sqrt(E1 * E2)));
 
-        //convert vertex to mm:
-        vtx = vtx * (1 / dd4hep::mm);
+  TLorentzVector n, g1, g2, lambda;
 
-        auto rec_lambda = out_lambdas->create();
-        rec_lambda.setPDG(3122);
+  for (int i = 0; i < m_cfg.iterations; i++) {
+    n      = { pn * (xn - vtx).Unit(), En };
+    g1     = { E1 * (x1 - vtx).Unit(), E1 };
+    g2     = { E2 * (x2 - vtx).Unit(), E2 };
+    lambda = n + g1 + g2;
 
-        rec_lambda.setEnergy(lambda.E());
-        rec_lambda.setMomentum({static_cast<float>(lambda.X()), static_cast<float>(lambda.Y()),
-                                static_cast<float>(lambda.Z())});
-        rec_lambda.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()),
-                                      static_cast<float>(vtx.Z())});
-        rec_lambda.setCharge(0);
-        rec_lambda.setMass(mass_rec);
+    const double theta_open = g1.Angle(g2.Vect());
+    if (theta_open > theta_open_expected)      f -= df;
+    else if (theta_open < theta_open_expected) f += df;
 
-        auto neutron_cm = out_decay_products->create();
-        neutron_cm.setPDG(2112);
-        neutron_cm.setEnergy(n.E());
-        neutron_cm.setMomentum(
-            {static_cast<float>(n.X()), static_cast<float>(n.Y()), static_cast<float>(n.Z())});
-        neutron_cm.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()),
-                                      static_cast<float>(vtx.Z())});
-        neutron_cm.setCharge(0);
-        neutron_cm.setMass(m_neutron);
-        //link the reconstructed lambda to the input neutron,
-        // and the cm neutron to the input neutron
-        rec_lambda.addToParticles(neutrons[i_n]);
-        neutron_cm.addToParticles(neutrons[i_n]);
+    vtx = lambda.Vect() * (f * m_zMax / lambda.Z());
+    df /= 2.0;
+  }
 
-        auto gamma1_cm = out_decay_products->create();
-        gamma1_cm.setPDG(22);
-        gamma1_cm.setEnergy(g1.E());
-        gamma1_cm.setMomentum(
-            {static_cast<float>(g1.X()), static_cast<float>(g1.Y()), static_cast<float>(g1.Z())});
-        gamma1_cm.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()),
-                                     static_cast<float>(vtx.Z())});
-        gamma1_cm.setCharge(0);
-        gamma1_cm.setMass(0);
-        rec_lambda.addToParticles(gammas[i_1]);
-        gamma1_cm.addToParticles(gammas[i_1]);
+  const double mass_rec = lambda.M();
+  if (std::abs(mass_rec - m_lambda) > m_cfg.lambdaMassWindow) return false;
 
-        auto gamma2_cm = out_decay_products->create();
-        gamma2_cm.setPDG(22);
-        gamma2_cm.setEnergy(g2.E());
-        gamma2_cm.setMomentum(
-            {static_cast<float>(g2.X()), static_cast<float>(g2.Y()), static_cast<float>(g2.Z())});
-        gamma2_cm.setReferencePoint({static_cast<float>(vtx.X()), static_cast<float>(vtx.Y()),
-                                     static_cast<float>(vtx.Z())});
-        gamma2_cm.setCharge(0);
-        gamma2_cm.setMass(0);
-        rec_lambda.addToParticles(gammas[i_2]);
-        gamma2_cm.addToParticles(gammas[i_2]);
-      }
+  // rotate everything back to lab coordinates
+  vtx.RotateY(m_cfg.globalToProtonRotation);
+  lambda.RotateY(m_cfg.globalToProtonRotation);
+  n.RotateY(m_cfg.globalToProtonRotation);
+  g1.RotateY(m_cfg.globalToProtonRotation);
+  g2.RotateY(m_cfg.globalToProtonRotation);
+
+  // boost decay products to Lambda rest frame
+  auto b = -lambda.BoostVector();
+  n.Boost(b);
+  g1.Boost(b);
+  g2.Boost(b);
+
+  // vertex back to EDM units (mm)
+  vtx = vtx * (1.0 / dd4hep::mm);
+
+  // saving reco lambda
+  auto rec_lambda = out_lambdas->create();
+  rec_lambda.setPDG(3122);
+  rec_lambda.setEnergy(lambda.E());
+  rec_lambda.setMomentum({static_cast<float>(lambda.X()),
+                          static_cast<float>(lambda.Y()),
+                          static_cast<float>(lambda.Z())});
+  rec_lambda.setReferencePoint({static_cast<float>(vtx.X()),
+                                static_cast<float>(vtx.Y()),
+                                static_cast<float>(vtx.Z())});
+  rec_lambda.setCharge(0);
+  rec_lambda.setMass(mass_rec);
+
+  // --- cm neutron
+  auto neutron_cm = out_decay_products->create();
+  neutron_cm.setPDG(2112);
+  neutron_cm.setEnergy(n.E());
+  neutron_cm.setMomentum({static_cast<float>(n.X()),
+                          static_cast<float>(n.Y()),
+                          static_cast<float>(n.Z())});
+  neutron_cm.setReferencePoint({static_cast<float>(vtx.X()),
+                                static_cast<float>(vtx.Y()),
+                                static_cast<float>(vtx.Z())});
+  neutron_cm.setCharge(0);
+  neutron_cm.setMass(m_neutron);
+
+  // --- cm gammas
+  auto gamma1_cm = out_decay_products->create();
+  gamma1_cm.setPDG(22);
+  gamma1_cm.setEnergy(g1.E());
+  gamma1_cm.setMomentum({static_cast<float>(g1.X()),
+                         static_cast<float>(g1.Y()),
+                         static_cast<float>(g1.Z())});
+  gamma1_cm.setReferencePoint({static_cast<float>(vtx.X()),
+                               static_cast<float>(vtx.Y()),
+                               static_cast<float>(vtx.Z())});
+  gamma1_cm.setCharge(0);
+  gamma1_cm.setMass(0);
+
+  auto gamma2_cm = out_decay_products->create();
+  gamma2_cm.setPDG(22);
+  gamma2_cm.setEnergy(g2.E());
+  gamma2_cm.setMomentum({static_cast<float>(g2.X()),
+                         static_cast<float>(g2.Y()),
+                         static_cast<float>(g2.Z())});
+  gamma2_cm.setReferencePoint({static_cast<float>(vtx.X()),
+                               static_cast<float>(vtx.Y()),
+                               static_cast<float>(vtx.Z())});
+  gamma2_cm.setCharge(0);
+  gamma2_cm.setMass(0);
+
+  // --- links
+  rec_lambda.addToParticles(n_in);
+  rec_lambda.addToParticles(g1_in);
+  rec_lambda.addToParticles(g2_in);
+
+  neutron_cm.addToParticles(n_in);
+  gamma1_cm.addToParticles(g1_in);
+  gamma2_cm.addToParticles(g2_in);
+
+  return true;
+}
+
+// selection of the best triplets n+g+g
+
+void FarForwardLambdaReconstruction::process(
+    const FarForwardLambdaReconstruction::Input& input,
+    const FarForwardLambdaReconstruction::Output& output) const
+{
+  const auto [neutralsHcal, neutralsB0, neutralsEcalEndcapP, neutralsLFHCAL] = input;
+  auto [out_lambdas, out_decay_products] = output;
+
+  const double m_pi0    = m_particleSvc.particle(111).mass;
+  const double m_lambda = m_particleSvc.particle(3122).mass;
+
+  // (A) collect by detector category
+  std::vector<edm4eic::ReconstructedParticle> neutrons_zdc, neutrons_other;
+  std::vector<edm4eic::ReconstructedParticle> gammas_zdc, gammas_other;
+
+  // ZDC Hcal
+  for (auto part : *neutralsHcal) {
+    if (part.getPDG() == 2112) neutrons_zdc.push_back(part);
+    if (part.getPDG() == 22)   gammas_zdc.push_back(part);
+  }
+
+  // B0 Ecal photons
+  for (auto part : *neutralsB0) {
+    if (part.getPDG() == 22) gammas_other.push_back(part);
+  }
+
+  // EndcapP Ecal photons
+  for (auto part : *neutralsEcalEndcapP) {
+    if (part.getPDG() == 22) gammas_other.push_back(part);
+  }
+
+  // LFHCAL neutrons
+  for (auto part : *neutralsLFHCAL) {
+    if (part.getPDG() == 2112) neutrons_other.push_back(part);
+  }
+
+  if (neutrons_zdc.empty() && neutrons_other.empty()) return;
+
+  // gamma pool
+  using GammaT = typename std::decay_t<decltype(gammas_zdc)>::value_type;
+  std::vector<GammaT> gamma_pool;
+  std::vector<uint8_t> gamma_is_zdc;
+
+  gamma_pool.reserve(gammas_zdc.size() + gammas_other.size());
+  gamma_is_zdc.reserve(gammas_zdc.size() + gammas_other.size());
+
+  for (const auto& g : gammas_zdc)   { gamma_pool.push_back(g); gamma_is_zdc.push_back(1); }
+  for (const auto& g : gammas_other) { gamma_pool.push_back(g); gamma_is_zdc.push_back(0); }
+
+  if (gamma_pool.size() < 2) return;
+
+  // invariant mass helpers (ranking only)
+
+  auto invMass2 = [](const edm4eic::ReconstructedParticle& a,
+                     const edm4eic::ReconstructedParticle& b) {
+    const auto pa = a.getMomentum();
+    const auto pb = b.getMomentum();
+    const double Ea = a.getEnergy();
+    const double Eb = b.getEnergy();
+    const double px = pa.x + pb.x;
+    const double py = pa.y + pb.y;
+    const double pz = pa.z + pb.z;
+    const double E  = Ea + Eb;
+    return E*E - (px*px + py*py + pz*pz);
+  };
+
+  auto invMass2_3 = [](const edm4eic::ReconstructedParticle& a,
+                       const edm4eic::ReconstructedParticle& b,
+                       const edm4eic::ReconstructedParticle& c) {
+    const auto pa = a.getMomentum();
+    const auto pb = b.getMomentum();
+    const auto pc = c.getMomentum();
+    const double Ea = a.getEnergy();
+    const double Eb = b.getEnergy();
+    const double Ec = c.getEnergy();
+    const double px = pa.x + pb.x + pc.x;
+    const double py = pa.y + pb.y + pc.y;
+    const double pz = pa.z + pb.z + pc.z;
+    const double E  = Ea + Eb + Ec;
+    return E*E - (px*px + py*py + pz*pz);
+  };
+
+  // (1) pi0 candidates
+  struct Pi0Pair {
+    int i, j;
+    int cat;             // 0:2ZDC, 1:1ZDC, 2:0ZDC
+    double dmpi0_cand;
+  };
+
+  std::vector<Pi0Pair> pi0_pairs;
+  pi0_pairs.reserve(gamma_pool.size() * (gamma_pool.size() - 1) / 2);
+
+  for (size_t i = 0; i < gamma_pool.size(); ++i) {
+    for (size_t j = i + 1; j < gamma_pool.size(); ++j) {
+      const auto& g1 = gamma_pool[i];
+      const auto& g2 = gamma_pool[j];
+
+      const double m2 = invMass2(g1, g2);
+      if (m2 <= 0.0) continue;
+
+      const double m  = std::sqrt(m2);
+      const double dm = std::abs(m - m_pi0);
+      if (dm > m_cfg.pi0Window * m_pi0) continue;
+
+      const bool z1 = gamma_is_zdc[i];
+      const bool z2 = gamma_is_zdc[j];
+      const int cat = (z1 && z2) ? 0 : ((z1 || z2) ? 1 : 2);
+
+      pi0_pairs.push_back({(int)i, (int)j, cat, dm});
     }
   }
+  if (pi0_pairs.empty()) return;
+
+  // (2) Lambda candidates pool
+  struct LambdaCand {
+    int g_i, g_j;
+    int n_idx;
+    int n_cat;   // 0: ZDC neutron, 1: other neutron
+    int g_cat;   // 0:2ZDC, 1:1ZDC, 2:0ZDC
+    double chi2;
+    double E;
+    double pz;
+  };
+
+  std::vector<LambdaCand> cands;
+  cands.reserve(64);
+
+  auto try_neutrons = [&](const auto& neutron_list, int n_cat) {
+    for (const auto& pp : pi0_pairs) {
+      const auto& g1 = gamma_pool[pp.i];
+      const auto& g2 = gamma_pool[pp.j];
+
+      for (size_t in = 0; in < neutron_list.size(); ++in) {
+        const auto& n = neutron_list[in];
+
+        const double m2L = invMass2_3(n, g1, g2);
+        if (m2L <= 0.0) continue;
+
+        const double mL = std::sqrt(m2L);
+        if (std::abs(mL - m_lambda) > m_cfg.lambdaMassWindow * m_lambda) continue;
+
+        const double dL = (mL - m_lambda);
+
+        // same score as your new code
+        const double chi2 =
+          (pp.dmpi0_cand / (m_cfg.pi0Window * m_pi0)) * (pp.dmpi0_cand / (m_cfg.pi0Window * m_pi0)) +
+          (dL / (m_cfg.lambdaMassWindow * m_lambda)) * (dL / (m_cfg.lambdaMassWindow * m_lambda));
+
+        const double E  = n.getEnergy() + g1.getEnergy() + g2.getEnergy();
+
+        const auto pn  = n.getMomentum();
+        const auto pg1 = g1.getMomentum();
+        const auto pg2 = g2.getMomentum();
+
+        const double px = pn.x + pg1.x + pg2.x;
+        const double py = pn.y + pg1.y + pg2.y;
+        const double pz = pn.z + pg1.z + pg2.z;
+
+        cands.push_back({pp.i, pp.j, (int)in, n_cat, pp.cat, chi2, E, pz});
+      }
+    }
+  };
+
+  if (!neutrons_zdc.empty())   try_neutrons(neutrons_zdc,   0);
+  if (!neutrons_other.empty()) try_neutrons(neutrons_other, 1);
+
+  if (cands.empty()) return;
+
+  // (3) best candidate selection (your policy)
+  auto better = [&](const LambdaCand& a, const LambdaCand& b) -> bool {
+    if (a.n_cat != b.n_cat) return a.n_cat < b.n_cat;  // prefer ZDC neutron
+    if (a.g_cat != b.g_cat) return a.g_cat < b.g_cat;  // prefer 2ZDC gammas
+    if (a.chi2  != b.chi2)  return a.chi2  < b.chi2;
+    if (a.pz    != b.pz)    return a.pz > b.pz;
+    return a.E > b.E;
+  };
+
+  int best_k = 0;
+  for (int k = 1; k < (int)cands.size(); ++k) {
+    if (better(cands[k], cands[best_k])) best_k = k;
+  }
+
+  const auto& best = cands[best_k];
+
+  const auto& g1 = gamma_pool[best.g_i];
+  const auto& g2 = gamma_pool[best.g_j];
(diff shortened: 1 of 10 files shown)


src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -558,9 +558,12 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
...
       "HcalFarForwardZDCTruthClusterAssociations",
-      "ReconstructedFarForwardZDCNeutrals",
-      "ReconstructedFarForwardZDCLambdas",
-      "ReconstructedFarForwardZDCLambdaDecayProductsCM",
+      "ReconstructedHcalFarForwardZDCNeutrals",
+      "ReconstructedB0EcalNeutrals",
+      "ReconstructedEcalEndcapPNeutrals",
+      "ReconstructedLFHCALNeutrals",
+      "ReconstructedFarForwardLambdas",
+      "ReconstructedFarForwardLambdaDecayProductsCM",
```
veprbl: EcalEndcapP/LFHCAL is not FarForward. If we do a rename, at this point should we call this `ReconstructedLambdas`? Is this reasonably unspecific of a name?

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
- */
-
 namespace eicrecon {
 
 void FarForwardNeutralsReconstruction::init() {
 
   try {
     m_gammaZMax =
-        m_cfg.gammaZMaxOffset + m_detector->constant<double>(m_cfg.offsetPositionName) / dd4hep::mm;
+        400 * dd4hep::mm + m_detector->constant<double>(m_cfg.offsetPositionName) / dd4hep::mm;
```
veprbl:
```suggestion
        m_cfg.gammaZMaxOffset + m_detector->constant<double>(m_cfg.offsetPositionName) / dd4hep::mm;
```

src/global/reco/reco.cc
```diff
@@ -216,27 +216,35 @@ void InitPlugin(JApplication* app) {
...
-      {"HcalFarForwardZDCClusters"},          // edm4eic::ClusterCollection
-      {"ReconstructedFarForwardZDCNeutrals"}, // edm4eic::ReconstrutedParticleCollection,
-      {.offsetPositionName        = "HcalFarForwardZDC_SiPMonTile_r_pos",
-       .neutronScaleCorrCoeffHcal = {-0.11, -1.5, 0},
-       .gammaScaleCorrCoeffHcal   = {0, -.13, 0},
-       .globalToProtonRotation    = -0.025,
-       .gammaZMaxOffset           = 300 * dd4hep::mm,
-       .gammaMaxLength            = 100 * dd4hep::mm,
-       .gammaMaxWidth             = 27 * dd4hep::mm},
+      "ReconstructedFarForwardZDCNeutrals",
```
veprbl:
```suggestion
      "ReconstructedFarForwardNeutrals",
```
(In the current naming)

src/global/reco/reco.cc
```diff
@@ -216,27 +216,35 @@ void InitPlugin(JApplication* app) {
...
+       "ReconstructedEcalEndcapPNeutrals", "ReconstructedLFHCALNeutrals"},
+      {.offsetPositionName               = "HcalFarForwardZDC_SiPMonTile_r_pos",
+       .neutronScaleCorrCoeffHcal        = {2.4, 0.89},
+       .gammaScaleCorrCoeffHcal          = {1.1, 0.98},
+       .neutronScaleCorrCoeffLFHCAL      = {2.55, 0.95},
+       .gammaScaleCorrCoeffLFHCAL        = {0., 0.},
+       .neutronScaleCorrCoeffB0Ecal      = {0., 0.},
+       .gammaScaleCorrCoeffB0Ecal        = {0.99, 1.14},
+       .neutronScaleCorrCoeffEcalEndcapP = {0., 0.},
+       .gammaScaleCorrCoeffEcalEndcapP   = {1.05, 1.01},
```
veprbl: This hardcodes 4 detectors, could we just have used 4 instances of the factory/algorithm?

My main concern is that with this change the https://github.com/eic/detector_benchmarks/tree/master/benchmarks/zdc_lambda benchmark will be broken. What is the path we should be taking? Do we need to recover ZDC-only lambda reconstruction to keep ZDC performance studied in isolation? Or the benchmark can use the new collection, but apply a selection cut?
cc @sebouh137

src/global/reco/reco.cc
```diff
@@ -216,27 +216,35 @@ void InitPlugin(JApplication* app) {
...
-       .gammaZMaxOffset           = 300 * dd4hep::mm,
-       .gammaMaxLength            = 100 * dd4hep::mm,
-       .gammaMaxWidth             = 27 * dd4hep::mm},
+      "ReconstructedFarForwardZDCNeutrals",
+      {"HcalFarForwardZDCClusters", "B0ECalClusters", "EcalEndcapPClusters", "LFHCALClusters"},
+      {"ReconstructedHcalFarForwardZDCNeutrals", "ReconstructedB0EcalNeutrals",
+       "ReconstructedEcalEndcapPNeutrals", "ReconstructedLFHCALNeutrals"},
+      {.offsetPositionName               = "HcalFarForwardZDC_SiPMonTile_r_pos",
+       .neutronScaleCorrCoeffHcal        = {2.4, 0.89},
+       .gammaScaleCorrCoeffHcal          = {1.1, 0.98},
```
veprbl:
```suggestion
       .neutronScaleCorrCoeffZDC        = {2.4, 0.89},
       .gammaScaleCorrCoeffZDC          = {1.1, 0.98},
```
If we keep 4 detectors?

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+  n_neutrons += processNeutralCalo(clustersLFHCAL, out_neutralsLFHCAL,
+                                   /*gammaCorrCoefs=*/m_cfg.gammaScaleCorrCoeffLFHCAL,
+                                   /*neutronCorrCoefs=*/m_cfg.neutronScaleCorrCoeffLFHCAL,
+                                   /*canDetectGammas=*/false,
+                                   /*canDetectNeutrons=*/true,
+                                   /*gammaCorr=*/corr_power,
+                                   /*neutronCorr=*/corr_power,
+                                   /*gammaMode=*/GammaMode::None,
+                                   /*gammaLeaderFracMin=*/0.0,
+                                   /*clusterEmin=*/7.0,
```
veprbl: What are the units for clusterEmin?
```suggestion
                                   /*clusterEmin=*/7.0 * edm4eic::unit::GeV,
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+        Esum += E;
+        idx.push_back(i);
+      }
+
+      if (idx.empty() || Esum <= 0.0) {
+        return 0;
+      }
+
+      const size_t Nkeep = 4;
+      for (size_t k = 0; k < std::min(Nkeep, idx.size()); ++k) {
```
veprbl: If clusters are not ordered, what is the effect of this?

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+        }
+        Esum += E;
+        idx.push_back(i);
+      }
+
+      if (idx.empty() || Esum <= 0.0) {
+        return 0;
+      }
+
+      const size_t Nkeep = 4;
```
veprbl: What is the reason to use 4 here? Could you, please, add a code comment for that question.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
-                             "have at least 3 parameters");
-  }
   trace("gamma detection params:   max length={},   max width={},   max z={}", m_cfg.gammaMaxLength,
         m_cfg.gammaMaxWidth, m_gammaZMax);
 }
-/** calculates the correction for a given uncorrected total energy and a set of coefficients*/
+
 double FarForwardNeutralsReconstruction::calc_corr(double Etot, const std::vector<double>& coeffs) {
   return coeffs[0] + coeffs[1] / sqrt(Etot) + coeffs[2] / Etot;
 }
```
veprbl:
```suggestion
```
Can be removed now.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+      const auto pos = cl.getPosition();
+      const double E = gammaCorr(cl.getEnergy(), gammaScaleCoeff);
+
+      const double r = edm4hep::utils::magnitude(pos);
+      if (r > 0) {
+        rec.setMomentum(pos * (E / r));
+      }
+
+      rec.setEnergy(E);
+      rec.setReferencePoint(pos);
```
veprbl: I believe reference point is meant to be the origin point for momentum. If you select $\hat{r} * E$ or $\hat{r} * |\vec{p}|$, that implies you used {0,0,0} as the reference point and not $\vec{r}$ (which has no physics meaning).

veprbl (commented):
This is a great addition and very nice work. I have some comments after reviewing. Please feel free to push back if things are out of scope for this change.

This review is based on output of an experimental AI tool https://epic-devcloud.org/doc/page/186f46b1-581d-4b2e-8416-a510abac0de7/. The suggestions below are by me even though they are inspired.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
veprbl: The docstring is ambiguous
https://github.com/eic/EDM4eic/blob/7d7705d9bbd4794b1b875c04258e7d986aea8a47/edm4eic.yaml#L259
@ruse-traveler Do you agree with my interpretation?

ruse-traveler: Agreed... It definitely is ambiguous :/

Using it as the origin for the momentum measurement makes sense to me (esp. since we have an explicit `startVertex`). But -- given that EDM4hep updated their `ReconstructedParticle` to drop `startVertex` in favor of `decayVertex` -- I could _also_ see reading it as the production vertex...

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+      if (desc.use_gamma && part.getPDG() == 22) {
+        (desc.is_zdc ? gammas_zdc : gammas_other).push_back(part);
+      }
     }
-    if (part.getPDG() == 22) {
-      gammas.push_back(part);
+  }
+
+  if (neutrons_zdc.empty() && neutrons_other.empty()) {
+    return;
```
ruse-traveler, line 228: Could be useful to pepper in some debugging statements! e.g.
```suggestion
    debug("No neutrons in input collection");
    return;
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+    gamma_pool.push_back(g);
+    gamma_is_zdc.push_back(1);
+  }
+  for (const auto& g : gammas_other) {
+    gamma_pool.push_back(g);
+    gamma_is_zdc.push_back(0);
+  }
+
+  if (gamma_pool.size() < 2) {
+    return;
```
ruse-traveler, line 255: Another good debug spot!
```suggestion
    debug("Less than 2 gammas found ({})", gamma_pool.size());
    return;
```

src/global/reco/reco.cc
DraTeots: > This hardcodes 4 detectors, could we just have used 4 instances of the factory/algorithm?
> 
> My main concern is that with this change the https://github.com/eic/detector_benchmarks/tree/master/benchmarks/zdc_lambda benchmark will be broken. What is the path we should be taking? Do we need to recover ZDC-only lambda reconstruction to keep ZDC performance studied in isolation? Or the benchmark can use the new collection, but apply a selection cut? cc @sebouh137

I worked with zdc_lambda benchmark. The caveat is that it is very tightly coupled with the idea of generating lambdas via particle gun for different momentums and then building plots foe each particular slot. My attempts to make it "general" so that it can process our EG based data lead to rewriting the scripts. Then @baptistefraisse wrote a thorough framework to analyze his work on lambda reconstruction  by multiple calorimeters. This includes the dedicated case where everything is captured by ZDC only  - doing the same as zdc_lambda_benchmark. Previously @sebouh137 told me he will likely not be working on this anymore, i.e. the current zdc_lambda benchmark is without the maintainer (unless now Paul will write otherwise). 

TL;DR; I propose to replace the framework with Baptiste analyses (or both). 

P.S. Definitely @sebouh137 credits for initial ZDC lambdas will be saved.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+  std::vector<edm4eic::ReconstructedParticle> neutrons_other;
+  std::vector<edm4eic::ReconstructedParticle> gammas_zdc;
+  std::vector<edm4eic::ReconstructedParticle> gammas_other;
+
+  struct NeutralInputDesc {
+    const edm4eic::ReconstructedParticleCollection* coll;
+    bool use_gamma;
+    bool use_neutron;
+    bool is_zdc;
+  };
```
ruse-traveler: Just for clarity, I'd suggest not abbreviating the struct name!
```suggestion
  struct NeutralInputDescription {
    const edm4eic::ReconstructedParticleCollection* coll;
    bool use_gamma;
    bool use_neutron;
    bool is_zdc;
  };
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+
+  // --------------------------------------------------------------------------
+  // Step 2: build a unified photon pool while retaining whether each photon
+  // comes from the ZDC-preferred category.
+  // --------------------------------------------------------------------------
+
+  using ParticleT = edm4eic::ReconstructedParticle;
+
+  std::vector<ParticleT> gamma_pool;
+  std::vector<uint8_t> gamma_is_zdc;
```
ruse-traveler: For my own understanding: what's the reason to prefer `uint8_t` over a `bool` here?
```suggestion
  std::vector<bool> gamma_is_zdc;
```

src/algorithms/reco/FarForwardLambdaReconstruction.h
```diff
@@ -19,18 +19,44 @@
...
+ * - collect photon and neutron candidates from the configured far-forward
+ *   neutral collections and split them into detector categories;
+ * - build pi0 candidates from all photon pairs passing the configured pi0
+ *   invariant-mass window;
+ * - combine accepted pi0 candidates with neutron candidates to form Lambda
+ *   candidates passing the configured Lambda invariant-mass window;
+ * - rank the surviving candidates according to detector-preference and
+ *   mass-compatibility criteria;
+ * - run the final Lambda reconstruction on the best candidate and store the
+ *   resulting Lambda and decay products.
```
ruse-traveler: Super minor point for clarity, I'd suggest an ordered list here!
```suggestion
 * 1. collect photon and neutron candidates from the configured far-forward
 *    neutral collections and split them into detector categories;
 * 2. build pi0 candidates from all photon pairs passing the configured pi0
 *    invariant-mass window;
 * 3. combine accepted pi0 candidates with neutron candidates to form Lambda
 *    candidates passing the configured Lambda invariant-mass window;
 * 4. rank the surviving candidates according to detector-preference and
 *    mass-compatibility criteria;
 * 5. run the final Lambda reconstruction on the best candidate and store the
 *    resulting Lambda and decay products.
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+    int n_idx;
+    NeutronCategory n_cat;
+    GammaCategory g_cat;
+    double chi2;
+    double E;
+    double pz;
+  };
+
+  std::vector<LambdaCand> cands;
+  cands.reserve(64);
```
ruse-traveler: I'd suggest a short note here indicating where the 64 comes from!
```suggestion
  cands.reserve(64);  // NB <= 4^3 candidates when searching for 3 daughters across 4 calos
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+                                : (z1 || z2) ? GammaCategory::OneZDC
+                                             : GammaCategory::ZeroZDC;
+
+      pi0_pairs.push_back({static_cast<int>(i), static_cast<int>(j), cat, dm});
     }
   }
 
-  if (neutrons.empty() || gammas.size() < 2) {
+  if (pi0_pairs.empty()) {
     return;
```
ruse-traveler, line 338: Another good debug spot!
```suggestion
    debug("No pi0s candidates found");
    return;
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
-  static const double m_neutron = m_particleSvc.particle(2112).mass;
-  static const double m_pi0     = m_particleSvc.particle(111).mass;
-  static const double m_lambda  = m_particleSvc.particle(3122).mass;
+  // --------------------------------------------------------------------------
+  // Step 4: build Lambda candidate pool from pi0-neutron combinations
+  // --------------------------------------------------------------------------
+
+  enum class NeutronCategory : uint8_t { ZDC = 0, Other = 1 };
+
+  struct LambdaCand {
```
ruse-traveler: Like above, I'd suggest avoiding abbreviations in struct names just for clarity!
```suggestion
  struct LambdaCandidate {
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+
+  rec_lambda.addToParticles(n_in);
+  rec_lambda.addToParticles(g1_in);
+  rec_lambda.addToParticles(g2_in);
+
+  neutron_cm.addToParticles(n_in);
+  gamma1_cm.addToParticles(g1_in);
+  gamma2_cm.addToParticles(g2_in);
+
+  return true;
```
ruse-traveler, line 174: Another good debugging spot!
```suggestion
  debug("Reconstructed lambda: E = {}, m = {}", rec_lambda.getEnergy(), rec_lambda.getMass());
  return true;
```

src/algorithms/reco/FarForwardLambdaReconstruction.cc
```diff
@@ -1,182 +1,446 @@
...
+  if (!neutrons_zdc.empty()) {
+    try_neutrons(neutrons_zdc, NeutronCategory::ZDC);
+  }
+  if (!neutrons_other.empty()) {
+    try_neutrons(neutrons_other, NeutronCategory::Other);
+  }
+
+  if (cands.empty()) {
+    return;
   }
```
ruse-traveler, line 408: Another good debugging spot!
```suggestion
  }
  debug("{} lambda candidates found", cands.size());
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
 namespace eicrecon {
 
 void FarForwardNeutralsReconstruction::init() {
 
   try {
     m_gammaZMax =
-        m_cfg.gammaZMaxOffset + m_detector->constant<double>(m_cfg.offsetPositionName) / dd4hep::mm;
+        400 * dd4hep::mm + m_detector->constant<double>(m_cfg.offsetPositionName) / dd4hep::mm;
   } catch (std::runtime_error&) {
     m_gammaZMax = m_cfg.gammaZMaxOffset + 35800;
```
ruse-traveler: I'd suggest adding a comment to explain where the 35800 comes from!
```suggestion
    m_gammaZMax = m_cfg.gammaZMaxOffset + 35800; // NB ZDC face at 35.8 m from IP
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
-      edm4hep::Vector3f momentum = position * (p / r);
-      rec_part.setEnergy(E);
-      rec_part.setMomentum(momentum);
-      rec_part.setReferencePoint(position);
-      rec_part.setCharge(0);
-      rec_part.setMass(0);
-      rec_part.addToClusters(cluster);
-      continue;
+
+  // Unpacking
```
ruse-traveler: For consideration:
```suggestion
  // Unpacking inputs/outputs
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+  const auto [clustersHcal, clustersB0, clustersEcalEndcapP, clustersLFHCAL]           = input;
+  auto [out_neutralsHcal, out_neutralsB0, out_neutralsEcalEndcapP, out_neutralsLFHCAL] = output;
+
+  // Global
```
ruse-traveler: For consideration:
```suggestion
  // Global constants
```

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+  enum class GammaMode { None, LeaderOnly, AllPassing };
+  enum class NeutronMode { None, SumAll, LeaderOnly };
+
+  using CorrFunc = std::function<double(double, const std::vector<double>&)>;
+
+  CorrFunc corr_power = [](double E, const std::vector<double>& coeffs) {
+    if (coeffs.size() < 2) {
+      return E;
     }
+    return coeffs[0] * std::pow(E, coeffs[1]);
```
ruse-traveler: For my own understanding: is this the functional form we'd expect for the B0, FEMC, and/or LFHCAL? I ask because it gives me a bit of a pause that the functional form is being changed here...

Building off @veprbl's comments below, we could potentially run this algorithm on 4 separate inputs (FEMC, LFHCAL, B0, ZDC) and make the correction a configurable parameter (maybe similar to what was done for the [CoG weighting](https://github.com/eic/EICrecon/blob/31efcb2158fcd1520a09fc93b772727f982808ec/src/algorithms/calorimetry/CalorimeterClusterRecoCoG.h#L48) or as a lambda).

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+  n_neutrons += processNeutralCalo(clustersEcalEndcapP, out_neutralsEcalEndcapP,
+                                   /*gammaCorrCoefs=*/m_cfg.gammaScaleCorrCoeffEcalEndcapP,
+                                   /*neutronCorrCoefs=*/m_cfg.neutronScaleCorrCoeffEcalEndcapP,
+                                   /*canDetectGammas=*/true,
+                                   /*canDetectNeutrons=*/false,
+                                   /*gammaCorr=*/corr_power,
+                                   /*neutronCorr=*/corr_power,
+                                   /*gammaMode=*/GammaMode::LeaderOnly,
+                                   /*gammaLeaderFracMin=*/0.0,
+                                   /*clusterEmin=*/1.0,
```
ruse-traveler: Again building on the idea of running the algorithm on individual collections: `clusterEmin` could be converted to a configurable parameter, which could help with tuning down the road.

src/global/reco/reco.cc
veprbl: Yes, we definitely would like to get @baptistefraisse's benchmark added. We could disable the zdc_lambda benchmark for now (anyone submits a PR to do that?). And I agree, all the nice work needs to be credited appropriately.

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
```diff
@@ -1,148 +1,322 @@
...
+
+  CorrFunc corr_power = [](double E, const std::vector<double>& coeffs) {
+    if (coeffs.size() < 2) {
+      return E;
     }
+    return coeffs[0] * std::pow(E, coeffs[1]);
+  };
+
+  // helper for processing detectors
+  auto processNeutralCalo =
```
ruse-traveler: Another question for my own understanding: is there a reason to make this and some of the other methods below a lambda rather than as members of the class? Algorithmically, everything looks good, but we might be able to do a light reorganization to make the flow of the calculation a little easier to follow...

src/factories/reco/FarForwardLambdaReconstruction_factory.h
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2025 Sebouh Paul
+// Copyright (C) 2025 Sebouh Paul, Baptiste Fraisse
```
ruse-traveler: Copyright year can be updated here!
```suggestion
// Copyright (C) 2026 Sebouh Paul, Baptiste Fraisse
```

src/algorithms/reco/FarForwardNeutralsReconstructionConfig.h
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2025 Sebouh Paul
+// Copyright (C) 2025 Sebouh Paul, Baptiste Fraisse
```
ruse-traveler: Also here!
```suggestion
// Copyright (C) 2026 Sebouh Paul, Baptiste Fraisse
```

src/algorithms/reco/FarForwardNeutralsReconstruction.h
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2025 Sebouh Paul
+// Copyright (C) 2025 Sebouh Paul, Baptiste Fraisse
```
ruse-traveler: And here!
```suggestion
// Copyright (C) 2026 Sebouh Paul, Baptiste Fraisse
```

ruse-traveler (commented):
Hi Baptiste, thanks for integrating the feedback on the previous PR! I think this looks great! I second the suggestions @veprbl left and have some of my own. Overall, I think this is in a great state, and most of my comments are more aimed at clarifying certain points. After this round, I'll be ready to sign-off!

That being said, I do think it's worth strongly considering applying the neutral reconstruction to individual cluster collections rather than all 4 at once. While this won't be a huge change -- largely just updating the inputs/outputs and deduplicating/maybe adding a few parameters -- I do think it will help significantly reduce the complexity of the neutral reconstruction!

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
baptistefraisse: Indeed. I added an explicit sort of the selected cluster indices by decreasing energy before keeping the leading clusters, so the selection is now deterministic and based on cluster energy. Thank you

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
baptistefraisse: Good point, thank you. I added a comment. The purpose of keeping only 4 clusters is to limit the combinatorics of low-energy secondary gammas while keeping leading candidates used for the neutral reconstruction (we are looking for 2 gammas).

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
baptistefraisse: Thanks, I see the ambiguity. In this reconstruction, the referencePoint is currently used in the Lambda reconstruction as the neutral cluster position (neutron and gammas). So changing it to {0,0,0} would break the current Lambda reconstruction logic. I agree that this may not be the ideal semantic use of referencePoint; I can either add a comment clarifying this usage for now, or move the cluster position to a more appropriate field if there is a recommended EDM4eic convention for this?

veprbl: But can't you just access the original cluster position like `part.getCluster(0).getPosition()`?

baptistefraisse: Indeed, I will do that. Thank you

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
baptistefraisse: Good point, I change the structure to make it easier to follow.

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Done

src/algorithms/reco/FarForwardLambdaReconstruction.cc
baptistefraisse: Done

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
veprbl: I was also looking at this one. I concluded that since the old coefficient was 0, so function form did not change. The change into 4 inputs, perhaps, we can do on next iteration.

src/global/reco/reco.cc
veprbl: Looks like the existing benchmark continues to run without failure, producing only partial output. Let's follow up on this in https://github.com/eic/detector_benchmarks/issues/309.

veprbl (approved):
LGTM

src/algorithms/reco/FarForwardNeutralsReconstruction.cc
ruse-traveler: Sounds good!

ruse-traveler (approved):
Thanks for integrating my feedback, Baptiste! Looks good to me!

wdconinc:
This PR has introduced irreproducible behavior in multi-threaded running: there are differences in outcomes depending on how many threads are used. See https://github.com/eic/EICrecon/actions/runs/26528753188/job/78145149238#step:5:202 (the last pipeline in this PR). It also shows up in new PR capybara reports. This needs to get fixed or this PR reverted.

baptistefraisse:
@wdconinc thanks for reporting the irreproductibility issue in multi-threaded running. I started investigating the issue and focused on the ZDC neutral reconstruction, since this is where the CI differences appear.

My current understanding is that the problem originates from the way clusters are classified as photons or neutrons and tracked for the subsequent Lambda reconstruction. I found that this bookkeeping relied on pointer comparisons between PODIO cluster objects, which may not be robust in a multi-threaded environment. I have therefore replaced it with index-based bookkeeping.

To validate the change, I implemented an event-by-event comparison between single and multi-threaded outputs. After matching events using EventHeader.eventNumber, the reconstructed neutral and Lambda collections are identical in my local tests.

I will push the proposed fix and continue validating it, but I would also be interested in your thoughts. I may still be missing part of the picture, especially since I do not have access to the exact reproducibility test used in CI.

wdconinc:
> @wdconinc thanks for reporting the irreproductibility issue in multi-threaded running. I started investigating the issue and focused on the ZDC neutral reconstruction, since this is where the CI differences appear.

Also LFHCAL.

> My current understanding is that the problem originates from the way clusters are classified as photons or neutrons and tracked for the subsequent Lambda reconstruction. I found that this bookkeeping relied on pointer comparisons between PODIO cluster objects, which may not be robust in a multi-threaded environment. I have therefore replaced it with index-based bookkeeping.

Yes but they are pointer equality tests so I dont imagine that affects this. Usually we see issues due to implicit or explicit pointer ordering. Those happen when relying on implicit ordering in std::map of pointers. I didn't see those used though. 

> To validate the change, I implemented an event-by-event comparison between single and multi-threaded outputs. After matching events using EventHeader.eventNumber, the reconstructed neutral and Lambda collections are identical in my local tests.
> 
> I will push the proposed fix and continue validating it, but I would also be interested in your thoughts. I may still be missing part of the picture, especially since I do not have access to the exact reproducibility test used in CI.

The whole point about CI is exactly that everyone has access to the same tests. Can you explain what you don't have access to?

baptistefraisse:
@wdconinc Thanks for the clarification. You are right, I was referring to the exact environment used to produce the failing artifact rather than the comparison test itself.

I downloaded the single-threaded (ST) and multi-threaded (MT) artifacts from the CI workflow, and reproduced the reported differences with Capybara.

I then checked out the merged commit (a48dc93), rebuilt EICrecon locally, and reran the reconstruction from the same DIS input file using 1 and 4 threads in the current eic_xl:nightly container. In this setup I do not reproduce the issue: Capybara reports no differences.

I also compared the downloaded artifacts against my locally regenerated outputs. The locally regenerated ST and MT outputs are consistent with each other, and the locally regenerated MT output agrees with the downloaded ST artifact. The discrepancy only appears in the downloaded MT artifact.

So at this point, I have not yet been able to reproduce the failure by rerunning the reconstruction from source. I am still investigating what could explain that difference. Is there another reproducibility test or environment detail that you think I should check?

wdconinc:
> So at this point, I have not yet been able to reproduce the failure by rerunning the reconstruction from source. I am still investigating what could explain that difference. Is there another reproducibility test or environment detail that you think I should check?

Yeah, that's the tricky part about this: it sometimes depends on the running conditions and where on which thread what runs in what order... I don't have any good advice, other than you can sometimes reproduce it by reducing the algorithm graph and running multiple times.

sebouh137:
@baptistefraisse @wdconinc  I currently have a local workflow for testing the reconstruction of particles in the ZDC that appears to be broken by these recent changes.  It appears as though the far forward neutrals reconstruction crashes for files simulated with the standalone ZDC geometry;  is this correct?  Would I need to rerun my simulations with the full epic geometry, or would just the far forward detectors be sufficient?  Or are there other ways to mitigate this issue?

veprbl:
I don't see why should it crash. Can you provide a log? It shouldn't work if lfhcal, femcal and b0ecal are missing, but not crash.

baptistefraisse:
Thank you for the feedback @sebouh137  ! The standalone ZDC setup may simply miss some of the collections/geometries expected by the current far-forward neutrals reconstruction, but I agree @veprbl it should not crash in that case. Could you please provide the log? Thank you

wdconinc:
> I currently have a local workflow for testing the reconstruction of particles in the ZDC

That sounds like an excellent target to add to a benchmark to ensure we catch any failures early and have access to logs.

sebouh137:
> Thank you for the feedback @sebouh137 ! The standalone ZDC setup may simply miss some of the collections/geometries expected by the current far-forward neutrals reconstruction, but I agree @veprbl it should not crash in that case. Could you please provide the log? Thank you

Exception in JFactory::Create, prefix=eicrecon::CalorimeterHitDigi_factory
Exception in JFactory::Create, prefix=eicrecon::CalorimeterHitReco_factory
Exception in JFactory::Create, prefix=eicrecon::CalorimeterIslandCluster_factory
Exception in JFactory::Create, prefix=eicrecon::CalorimeterClusterRecoCoG_factory
Exception in JFactory::Create, prefix=eicrecon::CalorimeterClusterShape_factory
Exception in JFactory::Create, prefix=eicrecon::FarForwardNeutralsReconstruction_factory
[e:0] [JEventProcessorPODIO] [error] Omitting PODIO collection 'ReconstructedHcalFarForwardZDCNeutrals' due to exception: JOmniFactory: Failed to get collection B0ECalClusters: JOmniFactory: Failed to get collection B0ECalClustersWithoutShapes: JOmniFactory: Failed to get collection B0ECalIslandProtoClusters: JOmniFactory: Failed to get collection B0ECalRecHits: JOmniFactory: Failed to get collection B0ECalRawHits: JOmniFactory: Failed to get collection B0ECalHits: Missing databundle with uniquename 'B0ECalHits'.
10:41:43.339 [fatal] Segfault detected!
10:41:43.340 [fatal] Hard exit due to segmentation fault! Backtrace:
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    0: TStreamerInfo::ForceWriteInfo(TFile*, bool)
10:41:43.340 [fatal]       /opt/software/linux-x86_64_v2/root-6.38.00-nm5dqbcfy2hwc6wzjjny4ezkl5pibj5i/lib/root/libRIO.so.6.38:1e7e77
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    1: TTree::BuildStreamerInfo(TClass*, void*, bool)
10:41:43.340 [fatal]       /opt/software/linux-x86_64_v2/root-6.38.00-nm5dqbcfy2hwc6wzjjny4ezkl5pibj5i/lib/root/libTree.so.6.38:125b60
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    2: TBranchElement::Unroll(char const*, TClass*, TClass*, char*, int, int, int)
10:41:43.340 [fatal]       /opt/software/linux-x86_64_v2/root-6.38.00-nm5dqbcfy2hwc6wzjjny4ezkl5pibj5i/lib/root/libTree.so.6.38:d4e03
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    3: TBranchElement::Init(TTree*, TBranch*, char const*, TVirtualCollectionProxy*, int, int, int)
10:41:43.340 [fatal]       /opt/software/linux-x86_64_v2/root-6.38.00-nm5dqbcfy2hwc6wzjjny4ezkl5pibj5i/lib/root/libTree.so.6.38:d9aa6
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    4: TBranchElement::TBranchElement(TTree*, char const*, TVirtualCollectionProxy*, int, int, int)
10:41:43.340 [fatal]       /opt/software/linux-x86_64_v2/root-6.38.00-nm5dqbcfy2hwc6wzjjny4ezkl5pibj5i/lib/root/libTree.so.6.38:d9e7a
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    5: TTree::BronchExec(char const*, char const*, void*, bool, int, int)
10:41:43.340 [fatal]       /opt/software/linux-x86_64_v2/root-6.38.00-nm5dqbcfy2hwc6wzjjny4ezkl5pibj5i/lib/root/libTree.so.6.38:13afd7
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    6: TTree::Branch(char const*, char const*, void*, int, int)
10:41:43.340 [fatal]       /opt/software/linux-x86_64_v2/root-6.38.00-nm5dqbcfy2hwc6wzjjny4ezkl5pibj5i/lib/root/libTree.so.6.38:123581
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    7: podio::ROOTWriter::initBranches(podio::ROOTWriter::CategoryInfo&, std::vector<std::tuple<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, podio::CollectionBase*>, std::allocator<std::tuple<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, podio::CollectionBase*> > > const&, podio::GenericParameters const&)
10:41:43.340 [fatal]       /opt/local/lib/libpodioRootIO.so:216bb
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    8: podio::ROOTWriter::writeFrame(podio::Frame const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)
10:41:43.340 [fatal]       /opt/local/lib/libpodioRootIO.so:25083
10:41:43.340 [fatal] 
10:41:43.340 [fatal]    9: JEventProcessorPODIO::Process(std::shared_ptr<JEvent const> const&)
10:41:43.340 [fatal]       /Users/spaul/EIC/EICrecon/install/lib/EICrecon/plugins/podio.so:7a40a
10:41:43.340 [fatal] 
10:41:43.340 [fatal]   10: void jana::components::JComponent::CallWithJExceptionWrapper<JEventProcessor::DoLegacyProcess(std::shared_ptr<JEvent const> const&)::{lambda()#3}>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, JEventProcessor::DoLegacyProcess(std::shared_ptr<JEvent const> const&)::{lambda()#3})
10:41:43.340 [fatal]       /opt/local/lib/libJANA.so:aa835
10:41:43.340 [fatal] 
10:41:43.340 [fatal]   11: JEventProcessor::DoLegacyProcess(std::shared_ptr<JEvent const> const&)
10:41:43.340 [fatal]       /opt/local/lib/libJANA.so:aacdd
10:41:43.340 [fatal] 
10:41:43.340 [fatal]   12: JEventMapArrow::fire(JEvent*, std::array<std::pair<JEvent*, int>, 2ul>&, unsigned long&, JArrow::FireResult&)
10:41:43.340 [fatal]       /opt/local/lib/libJANA.so:8e723
10:41:43.340 [fatal] 
10:41:43.340 [fatal]   13: JExecutionEngine::RunWorker(JExecutionEngine::Worker)
10:41:43.340 [fatal]       /opt/local/lib/libJANA.so:80e2e
10:41:43.340 [fatal] 
10:41:43.340 [fatal]   14: ???
10:41:43.340 [fatal]       /opt/local/lib/libstdc++.so.6:e1224
10:41:43.340 [fatal] 
10:41:43.340 [fatal]   15: ???
10:41:43.340 [fatal]       /lib/x86_64-linux-gnu/libc.so.6:92b7b
10:41:43.340 [fatal] 
10:41:43.340 [fatal]   16: __clone
10:41:43.340 [fatal]       /lib/x86_64-linux-gnu/libc.so.6:110630
10:41:43.340 [fatal] 
10:41:43.340 [fatal]

veprbl:
I suggest you to try recompiling your EICrecon from scratch.

baptistefraisse:
@sebouh137 @veprbl The failure seems to be that the second input of FarForwardNeutralsReconstructionAlgorithm is B0ECalClusters, which requires B0ECalHits. For a standalone ZDC file this collection is absent. It could make sense to make the non-ZDC calorimeter inputs **optional** in the reconstruction function. However, in the current implementation, these collections are declared as algorithm inputs, so they are resolved by the framework before the reconstruction. As a result, missing collections trigger an exception before entering the code.

sebouh137:
> @sebouh137 @veprbl The failure seems to be that the second input of FarForwardNeutralsReconstructionAlgorithm is B0ECalClusters, which requires B0ECalHits. For a standalone ZDC file this collection is absent. It could make sense to make the non-ZDC calorimeter inputs **optional** in the reconstruction function. However, in the current implementation, these collections are declared as algorithm inputs, so they are resolved by the framework before the reconstruction. As a result, missing collections trigger an exception before entering the code.

In that case, it should work if we make all inputs optional, and that way it can do stand-alone reconstruction on any far forward detector.  Will this be a viable solution?

veprbl:
We can. I hope it would be also possible to split into 4 instances of the factory.

ruse-traveler:
> We can. I hope it would be also possible to split into 4 instances of the factory.

Agreed! Making the inputs optional would definitely address the issue, but splitting the factory in 4 would be ideal. The latter option will help keep things modular, and make it easier to extend down the road!

DraTeots:
Hi All, run Mythos class (famously marketed for finding troubles) Fable 5 model on MT catching for this PR. It worked 6 (six!!!!) hours. Here is the report: 
[PR2622_MT_audit.md](https://github.com/user-attachments/files/28860987/PR2622_MT_audit.md)

Report saved to PR2622_MT_audit.md. Key findings:

**1. One real, serious bug — but it's not the MT mechanism.** The "cluster already used as gamma" bookkeeping in [[FarForwardNeutralsReconstruction.cc:111-167](https://claude.ai/epitaxy/src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L111)](src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L111) stores `&cl` where `cl` is a podio handle returned **by value** (verified in the generated edm4eic headers: `using reference = Cluster`). Every stored pointer is the address of the same reused, lifetime-extended stack temporary — dangling and compared against a different loop's temporary. It's undefined behavior, and in a minimal reproduction of the exact pattern GCC's behavior flips with optimization level: at `-O0` every cluster is excluded (no ZDC neutron ever produced), at `-O2` the dangling comparison folds to `false` so *nothing* is excluded — meaning the released build most likely **double-counts photon clusters into the ZDC neutron energy**. However, this is fixed at compile time and identical on every thread, so it cannot by itself produce an MT-vs-ST difference — wdconinc's skepticism about the author's diagnosis was justified.

**2. No shared-state races in the PR's own code.** Factories hold per-event-slot algorithm instances; `m_gammaZMax`/`m_zMax` are write-once-in-init; the magic statics in [[LambdaReconstruction.cc:41-43](https://claude.ai/epitaxy/src/algorithms/reco/LambdaReconstruction.cc#L41)](src/algorithms/reco/LambdaReconstruction.cc#L41) are thread-safe; `ParticleSvc` is read-only; there's no RNG; the `JEventProcessorPODIO.cc` and `reco.cc` changes are pure wiring. Per event, the new code is a pure function of its inputs.

**3. Most likely root cause of the CI diff: upstream perturbation amplified by the PR's winner-takes-all selections.** The LFHCAL neutral output depends on nothing but `LFHCALClusters` — so if it differed MT-vs-ST (wdconinc: "Also LFHCAL"), the input itself must have differed in that run, placing the root nondeterminism upstream of the PR. The PR's new reductions — non-stable sort + top-4 truncation ([[:135-144](https://claude.ai/epitaxy/src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L135)](src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L135)), strict-`>` leader selection ([[:188](https://claude.ai/epitaxy/src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L188)](src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L188)), order-dependent float sum ([[:185](https://claude.ai/epitaxy/src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L185)](src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L185)), exact-FP best-candidate ranking ([[LambdaReconstruction.cc:419-440](https://claude.ai/epitaxy/src/algorithms/reco/LambdaReconstruction.cc#L419)](src/algorithms/reco/LambdaReconstruction.cc#L419)) — turn an invisible ulp/tie-order wobble into a wholly different candidate, visible in capybara's histograms while the upstream wobble isn't. That fits all the evidence: the comparison job was green for a year before this PR, the failure appeared once on a loaded CI runner, and nobody (including the author) can reproduce it from source. Known shared-layer races exist (cf. the ROOT TSAN suppressions in a8857b2d), and the PR's new four-chain join also perturbs the JANA2 scheduling DAG.

**4. The author's fix: necessary but probably insufficient — and not actually pushed.** `main` (af0e8fbb) still contains the pointer bookkeeping; the only post-merge change is the #2679 rename, and the author's fork branch is stale. The index-based fix should be merged urgently (it cures the UB/double-counting), but it likely won't cure the CI flakiness. The report recommends additionally making the sorts/selections tie-stable, splitting the factory into four (as already proposed in the thread — this also fixes the standalone-ZDC crash), and hunting the residual flake with per-event exact comparisons rather than histograms.

veprbl:
Reading the summary above, at least to me, it looks like it's wrong on every point. Have you given it ability to actually run and debug the code?

wdconinc:
Rather than posting AI output here (with some vague expectation that we'll read it and fix it for you? we can also use AI...), it would be more useful to have your model test the hypotheses, generate working solutions, and contribute them as a PR after testing that they work and solve the problem. We don't want to get into a mode where people just have their favorite AI model post some analysis and then leave it there for others to figure out what to do with it (and whether it's even correct or relevant).

This is a prime example of what is called an "extractive contribution": a contribution that actually takes time *away* from the people you are sharing it with, because it now takes us more time to figure out whether it makes sense and what to do with it than it took you to put in the prompt and copy-paste the response.

(Try doing this in a project like llvm and it's your first of three warnings before you're banned: pure AI comments without any human interpretation or curation are not tolerated in any large project I know of.)

DraTeots:
Wouter, please, breathe out. Everything is fine. Running it autonomous is a part of a plan. And yes, there is a plan and it is even a was discussed on a meeting. As well as yo try to run massive parallel campaign and compare the plots. But. Wasn't it interesting how really Mythos behave after all that hype? Especially what it produces after 6 hpurs of thinking? 

I explicitly mentioned the text is ai generated. You could not read it further. And stop flooding this topic with your stories about llvm or I tell you about zig.

PR #2630, opened 2026-04-21
fardetectors_library: don't link to TMVA

veprbl:
TMVA is out since b5c3eedaf4039ef925c928f96a138505437e917b, we don't need to link to it anymore.

### DIFF
--- a/src/algorithms/fardetectors/CMakeLists.txt
+++ b/src/algorithms/fardetectors/CMakeLists.txt
@@ -15,4 +15,4 @@ plugin_add_eigen3(${PLUGIN_NAME})
 plugin_add_event_model(${PLUGIN_NAME})
 plugin_add_cern_root(${PLUGIN_NAME})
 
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
Thanks!

Where's that LWYU check? Should that have flagged this or is it not enforced?

veprbl:
> Where's that LWYU check? Should that have flagged this or is it not enforced?

Is it checking for unused symbols? I thought it was checking for unlinked libraries.

wdconinc:
> > Where's that LWYU check? Should that have flagged this or is it not enforced?
> 
> Is it checking for unused symbols? I thought it was checking for unlinked libraries.

It is checking for libraries that linked but of which no symbols are used, https://cmake.org/cmake/help/latest/prop_tgt/LINK_WHAT_YOU_USE.html

So it should have flagged ROOT::TMVA as a library without used symbols...

PR #2667, opened 2026-05-14
Random noise injection for silicon tracker 

jsobajic:
### Briefly, what does this PR introduce? Please link to any relevant presentations or discussions.
This PR implements noise in the SVT subdetectors (BVTX, BTRK, ECTRK) uniformly and with Poisson distribution about a given number of noise hits per layer.  

### What is the urgency of this PR?
- [ ] High (please describe reason below)
- [X] Medium
- [] Low

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Optimization (issue #__)
- [ ] Updated parameters, constants (issue #__)
- [ ] Updated documentation
- [ ] other: __

### Please check if any of the following apply
- [ ] This PR requires changes to geometry (epic PR: __)
- [ ] This PR requires changes to EDM4eic (EDM PR: __)
- [ ] This PR introduces breaking changes. Please describe changes users need to make below.
- [ ] This PR changes default behavior. Please describe changes below.
- [ ] AI was used in preparing this PR. Please describe usage below.

PR #2670, opened 2026-05-18
SVT random noise injection 

ShujieL:

Inject random sensor noise to SVT (ECTRK, BVTX, BTRK). With provided number of noise hits per layer, the code uses a for loop to randomly select a sensitive module, then a position --> cell ID within that module to inject the noise hit.  This mechanism works only if **the layer has one uniform module shape and size.** 

Noise are created at RawHits level, then get combined with real hits before RecHits conversion.

To enable noise injection: 
```
-Pectrk:siendcaptrackernoiserawhits:addnoise=1
-Pbvtx:sibarrelvertexnoiserawhits:addnoise=1
-Pbtrk:sibarrelnoiserawhits:addnoise=1
```

This PR based on the previous developments at https://github.com/eic/EICrecon/pull/1988 and https://github.com/eic/EICrecon/pull/2667.  It has relevant changes cherry-picked for rebase, refactoring, and a few more updates to ensure the code serves the purpose. 

- [x ] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [ ] Updated documentation
- [ ] other: __

- [x ] AI was used in preparing this PR. Please describe usage below.

CODEX was used for refactoring and debugging.


ShujieL:
@veprbl this PR passed all checks except eic-container. Can you take a look?

src/algorithms/digi/RandomNoise.cc
```diff
@@ -0,0 +1,738 @@
...
+
+  // Normalize DD4hep field names before matching layer-like volume IDs.
+  std::string lower(std::string s) {
+    for (auto& c : s) {
+      c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+    }
+    return s;
+  }
+
+  // Last-resort layer extraction for geometries where the layer ID is encoded
```
wdconinc: For which detector is this used?

src/algorithms/digi/RandomNoiseModule.cc
```diff
@@ -0,0 +1,738 @@
...
+    }
+    for (const auto& [field, value] : pv.volIDs()) {
+      const auto name = lower(field);
+      if (name == "layer" || name.find("lay") != std::string::npos) {
+        return value;
+      }
+    }
+    return std::nullopt;
+  }
+
```
wdconinc: Why not resolve layer id from layer DetElement in all cases? Where is the fallback necessary? Better to fail and provide error message to fix the geometry than resort to parsing names.

src/algorithms/digi/RandomNoise.cc
```diff
@@ -0,0 +1,738 @@
...
+    case RandomNoise::SensitiveShapeKind::Cone:
+      return d.size() >= 5;
+    }
+    return false;
+  }
+
+  // Estimate the sensitive solid volume used to weight multiple sensitive pieces
+  // inside the same module. For thin sensors this is equivalent to area weighting
+  // when all sensitive pieces have the same thickness.
+  double sensitiveVolume(RandomNoise::SensitiveShapeKind shape, const std::vector<double>& d) {
```
wdconinc: There's no volume function in TGeo to use directly instead of this roundtrip through custom types?

wdconinc (changes requested):
I think this needs a bit of simplification. It seems you are duplicating a lot of and ROOT TGeo already provides in terms of random point sampling and evaluation. We should build the rejection loop entirely around TGeoVolume RandomPoints and Contains.

src/algorithms/digi/RandomNoise.cc
ShujieL: It's tested for SVT, but should be applicable for other detectors with similar noise mechanism.

src/algorithms/digi/RandomNoiseModule.cc
ShujieL: This fall back function is removed.

src/algorithms/digi/RandomNoise.cc
ShujieL: You are right. Capacity() is used instead.

ShujieL:
> I think this needs a bit of simplification. It seems you are duplicating a lot of and ROOT TGeo already provides in terms of random point sampling and evaluation. We should build the rejection loop entirely around TGeoVolume RandomPoints and Contains.

Thanks for pointing this out. I switched to TGeo Contains for rejection loop. The RandomPoints function supports visualization but I can't see a clear way to use it for our event-by-event sampling, so I kept our random seeds approach.

src/algorithms/digi/RandomNoise.cc
```diff
@@ -0,0 +1,689 @@
...
+// sensitive component local coordinate frame. Dimensions follow DD4hep conventions.
+dd4hep::Position RandomNoise::randomPointInComponent(const SensitiveComponent& component,
+                                                     std::mt19937_64& rng) const {
+  auto uniform = [&](double lo, double hi) {
+    return std::uniform_real_distribution<double>{lo, hi}(rng);
+  };
+
+  const auto& d = component.dimensions;
+  switch (component.shape) {
+  case SensitiveShapeKind::Box:
```
wdconinc: Even if you don't use RandomPoints, you can still do the same thing it does, and it works for all volumes. This `SensitiveShapeKind` needs to disappear.

Algorithm:
- throw points uniformly in the oriented bounding box (all our volumes inherit TGeoBBox),
- accept/reject by Contains call

wdconinc (dismissed):

wdconinc:
> I switched to TGeo Contains for rejection loop.

Didn't see any use of Contains. Did you push those changes?

src/algorithms/digi/RandomNoise.cc
ShujieL: obviously I forgot to push. Please check again ;)

src/algorithms/digi/RandomNoise.cc
```diff
@@ -0,0 +1,553 @@
...
+// Convert one random sensitive-local point to a cell ID:
+// component-local point -> module-local transform -> global position -> DD4hep cell ID.
+std::optional<std::uint64_t> RandomNoise::randomCellID(const ModuleGeometry& module,
+                                                       const SensitiveComponent& component,
+                                                       std::mt19937_64& rng) const {
+  const auto componentLocal = randomPointInComponent(component, rng);
+  const auto& rotation      = component.localToModuleRotation;
+  const auto& translation   = component.localToModuleTranslation;
+  const double local[3]     = {componentLocal.x(), componentLocal.y(), componentLocal.z()};
+  const dd4hep::Position moduleLocal{
```
ShujieL: to do: use TGEO

ShujieL:
@wdconinc @veprbl 
I re-wrote this algorithm into RandomNoisePixel. Instead of distributing a pre-calculated number of noise hits for one layer which assumed one uniform sensitive component size, it now will cache all sensitive component dimensions once, calculate the number of total noise hits per layer, and distribute according to the size. The only configuration input needed is noise_rate_per_pixel_per_event which is 2e-7 per pixel per 2us by default.

simonge:
A question and suggestion without looking closely at the code.

Are we planning on including the noise in all reconstructed samples by default once this PR is merged? Or will we keep the main sample clean as with other (e.g.) beam backgrounds? It would be nice to have a simple global configuration to turn them all on and off.

I would changing the configuration from `noise_rate_per_pixel_per_event` to a more fundamental and explicit `noise_rate_per_pixel_per_us` (or other time unit). Then pass the time period as a config separately so that when available through a global configuration/metadata this is easily swapped out.

ShujieL:
> A question and suggestion without looking closely at the code.
> 
> Are we planning on including the noise in all reconstructed samples by default once this PR is merged? Or will we keep the main sample clean as with other (e.g.) beam backgrounds? It would be nice to have a simple global configuration to turn them all on and off.
> 
> I would changing the configuration from `noise_rate_per_pixel_per_event` to a more fundamental and explicit `noise_rate_per_pixel_per_us` (or other time unit). Then pass the time period as a config separately so that when available through a global configuration/metadata this is easily swapped out.

Thanks for the comments.

The noise injection is controlled by an "addNoise" flag which is off by default. The noise injection is heavy in storage and memory, I'd suggest to leave it off for general simulations, only turn it on with options like:
`        -PSVT:noise_rate_per_pixel_per_event="${noise_rate}"
        -PBVTX:SiBarrelVertexNoiseRawHits:addNoise=true
        -PBTRK:SiBarrelNoiseRawHits:addNoise=true
        -PECTRK:SiEndcapTrackerNoiseRawHits:addNoise=true`
for dedicated background+noise performance studies. 

I thought about micro-second vs event for the noise rate unit. I decided to go with "per_event" because our reconstruction is still event based. The code does not know how to translate rate per us to a certain number of noise hits per event unless the time window is also provided. So it's indeed easier (for now) if we use one parameter for rate per event, with the clear understanding that the user is responsible to provide the noise rate that is consistent with our time window choice. I will add a note on that.

veprbl:
What is the difference between RandomNoiseModule and RandomNoisePixel? Do we need both?

ShujieL:
> What is the difference between RandomNoiseModule and RandomNoisePixel? Do we need both?

NoisePixel is a newer approach. I removed NoiseModule so no more confusions.

docs/design/svt_pixel_noise_injection.md
```diff
@@ -0,0 +1,373 @@
...
+1. Verify that `sum(component.pixelCount)` equals `layer.totalPixels`.
+2. For a rectangle, compare the cache with the product of the two index counts.
+3. For a trapezoid, plot `firstMin` and `firstMax` versus row and compare them with sensor edges.
+4. Decode generated IDs and check their detector, layer, component, and pixel fields.
+5. Over many events, confirm that the mean hit count approaches `p * N_l`.
+6. Confirm that component hit fractions approach `N_m/N_l`.
+7. Repeat the same event and configuration and confirm identical cell IDs.
+
+These tests separately check geometry interpretation, statistical normalization, and
+reproducibility: the three main correctness requirements of the algorithm.
```
veprbl:
```suggestion
```

veprbl (approved):
Looked at the code and the accompanying documentation. It's hard to validate without a detailed study on specific shapes. I only checked that outputs are somewhat reasonable. The performance and memory consumption actually appear to be very good (too good?).

ShujieL:
> Looked at the code and the accompanying documentation. It's hard to validate without a detailed study on specific shapes. I only checked that outputs are somewhat reasonable. The performance and memory consumption actually appear to be very good (too good?).

Thanks. You can find some validation at https://indico.bnl.gov/event/31808/contributions/126718/

PR #2683, opened 2026-06-02
Fix eigenvalue calculation in CalorimeterClusterShape

akioogawa:

- Change float to double for 3d cluster shape covariance matrix calculation because floating point rounding error was giving ⟨z²⟩ − ⟨z⟩² = none-zero thus sigma_Z often to be around ~2 mm for Forward Ecal where z of hits are always exactly the same and sigma-z should be always 0.
- Replace Eigen::EigenSolver (general) with Eigen::SelfAdjointEigenSolver (for symmetric matrix). This also fixes that 2d and 3d sigmas were not sorted by its size. We'll need to fix https://github.com/eic/EDM4eic/blob/main/edm4eic.yaml on cluster shapeParams  : "3 entries for x-y-z widths [mm]." to something like "3 entries for 3D max, middle and min widths [mm]".
- Also they needed to be sqrt of egen values to make them [rad] and [mm] unit for shape parameters as specified in edm4eic.yaml.
- Always do intrinsic theta & phi calculations, since phi is indeed 2d "theta" (cluster long axis angle in xy) for detectors like Forward Ecal and useful even if there is no depth information.

- [ ] Medium
- [ x] Low

- [ x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterClusterShape.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterShape.cc
@@ -190,11 +190,11 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
       // set shape parameters
       out_clust.addToShapeParameters(radius);
       out_clust.addToShapeParameters(dispersion);
-      out_clust.addToShapeParameters(eigenValues_2D[0].real()); // 2D theta-phi out_cluster width 1
-      out_clust.addToShapeParameters(eigenValues_2D[1].real()); // 2D theta-phi out_cluster width 2
-      out_clust.addToShapeParameters(eigenValues_3D[0]);        // 3D x-y-z out_cluster width 1
-      out_clust.addToShapeParameters(eigenValues_3D[1]);        // 3D x-y-z out_cluster width 2
-      out_clust.addToShapeParameters(eigenValues_3D[2]);        // 3D x-y-z out_cluster width 3
+      out_clust.addToShapeParameters(std::sqrt(std::abs(eigenValues_2D[0].real()))); // 2D theta-phi out_cluster width 1 [rad]
+      out_clust.addToShapeParameters(std::sqrt(std::abs(eigenValues_2D[1].real()))); // 2D theta-phi out_cluster width 2 [rad]
+      out_clust.addToShapeParameters(std::sqrt(std::abs(eigenValues_3D[0])));        // 3D x-y-z out_cluster width 1 [mm]
+      out_clust.addToShapeParameters(std::sqrt(std::abs(eigenValues_3D[1])));        // 3D x-y-z out_cluster width 2 [mm]
+      out_clust.addToShapeParameters(std::sqrt(std::abs(eigenValues_3D[2])));        // 3D x-y-z out_cluster width 3 [mm]
 
       // check axis orientation
       double dot_product = out_clust.getPosition() * axis;
@@ -209,13 +209,16 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
       out_clust.setIntrinsicPhi(intrinsicPhi);
       // TODO intrinsicDirectionError
 
-      trace("ClusterShape: radius={:.3f} dispersion={:.3f} "
-            "2D_w1={:.4f} 2D_w2={:.4f} "
-            "3D_w1={:.3f} 3D_w2={:.3f} 3D_w3={:.3f} "
+      trace("ClusterShape: radius={:.3f} mm dispersion={:.3f} mm "
+            "2D_w1={:.4f} rad 2D_w2={:.4f} rad "
+            "3D_w1={:.3f} mm 3D_w2={:.3f} mm 3D_w3={:.3f} mm "
             "intrinsicTheta={:.4f} intrinsicPhi={:.4f}",
             radius, dispersion,
-            eigenValues_2D[0].real(), eigenValues_2D[1].real(),
-            eigenValues_3D[0], eigenValues_3D[1], eigenValues_3D[2],
+            std::sqrt(std::abs(eigenValues_2D[0].real())),
+            std::sqrt(std::abs(eigenValues_2D[1].real())),
+            std::sqrt(std::abs(eigenValues_3D[0])),
+            std::sqrt(std::abs(eigenValues_3D[1])),
+            std::sqrt(std::abs(eigenValues_3D[2])),
             intrinsicTheta, intrinsicPhi);
     } // end shape parameter calculation
 
--- a/src/algorithms/calorimetry/CalorimeterClusterShape.cc
+++ b/src/algorithms/calorimetry/CalorimeterClusterShape.cc
@@ -101,7 +101,7 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
       // create addresses for quantities we'll need later
       double radius     = 0;
       double dispersion = 0;
-      double w_sum     = 0;
+      double w_sum      = 0;
       // set up matrices/vectors — all double to avoid float32 cancellation
       Eigen::Matrix2d sum2_2D        = Eigen::Matrix2d::Zero();
       Eigen::Matrix3d sum2_3D        = Eigen::Matrix3d::Zero();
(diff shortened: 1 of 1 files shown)


src/algorithms/calorimetry/CalorimeterClusterShape.cc
```diff
@@ -102,15 +102,19 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
       // create addresses for quantities we'll need later
       float radius     = 0;
       float dispersion = 0;
-      float w_sum      = 0;
+      double w_sum     = 0;
 
       // set up matrices/vectors
+      // 2D (theta-phi) uses float — angles are O(1), no precision issue.
+      // 3D (x-y-z) uses double — hit positions are O(1000) mm; float32 causes
+      //   catastrophic cancellation in cov[2,2] = <z²> - <z>² for detectors
```
veprbl: Would it be a non-cancelation?

src/algorithms/calorimetry/CalorimeterClusterShape.cc
```diff
@@ -124,8 +128,8 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
           // theta, phi
           Eigen::Vector2f pos2D(edm4hep::utils::anglePolar(hit.getPosition()),
                                 edm4hep::utils::angleAzimuthal(hit.getPosition()));
-          // x, y, z
-          Eigen::Vector3f pos3D(hit.getPosition().x, hit.getPosition().y, hit.getPosition().z);
+          // x, y, z — double to avoid float32 cancellation for large z
```
veprbl: Here as well.

src/algorithms/calorimetry/CalorimeterClusterShape.cc
```diff
@@ -102,15 +102,19 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
       // create addresses for quantities we'll need later
       float radius     = 0;
       float dispersion = 0;
-      float w_sum      = 0;
+      double w_sum     = 0;
 
       // set up matrices/vectors
+      // 2D (theta-phi) uses float — angles are O(1), no precision issue.
```
wdconinc: I'd probably just change this to double as well, for consistency with the 3D case.

src/algorithms/calorimetry/CalorimeterClusterShape.cc
akioogawa: I'm not sure what exactly your question is. I was seeing sigma-z = 1~2mm while input hit's z are all exactly the same value (3522.3mm). After changing it to double, I'm getting much better zero. 32bit float only has ~7 digits. It causes issue in z since it has a large offset and loosing accuracy at z^2 in the matrix. So if you are asking about 2d theta/phi need double, 32 bit seems to be enough since it has no large offsets.

src/algorithms/calorimetry/CalorimeterClusterShape.cc
akioogawa: Starting with float 3 vector is ok. Just matrix need to be in double. Same rounding error happens when you try to get mass at forward using Lorenz vector in float (E^2-Pz^2 needs more than 7 digits, but we can start with E and Pz in float). But... we are using 64bit hardware and software everywhere (right?) , and we can use double everywhere as well, except for storage IO.

src/algorithms/calorimetry/CalorimeterClusterShape.cc
akioogawa: Chaged 1d (radius/deisparsion) and 2d (theta/phi) also to double.

src/algorithms/calorimetry/CalorimeterClusterShape.cc
veprbl: I'm asking about the wording of the comment. If <z^2> - <z>^2 is not zero that means those values did not cancel the way they should have in <(<z> - z)^2>. I see you removed the comment, so this is resolved.

src/algorithms/calorimetry/CalorimeterClusterShape.cc
```diff
@@ -100,17 +97,16 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
...
-      Eigen::Matrix2f sum2_2D         = Eigen::Matrix2f::Zero();
-      Eigen::Matrix3f sum2_3D         = Eigen::Matrix3f::Zero();
-      Eigen::Vector2f sum1_2D         = Eigen::Vector2f::Zero();
-      Eigen::Vector3f sum1_3D         = Eigen::Vector3f::Zero();
-      Eigen::Vector2cf eigenValues_2D = Eigen::Vector2cf::Zero();
-      Eigen::Vector3cf eigenValues_3D = Eigen::Vector3cf::Zero();
+      double radius     = 0;
+      double dispersion = 0;
+      double w_sum      = 0;
+      // set up matrices/vectors — all double to avoid float32 cancellation
```
veprbl:
```suggestion
      // set up matrices/vectors
```

src/algorithms/calorimetry/CalorimeterClusterShape.cc
```diff
@@ -153,60 +148,74 @@ void CalorimeterClusterShape::process(const CalorimeterClusterShape::Input& inpu
...
+      out_clust.addToShapeParameters(
+          std::sqrt(std::abs(eigenValues_2D[0]))); // 2D theta-phi out_cluster width 1 [rad]
+      out_clust.addToShapeParameters(
+          std::sqrt(std::abs(eigenValues_2D[1]))); // 2D theta-phi out_cluster width 2 [rad]
+      out_clust.addToShapeParameters(
+          std::sqrt(std::abs(eigenValues_3D[0]))); // 3D x-y-z out_cluster width 1 [mm]
+      out_clust.addToShapeParameters(
+          std::sqrt(std::abs(eigenValues_3D[1]))); // 3D x-y-z out_cluster width 2 [mm]
+      out_clust.addToShapeParameters(
+          std::sqrt(std::abs(eigenValues_3D[2]))); // 3D x-y-z out_cluster width 3 [mm]
```
veprbl:
```suggestion
      out_clust.addToShapeParameters(eigenValues_2D[0]); // 2D theta-phi out_cluster width 1 [rad^2]
      out_clust.addToShapeParameters(eigenValues_2D[1]); // 2D theta-phi out_cluster width 2 [rad^2]
      out_clust.addToShapeParameters(eigenValues_3D[0]); // 3D x-y-z out_cluster width 1 [mm^2]
      out_clust.addToShapeParameters(eigenValues_3D[1]); // 3D x-y-z out_cluster width 2 [mm^2]
      out_clust.addToShapeParameters(eigenValues_3D[2]); // 3D x-y-z out_cluster width 3 [mm^2]
```
If you don't mind, I'd like to postpone the introduction of sqrt. While I agree with it, we would need to announce the breaking change (e.g. needs retraining of ecal_pid ONNX files). We can get rest of the changes now.

veprbl: I will also initiate EDM4eic change to provide named fields.

veprbl: (those can be implemented with sqrt straight away)

ruse-traveler (commented):
Thanks for catching this, Akio! I'd second @veprbl's suggestion to hold off on the sqrt's until the datamodel change: I agree with it, but it's breaking so it'll need some announcements...

src/algorithms/calorimetry/CalorimeterClusterShape.cc
akioogawa: I don't mind staying without sqrt, if comments are consistent and clearly says its [rad]^2 or [mm]^2 etc.

akioogawa: Yes go ahead for rolling back without sqrt for now. When you change EDM4eic (adding each variables instead of using ShapeParam array?), just make it consistent with or without sqrt.

veprbl (approved):
LGTM

PR #2721, opened 2026-06-25
JEventProcessorManagedPODIO: use JEventSourcePODIO::IsOpenable instead of std::filesystem::exists

veprbl:

The current check requires file to be local and does not work for remote files such as those on XRootD.

- [ ] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __


src/services/io/podio/JEventSourcePODIO.cc
```diff
@@ -254,27 +275,7 @@ std::string JEventSourcePODIO::GetDescription() {
 //------------------------------------------------------------------------------
 template <>
 double JEventSourceGeneratorT<JEventSourcePODIO>::CheckOpenable(std::string resource_name) {
```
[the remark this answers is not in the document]
veprbl: Yes, this is a concern, but not an issue on platforms that we support.

veprbl:
@wdconinc Does this look reasonable to you?
The side problem I'm trying to overcome is that `JEventSourceGeneratorT<JEventSourcePODIO>::CheckOpenable` is not static in JANA2. We could ask to change that upstream, but I'm not sure if there are any implications wrt compilation of those templates.

src/services/io/podio/JEventProcessorManagedPODIO.cc
```diff
@@ -155,10 +156,10 @@ void JEventProcessorManagedPODIO::ProcessFileRequest(const nlohmann::json& reque
 
     m_log->info("Processing request: {} -> {}", input_file, output_file);
 
-    // Check if input file exists
-    if (!std::filesystem::exists(input_file)) {
+    // Check if input file can be opened
+    if (!JEventSourcePODIO::IsOpenable(input_file)) {
```
wdconinc, line 160: Probably unrelated to this PR's goal, but why does the event processor need to check the input file? Isn't the event source where that is supposed to happen?

src/services/io/podio/JEventSourcePODIO.cc
```diff
@@ -240,41 +240,29 @@ std::string JEventSourcePODIO::GetDescription() {
...
-  return 0.03;
+  return tree != nullptr;
+}
+
+//------------------------------------------------------------------------------
+// CheckOpenable
+//------------------------------------------------------------------------------
+template <>
+double JEventSourceGeneratorT<JEventSourcePODIO>::CheckOpenable(std::string resource_name) {
+  return JEventSourcePODIO::IsOpenable(resource_name) ? 0.03 : 0.0;
```
wdconinc, line 265: I wish I could unsee this use of "probability"...

src/services/io/podio/JEventSourcePODIO.cc
```diff
@@ -240,41 +240,29 @@ std::string JEventSourcePODIO::GetDescription() {
...
-  }
-  return 0.03;
+  return tree != nullptr;
+}
+
+//------------------------------------------------------------------------------
+// CheckOpenable
+//------------------------------------------------------------------------------
+template <>
+double JEventSourceGeneratorT<JEventSourcePODIO>::CheckOpenable(std::string resource_name) {
```
wdconinc, line 264: While this is being changed...
```suggestion
double JEventSourceGeneratorT<JEventSourcePODIO>::CheckOpenable(const std::string& resource_name) {
```

wdconinc (approved):
LGTM.

src/services/io/podio/JEventSourcePODIO.cc
veprbl: But that would change the call interface and the mangled name.

src/services/io/podio/JEventSourcePODIO.cc
veprbl: Same. I think, you can just train yourself to misread it as priority.

src/services/io/podio/JEventProcessorManagedPODIO.cc
veprbl: I convinced myself that early check is the only way to report a failure back. But yeah, let me check, if we have an alternative.

veprbl: #2725 might be actually a cleaner way to do this

veprbl:
Closing in favor of the other PR.

PR #2725, opened 2026-06-29
JEvent*ManagedPODIO: remove premature file checks

veprbl:

Closes: #2721 

This is an alternative take on #2721 instead of trying to check the existence, let's not do it and let exception come from deeper in the call stack. The trade off is that cleanup logic becomes slightly more involved.

- [ ] Medium
- [x] Low

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

### DIFF
--- a/src/services/io/podio/JEventProcessorManagedPODIO.cc
+++ b/src/services/io/podio/JEventProcessorManagedPODIO.cc
@@ -159,13 +159,6 @@ void JEventProcessorManagedPODIO::ProcessFileRequest(const nlohmann::json& reque
     m_log->info("Processing request: {} -> {} (nskip={}, nevents={})", input_file, output_file,
                 nskip, nevents == 0 ? std::string("0 [all]") : std::to_string(nevents));
 
-    // Check if input file exists
-    if (!std::filesystem::exists(input_file)) {
-      SendResponse({{"status", "error"},
-                    {"message", fmt::format("Input file does not exist: {}", input_file)}});
-      return;
-    }
-
     {
       std::lock_guard<std::mutex> lock(m_file_mutex);
       m_current_input_file  = input_file;
--- a/src/services/io/podio/JEventProcessorManagedPODIO.cc
+++ b/src/services/io/podio/JEventProcessorManagedPODIO.cc
@@ -181,9 +181,11 @@ void JEventProcessorManagedPODIO::ProcessFileRequest(const nlohmann::json& reque
     try {
       NotifySourceNewFile(input_file, nskip, nevents);
     } catch (...) {
-      std::lock_guard<std::mutex> lock(m_file_mutex);
-      m_file_processing_active = false;
-      m_writer.reset(); // finalizes and closes the (empty) output file
+      {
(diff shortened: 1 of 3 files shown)


src/services/io/podio/JEventProcessorManagedPODIO.cc
```diff
@@ -180,8 +173,15 @@ void JEventProcessorManagedPODIO::ProcessFileRequest(const nlohmann::json& reque
       m_file_processing_active = true;
     }
 
-    // Signal the event source that a new file is available
-    NotifySourceNewFile(input_file);
+    // Signal the event source that a new file is available.
+    try {
+      NotifySourceNewFile(input_file);
+    } catch (...) {
+      std::lock_guard<std::mutex> lock(m_file_mutex);
```
wdconinc: Can this fall back to `CloseOutputFile` (since by now we have already called `OpenOutputFile`).

veprbl: We can, but I think it needs a refactor. You want it in this PR?

wdconinc: Follow-up PR is fine.

veprbl: Okay, there will be multiple. First: #2730

veprbl: Now done.

src/services/io/podio/JEventProcessorManagedPODIO.cc
```diff
@@ -184,15 +177,25 @@ void JEventProcessorManagedPODIO::ProcessFileRequest(const nlohmann::json& reque
...
     }
 
-    // Signal the event source that a new file is available (with nskip/nevents)
-    NotifySourceNewFile(input_file, nskip, nevents);
+    // Signal the event source that a new file is available.
+    try {
+      NotifySourceNewFile(input_file, nskip, nevents);
+    } catch (...) {
+      {
+        std::lock_guard<std::mutex> lock(m_file_mutex);
```
wdconinc, line 185: Curious why you lock this. It seems you are using atomic bools for the flags.

wdconinc (approved):
Looks good to me.

wdconinc:
> The trade off is that cleanup logic becomes slightly more involved.

Yeah, that's the only thing I was going to comment on here, but I don't think it needs to hold up the PR.

src/services/io/podio/JEventProcessorManagedPODIO.cc
veprbl: At very least that one is not atomic:
https://github.com/eic/EICrecon/blob/32a135c765dade530d5e4f2a1b4cdf695c89bb17/src/services/io/podio/JEventProcessorManagedPODIO.h#L50
Making it atomic would not help with data races. Not saying that the current mutex is perfect, but it's supposed to do things beyond protecting against UB.

PR #2732, opened 2026-07-05
Implementation of an algorithm for Edep-to-Npe conversion

### LINKED ISSUE #2728: Conversion from energy deposit to the number of photoelectrons
This issue is intended to add a new algorithm that models the conversion of energy deposit into photoelectrons before the signal pulse is generated.

mhkim-anl:
This PR is to implement a separate algorithm to convert the energy deposit into the number of photoelectrons before pulse generation. The conversion can be performed using either a constant conversion factor or a LUT file, after which Poisson smearing is applied. The validation of this algorithm can be found at the following link:
[https://www.overleaf.com/read/mcvmncnpcwfk#41d7e1](https://www.overleaf.com/read/mcvmncnpcwfk#41d7e1)

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #2728 )
- [ ] Updated documentation
- [ ] other: __


sly2j (dismissed):

src/detectors/BEMC/BEMC.cc
```diff
@@ -62,11 +64,16 @@ void InitPlugin(JApplication* app) {
   decltype(SimCalorimeterHitProcessorConfig::timeWindow) EcalBarrelScFi_timeWindow = {
       100 * edm4eic::unit::ns};
 
+  decltype(EdepToNpeConversionConfig::edep_to_npe_fields) EcalBarrelScFi_edep_to_npe_fields = {
+      "layer"};
+  decltype(EdepToNpeConversionConfig::edep_to_npe_filename) EcalBarrelScFi_edep_to_npe_filename = {
+      "bic_edepToNpe_layer.lut"};
```
veprbl: Where one can get this file?
Needs to be added to https://github.com/eic/epic-data and registered in https://github.com/eic/epic/blob/main/compact/calibrations.xml
```suggestion
      "calibrations/bic_edepToNpe_layer.lut"};
```

sly2j: Already present in epic-data (https://github.com/eic/epic-data/blob/main/bic_edepToNpe_layer.lut), @mhkim-anl can you please update https://github.com/eic/epic/blob/main/compact/calibrations.xml to add the file?

mhkim-anl: @veprbl @sly2j 
Yes, I uploaded the LUT file to `epic-data`, but I wasn't aware that I also need to update `calibrations.xml`. Thanks for letting me know. I've submitted a PR to update it and added you as reviewers.

src/algorithms/calorimetry/EdepToNpeConversion.cc
```diff
@@ -0,0 +1,147 @@
...
+    try {
+      m_field_idxs.push_back(m_id_dec->index(field));
+    } catch (...) {
+      throw std::runtime_error(
+          fmt::format("Field {} not found in idSpec of readout {}", field, m_cfg.readout));
+    }
+  }
+
+  // Load the LUT and parses each line into a lookup key and a conversion factor
+  std::string filename = fmt::format("calibrations/{}", m_cfg.edep_to_npe_filename);
```
ssedd1123: Since the directory "calibrations/" is included in EcalBarrelScFi_edep_to_npe_filename, there is no need to append "calibrations/{}" here. Just `std::string filename = m_cfg.edep_to_npe_filename` is sufficient.

mhkim-anl: Thank you for pointing that out. I’ve updated the code accordingly.

wdconinc: Treating filename string as something that `calibrations/` can simply be prepended to is fragile. Absolute file path are semantically acceptable but would lead to likely failure.

src/algorithms/calorimetry/EdepToNpeConversion.h
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2026 Minho Kim
+
+#pragma once
+
+#include <DD4hep/Detector.h>
+#include <DD4hep/IDDescriptor.h>
+#include <Parsers/Primitives.h>
+#include <algorithms/algorithm.h>
+#include <algorithms/geo.h>
```
ssedd1123, line 10:
```suggestion
#include <algorithms/geo.h>
#include <edm4hep/CaloHitContributionCollection.h>
```

src/algorithms/calorimetry/EdepToNpeConversion.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "algorithms/calorimetry/EdepToNpeConversionConfig.h"
+#include "algorithms/interfaces/UniqueIDGenSvc.h"
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+using EdepToNpeConversionAlgorithm = algorithms::Algorithm<
+    algorithms::Input<edm4hep::EventHeaderCollection, edm4hep::SimCalorimeterHitCollection>,
+    algorithms::Output<edm4hep::SimCalorimeterHitCollection>>;
```
ssedd1123, line 28:
```suggestion
    algorithms::Output<edm4hep::SimCalorimeterHitCollection,
       edm4hep::CaloHitContributionCollection>>;
```

src/algorithms/calorimetry/EdepToNpeConversion.h
```diff
@@ -0,0 +1,51 @@
...
+
+class EdepToNpeConversion : public EdepToNpeConversionAlgorithm,
+                            public WithPodConfig<EdepToNpeConversionConfig> {
+
+public:
+  EdepToNpeConversion(std::string_view name)
+      : EdepToNpeConversionAlgorithm{name, {"EventHeader", "inputHits"}, {"outputHits"}, {}} {}
```
ssedd1123, line 35:
```suggestion
      : EdepToNpeConversionAlgorithm{name, {"EventHeader", "inputHits"}, {"outputHits", "outputHitContributions"}, {}} {}
```

src/algorithms/calorimetry/EdepToNpeConversion.cc
```diff
@@ -0,0 +1,147 @@
...
+  }
+  if (m_edep_to_npe_lut.empty()) {
+    throw std::runtime_error(fmt::format("LUT file {} contains no entries", filename));
+  }
+} // EdepToNpeConversion:init
+
+void EdepToNpeConversion::process(const EdepToNpeConversion::Input& input,
+                                  const EdepToNpeConversion::Output& output) const {
+  const auto [headers, inhits] = input;
+  auto [outhits]               = output;
```
ssedd1123, line 99:
```suggestion
  auto [outhits, outcontribs]               = output;
```

src/algorithms/calorimetry/EdepToNpeConversion.cc
```diff
@@ -0,0 +1,147 @@
...
+    if (npe == 0) {
+      continue;
+    }
+
+    auto out_hit = outhits->create();
+    out_hit.setCellID(hit.getCellID());
+    out_hit.setEnergy(static_cast<float>(npe));
+    out_hit.setPosition(hit.getPosition());
+    for (const auto& contrib : hit.getContributions()) {
+      out_hit.addToContributions(contrib);
```
ssedd1123:
```suggestion
      auto out_contrib = outcontribs->create();
      out_contrib.setPDG(contrib.getPDG());
      out_contrib.setEnergy(contrib.getEnergy());
      out_contrib.setTime(contrib.getTime());
      out_contrib.setStepPosition(contrib.getStepPosition());
      out_contrib.setStepLength(contrib.getStepLength());
      out_contrib.setParticle(contrib.getParticle());
      out_hit.addToContributions(out_contrib);
```

src/factories/calorimetry/EdepToNpeConversion_factory.h
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2026 Minho Kim
+
+#pragma once
+
```
ssedd1123:
```suggestion
#include <edm4hep/CaloHitContributionCollection.h>
```

src/factories/calorimetry/EdepToNpeConversion_factory.h
```diff
@@ -0,0 +1,47 @@
...
+public:
+  using AlgoT = eicrecon::EdepToNpeConversion;
+
+private:
+  std::unique_ptr<AlgoT> m_algo;
+
+  PodioInput<edm4hep::EventHeader> m_in_headers{this};
+  PodioInput<edm4hep::SimCalorimeterHit> m_in_hits{this};
+  PodioOutput<edm4hep::SimCalorimeterHit> m_out_hits{this};
```
ssedd1123, line 22:
```suggestion
  PodioOutput<edm4hep::SimCalorimeterHit> m_out_hits{this};
  PodioOutput<edm4hep::CaloHitContribution> m_out_contribs{this};
```

src/factories/calorimetry/EdepToNpeConversion_factory.h
```diff
@@ -0,0 +1,47 @@
...
+public:
+  void Configure() {
+    m_algo = std::make_unique<AlgoT>(GetPrefix());
+    m_algo->level(static_cast<algorithms::LogLevel>(logger()->level()));
+    m_algo->applyConfig(config());
+    m_algo->init();
+  }
+
+  void Process(int32_t /* run_number */, uint64_t /* event_number */) {
+    m_algo->process({m_in_headers(), m_in_hits()}, {m_out_hits().get()});
```
ssedd1123, line 42:
```suggestion
    m_algo->process({m_in_headers(), m_in_hits()}, {m_out_hits().get(), m_out_contribs().get()});
```

src/detectors/BEMC/BEMC.cc
```diff
@@ -125,8 +130,28 @@ void InitPlugin(JApplication* app) {
       },
       app // TODO: Remove me once fixed
       ));
+  app->Add(new JOmniFactoryGeneratorT<EdepToNpeConversion_factory>(
+      "EcalBarrelScFiPNpeHits", {"EventHeader", "EcalBarrelScFiPAttenuatedHits"},
+      {"EcalBarrelScFiPNpeHits"},
```
ssedd1123, line 131:
```suggestion
      {"EcalBarrelScFiPNpeHits", "EcalBarrelScFiPNpeHitContributions"},
```

src/detectors/BEMC/BEMC.cc
```diff
@@ -125,8 +130,28 @@ void InitPlugin(JApplication* app) {
...
+      {
+          .readout              = "EcalBarrelScFiHits",
+          .edep_to_npe_fields   = EcalBarrelScFi_edep_to_npe_fields,
+          .edep_to_npe_filename = EcalBarrelScFi_edep_to_npe_filename,
+      },
+      app // TODO: Remove me once fixed
+      ));
+  app->Add(new JOmniFactoryGeneratorT<EdepToNpeConversion_factory>(
+      "EcalBarrelScFiNNpeHits", {"EventHeader", "EcalBarrelScFiNAttenuatedHits"},
+      {"EcalBarrelScFiNNpeHits"},
```
ssedd1123, line 141:
```suggestion
      {"EcalBarrelScFiNNpeHits", "EcalBarrelScFiNNpeHitContributions"},
```

ssedd1123 (changes requested):
Please correct me if I am wrong, but I think you need to generate a new hit contribution as output. In `EdepToNpeConversion::process()`, you discard hits when `npe==0`, so the hit contributions no longer align with that from the attenuated hits. I have a hard time doing reco-MC associations without the conversion class generating it's own hit contributions. 

I suggest that we add hit contributions to the conversion class.

src/detectors/BEMC/BEMC.cc
ssedd1123, line 225: In the P side when you create `EcalBarrelScFiPCALOROCHits`, you take `EcalBarrelScFiPCombinedPulsesWithNoise` as input. However, in the N side, you take `EcalBarrelScFiNCombinedPulses` without noise. Do you want to incorporate noise at this stage? I am fine either way, but I feel like you need to be consistent on both P and N side.

src/algorithms/calorimetry/EdepToNpeConversion.cc
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    if (npe == 0) {
+      continue;
+    }
+
+    auto out_hit = outhits->create();
+    out_hit.setCellID(hit.getCellID());
+    out_hit.setEnergy(static_cast<float>(npe));
+    out_hit.setPosition(hit.getPosition());
```
ruse-traveler: For consideration: you can also make use of `clone()` to duplicate a hit, and then modify only the fields you need! For example (c.f. [PFA1a](https://github.com/eic/EICrecon/blob/c7e1abd753c9984bf9792e6d6d50b2e9e7307af0/src/algorithms/particle_flow/TrackClusterSubtractor.cc#L118)):
```suggestion
    auto out_hit = hit.clone();
    out_hit.setEnergy(static_cast<float>(npe));
    outhits->push_back(out_hit);
```
This should also preserve the relations...

src/algorithms/calorimetry/EdepToNpeConversion.cc
```diff
@@ -0,0 +1,155 @@
...
+    for (const auto& contrib : hit.getContributions()) {
+      auto out_contrib = outcontribs->create();
+      out_contrib.setPDG(contrib.getPDG());
+      out_contrib.setEnergy(contrib.getEnergy());
+      out_contrib.setTime(contrib.getTime());
+      out_contrib.setStepPosition(contrib.getStepPosition());
+      out_contrib.setStepLength(contrib.getStepLength());
+      out_contrib.setParticle(contrib.getParticle());
+      out_hit.addToContributions(out_contrib);
+    }
```
ruse-traveler: Like I mention in response to @ssedd1123's comment, you shouldn't need to duplicate the original contributions. I think it would suffice to either add the old contribution as a relation (`out_hit.addToContributions(contrib);`) or make use of `clone()`!

src/algorithms/calorimetry/EdepToNpeConversion.cc
```diff
@@ -0,0 +1,155 @@
...
+  for (const auto& field : m_cfg.edep_to_npe_fields) {
+    try {
+      m_field_idxs.push_back(m_id_dec->index(field));
+    } catch (...) {
+      throw std::runtime_error(
+          fmt::format("Field {} not found in idSpec of readout {}", field, m_cfg.readout));
+    }
+  }
+
+  // Load the LUT and parses each line into a lookup key and a conversion factor
```
ruse-traveler, line 61: So this isn't a show-stopper and I'm fine shipping the code as-is, but wanted to flag this as a potential improvement for later: this likely won't be the last LUT we use, so it might be worthwhile to eventually expand our LUT service into something more general beyond the [`PIDLookupTable`](https://github.com/eic/EICrecon/blob/main/src/services/pid_lut/PIDLookupTable.h). A lot of the boilerplate here might be able to moved into a central service....

ruse-traveler:
> Please correct me if I am wrong, but I think you need to generate a new hit contribution as output. In `EdepToNpeConversion::process()`, you discard hits when `npe==0`, so the hit contributions no longer align with that from the attenuated hits. I have a hard time doing reco-MC associations without the conversion class generating it's own hit contributions.
> 
> I suggest that we add hit contributions to the conversion class.

You shouldn't need to generate new contributions, all that's needed is to make sure the surviving hits' relations are preserved (unless I'm missing something in the reco-MC associations). You could do this by either manually adding the contributions from the old hit to the new one, or by using `clone()`!

ruse-traveler (dismissed):
Thanks for the development @mhkim-anl! This is very good to see! Overall, I'm pretty happy with this, and have just a couple minor comments to consider!

src/algorithms/calorimetry/EdepToNpeConversion.cc
ssedd1123: I can explain why I wanted a cloned contribution list. Currently this class generates `EcalBarrelScFi{P,N}NpeHits.contributions_begin`, `EcalBarrelScFi{P,N}NpeHits.contributions_end` and `_EcalBarrelScFi{P, N}NpeHitContributions_particle.index`. The last one is part of the cloned contribution list. 

What I observed is that if we cloned the list, "contributions_begin" and "contributions_end" will corresponds to the index of `_EcalBarrelScFi{P, N}NpeHitContributions_particle.index`. I can obtain the index of the MC particle that give rise to the Npe hits via,

```
for(int i = EcalBarrelScFi{P,N}NpeHits.contributions_begin[0]; i < EcalBarrelScFi{P,N}NpeHits.contributions_end[0]; ++i)  {
   int mcID = _EcalBarrelScFi{P, N}NpeHitContributions_particle.index[i];
   // mcID is the index of MC particles that give rise to EcalBarrelScFi{P,N}NpeHits[0]
}
```

But if the contribution list is not cloned here, then I can only use `_EcalBarrelScFi{P, N}AttenuatedHitContributions_particle.index`, but that list order hits differently. I can't just replace `_EcalBarrelScFi{P, N}NpeHitContributions_particle.index` with `_EcalBarrelScFi{P, N}AttenuatedHitContributions_particle.index` and expect the loop to be correct. I think that's because this class skips all hits with `npe == 0`, which messes up with all the index ordering.

wdconinc: Yeah, this should not be needed, and certainly not for reasons of convenience and not for the largest collection we have in all of epic...

ssedd1123: I guess I am not reading things from the right branch. Sorry for my ignorance, but which branch should I read to translate Npe contribution indices to MC indices? I genuinely don't know, I really appreciate your help.

wdconinc: Any contribution (a truth quantity) should have a direct link to the MCParticle: https://github.com/key4hep/EDM4hep/blob/8d00b4cd1002ba9200bfc8fed537b3f1b5bb7f74/edm4hep.yaml#L346

ruse-traveler: > I guess I am not reading things from the right branch. Sorry for my ignorance, but which branch should I read to translate Npe contribution indices to MC indices? I genuinely don't know, I really appreciate your help.

Sure! So if you don't create a new contribution collection, you would indeed use the `_EcalBarrelScFi{P,N}AttenuatedHitContributions*` branches. However, you would need to use the values in `EcalBarrelScFi{P,N}NpeHits.contributions_{begin,end}` when looping over the contributions. After calling `clone()`, the values stored in those branches correspond to the indices of the objects in the collection you're relating to.
```c++
for(int i = EcalBarrelScFi{P,N}NpeHits.contributions_begin[0]; i < EcalBarrelScFi{P,N}NpeHits.contributions_end[0]; ++i)  {
   int mcID = _EcalBarrelScFi{P,N}AttenuatedHitContributions_particle.index[i];
   // mcID is the index of MC particles that give rise to EcalBarrelScFi{P,N}NpeHits[0]
}
```

(BTW you can always double check which collection a relation is in using the `podio_metadata` tree. There's an example of how to work with it in the Jet/HF snippets area)
https://github.com/eic/snippets/blob/f2f164f405ef09df6252548ff371e830af6e1beb/JetsAndHF/eFinder/find_ScatElectron.C#L98

ruse-traveler: Looking back at the attenuation algorithm, though, I didn't catch that the contributions are being duplicated there as well... Flagging this as something we should fix later: unless we need to modify the contributions somehow, it would be ideal for the relations to always point back to the `EcalBarrelHitContributions` collection.

ssedd1123: I tried what you've suggested, but it doesn't work. `EdepToNpeConversion` skips hit when `npe == 0`, so it messes up the contiguous index. I can see that `EcalBarrelScFi{P,N}NpeHits.contributions_begin` and `EcalBarrelScFi{P,N}AttenuatedHits.contributions_begin` doesn't always agree even when they corresponds to the same underlying hit. 

Let me double check. That's what I saw last time I tried this.

ssedd1123: @ruse-traveler I think making use of `clone` is fine, but my test results seems to indicate that "add the old contribution as a relation (out_hit.addToContributions(contrib);)" doesn't work. 

Below is what I see on my machine. Unfortunately I can't provide a fully reproducible example because the DIS files I'm using are local to my machine, but I run a few DIS events with eicrecon and compare the resultant `EcalBarrelScFiNAttenuatedHits` and `EcalBarrelScFiNNpeHits` to check whether their contribution indices stay aligned.

In your suggestion, the only change is to replace `_EcalBarrelScFi{P,N}NpeHitContributions` with `_EcalBarrelScFi{P,N}AttenuatedHitContributions`, while continuing to use `EcalBarrelScFi{P,N}NpeHits.contributions_begin/end`. For this to work, `EcalBarrelScFi{P,N}NpeHits.contributions_begin` would need to match `EcalBarrelScFi{P,N}AttenuatedHits.contributions_begin` for corresponding hits. We can identify corresponding hits by comparing their `cellID`.

To test this, I removed all of the `out_contrib` logic and simply used

```cpp
out_hit.addToContributions(contrib);
```

I recompile and run `eicrecon` on my DIS file. In the output file, the first two entries of

```cpp
events->Scan("EcalBarrelScFiNAttenuatedHits.contributions_begin:EcalBarrelScFiNAttenuatedHits.cellID")
```

are

```text
***********************************************
*    Row   * Instance * EcalBarre * EcalBarre *
***********************************************
*        0 *        0 *         0 *  18902377 *
*        0 *        1 *         1 *  18894441 *
```

while the first two entries of

```cpp
events->Scan("EcalBarrelScFiNNpeHits.contributions_begin:EcalBarrelScFiNNpeHits.cellID")
```

are

```text
***********************************************
*    Row   * Instance * EcalBarre * EcalBarre *
***********************************************
*        0 *        0 *         0 *  18894441 *
*        0 *        1 *         1 *  69300329 *
```

Notice that the hit with `cellID = 18894441` appears as instance **1** in `EcalBarrelScFiNAttenuatedHits` but instance **0** in `EcalBarrelScFiNNpeHits`. Their `contributions_begin` values are different (`1` vs. `0`), even though they should correspond to the same underlying MC hits.

Because of this, I don't think we can simply replace `_EcalBarrelScFi{P,N}NpeHitContributions` with `_EcalBarrelScFi{P,N}AttenuatedHitContributions` while continuing to use the indices from `EcalBarrelScFi{P,N}NpeHits.contributions_begin/end`. Unless I'm missing something, it looks like the filtering of `npe == 0` hits changes the ordering of the hit collection, which also shifts the contribution indices.

ruse-traveler: Oh shoot! I'm a dummy: I forgot that you need to go through the `_EcalBarrelScFi{P,N}NpeHits_contributions` branch 😩 . I think the snippet might actually be this:
```c++
for(int i = EcalBarrelScFi{P,N}NpeHits.contributions_begin[0]; i < EcalBarrelScFi{P,N}NpeHits.contributions_end[0]; ++i)  {
   int conID = _EcalBarrelScFi{P,N}NpeHitContributions.index[i]; 
   int mcID = _EcalBarrelScFi{P,N}AttenuatedHitContributions_particle.index[conID];
   // mcID is the index of MC particles that give rise to EcalBarrelScFi{P,N}NpeHits[0]
}
```

(Note that the `_EcalBarrelScFi{P,N}NpeHits_contributions` branch will be generated regardless of whether you're creating a new collection of contributions, though)

ssedd1123: @ruse-traveler Thank you very much for your help! I think that's what I am missing. I think it is working now, let's remove the duplicated contributions.

ruse-traveler: 🥳Glad to hear! And apologies for the confusion!!

src/detectors/BEMC/BEMC.cc
mhkim-anl: Thank you for pointing that out. I agree that both should be consistent. I'll make them consistent when I update the `CALOROCDigitization` soon.

src/algorithms/calorimetry/EdepToNpeConversion.cc
mhkim-anl: Thank you for the suggestion. It helped make the code more correct and simpler. I've applied your comments.

src/algorithms/calorimetry/EdepToNpeConversion.cc
mhkim-anl: Thank you for the discussion. Following your conclusion, the algorithm now uses `clone()` to preserve the hit contribution relations, and the separate `NpeHitContributions` has been removed accordingly.

src/detectors/BEMC/BEMC.cc
```diff
@@ -296,46 +321,6 @@ void InitPlugin(JApplication* app) {
       {"EcalBarrelScFiClusters", "EcalBarrelScFiClusterLinks", "EcalBarrelScFiClusterAssociations"},
       {.longitudinalShowerInfoAvailable = true, .energyWeight = "log", .logWeightBase = 6.2}, app));
 
-  // Imaging TopoClustering on ScFi
```
wdconinc, line 293: Please fixup the code segments that were reverted in a merge from main into this branch. This may not be the only instance.

wdconinc (dismissed):

src/algorithms/calorimetry/EdepToNpeConversion.cc
mhkim-anl: Thank you for the suggestion. I agree with your point, and I'll keep this in mind so that the Edep-to-Npe conversion can also eventually be handled through a central service.

src/detectors/BEMC/BEMC.cc
mhkim-anl: Thank you for catching this. I'm sorry for my mistake. I've restored changes from main in `BEMC.cc` and `JEventProcessorPODIO.cc`.

mhkim-anl:
Sorry for the late updates to the remaining comments, and it caused the previous approval to be dismissed. I think I've now addressed all the remaining comments. If everything looks good, I would appreciate it if you could approve this PR. Thanks!

wdconinc:
I'm a bit confused by the EcalBarrelScFiNPulses results in the capybara plots for this PR: https://eicrecon.epic-eic.org/pr/2732/capybara/rec_dis_10x130_minQ2=1_craterlake_10x130/index.html#EcalBarrelScFiNPulses

It seems the number of EcalBarrelScFiNPulses goes up with this PR, but the new Edep-to-Npe algorithm that is inserted can only effect a reduction of the number of hits (through the npe == 0 continuation), not an increase in number of hits (and https://eicrecon.epic-eic.org/pr/2732/capybara/rec_dis_10x130_minQ2=1_craterlake_10x130/index.html#EcalBarrelScFiNAttenuatedHits indeed decreases to https://eicrecon.epic-eic.org/pr/2732/capybara/rec_dis_10x130_minQ2=1_craterlake_10x130/index.html#EcalBarrelScFiNNpeHits by a factor of about 6).

(And the PulseGeneration algorithm should create also at most one EcalBarrelScFiNPulses entry per EcalBarrelScFiNNpeHits entry.)

Do you have an explanation for the EcalBarrelScFiNPulses changes?

ssedd1123 (approved):

mhkim-anl:
Sorry for the delayed response. The reason is `ignore_thres` in `PulseGeneration` algorithm. Previously, I didn't consider the relation between the energy deposit and Npe carefully, so `ignore_thres` was set relatively higher than the level of 1 Npe. Now, with the Edep-to-Npe conversion, we can set `ignore_thres` more reasonably as 1 Npe because the parameters in the `PulseGeneration` algorithm was set so that `x` Npe produces a pulse with a height of `x`.

wdconinc:
@sly2j Please don't add PRs to the merge queue when there are open question, e.g. https://github.com/eic/EICrecon/pull/2732#issuecomment-5245384720.

wdconinc:
> Sorry for the delayed response. The reason is `ignore_thres` in `PulseGeneration` algorithm. Previously, I didn't consider the relation between the energy deposit and Npe carefully, so `ignore_thres` was set relatively higher than the level of 1 Npe. Now, with the Edep-to-Npe conversion, we can set `ignore_thres` more reasonably as 1 Npe because the parameters in the `PulseGeneration` algorithm was set so that `x` Npe produces a pulse with a height of `x`.

I see now what happened there. For future reference, please(!) factorize your commits so things like this are either avoided by not having unrelated or un-called-out changes in a single PR, or ensure that they are explicitly called out in the PR description.

wdconinc (approved):

mhkim-anl:
I'm sorry, I underestimated the impact of changing `ignore_thres`. I'll not only factorize my commits more carefully but also add more detailed explanations in the PR description to avoid any code change that isn't clearly explained in the future.

Issue #2689, opened 2026-06-03
Anti-pattern in FarForwardNeutralsReconstruction


wdconinc:
https://github.com/eic/EICrecon/blob/195966d41899800eaeca6fbafd63cd53735e6543/src/algorithms/reco/FarForwardNeutralsReconstruction.cc#L246

The process function in FarForwardNeutralsReconstruction should be 4 instances of a single factory to avoid bundling unnecessary sequencing in a single algorithm.

PR #2736, opened 2026-07-08
Refactor far-forward neutrals reconstruction 

baptistefraisse:
# Summary

This PR refactors neutral and Lambda reconstruction.

## 1. Factorize `FarForwardNeutralsReconstruction` (issue #2689)

Following the discussions in PR #2622 followed by issue #2689, the neutral reconstruction has been refactored so that the same algorithm is used for each calorimeter independently instead of processing everything in a single instance.

~~## 2. Use index-based cluster loops~~

~~The neutral reconstruction now iterates over clusters using index-based loops instead of object-based iteration. This was motivated by the investigation of the Capybara single-threaded / multi-threaded mismatch, as it removes one possible source of non-deterministic behavior.~~

## 3. Improve Lambda candidate selection

The Lambda selection has been slightly modified to prioritize the candidate with the lowest chi-2. The selection also continues scanning the remaining candidates when appropriate instead of stopping after the first valid one.

@ruse-traveler @veprbl @DraTeots @wdconinc


wdconinc:
Range-based for loops are deterministic, so the motivation doesn't seem to make sense. Indices are prone to abuse (and incorrect typing as `int` when they should likely be `size_t` throughout, removing the needs for static casts).

wdconinc:
Please fix the merge conflicts first. They will remove the need for us to comment on things that pre-commit will fix.

baptistefraisse:
Sorry, my branch was not up to date with main. I restarted the branch from the current main and reapplied the changes cleanly. @wdconinc yes, the range-based loops themselves are deterministic. The motivation behind index-based loops was to ensure that cluster objects / their paths were not involved in the ST/MT discrepancy. But okay, if there's no benefit to keeping it this way, I stick to '&cluster' loops.

Issue #2323, opened 2026-01-12
Calorimeters hits associations are not filled


DraTeots:
Bump. So far I see problems in calorimeters: 

### Environment: (where does this bug occur, have you tried other environments)
- the latest simulation campaign background data

e.g. 

```
root://dtn-eic.jlab.org//volatile/eic/EPIC/RECO/25.10.4/epic_craterlake/Bkg_Exactly1SignalPer2usFrame/SIDIS/D0_ABCONV/pythia8.306-1.1/10x100/q2_1/hiDiv/pythia8.306-1.1_D0_10x100_hiDiv_q2_1_run10.2110.eicrecon.edm4eic.root ./
```

### Actual Result: (the problem)

When one pulls ```edm4eic::MCRecoCalorimeterHitAssociationCollection```

You basically has: 

```c++
hit_assoc.getSimHit().isAvailable() == false
```

For calorimeters (collection name): 

- LFHCALRawHitAssociations 
- HcalEndcapPInsertRawHitAssociations
- HcalEndcapNRawHitAssociations
- EcalFarForwardZDCRawHitAssociations
- EcalEndcapPRawHitAssociations
- EcalEndcapNRawHitAssociations
- EcalBarrelScFiRawHitAssociations
- EcalBarrelImagingRawHitAssociations
- B0ECalRawHitAssociations

veprbl:
Or we could make another collection to summarize the original edm4hep::SimCalorimeterHit, but withtout the contributions. But you need to explain your use case.

veprbl:
You could also try to approach this using real island clusters and rely on weighted associations.

You have an interesting plot. What are the hits with $1<z<2$ m?

ruse-traveler:
BTW I've opened a PR to partially address #2072 in #2336!

PR #2770, opened 2026-07-23
Add CalorimeterCALOROCCalibration

ssedd1123:

Added CalorimeterCALOROCCalibration class to convert the CALOROC ADC/TDC values back to CalorimeterHits, which replaces CalorimeterHitReco and CalorimeterHitDigi. Downstream clusterization class is now taking output from the full digitization chain for a more realistic simulation.

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

First iteration of the CalorimeterCALOROCCalibration is written manually, but AI is used to fix bugs multiple times during development stage when it fails to compile.


src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
```diff
@@ -0,0 +1,491 @@
...
+    // link to parents
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitLink> links_staging;
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitAssociation>
+        rawassocs_staging;
+
+    for (bool NSide : std::vector<bool>{true, false}) {
+      const auto& hits = NSide ? pulseN.getCalorimeterHits() : pulseP.getCalorimeterHits();
+      for (const auto& hit : hits) {
+        // if link is already covered, don't add again
+        if (links_staging.find(hit.getObjectID()) != links_staging.end())
```
wdconinc: Can you run this process function through an AI model with a prompt that instructs it to pinpoint where different data structures can be used to avoid inefficiencies? Running a find in an inner loop is polynomially inefficient. A set with erase may naturally remove already found links.

ssedd1123: links_staging is an unordered_map. It uses hash map, so the time complexity of a find should be constant on average and not polynomial. It may get worse if there are a lot of hash collisions, but for such simple class link ObjectID, I don't think it's a concern.

wdconinc: My comment wasn't specific to this one line (where podio::LinkNavigator should be considered). For example, summing stuff in a for loop is also not ideal and should consider accumulate.

wdconinc: Also, even a constant search term is more expensive than a search that does not have to be done. Pulse sample iteration has in the past been a resource limiting factor in the TOF and the original implementation needed changes by the software group to make it work faster. I'd like to avoid us having to do the fixes after the fact this time around.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+  m_slope     = m_cfg.slope;
+  m_intercept = m_cfg.intercept;
+
+  info("calibration slope = {}, intercept = {}", m_slope, m_intercept);
+
+  // do not get the layer/sector ID if no readout class provided
+  if (m_cfg.readout.empty()) {
+    error("You MUST provide the name of the readout to decode CellID.");
+  }
+
```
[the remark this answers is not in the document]
ssedd1123: @copilot I think the "error()" function also terminates the routine by throwing exceptions. I don't think it's necessary to throw another std::runtime_error.

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
ssedd1123: I believe your concern is that SimPulse is a heavy object and that accessing its contents may be slow. If that is the case, I don’t think it is an issue here, because I am not looping over the pulse amplitudes. I only read the contributing SimCalorimeterHits from the pulse class for MC-reco association. I never access the pulse amplitude itself. The call to pulse.getCalorimeterHits() only returns the MC hits responsible for generating that hit.

If the goal is to construct a link between CalorimeterRawHits and SimCalorimeterHits, then the algorithm has to be at least O(N(SimCalorimeterHits)), right? Even filling a vector with all SimCalorimeterHits would already have time complexity O(N(SimCalorimeterHits)). So from a time-complexity perspective, I think this is already as fast as possible.

Although std::accumulate might improve the leading constant factor, the overall time complexity would still be O(N(SimCalorimeterHits)). We should also avoid premature optimization, so if the time complexity remains unchanged, I think this loop can stay, especially since I don't think CalorimeterCALOROCCalibration i the bottleneck in the BEMC chain? I haven't ran a full blown profiler on the code, but running EICrecon with and without CalorimeterCALOROCCalibration doesn't affect runtime that much.

ssedd1123: I guess I can read upstream classes such as `EcalBarrelScFiNNpeHits` for MC information if you want to get rid of reading from pulses, but I think the time complexity will stay the same.

src/algorithms/calorimetry/CalorimeterCALOROCCalibrationConfig.h
```diff
@@ -0,0 +1,87 @@
...
+  std::vector<std::string> edep_to_npe_fields{""};
+
+  enum class ProxyType { sum = 0, simpson = 1, templateFit = 2 } proxy_type = ProxyType::sum;
+
+  bool timeWalkCor;
+  bool usePulsePos;
+  bool usePulseNPE;
+
+  uint16_t highGainDR; // high gain dynamic range
+  double gainRatio;    // gain ratio of high gain to low gain, it should be < 1
```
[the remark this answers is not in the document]
ssedd1123: Value initialized. Also the comment is wrong, it should be low/high gain, not the other way round.

src/detectors/BEMC/BEMC.cc
```diff
@@ -239,47 +233,41 @@ void InitPlugin(JApplication* app) {
...
+          .maskPos              = "xy",
+          .maskPosFields        = {"fiber", "z"},
+          .edep_to_npe_filename = EcalBarrelScFi_edep_to_npe_filename,
+          .edep_to_npe_fields   = EcalBarrelScFi_edep_to_npe_fields,
+          .timeWalkCor          = true,
+          .usePulsePos          = false,
+          .usePulseNPE          = false,
+          .highGainDR           = 1023,
+          .gainRatio = EcalBarrelScFi_dyRangeLowGainADC / EcalBarrelScFi_dyRangeHighGainADC,
+          .attenuationReferencePositionNamePos = "EcalBarrel_LightGuide_PositivePosZ",
```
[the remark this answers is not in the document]
ssedd1123: The document is wrong, but the code is right.

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
```diff
@@ -0,0 +1,502 @@
...
+      m_cfg.edep_to_npe_filename.empty())
+    error("You MUST provide edep_to_npe files and filename for layer-by-layer calibration.");
+
+  for (const auto& field : m_cfg.edep_to_npe_fields) {
+    auto field_idx = id_dec->index(field);
+    m_field_idxs.push_back(field_idx);
+  }
+
+  std::string filename = fmt::format("calibrations/{}", m_cfg.edep_to_npe_filename);
+  std::ifstream infile(filename);
```
veprbl:
```suggestion
  std::ifstream infile(m_cfg.edep_to_npe_filename);
```
Move `calibrations/` into the default path

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
veprbl, line 1: Can we get a unit test for this?

src/algorithms/calorimetry/CalorimeterCALOROCRecoConfig.h
```diff
@@ -0,0 +1,88 @@
...
+  std::string sectorField{""};
+
+  // name of detelment or fields to find the local detector (for global->local transform)
+  // if nothing is provided, the lowest level DetElement (from cellID) will be used
+  std::string localDetElement{""};
+  std::vector<std::string> localDetFields{};
+  std::string maskPos{""};
+  std::vector<std::string> maskPosFields{};
+
+  std::string edep_to_npe_filename{""};
```
veprbl, line 26: Same comment as in https://github.com/eic/EICrecon/pull/2732#discussion_r3639988681

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
```diff
@@ -0,0 +1,502 @@
...
+          break;
+        case CalorimeterCALOROCCalibrationConfig::ProxyType::templateFit:
+          error("Proxy type not implemented.");
+          break;
+        case CalorimeterCALOROCCalibrationConfig::ProxyType::simpson:
+          error("Proxy type not implemented.");
+          break;
+        default:
+          error("Proxy type not implemented.");
+        }
```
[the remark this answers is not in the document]
ssedd1123: @copilot I think the "error()" function also terminates the routine by throwing exceptions. I don't think it's necessary to throw another std::runtime_error.

src/algorithms/calorimetry/CalorimeterCALOROCRecoConfig.h
ssedd1123: Updated to load lut files from the home directory as advised.

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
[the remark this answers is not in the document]
veprbl: No, it does not.

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
ssedd1123: I have updated the pull request so the class now reads from `EcalBarrelScFi*NpeHits` and not the pulse class. Hopefully that eliminates the concern on pulse class taking up too much resources. 

`RawCALORROCHit` class do not have MC-association information, therefore we cannot read `SimCalorimeterHit` contributions from just `EcalBarrelScFi*CALOROCHits`.

veprbl: So we don't plan to use pulse digitization now?
Asking in context of my question in https://chat.epic-eic.org/main/pl/3rp5kdtfji8zi8uuoud94mqexc

ssedd1123: We still use pulse digitization, since the calibration class still takes `EcalBarrelScFi*CALOROCHits` which are generated from digitized pulse. I am under the impression that @wdconinc don't want the calibration class to take the pulse class as input, that's why I modify it this way. 

However, I think we can drop the pulses from the default output. @mhkim-anl please confirm if that's the case.

mhkim-anl: I'm very sorry for the delayed response. Yes, Removing `EcalBarrelScFi{P, N}Pulses` from the default output would be fine, as long as `EcalBarrelScFi{P, N}CombinedPulses` and `EcalBarrelScFi{P, N}CombinedPulsesWithNoise` are retained.

veprbl: I apologize if I missed some context. My understanding was that the "calibration" factory was meant to take detector signals (recorded pulse or some other reduced form), and apply corrections to get the hits ready to be used for physics analysis. If pulse factory is/was exercised by this, then there is no need to store pulses directly and we could remove those from the output.

@mhkim-anl But do you actually analyze `CombinedPulses(WithNoise)` directly from the campaign files? If not, those can be removed, and you can always re-enable them for local studies.

mhkim-anl: @veprbl 
I apologize for the delayed response. Actually, we now mainly use locally produced output for our studies rather than campaign output. So I agree that they can be removed from the default output for now. If we need them to be included in the campaign output later, I will ask for them to be added back.

wdconinc: Why are you not using the production campaigns? Is there some problem or limitation with them? It doesn't help the rest of the collaboration if you only develop things in your own sandbox and don't feed them back to the production campaigns.

src/detectors/BEMC/BEMC.cc
```diff
@@ -239,47 +233,39 @@ void InitPlugin(JApplication* app) {
...
-      ));
-  app->Add(new JOmniFactoryGeneratorT<CalorimeterHitReco_factory>(
-      "EcalBarrelScFiRecHits", {"EcalBarrelScFiRawHits"}, {"EcalBarrelScFiRecHits"},
+
+  app->Add(new JOmniFactoryGeneratorT<CalorimeterCALOROCCalibration_factory>(
+      "EcalBarrelScFiCalibration",
+      {"EcalBarrelScFiPNpeHits", "EcalBarrelScFiPCALOROCHits", "EcalBarrelScFiNNpeHits",
+       "EcalBarrelScFiNCALOROCHits"},
+      {"EcalBarrelScFiRecHits", "EcalBarrelScFiRawHits", "EcalBarrelScFiRawHitLinks",
+       "EcalBarrelScFiRawHitAssociations"},
```
veprbl, line 237:
```
ERROR: branch _EcalBarrelScFiNCombinedPulsesWithNoise_calorimeterHits references unknown collectionID(s): [11030660]
ERROR: branch _EcalBarrelScFiNCombinedPulses_calorimeterHits references unknown collectionID(s): [11030660]
ERROR: branch _EcalBarrelScFiNPulses_calorimeterHits references unknown collectionID(s): [11030660]
ERROR: branch _EcalBarrelScFiPCombinedPulsesWithNoise_calorimeterHits references unknown collectionID(s): [2120591442]
ERROR: branch _EcalBarrelScFiPCombinedPulses_calorimeterHits references unknown collectionID(s): [2120591442]
ERROR: branch _EcalBarrelScFiPPulses_calorimeterHits references unknown collectionID(s): [2120591442]
```
The check performed by https://github.com/eic/EICrecon/blob/main/src/scripts/verify_for_dangling_references.py fails. Are we missing some entries in JEventProcessorPODIO.cc?

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
ssedd1123: I added a unit test on this class with the help of AI. Please take a look.

src/detectors/BEMC/BEMC.cc
ssedd1123: I added the Npe hits to JEventProcessorPODIO.cc. I hope this satisfies the requirement of this script.

wdconinc:
Should we talk about the name of this algorithm? I don't know if what this algorithm does is "calibration", and maybe what it actually does is hit "reconstruction" based on the results *from* calibration. In ePIC S&C we talk about calibration in two contexts: 1) the externally determined artifacts that are provided to interpret raw data into more physically interpretable data, 2) the phase after data taking when those artifacts are produced (and which notoriously has to happen within 3 weeks). So, should be reuse that word here in a context that is how we have traditionally use "reconstruction"?

ssedd1123:
> Should we talk about the name of this algorithm? I don't know if what this algorithm does is "calibration", and maybe what it actually does is hit "reconstruction" based on the results _from_ calibration. In ePIC S&C we talk about calibration in two contexts: 1) the externally determined artifacts that are provided to interpret raw data into more physically interpretable data, 2) the phase after data taking when those artifacts are produced (and which notoriously has to happen within 3 weeks). So, should be reuse that word here in a context that is how we have traditionally use "reconstruction"?

I have no qualms renaming it "CalorimeterCALOROCReco" class. If there are no objections to the name, I will rename it in my next push. 

Edit: The class is renamed.

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
mhkim-anl: The main reason we have recently relied on locally produced output is that most of our studies-validating each algorithm and developing this algorithm-are using the `EdepToNpe` algorithm, which has not yet been merged to main branch. I'll try to get it approved and merged as soon as possible and then return to using the campaign output.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+  m_reference_z_p = m_geo.detector()->constant<double>(m_cfg.attenuationReferencePositionNamePos) *
+                    edm4eic::unit::mm / dd4hep::mm;
+  m_reference_z_n = m_geo.detector()->constant<double>(m_cfg.attenuationReferencePositionNameNeg) *
+                    edm4eic::unit::mm / dd4hep::mm;
+
+  info("Pos reference z = {}", m_reference_z_p);
+  info("Neg reference z = {}", m_reference_z_n);
+
+  m_slope     = m_cfg.slope;
+  m_intercept = m_cfg.intercept;
```
ruse-traveler: Super minor point: is there a reason to store these as class members rather than just use the `m_cfg` members?

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+  try {
+    id_dec = id_spec.decoder();
+    if (!m_cfg.sectorField.empty()) {
+      sector_idx = id_dec->index(m_cfg.sectorField);
+      debug("Find sector field {}, index = {}", m_cfg.sectorField, sector_idx);
+    }
+    if (!m_cfg.layerField.empty()) {
+      layer_idx = id_dec->index(m_cfg.layerField);
+      debug("Find layer field {}, index = {}", m_cfg.layerField, sector_idx);
+    }
```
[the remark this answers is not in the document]
ruse-traveler: Second this suggestion!

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+    if (fields.empty()) {
+      local_mask = ~static_cast<decltype(local_mask)>(0);
+    }
+  }
+
+  // copy and pasted from Minho's custom PulseGenerator class
+  // Get the field indices and field-dependent conversion factors if necessary
+  if (m_cfg.readout.empty() || m_cfg.edep_to_npe_fields.empty() ||
+      m_cfg.edep_to_npe_filename.empty())
+    error("You MUST provide edep_to_npe files and filename for layer-by-layer calibration.");
```
ruse-traveler: Just for clarity: I'd suggest removing the readout check (since at this point it's confirmed the readout is not empty) and the reference to the custom class.
```suggestion
  // Get the field indices and field-dependent conversion factors if necessary
  if (m_cfg.edep_to_npe_fields.empty() || m_cfg.edep_to_npe_filename.empty()) {
    error("You MUST provide edep_to_npe files and filename for layer-by-layer calibration.");
  }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+    std::istringstream iss(line);
+    std::vector<int> keys;
+    for (std::size_t i = 0; i < m_cfg.edep_to_npe_fields.size(); i++) {
+      int value;
+      iss >> value;
+      keys.push_back(value);
+    }
+    double factor;
+    if (!(iss >> factor))
+      error("Malformed LUT file: {}", m_cfg.edep_to_npe_filename);
```
ruse-traveler: Looks like clang format missed some brackets!
```suggestion
    if (!(iss >> factor)) {
      error("Malformed LUT file: {}", m_cfg.edep_to_npe_filename);
    }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+
+double CalorimeterCALOROCReco::_sumADC(const edm4eic::RawCALOROCHit& ADC) const {
+  double sum = 0;
+
+  // check high gain ADC first. If it saturates, we switch to lowGainADC
+  for (const auto& bSample : ADC.getBSamples()) {
+    if (bSample.highGainADC >= m_cfg.highGainDR)
+      sum += static_cast<double>(bSample.lowGainADC);
+    else
+      sum += static_cast<double>(bSample.highGainADC) / m_cfg.gainRatio;
```
ruse-traveler: Same here!
```suggestion
    if (bSample.highGainADC >= m_cfg.highGainDR) {
      sum += static_cast<double>(bSample.lowGainADC);
    } else {
      sum += static_cast<double>(bSample.highGainADC) / m_cfg.gainRatio;
    }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+}
+
+double CalorimeterCALOROCReco::_timeWalkCorrection(double toa, double lowGainADC) const {
+  if (static_cast<double>(lowGainADC) - m_cfg.timeWalkCorrectionParameters[2] > 0)
+    return toa - (m_cfg.timeWalkCorrectionParameters[1] *
+                      pow(static_cast<double>(lowGainADC) - m_cfg.timeWalkCorrectionParameters[2],
+                          m_cfg.timeWalkCorrectionParameters[3]) +
+                  m_cfg.timeWalkCorrectionParameters[0]);
+  else
+    return toa;
```
ruse-traveler:
```suggestion
  if (static_cast<double>(lowGainADC) - m_cfg.timeWalkCorrectionParameters[2] > 0) {
    return toa - (m_cfg.timeWalkCorrectionParameters[1] *
                      pow(static_cast<double>(lowGainADC) - m_cfg.timeWalkCorrectionParameters[2],
                          m_cfg.timeWalkCorrectionParameters[3]) +
                  m_cfg.timeWalkCorrectionParameters[0]);
  } else {
    return toa;
  }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+  // find the relative toa
+  int samplePhase        = ADC.getSamplePhase();
+  int timeStamp          = ADC.getTimeStamp();
+  const auto& bSamples   = ADC.getBSamples();
+  size_t idx_toa         = 0;
+  uint16_t timeOfArrival = 0;
+  for (; idx_toa < bSamples.size(); ++idx_toa) {
+    timeOfArrival = bSamples.at(idx_toa).timeOfArrival;
+    if (timeOfArrival > 0)
+      break;
```
ruse-traveler:
```suggestion
    if (timeOfArrival > 0) {
      break;
    }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+    cellID2ADCNID[ADCN.getCellID()] = i;
+  }
+
+  for (const auto& npeHitP : *npeHitsP) {
+    auto cellID = npeHitP.getCellID();
+
+    // ignore data point if the hit is not showing up on all ADC and NpeHit collections
+    auto it = cellID2NpeHitNID.find(cellID);
+    if (it == cellID2NpeHitNID.end())
+      continue;
```
ruse-traveler:
```suggestion
    if (it == cellID2NpeHitNID.end()) {
      continue;
    }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+    const auto& npeHitN = npeHitsN->at(it->second);
+    it                  = cellID2ADCPID.find(cellID);
+    if (it == cellID2ADCPID.end())
+      continue;
```
ruse-traveler:
```suggestion
    if (it == cellID2ADCPID.end()) {
      continue;
    }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+    const auto& ADCP = ADCPs->at(it->second);
+    it               = cellID2ADCNID.find(cellID);
+    if (it == cellID2ADCNID.end())
+      continue;
```
ruse-traveler:
```suggestion
    if (it == cellID2ADCNID.end()) {
      continue;
    }
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+    // get layer and sector ID
+    const int lid = id_dec != nullptr && !m_cfg.layerField.empty()
+                        ? static_cast<int>(id_dec->get(cellID, layer_idx))
+                        : -1;
+    const int sid = id_dec != nullptr && !m_cfg.sectorField.empty()
+                        ? static_cast<int>(id_dec->get(cellID, sector_idx))
+                        : -1;
+
+    auto tP     = this->_toa(ADCP);
+    auto tN     = this->_toa(ADCN);
```
ruse-traveler: I'm not sure you need to use `this->` here (there's no potential for issues with templates), and it's not a convention we use anywhere else...
```suggestion
    auto tP     = _toa(ADCP);
    auto tN     = _toa(ADCN);
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+
+    // get position of the hit;
+    double zpos;
+
+    if (m_cfg.useNpeHitPos) {
+      zpos = static_cast<double>(npeHitP.getPosition().z);
+    } else {
+      if (m_cfg.timeWalkCor) {
+        tP = this->_timeWalkCorrection(tP, this->_sumADC(ADCP));
+        tN = this->_timeWalkCorrection(tN, this->_sumADC(ADCN));
```
ruse-traveler: Likewise here...
```suggestion
        tP = _timeWalkCorrection(tP, this->_sumADC(ADCP));
        tN = _timeWalkCorrection(tN, this->_sumADC(ADCN));
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+    const double eDep2NpeFactor = lut_it->second;
+    // find averaged energy
+    double npeP, npeN;
+    // get values from both sides in a loop
+    for (bool NSide : {true, false}) {
+      auto& ADC = NSide ? ADCN : ADCP;
+      auto& npe = NSide ? npeN : npeP;
+      switch (m_cfg.proxy_type) {
+      case CalorimeterCALOROCRecoConfig::ProxyType::sum:
+        npe = this->_sumADC(ADC);
```
ruse-traveler: Ditto...
```suggestion
        npe = _sumADC(ADC);
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,502 @@
...
+      }
+    }
+
+    // perform layer-by-layer energy correction
+    double chargeP = npeP / eDep2NpeFactor;
+    double chargeN = npeN / eDep2NpeFactor;
+
+    // attenuation correction
+    double corEP = this->_energyCor(m_reference_z_p, chargeP, zpos) * m_slope + m_intercept;
+    double corEN = this->_energyCor(m_reference_z_n, chargeN, zpos) * m_slope + m_intercept;
```
ruse-traveler:
```suggestion
    double corEP = _energyCor(m_reference_z_p, chargeP, zpos) * m_slope + m_intercept;
    double corEN = _energyCor(m_reference_z_n, chargeN, zpos) * m_slope + m_intercept;
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
```diff
@@ -0,0 +1,95 @@
...
+                               public WithPodConfig<CalorimeterCALOROCRecoConfig> {
+
+public:
+  CalorimeterCALOROCReco(std::string_view name)
+      : CalorimeterCALOROCRecoAlgorithm{
+            name,
+            {"inputNpeHitPCollection", "inputADCPCollection", "inputNpeHitNCollection",
+             "inputADCNCollection"},
+            {"outputRecHitCollection", "outputRawHitCollection", "outputRawLink", "outputRawAssoc"},
+            "Reconstruct hit from CALOROC ADC data with MC truth from NpeHits."} {}
```
ruse-traveler, line 48: I notice that this algorithm is very explicitly oriented towards the 2-sided hits of the BIC, which means this won't generalize to the other calorimeters which don't follow this readout scheme but still use the CALOROC.

There might be a way to reorganize this where the 2nd set of sim hits are optional, and if they're present then you extract the relevant info from both rather than just one. But I recognize that we're pretty late in development here, so a refactor in this PR might not be feasible.

My suggestion would be to consider clarifying the algorithm's scope in its name (maybe `CalorimeterFiberCALOROCReco` or something?) and flag generalization as a follow-up.

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
```diff
@@ -0,0 +1,95 @@
...
+
+  std::map<std::vector<int>, double> m_edep_to_npe_lut{};
+  std::vector<std::size_t> m_field_idxs{};
+
+  double _sumADC(const edm4eic::RawCALOROCHit& ADC) const;
+  double _simpson(const edm4eic::RawCALOROCHit& ADC) const;
+
+  double _energyCor(double referencePos, double energy, double z) const;
+  double _toa(const edm4eic::RawCALOROCHit& ADC) const;
+  double _timeWalkCorrection(double toa, double ADC) const;
```
ruse-traveler: Super minor point for readability: I'd suggest mirroring the function declaration order here in the implementation file!

src/algorithms/calorimetry/CalorimeterCALOROCRecoConfig.h
```diff
@@ -0,0 +1,86 @@
...
+  } else if (s == "simpson" or s == "1") {
+    proxyType = CalorimeterCALOROCRecoConfig::ProxyType::simpson;
+  } else if (s == "templateFit" or s == "2") {
+    proxyType = CalorimeterCALOROCRecoConfig::ProxyType::templateFit;
+  } else {
+    in.setstate(std::ios::failbit); // Set the fail bit if the input is not valid
+  }
+
+  return in;
+}
```
ruse-traveler, line 67: Not a show-stopper and I'm fine with shipping this code as-is, but I wanted to jot down my thoughts here: I assume this is following the patterns in `ImagingTopoClusterConfig` and `SiliconChargeSharingConfig`? Aside from a couple cases (those two, the old `IrtCherenkovParticleIdConfig`), our config classes are strict POD structures. It's definitely useful being able to map strings from the config file to an enum, so maybe there's a way to centralize this functionality and move it into the relevant algorithms?

ruse-traveler (commented):
This is very nice to see, @ssedd1123, @mhkim-anl! The functionality is extremely welcome! I have more specific comments below, but a few general thoughts:
- Building on some of @wdconinc's comments, the several loops to match hits based on cell ID suggest that we're missing some sim hit-raw hit links (namely the npe-to-raw connections). This would follow how we handle sim-to-raw connections elsewhere. Personally, though, I'm okay addressing this in a follow-up PR.
- The readout parsing in `init()` is _very_ verbose. It's not critical since almost all of it is behind `warning` or `info`, but I might suggest a pass to weed out some of the extra text.

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
ssedd1123: The function declaration order is revised. The function `_simpson` is removed as it is never implemented. Eventually though we do plan on switching to template fitting for pulse height, but that's something for the future.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: I removed `m_slope` and `m_intercept` as advised. I still keep `layer_idx` because I am a bit worried that the overhead of `id_dec->index(m_cfg.layerField)`. I inherited this part of code from `CalorimeterHitReco`, my thought process was if this part is good enough for that other class, it should be good enough for this class. I have no objections to cutting down the verbosity here though.

ruse-traveler: Makes sense! Thanks, @ssedd1123 !

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
ruse-traveler: Good catch! And yep! That's something for another day...

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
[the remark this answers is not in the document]
ssedd1123: `break` is now inserted in each switch case.

src/algorithms/calorimetry/CalorimeterCALOROCCalibration.cc
```diff
@@ -0,0 +1,502 @@
...
+                  m_cfg.timeWalkCorrectionParameters[0]);
+  else
+    return toa;
+}
+
+void CalorimeterCALOROCCalibration::process(
+    const CalorimeterCALOROCCalibration::Input& input,
+    const CalorimeterCALOROCCalibration::Output& output) const {
+
+  const auto [pulsesP, ADCPs, pulsesN, ADCNs]         = input;
```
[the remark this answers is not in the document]
ssedd1123: New unit test is implemented.

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
ssedd1123: I agree that cutting the class in two is a good idea. But to do that, I would like to know exactly where should I divide the class. Here's a list of things this class does,

1. Convert `RawCALOROCHits` to `CalorimeterRawHits` and `CalorimeterRecHits`.
2. Convert NPE to Edep via linear transformation and layer-by-layer gain correction using lut file.
3. Perform time-walk correction for time of arrival.
4. Estimate z-coordinate of a hit by comparing time of arrival on both ends.
5. Perform attenuation correction after knowing where the hit is along the length of the fiber.
6. Estimate energy with both ends E = sqrt(EL*ER).

If I have to divide it up now without known how any other calorimeter work, I would say job 1-3 can be a Reco class, and job 4-6 can be put in a separate `LRCombine` class. Can you confirm if that's a good way to separate the class? This job list is created entirely with BIC in mind, and other calorimeters may want different steps to be included. Please let me know if you have any ideas on this topic.

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,581 @@
...
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Test 1: Amplitude from ADC sum (no saturation)
+// ────────────────────────────────────────────────────────────────────────────────
+TEST_CASE("CalorimeterCALOROCReco: ADC amplitude in low-signal regime",
+          "[CalorimeterCALOROCReco][Amplitude]") {
+
+  // Write LUT
+  const std::string lut_path = "/tmp/caloroc_test_lut_amplitude.txt";
```
veprbl:
```suggestion
  const std::string lut_path = std::filesystem::temp_directory_path() / "caloroc_test_lut_amplitude.txt";
```
here and elsewhere

src/algorithms/calorimetry/CalorimeterCALOROCRecoConfig.h
ssedd1123: I agree with this idea. There are other instances where I want user to choose from a list and this should be more descriptive than just using integers.

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
ruse-traveler: TBH I wasn't thinking of separating anything out into a separate algorithm, more that you might move steps 4 - 6 into a subroutine. The default case would be where you have only one set of hits to process and can directly extract z and E from the individual hits, and then the "BIC" case would be where you have two sets of hits and you perform steps 4 - 6.

But it may make sense to split it into a separate algorithm! I could see the logic in doing steps 1 - 3 in an algorithm, where the z and E are set with only 1 set of hits, and then in the 2nd algorithm you take in two sets of hits and combine them via steps 4 -6.

Either way, in the interest of keeping momentum going, I think it's okay to keep the algorithm as-is! As long as we have an issue and a general plan for how we evolve this to accommodate the other calorimeters!

src/algorithms/calorimetry/CalorimeterCALOROCRecoConfig.h
ruse-traveler: Agreed! I'll open an issue to keep this on our horizon...

ruse-traveler: Here's the issue!
https://github.com/eic/EICrecon/issues/2860

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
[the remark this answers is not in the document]
ssedd1123: It is modified as suggested.

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+  REQUIRE(actual_amplitude == 100);
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Test 2: Amplitude uses lowGainADC when highGain saturates
+// ────────────────────────────────────────────────────────────────────────────────
+TEST_CASE("CalorimeterCALOROCReco: ADC amplitude switches to low-gain on saturation",
+          "[CalorimeterCALOROCReco][Amplitude]") {
+
+  const std::string lut_path = "/tmp/caloroc_test_lut_saturation.txt";
```
veprbl:
```suggestion
  const std::string lut_path = std::filesystem::temp_directory_path() / "caloroc_test_lut_saturation.txt";
```

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+  REQUIRE(actual_amplitude == 75);
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Test 3: z-position from delta-t
+// ────────────────────────────────────────────────────────────────────────────────
+TEST_CASE("CalorimeterCALOROCReco: z-position from timing difference",
+          "[CalorimeterCALOROCReco][ZPosition]") {
+
+  const std::string lut_path = "/tmp/caloroc_test_lut_zpos.txt";
```
veprbl:
```suggestion
  const std::string lut_path = std::filesystem::temp_directory_path() / "caloroc_test_lut_zpos.txt";
```

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+  REQUIRE_THAT(actual_z, Catch::Matchers::WithinAbs(zpos_expected, 1e-4));
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Test 4: z-position from NpeHit (useNpeHitPos=true)
+// ────────────────────────────────────────────────────────────────────────────────
+TEST_CASE("CalorimeterCALOROCReco: z-position read from NpeHit position",
+          "[CalorimeterCALOROCReco][ZPosition]") {
+
+  const std::string lut_path = "/tmp/caloroc_test_lut_npepos.txt";
```
veprbl:
```suggestion
  const std::string lut_path = std::filesystem::temp_directory_path() / "caloroc_test_lut_npepos.txt";
```

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+  REQUIRE_THAT(actual_z, Catch::Matchers::WithinAbs(npe_z, 1e-4));
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Test 5: Time walk correction shifts the reconstructed z
+// ────────────────────────────────────────────────────────────────────────────────
+TEST_CASE("CalorimeterCALOROCReco: time walk correction changes z from delta-t",
+          "[CalorimeterCALOROCReco][TimeWalk]") {
+
+  const std::string lut_path = "/tmp/caloroc_test_lut_timewalk.txt";
```
veprbl:
```suggestion
  const std::string lut_path = std::filesystem::temp_directory_path() / "caloroc_test_lut_timewalk.txt";
```

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+  }
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Test 6: MC truth links – correct counts and energy-normalized weights
+// ────────────────────────────────────────────────────────────────────────────────
+TEST_CASE("CalorimeterCALOROCReco: MC truth link weights are energy-normalized",
+          "[CalorimeterCALOROCReco][MCLinks]") {
+
+  const std::string lut_path = "/tmp/caloroc_test_lut_links.txt";
```
veprbl:
```suggestion
  const std::string lut_path = std::filesystem::temp_directory_path() / "caloroc_test_lut_links.txt";
```

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+  REQUIRE_THAT(weights[2], Catch::Matchers::WithinAbs(70.0 / 150.0, 1e-5));
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Test 7: Average time stored in rawhit timestamp
+// ────────────────────────────────────────────────────────────────────────────────
+TEST_CASE("CalorimeterCALOROCReco: rawhit timestamp is average of P and N TOA",
+          "[CalorimeterCALOROCReco][Timing]") {
+
+  const std::string lut_path = "/tmp/caloroc_test_lut_time.txt";
```
veprbl:
```suggestion
  const std::string lut_path = std::filesystem::temp_directory_path() / "caloroc_test_lut_time.txt";
```

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+#include <stdint.h>
+#include <deque>
+#include <gsl/pointers>
+#include <initializer_list>
+#include <utility>
+#include <vector>
+#include <filesystem>
+
+#if EDM4EIC_VERSION_MAJOR > 8 || (EDM4EIC_VERSION_MAJOR == 8 && EDM4EIC_VERSION_MINOR >= 7)
+
```
veprbl:
```suggestion
```
Not needed since #2760

src/tests/algorithms_test/calorimetry_CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,583 @@
...
+  std::cout << "tP (formula)       = " << tP << " ns" << "\n";
+  std::cout << "tN (formula)       = " << tN << " ns" << "\n";
+  std::cout << "Expected timestamp = 0.5*(tP+tN) = " << expected << " ns" << "\n";
+  std::cout << "Number of rawhits  = " << rawhits->size() << "\n";
+  std::cout << "Actual timestamp   = " << actual_time << " ns" << "\n";
+  REQUIRE(rawhits->size() == 1);
+  REQUIRE_THAT(actual_time, Catch::Matchers::WithinAbs(expected, 1e-4));
+}
+
+#endif // EDM4EIC_VERSION_MAJOR
```
veprbl:
```suggestion
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,506 @@
...
+
+        rawassocs_staging[contrib.getObjectID()] = assoc;
+        links_staging[contrib.getObjectID()]     = link;
+      }
+    }
+
+    if (edep > 0.0) {
+      for (auto& [key, link] : links_staging) {
+        link.setWeight(link.getWeight() / edep);
+        rawhitsLink->push_back(link);
```
veprbl: We should switch to `link = rawhitsLink->create()` pattern, if possible, which, I believe, it is in this case.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,506 @@
...
+  m_reference_z_p = m_geo.detector()->constant<double>(m_cfg.attenuationReferencePositionNamePos) *
+                    edm4eic::unit::mm / dd4hep::mm;
+  m_reference_z_n = m_geo.detector()->constant<double>(m_cfg.attenuationReferencePositionNameNeg) *
+                    edm4eic::unit::mm / dd4hep::mm;
+
+  info("Pos reference z = {}", m_reference_z_p);
+  info("Neg reference z = {}", m_reference_z_n);
+
+  info("calibration slope = {}, intercept = {}", m_cfg.slope, m_cfg.intercept);
+
```
veprbl:
```suggestion
```
Printing back the configuration with init is not the way to go

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,506 @@
...
+    return;
+  }
+
+  id_spec = m_detector->readout(m_cfg.readout).idSpec();
+
+  // local detector name has higher priority
+  if (!m_cfg.localDetElement.empty()) {
+    try {
+      m_local = m_detector->detector(m_cfg.localDetElement);
+      info("local coordinate system from DetElement {}", m_cfg.localDetElement);
```
veprbl:
```suggestion
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,506 @@
...
+  for (const auto& field : m_cfg.edep_to_npe_fields) {
+    auto field_idx = id_dec->index(field);
+    m_field_idxs.push_back(field_idx);
+  }
+
+  std::ifstream infile(m_cfg.edep_to_npe_filename);
+  if (!infile) {
+    error("Unable to open LUT file: {}", m_cfg.edep_to_npe_filename);
+  }
+  info("LUT file: {}", m_cfg.edep_to_npe_filename);
```
veprbl:
```suggestion
```

src/detectors/BEMC/BEMC.cc
```diff
@@ -234,46 +228,38 @@ void InitPlugin(JApplication* app) {
...
+          .timeWalkCor          = true,
+          .useNpeHitPos         = false,
+          .highGainDR           = 1023,
+          .gainRatio = EcalBarrelScFi_dyRangeLowGainADC / EcalBarrelScFi_dyRangeHighGainADC,
+          .attenuationReferencePositionNamePos = "EcalBarrel_LightGuide_PositivePosZ",
+          .attenuationReferencePositionNameNeg = "EcalBarrel_LightGuide_NegativePosZ",
+          .attenuationParameters               = EcalBarrelScFi_attPars,
+          .timeWalkCorrectionParameters        = {-13.7915, 33.5238, 3.15088, -0.313885},
+          .lightSpeedParameters                = {83.3221, -417.8},
+          .slope                               = 13.02, //3.54e-2,
```
veprbl: How did you obtain these parameters? Is there a script?

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,506 @@
...
+}
+
+void CalorimeterCALOROCReco::process(const CalorimeterCALOROCReco::Input& input,
+                                     const CalorimeterCALOROCReco::Output& output) const {
+
+  const auto [npeHitsP, ADCPs, npeHitsN, ADCNs]       = input;
+  auto [recohits, rawhits, rawhitsLink, rawhitsAssoc] = output;
+
+  // match NpeHits and ADC hits by cellID
+  std::unordered_map<dd4hep::rec::CellID, size_t> cellID2NpeHitNID, cellID2NpeHitPID, cellID2ADCNID,
```
veprbl: `cellID2NpeHitPID` is unused?

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
```diff
@@ -0,0 +1,84 @@
...
+             "inputADCNCollection"},
+            {"outputRecHitCollection", "outputRawHitCollection", "outputRawLink", "outputRawAssoc"},
+            "Reconstruct hit from CALOROC ADC data with MC truth from NpeHits."} {}
+
+  void init() final;
+  void process(const Input&, const Output&) const final;
+
+private:
+  std::function<double(const edm4hep::RawCalorimeterHit& h)> sampFrac;
+
```
veprbl:
```suggestion
```
From different class, unused.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,506 @@
...
+
+    // get position of the hit;
+    double zpos;
+
+    if (m_cfg.useNpeHitPos) {
+      zpos = static_cast<double>(npeHitP.getPosition().z);
+    } else {
+      if (m_cfg.timeWalkCor) {
+        tP = _timeWalkCorrection(tP, this->_sumADC(ADCP));
+        tN = _timeWalkCorrection(tN, this->_sumADC(ADCN));
```
veprbl, line 292: Does `time` need this correction?

src/detectors/BEMC/BEMC.cc
```diff
@@ -234,46 +228,38 @@ void InitPlugin(JApplication* app) {
...
+          .intercept                           = 0,     //1.425e-3/3.54e-2,
```
veprbl: Also unclear provenance and meaning of values in the comment.

veprbl (commented):
Have not yet read everything, but have some comments.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,506 @@
...
+    rawhit.setTimeStamp(time);
+
+    double edep = 0;
+
+    // link to parents
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitLink> links_staging;
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitAssociation>
+        rawassocs_staging;
+
+    for (bool NSide : std::vector<bool>{true, false}) {
```
veprbl:
```suggestion
    for (bool NSide : {true, false}) {
```

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: It does. Nice catch, thanks for pointing it out. Time is now set after time-walk correction.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: You're right, that is indeed unused. The variable is removed.

src/detectors/BEMC/BEMC.cc
mhkim-anl: I'm sorry for the delayed response. Please find our calibration studies via this [link](https://indico.bnl.gov/event/33018/contributions/124850/attachments/70268/123726/BIC20260602_calibration.pdf).

Regarding `timeWalkCorrectionParameters`, as shown in Slide 2, we fitted `t_toa-t_fit vs. ADC sum from samples` distribution and obtained those parameters. Here, `t_fit` is the time reconstructed by fitting `ADC vs. Time` distribution (Slide 3 is an example for this fitting) and it was considered as a true time. `t_toa` is the time reconstructed by TOA value.

Regarding `lightSpeedParameters`, we obtained parameters by fitting `z_true vs. t_left-t_right` distribution as shown in Slide 4.

veprbl: Is there a script that you can upload, e.g. to https://github.com/eic/snippets ? That would be much appreciated if anyone else tries to update these.

mhkim-anl: Yes, I can upload the script. I'll upload it after cleaning it up so that it can be used easily.

src/detectors/BEMC/BEMC.cc
ssedd1123: The commented values are removed. The calibration script for slope will be uploaded shortly. Intercept is assumed to be 0. When I allowed the fit to show non-zero intercept, the intercept is very small, so I just set that to zero, both because of simplicity, and because we know that zero ADC = zero energy deposition.

src/detectors/BEMC/BEMC.cc
ssedd1123: I have uploaded the script to extract the slope parameter: https://github.com/eic/snippets/tree/main/Calorimetery/ECalCalibration. However, those scripts do not estimate `.timeWalkCorrectionParameters` nor `.lightSpeedParameters`. Those will be uploaded by @mhkim-anl.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: I am sorry but I am not sure what does this mean exactly. I tried `auto newLink = rawhitsLink->create(); newLink = link` but the information didn't get copy over. I now set the target and destination manually for `newLink`. I hope this is what you are asking for.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,496 @@
...
+        auto newLink = rawhitsLink->create();
+        newLink.setTo(link.getTo());
+        newLink.setFrom(link.getFrom());
+        newLink.setWeight(link.getWeight() / edep);
+      }
+      for (auto& [key, assoc] : rawassocs_staging) {
+        assoc.setWeight(assoc.getWeight() / edep);
+        rawhitsAssoc->push_back(assoc);
+      }
+    }
```
veprbl, line 394:
```suggestion
        edm4eic::MutableMCRecoCalorimeterHitAssociation assoc = rawhitsAssoc->create();
        assoc.setRawHit(rawhit);

        edm4eic::MutableMCRecoCalorimeterHitLink link =  rawhitsLink->create();
        link.setFrom(rawhit);

        assoc.setSimHit(npeHit);
        assoc.setWeight(contrib.getEnergy() / npeHit.getEnergy());

        link.setTo(npeHit);
        link.setWeight(contrib.getEnergy() / npeHit.getEnergy());

        rawassocs_staging[contrib.getObjectID()] = assoc;
      }
    }
```
Something like that doesn't work?

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
veprbl: I left a suggestion below.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: Sorry I have to undo this suggestion. `Sum(contrib.getEnergy())` is not equal to `npeHit.getEnergy()` because of attenuation. `edep` needs to be summed in a separate loop.

src/detectors/BEMC/BEMC.cc
veprbl: Upload?

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
```diff
@@ -0,0 +1,81 @@
...
+#include "algorithms/interfaces/WithPodConfig.h"
+
+namespace eicrecon {
+
+using CalorimeterCALOROCRecoAlgorithm = algorithms::Algorithm<
+    algorithms::Input<edm4hep::SimCalorimeterHitCollection, edm4eic::RawCALOROCHitCollection,
+                      edm4hep::SimCalorimeterHitCollection, edm4eic::RawCALOROCHitCollection>,
+    algorithms::Output<edm4eic::CalorimeterHitCollection, edm4hep::RawCalorimeterHitCollection,
+                       edm4eic::MCRecoCalorimeterHitLinkCollection,
+                       edm4eic::MCRecoCalorimeterHitAssociationCollection>>;
```
veprbl, line 36: I'm appreciating this only now that this outputs a "reference" RawCalorimeterHit just to enable truth links. That is not pleasant, because it also means that bits of the CalorimeterDigi ended up in this reconstruction algorithm. I remember in one meetings long-long ago (cc @ShujieL @ruse-traveler @wdconinc @sly2j) we concluded that a singular `rawHit` relation (https://github.com/eic/EDM4eic/blob/de9f6bd3bb8630030ace4e0eee83891954699625/edm4eic.yaml#L310) would serve for the purposes of the truth information.

Now we have a hit that is reconstructed from two digitized hits and two waveforms, which is not something we necessarily expect downstream algorithms to know about. I was thinking about introducing `edm4eic::MCRecoCalorimeterHitLink` (unfortunately the name was taken https://github.com/eic/EDM4eic/pull/180/) to link `edm4eic::CalorimeterHit` to `edm4hep::SimCalorimeterHit`. I don't think this is urgent, since we don't save the `edm4hep::SimCalorimeterHit`s (documented in https://github.com/eic/eicrecon/issues/2323) and this does not become public (only the other algorithms will see how this is organized).

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -381,6 +381,8 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
       "EcalBarrelScFiRawHits",
       "EcalBarrelScFiPCALOROCHits",
       "EcalBarrelScFiNCALOROCHits",
+      "EcalBarrelScFiRawHitLinks",
+      "EcalBarrelScFiRawHitAssociations",
```
veprbl, line 385:
```suggestion
```
These two should have dangling references since we don't store the `EcalBarrelScFiHits`.
cc https://github.com/eic/EDM4eic/pull/180

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,496 @@
...
+
+  for (size_t i = 0; i < ADCPs->size(); ++i) {
+    const auto& ADCP                = ADCPs->at(i);
+    cellID2ADCPID[ADCP.getCellID()] = i;
+  }
+
+  for (size_t i = 0; i < ADCNs->size(); ++i) {
+    const auto& ADCN                = ADCNs->at(i);
+    cellID2ADCNID[ADCN.getCellID()] = i;
+  }
```
veprbl, line 251: Are these cellIDs including the z segmentation? Are we supposed to be summing over contributions in z? Or eventually?

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,496 @@
...
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitLink> links_staging;
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitAssociation>
+        rawassocs_staging;
+
+    for (bool NSide : {true, false}) {
+      const auto& npeHit = NSide ? npeHitN : npeHitP;
+      for (const auto& contrib : npeHit.getContributions()) {
+        // if link is already covered, don't add again
+        if (links_staging.find(contrib.getObjectID()) != links_staging.end())
+          continue;
```
veprbl, line 363: Should this use cellID?

src/detectors/BEMC/BEMC.cc
mhkim-anl: I've uploaded two macros that provide parameters for time-walk correction and z reconstruction in `Calorimetry/EcalCalibration/scripts/`. I'm very sorry for the delay.

src/detectors/BEMC/BEMC.cc
```diff
@@ -234,46 +228,38 @@ void InitPlugin(JApplication* app) {
...
+          .edep_to_npe_filename = EcalBarrelScFi_edep_to_npe_filename,
+          .edep_to_npe_fields   = EcalBarrelScFi_edep_to_npe_fields,
+          .timeWalkCor          = true,
+          .useNpeHitPos         = false,
+          .highGainDR           = 1023,
+          .gainRatio = EcalBarrelScFi_dyRangeLowGainADC / EcalBarrelScFi_dyRangeHighGainADC,
+          .attenuationReferencePositionNamePos = "EcalBarrel_LightGuide_PositivePosZ",
+          .attenuationReferencePositionNameNeg = "EcalBarrel_LightGuide_NegativePosZ",
+          .attenuationParameters               = EcalBarrelScFi_attPars,
+          .timeWalkCorrectionParameters        = {-13.7915, 33.5238, 3.15088, -0.313885},
```
veprbl:
```suggestion
          // See https://github.com/eic/snippets/tree/main/Calorimetery/ECalCalibration
          .timeWalkCorrectionParameters        = {-13.7915, 33.5238, 3.15088, -0.313885},
```

src/detectors/BEMC/BEMC.cc
veprbl: Thank you.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: Yes, the cellIDs are z-segmented. However, all hits along a fiber are already merged in SimCalorimeterHitProcessor. The z-information of cellID is not used here other than for truth association and the merging of Npe hits from the N- and P-sides. z-position of the hit still come from the pulses.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: I think either is fine. I just want to make sure the same hit won't be double counted. `getObjectID()` should work just as well as `cellID`.

veprbl: This condition is never triggered. We've learned that from https://github.com/eic/EICrecon/pull/2918 and once I added https://github.com/eic/EICrecon/pull/2770/commits/408a968596a0d9165ebcdbd6d15ca31b15c4b5c6 the CI still succeeds. Any ideas why?

ssedd1123: I think you're right and ObjectID() changes every time they are copied over to contribution links. I should switch to cellID. I'll test more before I push another commit on this issue.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
veprbl: What happens if there are multiple hits on the same fiber bundle?

ssedd1123: I am under the assumption that the merging of hits in the same fiber bundle is handled by `SimCalorimeterHitProcessor`, therefore I did not write any routine to handle multiple hits. 

I can't confirm what `SimCalorimeterHitProcessor` did at the moment, but given that the output of that class will be converted to pulses, the NpeP and NpeN hits themselves should have been merged in a way that is consistent with the electronic design of BEMC readouts.

veprbl: Interesting, thanks.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: I can't find a `getCellID()` function in `CaloHitContribution`. However, I have changed it to read ID from the underlying MCParticle. I think that will resolve the ambiguity.

src/algorithms/calorimetry/CalorimeterCALOROCReco.h
ssedd1123: Are you thinking of skipping the `RawCalorimeterHit` and directly have the `CalorimeterHit` linked to `SimCalorimeterHit`? I guess that would work, but then what should the rawHit be set for the RecHit? I don't know what downstream classes are doing with calorimeter hits, would it be fine if a RecHit has no corresponding RawHit? Just want to make sure before we make a new class in EDM4eic.

veprbl: Yes, I was thinking about that. And, yes there are open questions about this.
* rawHit is a one-to-one relation, and, at least originally, the standard was that all one-to-one relations must be set to a valid object (and people do get confused when their code crashes on dereference of a missing object)
* The downstream algorithms expect `rawHit` and are not aware of `MCRecoCalorimeterHitLink`, we would need to support both. If we want to also solve the previous point, we should transition all uses to `MCRecoCalorimeterHitLink` and deprecate `rawHit` field

ssedd1123: While I agree with the sentiment, I also think this is out of the scope of this pull request since we have to update downstream classes for them to handle hits with no corresponding raw hits. If you agree, I will mark this as resolved, but I will keep this idea in mind and we shall switch when the time is right.

veprbl: It's not a sentiment. I'm looking to discuss technical details of how we should organize hit processing to support your use case. We can't punt the discussion, but should file an issue to track this question.

veprbl: Filed as https://github.com/eic/EICrecon/issues/2989
Marking as resolved

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
veprbl: Did you do the sanity check?

veprbl: Also, of course there is not a cellID in CaloHitContribution, one should compare at the level of hits.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -0,0 +1,497 @@
...
+
+    // link to parents
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitLink> links_staging;
+    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitAssociation>
+        rawassocs_staging;
+
+    for (bool NSide : {true, false}) {
+      const auto& npeHit = NSide ? npeHitN : npeHitP;
+      for (const auto& contrib : npeHit.getContributions()) {
+        if (links_staging.find(contrib.getParticle().getObjectID()) != links_staging.end()) {
```
veprbl, line 362: The links_staging map is global for the whole event. If one particle produces multiple hits in the detector (ionization loss), only one hit will receive link?

ssedd1123: That is a good point. This is especially true since `SimCalorimeterHitProcessor` recursively find the parents of the hit points, so all contributions in the shower will point to the same particle. 

In that case, I am out of idea on how to compare hits from both sides. We can set additional conditions where time and energy difference are smaller than a certain thresholds, but none of them are fool-proof. 

I think you are right and we should just sum energy contributions from both sides. That may not be a physically meaningful quantities, but it eliminates all of these ambiguities.

ssedd1123: Wait sorry I didn't think this through. I think the code is fine. The links_staging map is not global and is declared within the for loop on line 254, `for (const auto& npeHitP : *npeHitsP) {`. Therefore, when we move to another hit from the same particle with a different cellID (It has to have a different cellID because `SimCalorimeterHitProcessor` has already bunched up all hits within one readout unit), the map will be reset. 

I think this will work as is.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
ssedd1123: Yes I did the sanity check. This time the if statement is indeed triggered sometimes.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
veprbl: Okay, good. Thank you for checking.

veprbl:
This is mostly good, but also some things to point out:

https://eicrecon.epic-eic.org/pr/2770/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#EcalBarrelScFiRecHits

image

Sees some reduction in hits count, that might be explainable. It also has some weird peak at -50 cm too (seen in Q2>100 sample too).

https://eicrecon.epic-eic.org/pr/2770/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#EcalBarrelScFiTopoClusters

image

almost empty with this PR, which kills EcalBarrelClusters. If we merge this we probably need to follow up with clustering tuning?

ssedd1123:
> This is mostly good, but also some things to point out:
> 
> https://eicrecon.epic-eic.org/pr/2770/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#EcalBarrelScFiRecHits
> 
> image
> Sees some reduction in hits count, that might be explainable. It also has some weird peak at -50 cm too (seen in Q2>100 sample too).
> 
> https://eicrecon.epic-eic.org/pr/2770/capybara/rec_dis_18x275_minQ2=0_craterlake_18x275/index.html#EcalBarrelScFiTopoClusters
> 
> image
> almost empty with this PR, which kills EcalBarrelClusters. If we merge this we probably need to follow up with clustering tuning?

I have to look deeper before giving you a concrete answer, but I am ~70% sure that the disappearance of clusters is because https://github.com/eic/EICrecon/pull/2933 is not applied yet. It is known that the z-location of clusters are all wrong without updating `CALOROCDigiization`. In fact, when I showed the baseline performance of BIC in the weekly meetings, I always use my own branch that combines both #2933 and this pull request, otherwise I will see a significant drop in cluster efficiency. 

I'll keep you posted.

veprbl:
Okay, I will look at that one too.

PR #2789, opened 2026-07-26
Segmentation fault in PulseCombiner::sumPulses caused by a hit with an exceptionally large time

### LINKED ISSUE #2788: Segmentation fault in PulseCombiner::sumPulses caused by a hit with an exceptionally large time
In rare cases, a BIC hit has an exceptionally large time value. In `PulseCombiner::sumPulses`, `maxStep` is calculated using `float` precision, and this large time value causes `maxStep` to be calculated as 0, resulting in a segmentation fault.

mhkim-anl:

In rare cases, a BIC hit has an exceptionally large time value (> 10^10). In `PulseCombiner::sumPulses`, `maxStep` is calculated using `float` precision, and this large time value causes `maxStep` to be calculated as 0, resulting in a segmentation fault. This PR is to address the above problem. 

Actually, it is unusual for some BIC hits to have such large time values. These are likely hits that should be discarded. I will investigate why they are produced, and, if necessary, introduce a time threshold in the `SimCalorimeterHitProcessor` algorithm.

- [x] Medium
- [ ] Low

- [x] Bug fix (issue #2788 )
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __


src/algorithms/digi/PulseCombiner.cc
```diff
@@ -149,16 +149,13 @@ PulseCombiner::clusterPulses(const std::vector<PulseType> pulses) const {
...
+  int maxStep = 0;
+  for (const auto& pulse : pulses) {
+    int startStep = std::round((pulse.getTime() - pulses[0].getTime()) / pulses[0].getInterval());
+    maxStep       = std::max(maxStep, startStep + static_cast<int>(pulse.getAmplitude().size()));
   }
 
-  //Calculate maxTime in interval bins
-  int maxStep = std::round((maxTime - pulses[0].getTime()) / pulses[0].getInterval());
-
   std::vector<float> newPulse(maxStep, 0.0);
```
wdconinc: I think your analysis stops short of identifying why this leads to a segmentation fault. If I understand correctly, in the old code your `maxTime` is zero and `maxStep` becomes negative. Then you use that negative `maxStep` as a vector size, where it is interpreted as an unsigned int and wraps around to large values. That's then probably the cause of the segmentation fault. Is that's what's happening?

If that's what's happening, then you're still not guarding against the underlying cause, which is that you allow a signed integer to be used as a vector size. If `maxStep` is used to size the vector, it should be of type `size_t` or at least `unsigned`. You effectively restrict the domain already to positive values in the new code, but using types is safer (you'll also get helpful compiler warnings about any operations you do that could result in incorrect behavior about the signedness).

wdconinc: In this case you appear to have suppressed the compiler warnings the worse of two ways, by casting the unsigned size into a signed integer... Instead of casting the size to int, you should stay in size_t domain entirely when you'll use this as a size.

src/algorithms/digi/PulseCombiner.cc
ssedd1123: Since we are on this topic of `sumPulses`, I think we should also implement the bug fix that Zhiwan suggested where the endStep can some time go beyond the vector size. 
```suggestion
    int endStep   = std::min(startStep + pulseSize, maxStep);
```

simonge: I don't understand why that would be useful, please can you explain or link to the suggestion?

ssedd1123: There are events where `startStep + pulseSize > maxStep` which causes the for loop in line 166 to seg-fault, because `newPulse` only has size `maxStep`.

It only happens to a few events, and since `PulseCombiner` weren't enabled is most detector, it is hard to detect this error. We ran into this error when we tried to use `PulseCombiner` for BIC in branch 2728-edep-to-npe-conversion last week, but now with all the updates I failed to reproduce the seg-fault again.

ssedd1123: Now that I think about it, `maxStep` should have already accommodated the latest `pulseSize`. Perhapes this error that I ran into is intimately related to overflowing of `maxStep`, causing the vector to not be large enough.

src/algorithms/digi/PulseCombiner.cc
mhkim-anl: The old code causes a segmentation fault when a pulse has an exceptionally large start time. In this case, because of the `float` precision, `maxTime` ends up equal to `pulse.getTime()`, making `maxStep` 0. This results in an empty `newPulse` vector, which causes a segmentation fault when it is accessed in the following for-loop.

Regarding the types, I agree with your comments. My original update wasn't the right approach. I’ve updated the code to keep the calculations entirely in the `size_t` domain.

src/algorithms/digi/PulseCombiner.cc
```diff
@@ -149,16 +149,14 @@ PulseCombiner::clusterPulses(const std::vector<PulseType> pulses) const {
...
 
-  // Find maximum time of pulses in cluster
-  float maxTime = 0;
-  for (auto pulse : pulses) {
-    maxTime =
-        std::max(maxTime, pulse.getTime() + pulse.getInterval() * pulse.getAmplitude().size());
+  // Calculate the number of interval bins for the combined pulse
+  std::size_t maxStep = 0;
+  for (const auto& pulse : pulses) {
+    std::size_t startStep = static_cast<std::size_t>(
```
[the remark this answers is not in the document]
mhkim-anl: Followed github-actions[bot]'s comment

src/algorithms/digi/PulseCombiner.cc
simonge: I'm not sure this will help. Time values are only ever stored in the datamodel to float precision, once the value gets large enough, any addition of step time just gets lost. The best the algorithm would be able to do is combine pulses starting at the same "time". But as you suggested we hopefully won't ever care about correlations between our events and these hits so might be best to consider a way of carefully removing them sooner.

mhkim-anl (commented):
Followed github-actions[bot]'s suggestion

src/algorithms/digi/PulseCombiner.cc
mhkim-anl: I agree. If pulses have exceptionally large time values, the limited float precision makes it difficult to preserve small time differences between them accurately. I’ll first investigate the origin of those hits and then submit a separate PR addressing how they should be handled.

simonge (dismissed):
A good improvement.

src/algorithms/digi/PulseCombiner.cc
wdconinc: This is another loop that should be rewritten to be structurally safe, not reliant on having correctly calculated variables earlier.

ssedd1123:
```suggestion
auto startStep = static_cast<std::size_t>(
        std::round((pulse.getTime() - pulses[0].getTime()) / pulses[0].getInterval()));
auto pulseSize = pulse.getAmplitude().size();
auto endStep   = startStep + pulseSize;
for (auto i = startStep; i < endStep; i++) {
```

Not that I think anything will change, but you cast time step to size_t in line 156-157, you may as well do it here for the sake of consistency.

ruse-traveler (dismissed):
Thanks for the fix, @mhkim-anl! I think @ssedd1123 and @wdconinc's comments are good to consider, but otherwise think this is good to go!

src/algorithms/digi/PulseCombiner.cc
mhkim-anl: Thank you for your comment. I agree. I've updated the loop so that it no longer assumes `maxStep` is correct. Please let me know if I've misunderstood your comment.

mhkim-anl: Thank you for your suggestion. I applied it, though not exactly as suggested, since I also addressed Wouter's comment above.

mhkim-anl:
Sorry for the late updates to the remaining comments, and it caused the previous approval to be dismissed. I think I've now addressed all the remaining comments. If everything looks good, I would appreciate it if you could approve this PR. Thanks!

src/algorithms/digi/PulseCombiner.cc
wdconinc, line 67: We make a copy of all pulses.

wdconinc, line 77: We make a copy of all pulses (call by value, return by value (but probably tail call eliminated)).

wdconinc: We make a copy of all pulses (call by value, return by value (but probably tail call eliminated)).

wdconinc, line 121: We make a copy of all pulses.

src/algorithms/digi/PulseCombiner.cc
```diff
@@ -149,26 +150,24 @@ PulseCombiner::clusterPulses(const std::vector<PulseType> pulses) const {
...
-  // Find maximum time of pulses in cluster
-  float maxTime = 0;
-  for (auto pulse : pulses) {
-    maxTime =
-        std::max(maxTime, pulse.getTime() + pulse.getInterval() * pulse.getAmplitude().size());
+  // Calculate the number of interval bins for the combined pulse
+  std::size_t maxStep = 0;
+  for (const auto& pulse : pulses) {
+    auto startStep = static_cast<std::size_t>(
+        std::round((pulse.getTime() - pulses[0].getTime()) / pulses[0].getInterval()));
```
wdconinc: There is an implicit assumption for the correct operation of this function: that `pulses` is time-ordered (and has at least one entry). If that assumption is not guaranteed (in this case two frames away) this will start to return negative time differences. Can you make this a bit clearer in the name of this function, its documentation, and the calling context?

Also, we calculate `(pulse.getTime() - pulses[0].getTime()) / pulses[0].getInterval()` in two identical loops. In the first loop, it seems that all you calculate is actually:
```
std::ranges::max(pulses, [&int=pulses[0].getInterval()](auto a, auto b) {
    return a.getTime() + a.getAmplitude().size()*int < b.getTime() + b.getAmplitude().size()*int;
});
```
(where the `std::ranges::max` does not do worse than what you do, but is arguably clearer).

wdconinc (commented):
I looked again and the duplication of the loop in `sumPulses` struck me. Can we not avoid that explicit loop and at least make the first one more computationally efficient and avoid function calls?

Then I looked a bit further outside the immediate context and noticed a lot of pulse copying. Some of this is maybe inevitable (but I'd think you can avoid all copies here). Can we at least get the functions changes to call-by-reference? Each time we copy a pulse, it's not just a single number but a varying sequence of numbers. You don't seem to modify any of these pulses until the very end when you sum them.

src/algorithms/digi/PulseCombiner.cc
veprbl: Copy of handles? That should be cheap, and is welcome compared to messing with indices.

mhkim-anl: Grouping by cell is necessary to combine the pulses per cell in this algorithm, and since what gets copied here are handles as @veprbl points out, I think we can keep this part as it is.

mhkim-anl: I changed to take the pulses by const reference to avoid copying the vector on every call.

mhkim-anl: I changed to take the pulses by const reference here as well.

mhkim-anl: I think this copy is needed to sort the pulses without modifying the input, so I suggest keeping this part as it is.

src/algorithms/digi/PulseCombiner.cc
mhkim-anl: Thanks for your comments and suggestions. I agree that the duplicated loop was not efficient. I updated the function to use a single loop. Based on your comments, I also renamed `sumPulses` to `sumTimeOrderedPulses` and added comments explaining the assumption you mentioned. With the first loop removed, `maxStep` is no longer computed, so I didn't use `std::ranges::max` in the updated code.

mhkim-anl:
I've updated the code following the additional comments and suggestions. Please take another look and approve if it looks good. Thanks!

src/algorithms/digi/PulseCombiner.cc
```diff
@@ -111,21 +115,22 @@ void PulseCombiner::process(const PulseCombiner::Input& input,
...
+PulseCombiner::clusterPulses(const std::vector<PulseType>& pulses) const {
 
-  // Clone the pulses array of pointers so they aren't const
+  // Copied so they can be sorted
   std::vector<PulseType> ordered_pulses{pulses};
 
   // Sort pulses by time, greaty simplifying the combination process
-  std::ranges::sort(ordered_pulses,
-                    [](PulseType a, PulseType b) { return a.getTime() < b.getTime(); });
+  std::ranges::sort(ordered_pulses, [](const PulseType& a, const PulseType& b) {
```
wdconinc, line 124: Not for this PR, but I guess you really don't need to do a full time sort at O(N logN) and you're fine with just finding the minimum time pulse at O(N). Then you can pass that as a reference pulse (by reference) to sumPulses.

wdconinc (approved):
Looks good to me.

veprbl (approved):
LGTM. I think `const auto&` is excessive (handler objects should be on stack anyway, but it's going to work anyway.

epic-capybara:
- #2973

Issue #2746, opened 2026-07-14
Can't reconstruct pions produced from Lambda decays in B0 tracker


DraTeots:
**B0 tracking produces zero tracks in production: B0TrackerSeeds runs with
central-tracker seeding geometry, and truth seeding cannot seed displaced secondaries**

- 26-07 campaign is affected. 
- Any secondaries (e.g. pion from lambda (aha...)) are affected for truth finder. 

### Environment

- Both defects verified present on:
  - `main` at `8ff8dbee` (latest as of 2026-07-14
    `getGeneratorStatus() != 1` in TrackParamTruthInit.cc:44, default-config
    `B0TrackerSeeds` registration in tracking.cc:393-395
  - release branch `v1.37` (newest, as shipped in current eic_xl stable
    images): same two sites, no `acceptSecondaries` in
    TrackParamTruthInitConfig.h, `maxVertexZ = 200*dd4hep::mm`
- Empty B0 collections observed directly in edm4eic outputs of a full
  production with eic_xl `26.07-stable` and `26.07-stable`

### Symptom

Every B0 track collection is empty in every event, at every beam energy:

| collection | entries over 30,000 events |
|---|---|
| `B0TrackerRecHits` | populated (present in 30–73% of events) |
| `B0TrackerMeasurements` | populated (tracks RecHits 1:1) |
| `B0TrackerSeeds` | **0** |
| `B0TrackerCKFTracks` / `...Unfiltered` | **0** |
| `B0TrackerCKFTruthSeededTracks` / `...Unfiltered` | **0** |

So the chain goes dark exactly at seeding, for both the real-seeded and the
truth-seeded branch. 

This one is run on the latest 26.07 campaign

```
epic:/RECO/26.07.0/epic_craterlake/Bkg_Exact1S_2us/GoldCt/10um/DIS/NC/10x275/minQ2=100
```

| collection | total | events > 0 |
|---|---|---|
| B0TrackerRecHits | 3588 | 98/99 |
| B0TrackerMeasurements | 3588 | 98/99 |
| **B0TrackerSeeds** | **0** | 0 |
| **B0TrackerCKFTracksUnfiltered** | **0** | 0 |
| **B0TrackerCKFTracks** | **0** | 0 |
| CentralTrackSeeds (control) | 50122 | 99/99 |
| CentralCKFTracks (control) | 900 | 99/99 |
| B0TrackerCKFTruthSeededTracks (control) | 89 | 63/99 |

[...]

PR #2797, opened 2026-07-28
Allow truth seeder to work with secondary particles

DraTeots:

Should fix part 1 of #2746 - allow TruthSeeder to seed secondary particles from e.g. lambdas

- [ ] Medium
- [ ] Low

Current setup makes B0 not working for any decay products, for which (arguably) it is intended. In our case, for example, we can't see the most of Sullivan lambdas via proton+pi- channel (which is ~75% of lambda decays that leave a trace in our detector)

- [x] Bug fix (issue #2746)

  Truth finder cuts 
   1. any particles with status != 1, i.e. primary particles coming from event generator.
   2. require close to interaction vertex
   3. require minimum momentum
   4. require minimum pseudorapidity
   So 1 and 2 must go and 4 probably be changed in order to start seeding secondaries. But this will probably change all pictures that use TruthSeeding and this must be especially seen in background populated frames reconstruction
- [ ] AI was used in preparing this PR. - not yet...

### DIFF
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -40,12 +40,6 @@ void TrackParamTruthInit::process(const Input& input, const Output& output) cons
   // Loop over input particles
   for (const auto& mcparticle : *mcparticles) {
 
-    // require generatorStatus == 1 for stable generated particles in HepMC3 and DDSim gun
-    if (mcparticle.getGeneratorStatus() != 1) {
-      trace("ignoring particle with generatorStatus = {}", mcparticle.getGeneratorStatus());
-      continue;
-    }
-
     // require close to interaction vertex
     auto v = mcparticle.getVertex();
     if (std::abs(v.x) * dd4hep::mm > m_cfg.maxVertexX ||
--- a/src/algorithms/tracking/TrackParamTruthInit.cc
+++ b/src/algorithms/tracking/TrackParamTruthInit.cc
@@ -13,13 +13,12 @@
 #include <edm4hep/Vector2f.h>
 #include <edm4hep/Vector3d.h>
 #include <edm4hep/Vector3f.h>
-#include <Eigen/Core>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


wdconinc:
FYI @AkshayaVijay this is a very similar kind of question as to what particle caused a cluster. Maybe the logic of navigating up the MCParticle graph can be shared between this and the cluster links determination.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -40,12 +39,6 @@ void TrackParamTruthInit::process(const Input& input, const Output& output) cons
   // Loop over input particles
   for (const auto& mcparticle : *mcparticles) {
 
-    // require generatorStatus == 1 for stable generated particles in HepMC3 and DDSim gun
-    if (mcparticle.getGeneratorStatus() != 1) {
```
veprbl, line 44: I wonder if there is a benefit to not relax the generator requirement completely, but instead also allow including particles with `getSimulatorStatus() != 0`.

`particleSvc` below should report quarks/diquarks as unknown, but we can be more explicit about intent here.

veprbl: Implemented.

src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -40,9 +39,11 @@ void TrackParamTruthInit::process(const Input& input, const Output& output) cons
...
-      trace("ignoring particle with generatorStatus = {}", mcparticle.getGeneratorStatus());
+    // accept generator-stable particles (HepMC3/DDSim gun) or Geant4-produced
+    // secondaries; reject generator intermediates (partons, resonances, beams)
+    if (!(mcparticle.getGeneratorStatus() == 1 || mcparticle.getSimulatorStatus() != 0)) {
```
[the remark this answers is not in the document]
veprbl: No, this is written like that on purpose.

veprbl (approved):

PR #2809, opened 2026-07-29
bugfix: Imaging cluster phi calculation

### LINKED ISSUE #2803: Cluster phi calculation bug in ImagingClusterReco
- Which branch (often `main` for latest released): main
- Which version (or `HEAD` for the most recent on git): HEAD
- Any specific OS or system where the issue occurs? n/a
- Any special versions of ROOT or Geant4? n/a

Moving from the discussion on MM to here. This concerns a bug in the determination of the BIC Imaging cluster phi value from its constituent hits. On L176 of ImagingClusterReco : https://github.com/eic/EICrecon/blob/main/src/algorithms/calorimetry/ImagingClusterReco.cc#L176 - the cluster phi is determined via an energy-weighted average of the phi values of its hits. However, this (obviously) does the wrong thing when hit phi values straddle the phi -pi/+pi boundaries and then one gets a strange cluster phi.

Here is example failure case summarized by Claude Code: 

```
Truth electron: E = 9.974 GeV, η = −0.537, φ = −3.136 rad (i.e. essentially sitting right on the ±π seam)

The mis-reconstructed cluster (cluster[0]): E = 9.782 GeV (98% response — this really is the electron's shower), 203 hits, spanning sectors 13 and 14.

Sector 14 │ 187 hits │ ≈ +3.12 rad (just below +π)
Sector 13 │ 16 hits  │ ≈ −3.13 rad (just above −π)

Physically these are all adjacent, tightly packed hits — same shower, same tiny angular region, just that atan2 reports them on opposite sides of its ±π branch cut.

What EICrecon reported as the cluster position:
φ = 2.694 rad, at (x,y,z) = (−754.6, 361.9, −485.1) mm — nowhere near the shower.

What it should have been (energy-weighted Cartesian mean of the same hits, or equivalently what SciFi's algorithm would give): φ = 3.118 rad, at (x,y,z) = (−939.6, 22.5, −544.8) mm — correctly on top of the truth direction.
```

I can suggest a PR fix if the experts prefer -- I believe it should be straightforward.

dvperepelitsa:
This PR fixes an issue with Imaging cluster reconstruction which returned a buggy value for clusters whose hit constituents spanned the phi = +pi -> -pi wraparound. Discussed briefly in this Mattermost thread: https://chat.epic-eic.org/main/pl/3d5g4tq9k7f8iqfsr4yq478iky and then in Issue 2803 referenced below.

Note this PR replaces the older PR from a branch outside eic/EICrecon per request from @wdconinc .

- [x] Low

- [x] Bug fix (issue #2803) https://github.com/eic/EICrecon/issues/2803 

- [x] This PR changes default behavior. The Imaging cluster phi will now be determined based on the energy-weighted cluster (x,y) coordinates.
- [x] AI was used in preparing this PR. I noticed this issue from diagnostic plots made while exploring the data using Claude Code, then used Claude Code to track down the source of the problem and discuss possible solutions.

### DIFF
--- a/src/algorithms/calorimetry/ImagingClusterReco.cc
+++ b/src/algorithms/calorimetry/ImagingClusterReco.cc
@@ -161,7 +161,8 @@ ImagingClusterReco::reconstruct_cluster(const edm4eic::ProtoCluster& pcl) const 
   double time        = 0.;
   double timeError   = 0.;
   double meta        = 0.;
-  double mphi        = 0.;
+  double mx          = 0.;
+  double my          = 0.;
   double r           = 9999 * dd4hep::cm;
   for (unsigned i = 0; i < hits.size(); ++i) {
     const auto& hit    = hits[i];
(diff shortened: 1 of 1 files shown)


wdconinc:
No diffs, so we probably don't hit this bug in the CI events (small set).

wdconinc (approved):

PR #2816, opened 2026-07-31
Sort electron candidates by pT

### LINKED ISSUE #2815: ScatteredElectronsEMinusPz should sort by pT
Recent version of the [Inclusive PWG's ElectronID code](https://github.com/eic/snippets/tree/main/ElectronID) take the highest pT particle in the case of multiple scattered electron candidates.

The ScatteredElectronEMinusPz branch in EICrecon is currently indirectly sorting by momentum -> this should be updated to more closely match the snippets code.

smaple1:
- [ ] Medium
- [X ] Low

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ X] other: (close issue #2815 )

- [ X] AI was used in preparing this PR. Please describe usage below.
opencode used in implementation.

### DIFF
--- a/src/algorithms/reco/ScatteredElectronsEMinusPz.cc
+++ b/src/algorithms/reco/ScatteredElectronsEMinusPz.cc
@@ -6,11 +6,11 @@
 #include <Math/Vector4Dfwd.h>
 #include <edm4eic/ReconstructedParticleCollection.h>
 #include <edm4hep/Vector3f.h>
-#include <fmt/core.h>
 #include <podio/ObjectID.h>
 #include <algorithm>
-#include <gsl/pointers>
 #include <cmath>
+#include <iterator>
+#include <tuple>
 #include <utility>
 #include <vector>
 
--- a/src/algorithms/reco/ScatteredElectronsEMinusPz.cc
+++ b/src/algorithms/reco/ScatteredElectronsEMinusPz.cc
@@ -107,7 +107,7 @@ void ScatteredElectronsEMinusPz::process(const ScatteredElectronsEMinusPz::Input
   } // electron loop
 
   // Sort by transverse momentum descending
-  std::ranges::sort(candidates, , [](const auto& a, const auto& b) { return a.first > b.first; });
+  std::ranges::sort(candidates, [](const auto& a, const auto& b) { return a.first > b.first; });
 
   for (const auto& [pT, particle] : candidates) {
     out_electrons->push_back(particle);
--- a/src/algorithms/reco/ScatteredElectronsEMinusPz.cc
+++ b/src/algorithms/reco/ScatteredElectronsEMinusPz.cc
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2024 Daniel Brandenburg, Dmitry Kalinkin
+// Copyright (C) 2026 Daniel Brandenburg, Dmitry Kalinkin, Stephen Maple
 
 #include <Math/GenVector/LorentzVector.h>
 #include <Math/GenVector/PxPyPzM4D.h>
@@ -8,10 +8,11 @@
 #include <edm4hep/Vector3f.h>
 #include <fmt/core.h>
 #include <podio/ObjectID.h>
-#include <functional>
 #include <gsl/pointers>
-#include <map>
+#include <algorithm>
+#include <cmath>
 #include <utility>
+#include <vector>
 
 #include "algorithms/reco/ScatteredElectronsEMinusPz.h"
 #include "algorithms/reco/ScatteredElectronsEMinusPzConfig.h"
@@ -41,10 +42,10 @@ void ScatteredElectronsEMinusPz::process(const ScatteredElectronsEMinusPz::Input
   static const auto m_electron = m_particleSvc.particle(11).mass;
   static const auto m_pion     = m_particleSvc.particle(211).mass;
 
-  // this map will store intermediate results
+  // this vector will store intermediate results
   // so that we can sort them before filling output
   // collection
-  std::map<double, edm4eic::ReconstructedParticle, std::greater<>> scatteredElectronsMap;
+  std::vector<std::pair<double, edm4eic::ReconstructedParticle>> candidates;
 
   out_electrons->setSubsetCollection();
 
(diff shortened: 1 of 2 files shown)


wdconinc (commented):
I have my concerns about making any guarantees about sorting in our collections (since it's a property of the collection that we don't enforce at the data model level and is implementation dependent in a way that podio does not commit to), and I have even stronger concerns about *changing* the sorting we use. That means that past analysis code that relied on this becomes invalid in a way that will not be apparent. In my opinion this discussion (i.e. my concerns about how we should guarantee and document sorting convents) warrants a wider discussion in a reco or compsw meeting.

smaple1:
My concern with the current implementation is that there is a sorting by the E-pz sum of all particles in the event for different hypotheses of electron and pion masses, which is not a meaningful sorting. The E-pz sum should be the same for any given set of ReconstructedParticles (and = ~2*Eebeam ) so the values we get out will be almost exactly the same except for differences after the decimal point due to the different masses assumed.

A back of the envelope calculation leads me to believe that this ending up approximately sorting by momentum, though I don't know if this is always the case - but it's very hard to tell what kind of sorting is actually happening here if you were only to look at the code.

veprbl:
@wdconinc Regarding ordering preservation not being guaranteed, can we file a question issue to PODIO?

Our data model has very little say on what collections we book and how they relate to each other, but we are able to implement new collections.

> The E-pz sum should be the same for any given set of ReconstructedParticles (and = ~2*Eebeam ) so the values we get out will be almost exactly the same except for differences after the decimal point due to the different masses assumed.

This makes sense to me, $(\sum p^-) / \sqrt{s}$ would be like $x$ of the electron, which in case of no collinear radiation is $\approx 1$. This ordering was there intentionally from the beginning in https://github.com/eic/EICrecon/pull/1277, so I agree with @wdconinc that this ideally to be discussed before we merge this.

veprbl:
> @wdconinc Regarding ordering preservation not being guaranteed, can we file a question issue to PODIO?

(I can file it myself, just asking in case you are more equipped with the lore to pose the question.)

wdconinc:
> Regarding ordering preservation not being guaranteed, can we file a question issue to PODIO?

At the start of August that's hardly a quick way to proceed...

> Our data model has very little say on what collections we book and how they relate to each other, but we are able to implement new collections.

and

> A back of the envelope calculation leads me to believe that this ending up approximately sorting by momentum, though I don't know if this is always the case - but it's very hard to tell what kind of sorting is actually happening here if you were only to look at the code.

I had tried earlier to reply to @smaple1's comment (but the OOM killer ate it), and it relates to @veprbl's comment. I think that sorting is something that should be applied at the place where sorting is needed. We can try(!) to provide a presorted collection, ideally through the design of the algorithm to make the downstream need for sorting close to a single pass confirmation.

The main concern I have (the lack of manifest sort) is apparent here: we switch the same collection from one sort to another sort without a way to invalidate old algorithms. If we take this upstream to podio, I'd like to suggest something like a sorted view (implemented as a subset collection) where we can maybe make stronger guarantees, including manifesting it in the name.

veprbl:
> At the start of August that's hardly a quick way to proceed...

Well, my idea on this is that we might hear back that there is no plan to break collection ordering, i.e. they are to stay array-like.

> If we take this upstream to podio, I'd like to suggest something like a sorted view (implemented as a subset collection) where we can maybe make stronger guarantees, including manifesting it in the name.

If one believes that the ordering is a guarantee, then subset collections are those views you are looking for. I do like the manifesting in the name part.

wdconinc:
> Well, my idea on this is that we might hear back that there is no plan to break collection ordering, i.e. they are to stay array-like.

Possibly relevant: https://github.com/AIDASoft/podio/issues/778

wdconinc:
> > Well, my idea on this is that we might hear back that there is no plan to break collection ordering, i.e. they are to stay array-like.
> 
> Possibly relevant: [AIDASoft/podio#778](https://github.com/AIDASoft/podio/issues/778)

That's actually more related to the other reason of why I don't like guarantees about ordering: it enforces a serial step, whether an explicit sort or forcing the entire algorithm to run serially. At this point we run 1 algorithm = 1 core, but there's not really anything that forces us to do that. We could at some point realize that maybe for clustering we can run some expensive loops with execution policy par_unseq on a few threads allocated to that algorithm.

veprbl:
Filed https://github.com/AIDASoft/podio/issues/997

wdconinc:
> I do like the manifesting in the name part.

Suggestion for a convention, then: a sorted subset collection will end with `By[Field]` where we use CamelCase for field (or expression, as the case may be). So, in this case, `ScatteredElectronsEMinusPz` stays unchanged, `ScatteredElectronsEMinusPzByPt` is a subset collection sorted by `Pt`. Instead of doing the sorting in each algorithm, we use a meta algorithm/factory that does the sorting and that takes a field accessor or lambda in the config.

wdconinc:
> > I do like the manifesting in the name part.
> 
> Suggestion for a convention, then: a sorted subset collection will end with `By[Field]` where we use CamelCase for field (or expression, as the case may be). So, in this case, `ScatteredElectronsEMinusPz` stays unchanged, `ScatteredElectronsEMinusPzByPt` is a subset collection sorted by `Pt`. Instead of doing the sorting in each algorithm, we use a meta algorithm/factory that does the sorting and that takes a field accessor or lambda in the config.

Ref https://github.com/eic/EICrecon/pull/2818

ruse-traveler (commented):
Thanks for this @smaple1! I second that we should discuss this before merging: would you be willing to discuss at an upcoming Reco WG meeting?

That being said, I appreciate the factoring out of the HFS (ish) calculation from the algorithm! Possibly one way to proceed along this direction would be to 1st compute the HFS for each electron candidate (`edm4eic::HadronicFinalState`) above a momentum threshold, which are then used in this algorithm. Unfortunately, I think we would need to modify the HFS algorithm 1st to work with electron candidates...

smaple1:
> Thanks for this @smaple1! I second that we should discuss this before merging: would you be willing to discuss at an upcoming Reco WG meeting?
> 
> That being said, I appreciate the factoring out of the HFS (ish) calculation from the algorithm! Possibly one way to proceed along this direction would be to 1st compute the HFS for each electron candidate (`edm4eic::HadronicFinalState`) above a momentum threshold, which are then used in this algorithm. Unfortunately, I think we would need to modify the HFS algorithm 1st to work with electron candidates...

@ruse-traveler I can be available for next week's reconstruction meeting if there's space on the agenda?

I think it would be preferable as a use to have an obvious mapping between (true or candidate) electrons and the corresponding HFS (where both exist - which reminds me that we should probably have a HFS collection that is produced for CCDIS events)

ruse-traveler:
> @ruse-traveler I can be available for next week's reconstruction meeting if there's space on the agenda?

Sure! The agenda's wide-open at the moment!

> I think it would be preferable as a use to have an obvious mapping between (true or candidate) electrons and the corresponding HFS (where both exist - which reminds me that we should probably have a HFS collection that is produced for CCDIS events)

Definitely!

veprbl:
Achieved by #2818

PR #2822, opened 2026-08-03
fix: remove *Associations from PODIO output collections

wdconinc:
## Dependencies
- [x] #2875 

## Description

Removes 60 unused *Associations entries from the default output collections list in `JEventProcessorPODIO.cc`. These association collections are no longer needed in the standard PODIO output, reducing file size and output verbosity. For each of the removed *Associations, the *Links are written.

Also, modifies IrtCherenkovParticleID algorithm/factory to use links instead of associations to avoid dangling references.

## Changes

- Removed all `*Associations` collection entries from `output_collections` vector
- Adds `*Links` collection outputs to `IrtCherenkovParticleID` algorithm/factory

This change affects the default PODIO output files generated by EICrecon, excluding these association collections unless explicitly requested via configuration.

## Impact

- Smaller output ROOT files by default
- Faster file I/O operations
- Users can still enable specific associations via `podio:output_collections` parameter if needed


src/algorithms/pid/IrtCherenkovParticleID.cc
```diff
@@ -227,10 +230,10 @@ void IrtCherenkovParticleID::process(const IrtCherenkovParticleID::Input& input,
...
         if (m_cfg.cheatPhotonVertex || m_cfg.cheatTrueRadiator) {
-          for (const auto& hit_assoc : *in_hit_assocs) {
-            if (hit_assoc.getRawHit().isAvailable()) {
-              if (hit_assoc.getRawHit().id() == raw_hit.id()) {
-                mc_photon       = hit_assoc.getSimHit().getParticle();
+          for (const auto& hit_link : *in_hit_links) {
+            if (hit_link.getFrom().isAvailable()) {
+              if (hit_link.getFrom().id() == raw_hit.id()) {
+                mc_photon       = hit_link.getTo().getParticle();
                 mc_photon_found = true;
```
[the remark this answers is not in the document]
wdconinc: Don't want to change this behavior.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -214,33 +196,30 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
...
       "ReconstructedChargedParticles",
       "ReconstructedChargedParticleLinks",
-      "ReconstructedChargedParticleAssociations",
-      "MCScatteredElectronAssociations",    // Remove if/when used internally
-      "MCNonScatteredElectronAssociations", // Remove if/when used internally
+      "ReconstructedChargedParticleAssociations", // Used by associations below
+      "MCScatteredElectronLinks",                 // Remove if/when used internally
+      "MCScatteredElectronAssociations",          // Remove if/when used internally
+      "MCNonScatteredElectronLinks",              // Remove if/when used internally
+      "MCNonScatteredElectronAssociations",       // Remove if/when used internally
```
veprbl, line 224: Can we split additions to a separate PR that we can lead with?

wdconinc: Sure, https://github.com/eic/EICrecon/pull/2853

I don't think it allows us to get rid of them here, though, or at least not for a little while.

src/algorithms/pid/IrtCherenkovParticleID.cc
```diff
@@ -463,10 +466,10 @@ void IrtCherenkovParticleID::process(const IrtCherenkovParticleID::Input& input,
...
       }
 
-      // relate hit associations
-      for (const auto& hit_assoc : *in_hit_assocs) {
-        out_cherenkov_pid.addToRawHitAssociations(hit_assoc);
-      }
+      // FIXME store raw hits when supported
+      //for (const auto& hit_link : *in_hit_links) {
+      //  out_cherenkov_pid.addToRawHits(hit_link.getFrom());
+      //}
```
wdconinc: @chchatte92 Do you care about what's stored in the `rawHitAssociations` relation of the IRT1 output? See also https://github.com/eic/EDM4eic/pull/174 .

src/algorithms/pid/IrtCherenkovParticleID.cc
```diff
@@ -134,14 +137,14 @@ void IrtCherenkovParticleID::init(CherenkovDetectorCollection* irt_det_coll) {
 
 void IrtCherenkovParticleID::process(const IrtCherenkovParticleID::Input& input,
                                      const IrtCherenkovParticleID::Output& output) const {
-  const auto [in_aerogel_tracks, in_gas_tracks, in_merged_tracks, in_raw_hits, in_hit_assocs] =
+  const auto [in_aerogel_tracks, in_gas_tracks, in_merged_tracks, in_raw_hits, in_hit_links] =
```
wdconinc: @veprbl It may also be possible that we here don't just switch from assocs to links and write both for a while. That allows us to keep assocs for now, have IRT1 users convert, and then we yank the assocs.

src/algorithms/pid/IrtCherenkovParticleID.cc
chchatte92: Hi @wdconinc I am looking at it.

src/algorithms/pid/IrtCherenkovParticleID.cc
wdconinc: Done.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -121,17 +117,14 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
...
 
       "CombinedTOFTruthSeededParticleIDs",
       "CombinedTOFParticleIDs",
 
       // DRICH
       "DRICHRawHits",
       "DRICHRawHitsLinks",
-      "DRICHRawHitsAssociations",
       "DRICHAerogelTracks",
       "DRICHGasTracks",
```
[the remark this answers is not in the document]
wdconinc: Restored.

src/algorithms/pid/IrtCherenkovParticleID.cc
wdconinc: In any case, I restored it. We'll deal with this later.

PR #2835, opened 2026-08-06
BUGFIX: save TOF hits collections in output

ShujieL:
Save TOF hits and associations in output rootfiles so that one can connect TOF hits back to mcparticles in background analysis. Would be good to have those collections for next simulation campaign.

- [x ] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [ ] Updated documentation
- [ ] other: __


veprbl (dismissed):

veprbl:
(We should convert hit snippet to a benchmark, so that this kind of analysis does not regress.)

ssedd1123 (dismissed):

ShujieL:
> (We should convert hit snippet to a benchmark, so that this kind of analysis does not regress.)

I added a truth relation test script, please check. 
Also, the eic-container check failed with no good reason, can you take a look?

veprbl:
> > (We should convert hit snippet to a benchmark, so that this kind of analysis does not regress.)
> 
> I added a truth relation test script, please check.

That is not what we need, the existing script already does the synthetic check across every collection. I meant taking https://github.com/eic/snippets/blob/main/Tracking/TrackHitContributions/hit_matching.C and putting it into https://github.com/eic/detector_benchmarks/tree/master/benchmarks/tracking_performances wired to run with Snakemake.

> Also, the eic-container check failed with no good reason, can you take a look?

Part of a significant outage at one of the labs.

ShujieL:
> > > (We should convert hit snippet to a benchmark, so that this kind of analysis does not regress.)
> > 
> > 
> > I added a truth relation test script, please check.
> 
> That is not what we need, the existing script already does the synthetic check across every collection. I meant taking https://github.com/eic/snippets/blob/main/Tracking/TrackHitContributions/hit_matching.C and putting it into https://github.com/eic/detector_benchmarks/tree/master/benchmarks/tracking_performances wired to run with Snakemake.

OK as long as it can trigger a failure signal when not satisfied by future EICrecon PRs, I am fine moving it to benchmarks.

veprbl (approved):

veprbl:
Would it be okay if this only lands in `26.08.0` (I'd rather not tag `26.07.3` this late)?

ShujieL:
> Would it be okay if this only lands in `26.08.0` (I'd rather not tag `26.07.3` this late)?

As long as that will go into the August simulation campaign ;)

Issue #2834, opened 2026-08-05
FarDetectorLinearTracking cannot handle 1000s of hits in each layer.


simonge:
The current approach to combinatorics does not extend well to a 2us timeframe including ~2000 Bremsstrahlung electrons. It appears memory isn't being well managed.

No timing information is used in the tracking at the moment so once the the hits are separated by bunch crossing this will return to a manageable 20ish tracks expected.

The simple, immediate solution to this is to sort the hits by time and use a moving time window on the hits in subsequent layers so that only hits associated with a bunch crossing have a track fit attempted. There are also potentially internal memory optimizations which should be fixed in addition.

However recent discussions and PRs on a meta sorting factory https://github.com/eic/EICrecon/pull/2818 suggest this could be moved out of the algorithm. However, naturally, this changes the algorithm internals and stops it from working on unsorted collections.

The other potentially related PR - timeframe splitting https://github.com/eic/EICrecon/pull/2824. This demonstration of the JANA unfolder always seems to attempt to achieve everything across the whole detector at once. If there isn't much overhead I would propose demonstrating the Unfolder/Folder topology here by slicing the hit collections in each tracker layer by time and passing them through the algorithm and handing the reconstructed tracks back to the Event/Timeslice. My concern here is that the data may be unnecessarily copied and the links might be lost?

wdconinc:
> The other potentially related PR - timeframe splitting [#2824](https://github.com/eic/EICrecon/pull/2824). This demonstration of the JANA unfolder always seems to attempt to achieve everything across the whole detector at once. If there isn't much overhead I would propose demonstrating the Unfolder/Folder topology here by slicing the hit collections in each tracker layer by time and passing them through the algorithm and handing the reconstructed tracks back to the Event/Timeslice. My concern here is that the data may be unnecessarily copied and the links might be lost?

If you pass sliced subset collections then there won't be any unnecessary copying, right? You'd essentially just be passing pointers.

I would very much like to see a small scale application of a JEventUnfolder/Folder algorithm before jumping in with a global algorithm and all the issues it may come with.

simonge:
If you can just pass slices subset collections that's good. However here collections of tracks would be created for every slice, we would want those merged back together as a full collection rather than subset if it's to be saved to the output. Even if the individual event/subevent collections were saved in a separate podio frame I don't believe there is a way to reference from one frame to another.

simonge:
[...]

PR #2841, opened 2026-08-07
Add track relationship to TaggerTrackerReconstructedParticles

simonge:
Adds the relationship between the ReconstructedParticle and the Track which was used to reconstruct it in the LOWQ2 workflow.
Currently these tracks do not have a trajectory associated with them so downstream algorithms which request the trajectory directly have had a protection added before them.

- [x] Medium
- [ ] Low

- [x] Bug fix (related to issue #2834)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

ReconstructedParticle-Track relationship will be populated for Low-Q2 Tagger electrons


src/algorithms/reco/Helix.cc
```diff
@@ -36,7 +36,9 @@ Helix::Helix(const edm4hep::Vector3f& p, const edm4hep::Vector3f& o, const doubl
 Helix::Helix(const edm4eic::ReconstructedParticle& p, const double b_field) {
   const auto& tracks = p.getTracks();
   for (const auto& trk : tracks) {
-    const auto& traj    = trk.getTrajectory();
+    const auto& traj = trk.getTrajectory();
+    if (!traj.isAvailable())
+      continue;
     const auto& trkPars = traj.getTrackParameters();
```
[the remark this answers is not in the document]
simonge: This behaviour matches the current behaviour where there may be no tracks associated with a particle at all.

src/algorithms/fardetectors/FarDetectorTransportationPostML.cc
```diff
@@ -111,6 +111,17 @@ void FarDetectorTransportationPostML::process(
     particle.setMass(m_mass);
     particle.setPDG(m_cfg.pdg_value);
 
+    if (tracks == nullptr) {
+      error("No tracks collection provided; cannot set ReconstructedParticle-Track relation");
+      throw std::runtime_error("No tracks collection provided");
+    }
```
veprbl:
```suggestion
```
Can this really happen? If this is not std::optional in the algorithms interface.

veprbl (dismissed):

src/algorithms/fardetectors/FarDetectorTransportationPostML.cc
simonge: Fair point, it's not optional so shouldn't need this protection.

src/algorithms/reco/Helix.cc
```diff
@@ -36,7 +36,10 @@ Helix::Helix(const edm4hep::Vector3f& p, const edm4hep::Vector3f& o, const doubl
 Helix::Helix(const edm4eic::ReconstructedParticle& p, const double b_field) {
   const auto& tracks = p.getTracks();
   for (const auto& trk : tracks) {
-    const auto& traj    = trk.getTrajectory();
+    const auto& traj = trk.getTrajectory();
+    if (!traj.isAvailable()) {
+      continue;
+    }
```
veprbl, line 42: What is the long-term vision for this? Are we adding the trajectory for far backward tracks and removing this?

simonge: I'll admit I still don't fully understand the layers of tracking objects in the data model. I don't think there is a need to add trajectories and track parameters to the far backwards code. Someone in the future (potentially myself down the line) might refactor to use more centralized acts methods which would presumably create them.

The short term need is for attaching an existing time to the output reconstructed particles. A vertex time would be better but will need some extra work.

simonge: Would this be ok to merge as is and continue the discussion as necessary elsewhere?

PR #2875, opened 2026-08-18
fix(pid): DRICH IRT link migration

wdconinc:
This PR isolates the DRICH IRT link migration from #2822 into a standalone change for focused review. It updates IrtCherenkovParticleID to consume `DRICHRawHitsLinks` (using `getFrom()/getTo()` for link access) while preserving `addToRawHitAssociations` output behavior by still using association objects. DRICH wiring/factory inputs are updated to pass both links and associations.

- [x] Medium
- [ ] Low

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

Used Copilot CLI to split the change, apply review-requested fixes, and update wording/comments for correctness.

### DIFF
--- a/src/detectors/DRICH/DRICH.cc
+++ b/src/detectors/DRICH/DRICH.cc
@@ -137,7 +137,7 @@ void InitPlugin(JApplication* app) {
   app->Add(new JOmniFactoryGeneratorT<IrtCherenkovParticleID_factory>(
       "DRICHIrtCherenkovParticleID",
       {"DRICHAerogelTracks", "DRICHGasTracks", "DRICHMergedTracks", "DRICHRawHits",
-       "DRICHRawHitsLinks"},
+       "DRICHRawHitsLinks", "DRICHRawHitsAssociations"},
       {"DRICHAerogelIrtCherenkovParticleID", "DRICHGasIrtCherenkovParticleID"}, irt_cfg, app));
 
   // merge aerogel and gas PID results
--- a/src/detectors/DRICH/DRICH.cc
+++ b/src/detectors/DRICH/DRICH.cc
@@ -137,7 +137,7 @@ void InitPlugin(JApplication* app) {
   app->Add(new JOmniFactoryGeneratorT<IrtCherenkovParticleID_factory>(
       "DRICHIrtCherenkovParticleID",
       {"DRICHAerogelTracks", "DRICHGasTracks", "DRICHMergedTracks", "DRICHRawHits",
-       "DRICHRawHitsAssociations"},
+       "DRICHRawHitsLinks"},
       {"DRICHAerogelIrtCherenkovParticleID", "DRICHGasIrtCherenkovParticleID"}, irt_cfg, app));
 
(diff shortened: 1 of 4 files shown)


wdconinc:
Correction (formatting only): addressed review feedback from pullrequestreview-4965526672 in commit 040dbfe61.

- clarified `in_hit_links` docs to describe link objects (`MCRecoTrackerHitLink`) rather than raw hits
- updated stale in-code comments to reference `in_hit_links` / link terminology
- fixed trace grammar: "hit" -> "hits"

Associations output behavior is unchanged: `addToRawHitAssociations` still uses association objects.

chchatte92 (commented):
LGTM. Also I have downloaded the artifact and I can print out the number of photons fror Aerogel and Gas. BTW, I noticed that now we have RNTuple. This would require to change the benchmarking. I believe in the benchmark we should have to have 
```
auto reader = podio::RNTupleReader();
```
and podio/RNTupleReader.h has to be added.

wdconinc:
> BTW, I noticed that now we have RNTuple. This would require to change the benchmarking.

The default remains TTree, and only here in EICrecon CI do we explicitly switch to RNTuple. If you want to use RNTuple in the DRICH benchmarks, then you can switch.

veprbl:
> LGTM. Also I have downloaded the artifact and I can print out the number of photons fror Aerogel and Gas. BTW, I noticed that now we have RNTuple. This would require to change the benchmarking. I believe in the benchmark we should have to have
> 
> ```
> auto reader = podio::RNTupleReader();
> ```
> 
> and podio/RNTupleReader.h has to be added.

There is a generic `Reader` interface that automatically handles this
https://github.com/eic/snippets/blob/96b1676aa938161405f5b8ea9cce207ce755a948/ElectronID/InclusiveSkim.C#L30-L31

src/detectors/DRICH/DRICH.cc
```diff
@@ -137,7 +137,7 @@ void InitPlugin(JApplication* app) {
   app->Add(new JOmniFactoryGeneratorT<IrtCherenkovParticleID_factory>(
       "DRICHIrtCherenkovParticleID",
       {"DRICHAerogelTracks", "DRICHGasTracks", "DRICHMergedTracks", "DRICHRawHits",
-       "DRICHRawHitsAssociations"},
+       "DRICHRawHitsLinks", "DRICHRawHitsAssociations"},
```
veprbl, line 140: Need to add this to JEventProcessorPODIO

chchatte92: Isn't it already added?
 In line 132? 
 https://github.com/eic/EICrecon/blob/3319124d3982ce95c797987c8b663a531030528d/src/services/io/podio/JEventProcessorPODIO.cc#L132

wdconinc: Confirmed.

veprbl (approved):

PR #2081, opened 2025-09-18
fix: Add debug logs for time cap exceedance in hits; don't skip them

wdconinc:
### Briefly, what does this PR introduce?
Added debug logging for hits retained despite exceeding `capTime`.

### What kind of change does this PR introduce?
- [x] Bug fix (issue #2080)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
Yes, multiple hits at different times inside a single calorimeter channel will be retained.

PR #2892, opened 2026-08-24
HEXPLIT: adjust timing cut to respect t0

### LINKED ISSUE #2802: Background-merged events rejected by ZDC tmax cut (269 ns)
# Default ZDC `tmax` rejects valid events from `SignalBackgroundMerger`

## Summary

When reconstructing files produced with the `SignalBackgroundMerger`, the default ZDC cluster time cut

```cpp
// src/detectors/ZDC/ZDC.cc (line 183)
.tmax = 269 * dd4hep::ns,
```

rejects valid signal events, even when **all background contributions are set to zero**. 

The `SignalBackgroundMerger` tool spreads events over a time frame to emulate the continuous collision stream, making the default `tmax` possibly incompatible with this workflow. Typically, the time window is of 1000-5000 ns. 

Increasing `tmax` appears to restore the expected reconstructions relying on the ZDC.

Possible limitation: the current value (269 ns) could have been chosen for a good reason. Is this documented? 

@DraTeots @veprbl @mdiefent @sebouh137

veprbl:

150 ns excess cut is relatively large for big chunk hadronic interactions, but we need to support case when t0 != 0, or this removes hits completely.

Related to #2081
Closes: #2802

  For meson structure group
- [ ] Medium
- [ ] Low

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

- [x] This PR introduces breaking changes. Please describe changes users need to make below.
  Old `plugin:factory:tmax` name is replaced by `plugin:factory:max_time_to_truth_t0`
  Claude

### DIFF
--- a/src/algorithms/calorimetry/HEXPLIT.cc
+++ b/src/algorithms/calorimetry/HEXPLIT.cc
@@ -102,17 +102,17 @@ void HEXPLIT::init() {
 void HEXPLIT::process(const HEXPLIT::Input& input, const HEXPLIT::Output& output) const {
 
   const auto [hits, mchitlinks] = input;
-  auto [subcellHits]             = output;
+  auto [subcellHits]            = output;
 
   std::optional<podio::LinkNavigator<edm4eic::MCRecoCalorimeterHitLinkCollection>> nav;
   if (mchitlinks != nullptr && !mchitlinks->empty()) {
     nav.emplace(*mchitlinks);
   }
 
-  double MIP      = m_cfg.MIP / dd4hep::GeV;
-  double delta    = m_cfg.delta_in_MIPs * MIP;
-  double Emin     = m_cfg.Emin_in_MIPs * MIP;
-  double max_dt   = m_cfg.max_time_to_truth_t0 / dd4hep::ns;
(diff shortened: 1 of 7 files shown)


DraTeots:
But... what happens if this timeout is fully removed? Are there analysis plots showing how bad everything is because of this? 

Because again, in spirit of background meeting, if events might appear close to each other in time this cut will do the mess. Or it is seems so.

veprbl:
> But... what happens if this timeout is fully removed? Are there analysis plots showing how bad everything is because of this?

Not much in ZDC, but there appears to be a slight energy cut in forward insert. I don't know why timescales are different.

**edit:** My testing was specifically on neutron benchmarks.

veprbl:
@sebouh137 I wonder if you can share some lore on this. I've implemented this new form of cut assuming this models future hardware cuts (i.e. cutoff with respect to a local t0) that we may be looking to implement. But also @DraTeots raises a valid point about if we could simply throw out the cut for now. I'm happy to go either way, but would like to know your opinion.

PR #2901, opened 2026-08-26
TrackParamTruthInit: implement relaxed mode to include secondaries in B0TrackerTruthSeeds

veprbl:

1. This continues from #2797
2. B0 truth seeding is in a different code path for which
   * The z cut is relaxed
   * Extrapolation to perigee surface is replaced with somewhat degenerate perigee case – surface is placed at the vertex itself
     * This does not seem to break anything, including the existing tracking for the primaries
3. There is one known inefficiency, the tracks are still extrapolated to perigee at (0,0,0), which will not always work for the secondaries. We would probably need to provide "global" track mode for that.

Addresses #2746

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

- [x] This PR introduces breaking changes. Please describe changes users need to make below.
  This change removes the legacy TrackerTruthSeeds / TrackerTruthSeedParameters collections from the default PODIO output list and replaces them with the new central/B0-specific names.
  B0 tracking works slightly differently, but there should not be a visible change to the user analyses
  Claude


src/algorithms/tracking/TrackParamTruthInit.cc
```diff
@@ -87,19 +87,34 @@ void TrackParamTruthInit::process(const Input& input, const Output& output) cons
...
 
-    auto xpca = v.x + linesurface_parameter * p.x;
-    auto ypca = v.y + linesurface_parameter * p.y;
-    auto zpca = v.z + linesurface_parameter * p.z;
+    // Choose the perigee reference point and compute local (d0, z0) coordinates.
+    //
+    // Default: back-extrapolate to the transverse PCA w.r.t. the beam axis at z=0.
+    // This is correct for IP-originating tracks but produces a reference point far
+    // behind the IP for displaced vertices (e.g. Lambda decay daughters at large z),
+    // which forces ACTS to propagate many metres before reaching the detector.
```
bschmookler, line 97: I'm trying to understand why the original approach fails for particles produced at a large z via decays.

If we define a perigee surface at (x,y,z) = (0,0,0) and a second perigee surface at (x,y,z) = (0,0,100cm), for example, I think we still end up with the same transverse PCA point for a given track fit. The only difference is that the second local coordinate is shifted by 100 cm. But in both cases the CFK will end up extrapolating to the same actual point in space.

For the case of particles produced at large z via, I guess we also get a reasonably large x value for the production position. due to the crossing angle. Is the large transverse production position the reason for the failures with a (0,0,z) perigee choice?

src/global/tracking/tracking.cc
```diff
@@ -49,19 +42,26 @@ void InitPlugin(JApplication* app) {
...
+      "CentralTrackerTruthSeeds", {"EventHeader", "MCParticles"},
+      {"CentralTrackerTruthSeeds", "CentralTrackerTruthSeedParameters"}, {}, app));
+
+  // Dedicated B0 truth seeder: relaxed vertex-z cut so that charged daughters of Lambda
+  // decays at large z (up to ~O(meters)) are also seeded for the B0 tracker CKF.
+  app->Add(new JOmniFactoryGeneratorT<TrackParamTruthInit_factory>(
+      "B0TrackerTruthSeeds", {"EventHeader", "MCParticles"},
+      {"B0TrackerTruthSeeds", "B0TrackerTruthSeedParameters"},
+      TrackParamTruthInitConfig{
+          .maxVertexX         = 120 * dd4hep::mm, // maxVertexZ * tan(20mrad) ~ 120mm
```
bschmookler: Why do we want tan(20mrad) here?

The crossing angle in the x-z plane is 25 mrad. So, the outgoing particles (e.g. the Lambdas) are mostly travelling with (px/pz) ~ -0.025.

So, I think the maximum allowed x vertex should be larger than maxVertexZ * tan(20mrad).

veprbl: I think you are right, this is now extended to 40 mrad / 240 mm.

image
image

src/algorithms/tracking/TrackParamTruthInit.cc
veprbl: This documents that the naive extrapolation resulted negative $z$ after extrapolation. The perigee surfaces that you mention are the same line surface, that is understood. My understanding for the effect is that decay protons simply have shallower angles than the crossing angle. But it's tricky because the naive formula that we have looks at transverse DCA, so for some decay configurations it may produce positive $z$ as well, or so I expect. The real problem is not that $z$ is positive or negative, and it's only alluded here, it's that the naive linear extrapolation that does not match with forward propagation and the seeds are not usable.

Let me know if this makes sense. Let's reword the comment in the code into something we both understand.

PR #2910, opened 2026-08-27
fix: switch from OrthogonalSeeding to Seeding2 TripletSeeder

wdconinc:

This PR updates the default seeding method selection logic so that the modern Seeding2 TripletSeeder is used with Acts 45.3+, and fixes a bug in the Seeding2 path that caused ~20% more seeds compared to the Orthogonal seeder.

**Changes:**
1. **`TrackSeedingConfig.h`**: Updated the preprocessor condition from `Acts_VERSION_MAJOR <= 46` to `Acts_VERSION_MAJOR < 45 || (Acts_VERSION_MAJOR == 45 && Acts_VERSION_MINOR < 3)`, so Seeding2 is the default for Acts ≥ 45.3.

2. **`TrackSeeding.cc`** (bug fix): The Seeding2 path was calling `createSeedsFromGroup` **twice** per middle space point — once for low-to-high z candidates (lh) and once for high-to-low z candidates (hl). Because `BroadTripletSeedFilter::filterTripletsMiddleFixed` clears its candidate collector after each call, `maxSeedsPerSpM` was applied **independently** per z-direction. With the default `maxSeedsPerSpM = 0` (= 1 seed allowed), this permitted up to **2 seeds per middle SP**, while the Orthogonal seeder accumulates candidates from both directions in a single filter pass allowing only 1 seed total.

   Fix: combine `bottom_lh_v + bottom_hl_v` and `top_lh_v + top_hl_v` into single subsets and call `createSeedsFromGroup` **once** per middle SP. This is physically valid since all bottom candidates (lh + hl) have `r < rM` and all top candidates have `r > rM`; the doublet finder's physics cuts (cotTheta, deltaR, impact parameter) naturally reject invalid combinations. This matches the Orthogonal seeder behavior and eliminates the ~20% excess seeds.

- [x] Medium
- [ ] Low

- [x] Bug fix (Seeding2 path produced ~20% more seeds than Orthogonal due to per-direction seed counting)
- [x] New feature (complete switch to Seeding2 for Acts ≥ 45.3)
- [ ] Updated documentation
- [ ] other: __

- [x] This PR changes default behavior. Seeding2 becomes the default for Acts ≥ 45.3 (was Acts ≥ 47).
- [x] AI was used in preparing this PR. Root cause identified and fix implemented with GitHub Copilot.

### DIFF
--- a/src/algorithms/tracking/TrackSeeding.cc
+++ b/src/algorithms/tracking/TrackSeeding.cc
@@ -363,25 +363,26 @@ void TrackSeeding::process(const Input& input, const Output& output) const {
     // Build KD-tree
     Acts::Experimental::CylindricalSpacePointKDTree kdTree = kdTreeBuilder.build();
 
-    // Configure KD-tree options for bottom and top candidate searches
-    Acts::Experimental::CylindricalSpacePointKDTree::Options bottomOptions;
-    bottomOptions.rMax   = m_cfg.rMax;
-    bottomOptions.zMin   = m_cfg.zMin;
-    bottomOptions.zMax   = m_cfg.zMax;
-    bottomOptions.phiMin = m_cfg.phiMin;
-    bottomOptions.phiMax = m_cfg.phiMax;
-    // Use bottom-specific deltaR parameters for bottom region
-    bottomOptions.deltaRMin          = m_cfg.deltaRMinBottomSP;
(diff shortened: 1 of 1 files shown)


wdconinc:
FYI @ShujieL @bschmookler Please study the new seeding algorithm on a deterministic timeframe so we can move towards it by default. Parameters can always be tweaked later.

PR #2918, opened 2026-08-28
CalorimeterCALOROCReco: convert to use create() instead of push_back()

veprbl:
PR against #2770 implementing my suggestion

### DIFF
--- a/src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
@@ -359,7 +359,7 @@ void CalorimeterCALOROCReco::process(const CalorimeterCALOROCReco::Input& input,
       const auto& npeHit = NSide ? npeHitN : npeHitP;
       for (const auto& contrib : npeHit.getContributions()) {
         // if contribution is already covered, don't add again
-        if (!seen_contribs.insert(contrib.getObjectID()).second)
+        if (!seen_contribs.insert(contrib.getObjectID()).second) {
           continue;
 
         edep += contrib.getEnergy();
--- a/src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
+++ b/src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
@@ -349,12 +349,10 @@ void CalorimeterCALOROCReco::process(const CalorimeterCALOROCReco::Input& input,
     rawhit.setAmplitude(npeP);
     rawhit.setTimeStamp(time);
 
-    double edep = 0;
+    const double edep = npeHitN.getEnergy() + npeHitP.getEnergy();
 
-    // gather unique contributions (dedup by contribution ObjectID), preserving order
+    // link to parents, deduplicating contributions by ObjectID
     std::unordered_set<podio::ObjectID> seen_contribs;
-    std::vector<std::pair<edm4hep::SimCalorimeterHit, double>> staged; // simHit, weight
-
     for (bool NSide : {true, false}) {
       const auto& npeHit = NSide ? npeHitN : npeHitP;
       for (const auto& contrib : npeHit.getContributions()) {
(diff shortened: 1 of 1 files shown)


src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
```diff
@@ -348,48 +349,27 @@ void CalorimeterCALOROCReco::process(const CalorimeterCALOROCReco::Input& input,
     rawhit.setAmplitude(npeP);
     rawhit.setTimeStamp(time);
 
-    double edep = 0;
-
-    // link to parents
-    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitLink> links_staging;
-    std::unordered_map<podio::ObjectID, edm4eic::MutableMCRecoCalorimeterHitAssociation>
-        rawassocs_staging;
+    const double edep = npeHitN.getEnergy() + npeHitP.getEnergy();
```
ssedd1123: `edep` isn't the sum of energy from both sides because the same hit could contributes to both the 'P' (i.e. +ve side in z-axis) and 'N' side. In the original loop, the addition of `edep` is done after checking for duplications with `unordered_map`. I recommend that we revert back to summing in the for loop.

veprbl: I'm not sure, somehow the current comparison shows zero diff.

veprbl: Let's rerun it again.

veprbl: The logic could be wrong. You compare object id's for hits from different collections. They will never match unless those are somehow subset collections, which I don't think they are.

veprbl: Maybe it would have worked with the original hit contributions from DD4hep, but this now has `EcalBarrelScFiNAttenuatedHitContributions` and such.

ssedd1123: I need to double check. Maybe you're right. My intention was to find all the underlying MC parent particles that give rise to `NpeNHits` and `NpePHits`, and then the weighting factor should be energy from a MC parent/total energy. Because the same MC parent give rise to both `NpeNHits` and `NpePHits`, I thought this would remove duplicated MC parents registered on both ends. 

Perhaps `contrib.getObjectID` isn't quite right. Something like `contrib.getMCParticles().getObjectID()`? Not exact code of course, but just some idea I have in my mind. I'll look it up.

veprbl: Please don't rely on truth information if you can help it. Look to do some projection of cellID's.

ssedd1123: May I asked what is the meaning of weight of each links please? `npeHitN.getEnergy() + npeHitP.getEnergy()` still bugs me because the calibrated energy isn't the sum of the two, but the geometric mean `sqrt(npeHitN.getEnergy()*npeHitP.getEnergy())` (if we ignore digitization and gain factor and only consider attenuation correction that is). I have trouble understanding the meaning of adding those two things together. If you can tell me what are the use cases of the association weights, that will give me a better idea of what to include. 

In my mind, using the truth information gives me the true proportion of how much each particle contribute to the calibrated hit, but now I am not quite sure. I appreciate your help.

veprbl:
Since the original code is copied from digi and does not even fully work, we can close this for now.

src/algorithms/calorimetry/CalorimeterCALOROCReco.cc
veprbl: The link weights quantify the level of causality. Using the fraction of energy is generally an easy way to do so. In your case, if you start summing the hits on the same pair of sipms the square root formula might break and association also becomes trickier.

veprbl: (And yes, I agree, summing is wrong, I did not understand this originally)

PR #2926, opened 2026-08-30
Add ChargedLambdaReconstruction: far-forward Lambda -> p pi- candidates

### LINKED ISSUE #2925: Implement Proton+Pion lambda reconstruction
- the decay proton (Lambda -> p pi-, BR 64%) in the Roman Pots, the
  off-momentum tracker, or the B0 tracker;
- the pi- essentially only in the B0 tracker.

EICrecon reconstructs the *neutral* channel (Lambda -> n pi0 in the
ZDC, `LambdaReconstruction`, collection `ReconstructedLambdas`), but there is no
algorithm for the charged channel.

### Proposal

Add a charged-channel sibling of `LambdaReconstruction`: build all (p, pi-)
pairings from the tracking + far-forward collections and keep those passing a
classic V0 cut chain computable from real data only (momenta and reference
points, no truth):

1. opening angle and pair polar angle (the Lambda points forward);
2. distance of closest approach of the two straight track lines, and the z of
   the line-line midpoint — a decay-vertex proxy;
3. momentum asymmetry (the proton carries ~ m_p/m_Lambda of the Lambda momentum);
4. invariant-mass window under the (m_p, m_pi) hypothesis.

The cut values were tuned on a dedicated Sullivan-process study (meson-structure
campaign, 30 files across 5x41...9x275) at 99% marginal signal efficiency per
cut; on that study's test split the chain keeps 92% of signal pairs while
rejecting 99.9% of combinatorial background. All values are 
configuration parameters.

### Demonstration on 9x130 Sullivan events

Image

Image

Image

Setup: EICrecon `main` (c1e62990) + the attached algorithm, run over 2000
meson-structure dd4hep events (`msf_9x130_1000evt_0001/0002`,
`epic_craterlake_9x130`). Every event contains a Sullivan Lambda; 938 of 2000
decay to p pi- with decay z < 30 m. Because B0 real seeding produces no tracks, the demonstration uses the truth-seeded chain, in two configurations: stock parameters, and the truth-seed vertex acceptance opened to the far-forward decay region (`MaxVertexX/Y = 1 m`, `MaxVertexZ = 30 m` -
"relaxed", gap 2 below).

| configuration | collection | candidates | truth-matched |
|---|---|---|---|
| stock | `ReconstructedChargedLambdas` | 0 | 0 |
| stock | `ReconstructedTruthSeededChargedLambdas` | 7 | 7 |
| relaxed | `ReconstructedChargedLambdas` | 0 | 0 |
| relaxed | `ReconstructedTruthSeededChargedLambdas` | **118** | **110** |

(Truth-matched: candidate momentum within 10% of the generated Lambda momentum
vector.) In the relaxed configuration the algorithm reconstructs the Lambda in
110 of the 938 decaying events — consistent with the acceptance chain measured
in the standalone study (pi- geometric acceptance x CKF efficiency x proton
acceptance ~ 0.14 at these energies) — with 93% candidate purity before any
mass fit. The invariant-mass peak sits at mu = 1117 MeV with sigma = 8.6 MeV;
the line-line vertex proxy tracks the true decay z up to the ~6 m reach of the
B0. At 9x130 both daughters are typically B0 tracks; Roman-Pot/off-momentum
proton legs become relevant at higher-rigidity settings (e.g. 9x275), which is
why those collections are inputs.

DraTeots:

Adds `ChargedLambdaReconstruction`, a charged-channel sibling of
`LambdaReconstruction` (Lambda -> n pi0). It reconstructs Lambda -> p pi-
candidates by pairing proton candidates (positive tracking particles, all
Roman-Pot and off-momentum candidates — both are transfer-matrix
reconstructions under the proton hypothesis) with pi- candidates (negative
tracking particles) and keeping pairs that pass a classic V0 selection:

- opening angle and pair polar angle (forward Lambda);
- distance of closest approach of the two straight track lines and the z of the
  closest-approach midpoint (decay-vertex proxy);
- momentum asymmetry (the proton carries ~m_p/m_Lambda of the Lambda momentum);
- invariant-mass window around the Lambda mass (default ±25 MeV) under the
  (m_p, m_pi) hypothesis.

Every surviving pairing is stored — no best-candidate choice — with the
daughters attached through the `particles` relation, so analyses keep the full
candidate set and the sidebands within the window. Two factories are wired:

| collection | inputs |
|---|---|
| `ReconstructedChargedLambdas` | `ReconstructedChargedParticles` + `ForwardRomanPotRecParticles` + `ForwardOffMRecParticles` |
| `ReconstructedTruthSeededChargedLambdas` | same with `ReconstructedTruthSeededChargedParticles` |

All cut values are configuration parameters. Defaults come from a
Sullivan-process study on the meson-structure campaign (5x41-9x275, cuts fixed
at 99% marginal signal efficiency each on a train split; 92% signal efficiency
at 99.9% background-pair rejection on the test split, before the mass window).

Closes #2925 (it carries the physics motivation,
validation plots and the two seeding gaps that currently limit the real-seeded
collection).

New files:
- `src/algorithms/reco/ChargedLambdaReconstruction.{h,cc}`,
  `ChargedLambdaReconstructionConfig.h`
- `src/factories/reco/ChargedLambdaReconstruction_factory.h`
- `src/tests/algorithms_test/reco_ChargedLambdaReconstruction.cc` — four catch2
  cases: a kinematically exact Lambda decay is found (mass, vertex proxy,
  daughter links), a wide-angle pairing is rejected, charge routing is enforced,
  the mass window follows the configuration.

Touched files: registration in `src/global/reco/reco.cc`, output list in
`JEventProcessorPODIO.cc`, test list in `src/tests/algorithms_test/CMakeLists.txt`.

Validation on meson-structure 9x130 (2000 events, truth-seeded chain with the
vertex acceptance opened to the far-forward decay region):

| configuration | `ReconstructedChargedLambdas` | `ReconstructedTruthSeededChargedLambdas` |
|---|---|---|
| stock parameters | 0 | 7 (7 truth-matched) |
| truth-seed vertex acceptance opened to 30 m | 0 | 118 (110 truth-matched) |

The truth-seeded collection shows a clean Lambda peak (mu = 1117 MeV,
sigma = 8.6 MeV) and its vertex proxy follows the true decay z up to the B0
reach (~6 m); plots are in the linked issue. The real-seeded collection is
empty because B0 seeding currently produces no tracks (see the issue) — the
factory is wired now so the collection fills as soon as seeding is fixed, with
no further changes here.

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #2925)
- [ ] Updated documentation
- [ ] other: __

AI usage: the algorithm, factory, tests and this description were prepared with
Claude Code under the author's direction; the cut selection and validation
numbers come from the author's meson-structure Lambda study, and all code was
reviewed, built and validated by the author.

Adds two new output collections (`ReconstructedChargedLambdas`,
`ReconstructedTruthSeededChargedLambdas`); no existing collection or default
changes.

### DIFF
--- a/src/algorithms/reco/ChargedLambdaReconstruction.cc
+++ b/src/algorithms/reco/ChargedLambdaReconstruction.cc
@@ -1,0 +1,171 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2026 Dmitry Romanov
+
+#include <TVector3.h>
+#include <edm4hep/Vector3f.h>
+#include <cmath>
+#include <vector>
+
+#include "ChargedLambdaReconstruction.h"
+
+namespace {
+
+/** Closest approach of two straight lines r + t*d (d normalized).
+ *
+ * Returns the distance and the midpoint between the two closest points. For
+ * near-parallel lines (a degenerate pairing, not a physical V0) the closest points
+ * are ill-defined; fall back to the perpendicular distance between the lines and the
+ * midpoint of the reference points. */
+struct LineApproach {
+  double distance;
+  TVector3 midpoint;
+};
+
+LineApproach closestApproach(const TVector3& r1, const TVector3& d1, const TVector3& r2,
+                             const TVector3& d2) {
+  const TVector3 cross = d1.Cross(d2);
+  const double denom   = cross.Mag2();
+  const TVector3 dr    = r2 - r1;
+  if (denom < 1e-12) {
+    const TVector3 perp = dr - dr.Dot(d1) * d1;
+    return {perp.Mag(), 0.5 * (r1 + r2)};
+  }
+  const double t1   = dr.Cross(d2).Dot(cross) / denom;
+  const double t2   = dr.Cross(d1).Dot(cross) / denom;
+  const TVector3 p1 = r1 + t1 * d1;
+  const TVector3 p2 = r2 + t2 * d2;
+  return {(p2 - p1).Mag(), 0.5 * (p1 + p2)};
+}
+
+} // namespace
+
+namespace eicrecon {
+
+void ChargedLambdaReconstruction::init() {
+  m_proton_mass = m_particleSvc.particle(2212).mass;
+  m_pion_mass   = m_particleSvc.particle(211).mass;
+  m_lambda_mass = m_particleSvc.particle(3122).mass;
+}
+
+void ChargedLambdaReconstruction::process(const ChargedLambdaReconstruction::Input& input,
+                                          const ChargedLambdaReconstruction::Output& output) const {
+  const auto [charged, roman_pots, off_momentum] = input;
+  auto [out_lambdas]                             = output;
+
+  using ParticleT = edm4eic::ReconstructedParticle;
+
+  // --------------------------------------------------------------------------
+  // Collect the daughter pools. Roman-Pot and off-momentum candidates are
+  // transfer-matrix reconstructions under the proton hypothesis, so all of them
+  // enter the proton pool; tracking candidates are routed by charge.
+  // --------------------------------------------------------------------------
+
+  std::vector<ParticleT> protons;
+  std::vector<ParticleT> pions;
+
+  for (const auto& part : *charged) {
+    if (part.getCharge() > 0) {
+      protons.push_back(part);
+    } else if (part.getCharge() < 0) {
+      pions.push_back(part);
+    }
+  }
+  for (const edm4eic::ReconstructedParticleCollection* coll :
+       {roman_pots.get(), off_momentum.get()}) {
+    for (const auto& part : *coll) {
+      protons.push_back(part);
+    }
+  }
+
+  if (protons.empty() || pions.empty()) {
+    debug("no pairs to build: {} proton candidates, {} pion candidates", protons.size(),
+          pions.size());
+    return;
+  }
+
+  // --------------------------------------------------------------------------
+  // Evaluate every (p, pi-) pairing with the V0 cut chain.
+  // --------------------------------------------------------------------------
+
+  for (const auto& proton : protons) {
+    const auto pp = proton.getMomentum();
+    const auto rp = proton.getReferencePoint();
+    const TVector3 p_mom(pp.x, pp.y, pp.z);
+    const TVector3 p_ref(rp.x, rp.y, rp.z);
+    const double p_mag = p_mom.Mag();
+    if (p_mag < 1e-9) {
+      continue;
+    }
+
+    for (const auto& pion : pions) {
+      const auto pip = pion.getMomentum();
+      const auto rip = pion.getReferencePoint();
+      const TVector3 pi_mom(pip.x, pip.y, pip.z);
+      const TVector3 pi_ref(rip.x, rip.y, rip.z);
+      const double pi_mag = pi_mom.Mag();
+      if (pi_mag < 1e-9) {
+        continue;
+      }
+
+      const double opening = p_mom.Angle(pi_mom);
+      if (opening > m_cfg.openingAngleMax) {
+        continue;
+      }
+
+      const TVector3 pair_mom = p_mom + pi_mom;
+      if (pair_mom.Theta() > m_cfg.pairThetaMax) {
+        continue;
+      }
+
+      const auto approach =
+          closestApproach(p_ref, p_mom.Unit(), pi_ref, pi_mom.Unit()); // [mm], EDM units
+      if (approach.distance > m_cfg.dcaMax) {
+        continue;
+      }
+      if (approach.midpoint.Z() < m_cfg.vertexZMin || approach.midpoint.Z() > m_cfg.vertexZMax) {
+        continue;
+      }
+
+      if (approach.midpoint.Mag() > 1e-9 && pair_mom.Angle(approach.midpoint) > m_cfg.pointingMax) {
+        continue;
+      }
+
+      const double mom_asym = (p_mag - pi_mag) / (p_mag + pi_mag);
+      if (mom_asym < m_cfg.momAsymMin) {
+        continue;
+      }
+
+      // invariant mass under the (m_p, m_pi) hypothesis; input energies are not used
+      // because tracking-based candidates carry no mass assignment
+      const double energy = std::sqrt(p_mag * p_mag + m_proton_mass * m_proton_mass) +
+                            std::sqrt(pi_mag * pi_mag + m_pion_mass * m_pion_mass);
+      const double mass2  = energy * energy - pair_mom.Mag2();
+      const double mass   = std::sqrt(std::max(0.0, mass2));
+      if (std::abs(mass - m_lambda_mass) > m_cfg.massWindow) {
+        continue;
+      }
+
+      auto lambda = out_lambdas->create();
+      lambda.setPDG(3122);
+      lambda.setCharge(0);
+      lambda.setMass(mass);
+      lambda.setEnergy(energy);
+      lambda.setMomentum({static_cast<float>(pair_mom.X()), static_cast<float>(pair_mom.Y()),
+                          static_cast<float>(pair_mom.Z())});
+      lambda.setReferencePoint({static_cast<float>(approach.midpoint.X()),
+                                static_cast<float>(approach.midpoint.Y()),
+                                static_cast<float>(approach.midpoint.Z())});
+      lambda.addToParticles(proton);
+      lambda.addToParticles(pion);
+
+      trace("Lambda candidate: m = {:.4f} GeV, p = {:.2f} GeV, decay-z proxy = {:.0f} mm, "
+            "dca = {:.1f} mm, opening = {:.1f} mrad",
+            mass, pair_mom.Mag(), approach.midpoint.Z(), approach.distance, opening * 1e3);
+    }
+  }
+
+  debug("built {} Lambda candidates from {} proton and {} pion candidates", out_lambdas->size(),
+        protons.size(), pions.size());
+}
+
+} // namespace eicrecon
--- a/src/algorithms/reco/ChargedLambdaReconstruction.cc
+++ b/src/algorithms/reco/ChargedLambdaReconstruction.cc
@@ -97,7 +97,7 @@ void ChargedLambdaReconstruction::process(const ChargedLambdaReconstruction::Inp
     const TVector3 p_mom(pp.x, pp.y, pp.z);
     const TVector3 p_ref(rp.x, rp.y, rp.z);
     const double p_mag = p_mom.Mag();
-    if (p_mag < 1e-9) {
+    if (p_mag < std::numeric_limits<float>::epsilon) {
       continue;
     }
 
--- a/src/algorithms/reco/ChargedLambdaReconstruction.cc
+++ b/src/algorithms/reco/ChargedLambdaReconstruction.cc
@@ -7,6 +7,7 @@
 #include <algorithm>
 #include <cmath>
 #include <initializer_list>
+#include <limits>
 #include <tuple>
 #include <vector>
 
@@ -30,7 +31,7 @@ LineApproach closestApproach(const TVector3& r1, const TVector3& d1, const TVect
   const TVector3 cross = d1.Cross(d2);
   const double denom   = cross.Mag2();
   const TVector3 dr    = r2 - r1;
-  if (denom < std::numeric_limits<float>::epsilon) {
+  if (denom < std::numeric_limits<float>::epsilon()) {
     const TVector3 perp = dr - dr.Dot(d1) * d1;
     return {perp.Mag(), 0.5 * (r1 + r2)};
   }
@@ -46,9 +47,10 @@ LineApproach closestApproach(const TVector3& r1, const TVector3& d1, const TVect
 namespace eicrecon {
 
 void ChargedLambdaReconstruction::init() {
-  m_proton_mass = m_particleSvc.particle(2212).mass;
-  m_pion_mass   = m_particleSvc.particle(211).mass;
-  m_lambda_mass = m_particleSvc.particle(3122).mass;
+  const double mass_min = std::max(0.0, m_lambda_mass - m_cfg.massWindow);
+  const double mass_max = m_lambda_mass + m_cfg.massWindow;
+  m_mass2_min           = mass_min * mass_min;
+  m_mass2_max           = mass_max * mass_max;
 }
 
 void ChargedLambdaReconstruction::process(const ChargedLambdaReconstruction::Input& input,
@@ -97,7 +99,7 @@ void ChargedLambdaReconstruction::process(const ChargedLambdaReconstruction::Inp
     const TVector3 p_mom(pp.x, pp.y, pp.z);
     const TVector3 p_ref(rp.x, rp.y, rp.z);
     const double p_mag = p_mom.Mag();
-    if (p_mag < std::numeric_limits<float>::epsilon) {
+    if (p_mag < std::numeric_limits<float>::epsilon()) {
       continue;
     }
 
@@ -107,7 +109,7 @@ void ChargedLambdaReconstruction::process(const ChargedLambdaReconstruction::Inp
       const TVector3 pi_mom(pip.x, pip.y, pip.z);
       const TVector3 pi_ref(rip.x, rip.y, rip.z);
       const double pi_mag = pi_mom.Mag();
-      if (pi_mag < 1e-9) {
+      if (pi_mag < std::numeric_limits<float>::epsilon()) {
         continue;
       }
 
@@ -130,7 +132,8 @@ void ChargedLambdaReconstruction::process(const ChargedLambdaReconstruction::Inp
         continue;
       }
 
-      if (approach.midpoint.Mag() > 1e-9 && pair_mom.Angle(approach.midpoint) > m_cfg.pointingMax) {
+      if (approach.midpoint.Mag() > std::numeric_limits<float>::epsilon() &&
+          pair_mom.Angle(approach.midpoint) > m_cfg.pointingMax) {
         continue;
       }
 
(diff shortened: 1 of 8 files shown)


src/algorithms/reco/ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,175 @@
...
+ * Returns the distance and the midpoint between the two closest points. For
+ * near-parallel lines (a degenerate pairing, not a physical V0) the closest points
+ * are ill-defined; fall back to the perpendicular distance between the lines and the
+ * midpoint of the reference points. */
+struct LineApproach {
+  double distance;
+  TVector3 midpoint;
+};
+
+LineApproach closestApproach(const TVector3& r1, const TVector3& d1, const TVector3& r2,
```
wdconinc, line 28: It would be better if this could use the secondary vertex finding algorithms instead of adding adhoc DCA algorithms inside this algorithm.

DraTeots:
To provide some broader context for everyone reviewing this PR on Helix and AMVF. The far-forward spectator Lambda requires a different approach for two fundamental physics and architectural reasons. I'm actually pretty careful about reusing rather than rewriting functionality, and I argue the 10-line DCA function is the right tool here.

1. **A spectator Lambda is not "just a geometric vertex"**

A generic vertex finder answers a purely geometric question (where do trajectories cross). But in EIC physics the spectator Lambda is an event-defining physics object, much like the scattered electron. Its reconstruction requires logic that is structurally outside the scope of a geometric vertexer (actually is reversed to vertexer) :

- **Mass hypothesis**: a vertex finder doesn't know about particle masses. With no far-forward hadron PID, we must assign masses by role (positive leg = proton, negative leg = pion) to even calculate an invariant mass. The hypothesis is part of the candidate, the same two tracks under a (pi+, pi-) hypothesis would be a K0s candidate.

- **Hypothesis-specific kinematics**: then we use momentum asymmetry (the proton carries ~m_p/m_lam of the Lambda momentum). A then we descent to other "physics aware" cuts. A generic vertexer doesn't have this concept.

- **This algorithm define candidates** - we do not choose between them. Each cut tests one pair against that pair's own measured properties: opening angle, DCA, momentum asymmetry, invariant mass (and every threshold is a configuration parameter). (!)All passing pairs are stored, and one proton may appear in several pairs(!). The choice between candidates belongs to the physics analysis. Only analysis knows choosing criteria: a yield measurement fits the mass spectrum and needs the sidebands. An exclusive analysis selects the pair consistent with the missing momentum of the e' K+ system. This information is not something this algorithm knows or operates on. Or supposed to know. A vertex-fit selection just doesn't work this way, and not intended, and not designed for.

- **A vertex fit adds nothing for these tracks**: a fit improves on closed-form geometry only when it has covariances or curvature to work with. RP/OffM tracks have neither, for two straight lines, the least-squares vertex is the closest-approach midpoint this function computes (exactly). Same number, no chi2, nothing gained. 

2. **The far-forward lambda children don't fit the standard ACTS track model**

AMVF consumes standard ACTS track objects and linearizes them with a locally helical model around a perigee near the beamline. They need ACTS objects to do some calculations. Putting this into the algorithm by itself would be so much an overcomplication. But, the far-forward region operates differently:

- **No trajectory**: Roman Pot and Off-Momentum candidates are not tracks. The transfer matrix maps a detector hit directly to a momentum vector at the IP. The only physical trajectory we have for these tracks is a straight line. There are no track states, no perigee, no covariance for any fitter to consume.

- **This DCA function is exact**: since the measurement restricts us to straight lines, the 10-line function is the closed-form, mathematically exact closest approach for that geometry, not an approximation a fit would improve.

- **The current seeding can't form the candidate anyway**: AMVF's seeder models each track as a Gaussian in the perigee parameters (d0, z0) evaluated on the beam axis, places the seed transversely at the beamspot constraint, and only associates tracks whose z-at-PCA is within `tracksMaxZinterval` of the seed, which is **3 mm** by default (`SecondaryVertexFinderConfig.h:22`), 10 mm in the `SecondaryVerticesAMVF` registration (`tracking.cc:486`). For far-forward daughters, z-at-PCA doesn't encode the decay position at all: tracks from a decay meters downstream (median ~3 m at 9x130) reach their beamline PCA at unrelated z values, differing from each other by far more than that window , so that the two daughters would never even be grouped into one candidate vertex.

In short, this 10-line function is the correct balanced (in terms of software) solution for the straight-line data the transfer matrices provide. Once true B0 seeding is implemented ([#2746](https://github.com/DraTeots/optim/issues/2746)) and B0×B0 pairs have standard ACTS tracks, it could be moved... I mean, those specific cases could be put into a proper two-track vertex fit. And it might be, that will be the case where a fit genuinely adds information (and a LOT of complications). For the current RP/OffM tracks, the analytical approach is the correct one.

src/algorithms/reco/ChargedLambdaReconstruction.cc
DraTeots: Answered below

wdconinc:
I don't think anyone is arguing that your algorithm *is* the same as geometry vertex reconstruction, but that your algorithm *does* (in part) geometry vertex reconstruction. `closestApproach` is a plainly geometric function on two ReconstructedParticles, in the same way any other resonance decay will need to determine the closest approach (thanks for giving the K0S example). That is what I think we should centralize.

DraTeots:
@wdconinc You created a question technically blocking the PR review process: 

> It would be better if this could use the secondary vertex finding algorithms instead of adding adhoc DCA algorithms inside this algorithm

I hope the response above clears the question. And there is nothing blocking this PR.

src/algorithms/reco/ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,175 @@
...
+                             const TVector3& d2) {
+  const TVector3 cross = d1.Cross(d2);
+  const double denom   = cross.Mag2();
+  const TVector3 dr    = r2 - r1;
+  if (denom < 1e-12) {
```
wdconinc:
```suggestion
  if (denom < std::numeric_limits<float>::epsilon) {
```

src/algorithms/reco/ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,175 @@
...
+  // Evaluate every (p, pi-) pairing with the V0 cut chain.
+  // --------------------------------------------------------------------------
+
+  for (const auto& proton : protons) {
+    const auto pp = proton.getMomentum();
+    const auto rp = proton.getReferencePoint();
+    const TVector3 p_mom(pp.x, pp.y, pp.z);
+    const TVector3 p_ref(rp.x, rp.y, rp.z);
+    const double p_mag = p_mom.Mag();
+    if (p_mag < 1e-9) {
```
wdconinc:
```suggestion
    if (p_mag < std::numeric_limits<float>::epsilon) {
```

src/algorithms/reco/ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,175 @@
...
+      continue;
+    }
+
+    for (const auto& pion : pions) {
+      const auto pip = pion.getMomentum();
+      const auto rip = pion.getReferencePoint();
+      const TVector3 pi_mom(pip.x, pip.y, pip.z);
+      const TVector3 pi_ref(rip.x, rip.y, rip.z);
+      const double pi_mag = pi_mom.Mag();
+      if (pi_mag < 1e-9) {
```
wdconinc, line 110:
```suggestion
      if (pi_mag < std::numeric_limits<float>::epsilon) {
```

src/algorithms/reco/ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      const double mom_asym = (p_mag - pi_mag) / (p_mag + pi_mag);
+      if (mom_asym < m_cfg.momAsymMin) {
+        continue;
+      }
+
+      // invariant mass under the (m_p, m_pi) hypothesis; input energies are not used
+      // because tracking-based candidates carry no mass assignment
+      const double energy = std::sqrt(p_mag * p_mag + m_proton_mass * m_proton_mass) +
```
wdconinc, line 144: Use `std::hypot` for overflow safety.

src/algorithms/reco/ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,175 @@
...
+                            std::sqrt(pi_mag * pi_mag + m_pion_mass * m_pion_mass);
+      const double mass2  = energy * energy - pair_mom.Mag2();
+      const double mass   = std::sqrt(std::max(0.0, mass2));
+      if (std::abs(mass - m_lambda_mass) > m_cfg.massWindow) {
+        continue;
+      }
```
wdconinc, line 150: Do the selection on squared masses, with init-stage squared massWindow.

src/algorithms/reco/ChargedLambdaReconstruction.h
```diff
@@ -0,0 +1,73 @@
...
+            "particles with V0 topology and kinematics cuts"} {}
+
+  void init() final;
+  void process(const Input&, const Output&) const final;
+
+private:
+  const algorithms::ParticleSvc& m_particleSvc = algorithms::ParticleSvc::instance();
+  double m_proton_mass{0};
+  double m_pion_mass{0};
+  double m_lambda_mass{0};
```
wdconinc, line 70: Make const and initialize to value from the service right away.

src/algorithms/reco/ChargedLambdaReconstructionConfig.h
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2026 Dmitry Romanov
+
+#pragma once
+
+#include <cmath>
+
+namespace eicrecon {
+
+struct ChargedLambdaReconstructionConfig {
```
wdconinc, line 10: All configs need units.

src/algorithms/reco/ChargedLambdaReconstructionConfig.h
```diff
@@ -0,0 +1,35 @@
...
+  double dcaMax = 90.0;
+  /** accepted z range of the two-line closest-approach midpoint (decay-vertex proxy) [mm] */
+  double vertexZMin = -1500.0;
+  double vertexZMax = 32000.0;
+  /** maximum angle between the pair momentum and the direction from the origin to the
+   *  decay-vertex proxy [rad]. Disabled by default (pi): Roman-Pot and off-momentum
+   *  candidates are transfer-matrix reconstructions whose track line passes through the
+   *  origin, which degenerates this classic V0 variable until far-forward candidates
+   *  carry an independent vertex estimate. */
+  double pointingMax = M_PI;
```
wdconinc, line 27: Use `std::numbers::pi`.

src/tests/algorithms_test/reco_ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,179 @@
...
+#include "algorithms/reco/ChargedLambdaReconstructionConfig.h"
+
+using eicrecon::ChargedLambdaReconstruction;
+using eicrecon::ChargedLambdaReconstructionConfig;
+
+namespace {
+
+constexpr double M_LAMBDA = 1.115683;
+constexpr double M_PROTON = 0.9382720813;
+constexpr double M_PION   = 0.13957039;
```
wdconinc, line 22: Use service in tests too.

src/tests/algorithms_test/reco_ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,179 @@
...
+}
+
+} // namespace
+
+TEST_CASE("ChargedLambdaReconstruction finds a true p pi- pair", "[ChargedLambdaReconstruction]") {
+  ChargedLambdaReconstruction algo("ChargedLambdaReconstruction");
+  setupAlgo(algo);
+
+  const edm4hep::Vector3f decay_vertex{0.F, 0.F, 6000.F}; // [mm], meters downstream of the IP
+  const auto daughters = lambdaDecayDaughters(100.0, M_PI / 2);
```
wdconinc, line 65: Units.

src/algorithms/reco/ChargedLambdaReconstruction.cc
```diff
@@ -0,0 +1,175 @@
...
+      if (approach.distance > m_cfg.dcaMax) {
+        continue;
+      }
+      if (approach.midpoint.Z() < m_cfg.vertexZMin || approach.midpoint.Z() > m_cfg.vertexZMax) {
+        continue;
+      }
+
+      if (approach.midpoint.Mag() > 1e-9 && pair_mom.Angle(approach.midpoint) > m_cfg.pointingMax) {
+        continue;
+      }
```
ruse-traveler: Small point, but I'd suggest using `epsilon` here as well unless the `1e-9` is specifically needed!
```suggestion
      if (approach.midpoint.Mag() > std::numeric_limits<float>::epsilon && pair_mom.Angle(approach.midpoint) > m_cfg.pointingMax) {
        continue;
      }
```

ruse-traveler (dismissed):
Hi @DraTeots! Thanks for the contribution! @wdconinc already made all of the suggestions I would have, so once those have been addressed I'm happy to merge!

Also, I appreciate the clarification on the differences between this and a general vertex finder! Building on the discussion so far, the idea wouldn't be that a vertex finder completely replaces this algorithm. Instead, a unified approach would likely be a sequence of algorithms: the 1st being the vertex finding step, and the following ones filtering vertices by potential decay channels and reconstructing kinematics.

Of course, we're still a far ways off from that right now! So I think it's a reasonable strategy to merge this now and refactor later!

veprbl:
I don't quite understand how this works. The momenta are parameterized at IP (for B0 through Acts extrapolation and for OMD/RP through transfer matrix), but a linear extrapolation is sufficient?

It's still unclear to me why this needs a full custom secondary vertex finder. One could be converting OMD/RP information to tracks and running AMVF, or even Helix, then apply the same analysis cuts as needed. It certainly simplifies life to contribute this, but we are also looking down the barrel of maintaining this, so I'd really like to understand it or at least know that someone else understands it.

> So I think it's a reasonable strategy to merge this now and refactor later!

Who will be doing this refactoring? Do we know that they will are equipped with necessary information to do a drastic transition?

ruse-traveler (approved):
Thanks for integrating our feedback @DraTeots! Still looks good to me! I do second @veprbl's question, though: it would be good to have somebody identified (at least provisionally) who will carry out the refactoring and validating it...

veprbl (changes requested):
See my other message above

DraTeots:
I start with the back. 

**Why not AMVF/Helix**. AMVF is a track-pool fitter: it takes tracks from a pool, assigns each to a vertex, and removes it from the pool(!). That is very bold point why it is not suitable for this reconstruction conceptually. There is no pool (and the pool should not be used!). A proton candidate may legitimately belong to several pairs, and all of them are kept. The decision must be taken on the analysis level. How this selection is made - is analysis specific (! can't stress enough how important this is). That is the standard way V0 candidates are built: CDF and D0 stored every opposite-charge pair passing topological cuts, daughters shared, and the analysis resolved the ambiguity (best candidate for exclusive measurements, sidebands for yields). ALICE and STAR V0 finders do the same. A fitter that consumes tracks makes that choice at reconstruction level, where the information to make it does not exist.

**The logic is the reverse of a vertexer.** A vertexer finds a point and then asks what decayed there. This algorithm starts from the hypothesis: Lambda => p pi-. Then masses assigned by role because there is no far-forward PID, and uses geometry only as a consistency cut on the hypothesis: opening angle, DCA, decay-z proxy, momentum asymmetry, invariant mass. ClosestApproach is not a vertex finder; it is just one of the cuts (and not the first one). 

**Linear extrapolation**. There is nothing else to extrapolate. An RP/OMD candidate is a transfer-matrix: the optics between IP and station is folded in. The output is a momentum at the origin under the proton hypothesi. No trajectory, no covariance. A B0 candidate is an Acts fit already propagated through the field to its perigee. In both cases the straight line is the tangent at the reference point, and for an on-axis spectator Lambda the pion's perigee is the decay point. For two lines without covariance the least-squares vertex is exactly the closest-approach midpoint; a fit returns the same point without a χ². The residual approximation is the IP-origin assumption inside the transfer matrix, and its effect is measured (sigma = 8.6 MeV) on the mass peak and the vertex proxy follows the true decay z to the B0 reach.

DraTeots:
Without AI knoweledge: we did it in Fermilab (basically the part of my thesis) and I know my previous group do the same at STAR and CERN. I also checked about Hera did it. If you want to dive into "AI slop". I asked Claude to spawn agents and check codebases/articles on how such reconstruction is made. According to it: 

**Every major collider experiment: ALICE, STAR, CMS, ATLAS, LHCb, Belle II, BaBar, CDF, D0, ZEUS, H1** - builds V0 candidates from all opposite-charge pairs, keeps every candidate, shares daughters, and resolves the ambiguity in the analysis. Where a fit is used it is a per-pair two-track fit, a pool fitter that consumes tracks is not used.

veprbl:
> AMVF is a track-pool fitter: it takes tracks from a pool, assigns each to a vertex, and removes it from the pool(!). That is very bold point why it is not suitable for this reconstruction conceptually.

You can instruct your AI to check how it's used in EICrecon.

> ClosestApproach is not a vertex finder; it is just one of the cuts (and not the first one).

Great! So, to avoid the confusion, we should reuse existing vertex finder and implement just the cuts?

> Linear extrapolation. 

I don't understand at all what you say. Where are the lambda vertices that you want to reconstruct along z? $\pm 1$ cm? +6 m?

mdiefent:
I agree that using the existing vertexing infrastructure would be the cleaner long-term solution. My understanding, however, is that this would first require integrating the far-forward detectors into ACTS, which is a substantial amount of additional work and for which there does not appear to be dedicated effort available.

Given that the DCA-based approach provides a working solution for the meson-structure studies and has been validated for the intended use case, I suggest accepting the PR rather than blocking the analysis on the larger ACTS integration effort.

It would be good to document clearly in the code that the DCA treatment is an approximation, along with the conditions under which it has been tested. I would also make it clear in the code comments that this approach should be revisited once a more complete far-forward vertexing solution becomes available.

wdconinc:
> My understanding, however, is that this would first require integrating the far-forward detectors into ACTS, which is a substantial amount of additional work and for which there does not appear to be dedicated effort available.

The Helix method is completely unrelated to ACTS, not affected by ACTS far forward geometry, and reverts to straight line intersections in the absence of fields. It is appropriate for use here and can be used today.

ruse-traveler:
Hi all, I'd like to elaborate a little on why I advocated (and still advocate) for merging this:
- I'm having a hard time seeing this introducing significant technical debt. We can easily drop this once we have a more unified approach in hand, even if it comes relatively quickly. I'm also not seeing significant impact on our performance or running stability. (Apologies if I'm missing something, of course!)
- Moreover, this provides functionality that the Exclusive WG has explicitly stated their interest in. I don't think we risk any precedent by merging stopgap algorithms like this if they (1) provide desired functionality and (2) don't introduce serious technical debt/performance degradation.
- If there is a plan to eventually deprecate this in favor of a unified approach and we have a contact in either a PWG/DSC who will be able to validate the change, I think it does us no harm to merge now.

Now all that being said, I did want to respond to one point you raised @DraTeots: 

> Every major collider experiment: ALICE, STAR, CMS, ATLAS, LHCb, Belle II, BaBar, CDF, D0, ZEUS, H1 - builds V0 candidates from all opposite-charge pairs, keeps every candidate, shares daughters, and resolves the ambiguity in the analysis.

Correct! Like I mentioned above, a unified approach would involve a sequence of algorithms (vertex finding, filtering by topology, etc.). The goal would be to make sure that this sequence preserves the combinatorics and makes it available to the user so that they can still select candidates according to their analysis needs.

Issue #1956, opened 2025-07-07
Create ReconstructedParticles from track cluster matches


tlprotzman:
**Is your feature request related to a problem? Please describe.**
Currently the `ReconstructedParticles` type is created by the `MatchClusters` algorithm making use of truth data.  To get a better representation of the performance of the ePIC detector, we would like to remove as much use of truth data as we are able to.  

**Describe the solution you'd like**
Instead of creating the `ReconstructedParticles` from associations between truth particles and clusters, we can now use the `TrackClusterMatch` factories.  This will both centralize the location where matching is done (currently multiple algorithms have their own implementations) as well as remove the dependency on truth data. 

**Describe alternatives you've considered**
The initial motivation for this change was to use reco track cluster matching as the input the the electron ID chain.  We could simply update that algorithm to accept `TrackClusterMatches` as an input instead, but this would not be sustainable in the long run as it would require updating every time we want to use more information in the selection process.

**Additional context**
Simply using the track cluster matching alone to create `ReconstructedParticles` would not be sufficient to be a drop in replacement to the current collections.  It does not handle any left over neutral components, nor does it incorporate any of the currently available PID lookup tables.  To have parity with the `MatchCluster` output an additional step to incorporate the neutral candidates is required.

@ruse-traveler has also pointed out that this is similar to the particle flow PFA1(b) step, the charged candidate maker.

PR #2937, opened 2026-09-01
Implement Reco-Only Equivalent to `MatchClusters` (EFZB)

ruse-traveler:

This PR implements the sequence of algorithms illustrated below, which is intended to be a reco-only equivalent to the `MatchClusters` algorithm.
```mermaid
flowchart TB
    tcm([Track-Cluster Matches]) --> fmc(Filter Matched Clusters)
    clu([Clusters]) --> fmc
    fmc --> umc([Unmatched Clusters])
    umc --> c2p(Convert Clusters to Particles)
    c2p --> neu([Neutral Reconstructed Particles])
    chr([Charged Reconstructed Particles]) --> col(Collect Collections)
    neu --> col
    col --> sub([Reconstructed Particles subset])
    sub --> clo(Clone)
    clo --> par([Reconstructed Particles])
```

For easier comparison, this workflow is referred to as "Energy-Flow Zero (EFZero)" since this is (1) a simple energy flow algorithm, and (2) a 0th order "baseline-to-the-baseline" comparison for PFAlpha. Outputs are labeled with "Zero" to distinguish them from defaults.

**Note:** as this is intended as the reco-only equivalent to `MatchClusters`, only EMCal clusters are utilized. Integrating HCal clusters is one of the goals of PFAlpha.

- [ ] Medium
- [x] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #1956 )
- [ ] Updated documentation
- [ ] other: __

The collections `ReconstructedNeutralParticlesZero`, `ReconstructedParticlesZero`, and relevant associations/links will now appear in the default output.

### DIFF
--- a/src/global/particle_flow/particle_flow.cc
+++ b/src/global/particle_flow/particle_flow.cc
@@ -47,52 +47,63 @@ void InitPlugin(JApplication* app) {
   // EFZ (B) using only reco info
   // --------------------------------------------------------------------
 
-  app->Add(new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::TrackClusterMatch, false>>(
-      "EcalTrackClusterMatches",
-      {"EcalEndcapNTrackClusterMatches", "EcalBarrelTrackClusterMatches", "EcalEndcapPTrackClusterMatches"},
-      {"EcalTrackClusterMatches"}, app));
+  app->Add(
+      new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::TrackClusterMatch, false>>(
+          "EcalTrackClusterMatches",
+          {"EcalEndcapNTrackClusterMatches", "EcalBarrelTrackClusterMatches",
+           "EcalEndcapPTrackClusterMatches"},
+          {"EcalTrackClusterMatches"}, app));
+
+  app->Add(
+      new JOmniFactoryGeneratorT<FilterMatching_factory<
+          edm4eic::Cluster, [](auto* obj) { return obj->getObjectID(); },
+          edm4eic::TrackClusterMatch, [](auto* obj) { return obj->getCluster().getObjectID(); }>>(
+          "MatchedEcalClusters", {"EcalClusters", "EcalTrackClusterMatches"},
+          {"MatchedEcalClusters", "UnmatchedEcalClusters"}, app));
 
   app->Add(new JOmniFactoryGeneratorT<FilterMatching_factory<
-               edm4eic::Cluster, [](auto* obj) { return obj->getObjectID(); },
-               edm4eic::TrackClusterMatch, [](auto* obj) { return obj->getCluster().getObjectID(); }>>(
-      "MatchedEcalClusters", {"EcalClusters", "EcalTrackClusterMatches"},
-      {"MatchedEcalClusters", "UnmatchedEcalClusters"}, app));
-
-  app->Add(new JOmniFactoryGeneratorT<FilterMatching_factory<
-               edm4eic::MCRecoClusterParticleAssociation, [](auto* obj) { return obj->getRec().getObjectID(); },
-               edm4eic::TrackClusterMatch, [](auto* obj) { return obj->getCluster().getObjectID(); }>>(
+               edm4eic::MCRecoClusterParticleAssociation,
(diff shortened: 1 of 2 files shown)


src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -483,6 +483,14 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
       "EcalEndcapNTrackClusterMatches",
       "HcalEndcapNTrackClusterMatches",
 
+      // energy flow
+      "ReconstructedNeutralParticlesZero",
+      "ReconstructedNeutralParticleZeroAssociations",
```
veprbl:
```suggestion
```
Let's not add new associations.

src/services/io/podio/JEventProcessorPODIO.cc
```diff
@@ -483,6 +483,14 @@ JEventProcessorPODIO::JEventProcessorPODIO() {
...
+      "ReconstructedNeutralParticleZeroLinks",
+      "ReconstructedParticlesZero",
+      "ReconstructedParticleZeroAssociations",
```
veprbl:
```suggestion
```

src/global/particle_flow/particle_flow.cc
```diff
@@ -32,6 +41,72 @@ void InitPlugin(JApplication* app) {
...
+      "ReconstructedParticlesZero", {"ReconstructedParticlesZeroSubset"},
+      {"ReconstructedParticlesZero"}, app));
+
+  app->Add(new JOmniFactoryGeneratorT<Cloner_factory<edm4eic::MCRecoParticleLink>>(
+      "ReconstructedParticleZeroLinks", {"ReconstructedParticleZeroLinksSubset"},
+      {"ReconstructedParticleZeroLinks"}, app));
+
+  app->Add(new JOmniFactoryGeneratorT<Cloner_factory<edm4eic::MCRecoParticleAssociation>>(
+      "ReconstructedParticleZeroAssociations", {"ReconstructedParticleZeroAssociationsSubset"},
+      {"ReconstructedParticleZeroAssociations"}, app));
```
veprbl, line 108: The links will still point to objects pointed by `ReconstructedParticlesZeroSubset`.

veprbl (commented):
This is a cute way to do it, but there is an edge case.

src/global/particle_flow/particle_flow.cc
```diff
@@ -32,6 +41,72 @@ void InitPlugin(JApplication* app) {
...
+  // ====================================================================
+  // EFZero: minimal reference EF
+  // ====================================================================
+
+  // --------------------------------------------------------------------
+  // EFZ (B) using only reco info
+  // --------------------------------------------------------------------
+
+  app->Add(
+      new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::TrackClusterMatch, false>>(
```
veprbl:
```suggestion
      new JOmniFactoryGeneratorT<CollectionCollector_factory<edm4eic::TrackClusterMatch, true>>(
```
Making those optional is fine too.

src/global/particle_flow/particle_flow.cc
ruse-traveler: Ahhhh I see! So these would still be between `MCParticles` and the original reconstructed particles. Hmmm...

Two approaches, then, that come to my mind are:
1. We extend the cloner algorithm to optionally accept a vector of link collections and create new links for each provided;
2. Or we could generalize the protocluster-track match promoter to take copy links between two (related) collections...

veprbl: good options, but I can add two more

3. not use cloners (for now?)
4. write a generic link promoter

ruse-traveler: Also good suggestions! I'd be happy grappling with the issue in a follow-up PR, so I'm fine not using the cloners for now 

(TBH those were there mostly for anyone analyzing the collections using older TTreeReader code. I don't think we're going to have anyone doing that anytime soon, though...)

ruse-traveler: Done in [e45143a](https://github.com/eic/EICrecon/pull/2937/commits/e45143a2152c61e24b8f9fe296c70e5aebcf6a99)!

veprbl (approved):

PR #2938, opened 2026-09-02
Change FF LGAD resolution from 8ns to 30ps

simonge:
Updates the FF tracker timing resolution from 8ns which was the default for all detectors when introduced to the 30ps expected by the sensor (referenced from the now quite old digitization spreadsheet https://docs.google.com/spreadsheets/d/1s8oXj36SqIh7TJeHFH89gQ_ayU1_SVEpWQNkx6sETKs/edit?gid=238482234#gid=238482234)

Studies by @Garypenman highlighted looking for coincidences between the FF protons and FB electrons was being smeared out significantly more than anticipated. @ajentsch 

A more detailed digitization approach and the correctness of the resolution value of other detectors are not being considered in this PR

- [x] Medium
- [ ] Low

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

### DIFF
--- a/src/detectors/B0TRK/B0TRK.cc
+++ b/src/detectors/B0TRK/B0TRK.cc
@@ -8,6 +8,7 @@
 #include <JANA/Utils/JTypeInfo.h>
 #include <string>
 #include <vector>
+#include <edm4eic/unit_system.h>
 
 #include "extensions/jana/JOmniFactoryGeneratorT.h"
 #include "factories/digi/SiliconTrackerDigi_factory.h"
@@ -25,7 +26,7 @@ void InitPlugin(JApplication* app) {
       {"B0TrackerRawHits", "B0TrackerRawHitLinks", "B0TrackerRawHitAssociations"},
       {
           .threshold      = 10.0 * dd4hep::keV,
-          .timeResolution = 30 * dd4hep::picosecond,
+          .timeResolution = 30 * edm4eic::unit::ps,
       },
       app));
 
@@ -33,7 +34,7 @@ void InitPlugin(JApplication* app) {
   app->Add(new JOmniFactoryGeneratorT<TrackerHitReconstruction_factory>(
       "B0TrackerRecHits", {"B0TrackerRawHits"}, {"B0TrackerRecHits"},
       {
-          .timeResolution = 30 * dd4hep::picosecond,
+          .timeResolution = 30 * edm4eic::unit::ps,
       },
       app));
 }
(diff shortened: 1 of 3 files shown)


src/detectors/B0TRK/B0TRK.cc
```diff
@@ -25,15 +25,15 @@ void InitPlugin(JApplication* app) {
       {"B0TrackerRawHits", "B0TrackerRawHitLinks", "B0TrackerRawHitAssociations"},
       {
           .threshold      = 10.0 * dd4hep::keV,
-          .timeResolution = 8,
+          .timeResolution = 30 * dd4hep::picosecond,
```
veprbl: Why is this in dd4hep:: units and not edm4eic::unit::?

I don't see the conversion in https://github.com/eic/EICrecon/blob/50d5375fa2526e76734120bc8bfb42d65a99bf3c/src/algorithms/digi/SiliconTrackerDigi.cc#L44

simonge: Aside from that they are the same thing here, should we not be using dd4hep units when interfacing with the dd4hep simhit outputs?

Would the TrackerHitReconstruction_factory config not be a more necessary place to ensure the edm4eic units are being used?

ajentsch: Yeah, I am not understanding the concern here. dd4hep:: units are used throughout our various geometry and detector definitions, why does it matter?

In any case, the change to 30ps is perfectly fine - it's around what we expect to deliver, we just have not been directly using any timing information up to this point, so updating the number is a good idea.

veprbl: When DD4hep writes EDM4hep, those are EDM4hep units, not DD4hep. The documentation of those exists in edm4eic::

> Aside from that they are the same thing here, should we not be using dd4hep units when interfacing with the dd4hep simhit outputs?

They are not! `dd4hep::picosecond = 1e-12`

veprbl: Also, coincidentally, see https://github.com/eic/EICrecon/pull/2939

veprbl: Also references from distant past: https://github.com/eic/EICrecon/pull/390 https://github.com/eic/EDM4eic/pull/35

The `dd4hep::keV` above exists basically from the time of porting from Gaudi - it should be fixed too.

simonge: Ok, looking around at the various files I agree, just assumed that whatever was coming out of dd4hep was dd4hep but its really just the geometry which isn't being automatically converted to edm4hep units, and dd4hep isn't compiled with `DD4HEP_USE_GEANT4_UNITS` although even if it was it would always be safer not to use...

simonge:
> In any case, the change to 30ps is perfectly fine - it's around what we expect to deliver, we just have not been directly using any timing information up to this point, so updating the number is a good idea.

Thanks @ajentsch

ajentsch (approved):

PR #2943, opened 2026-09-07
Remove -Wno-error=maybe-uninitialized from gcc build

wdconinc:
This PR addresses https://github.com/eic/epic/issues/1104 which is filed there but actually applies to code in https://github.com/eic/epic/pull/865 that is copied from here. So, it applies to here... It is likely not necessary anymore since we don't have gcc-12 systems we build on anymore.

- [x] Medium
- [ ] Low

- [x] Bug fix (issue: tech debt paydown)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

### Follow-up fix for GCC 14 CI failure
After removing `-Wno-error=maybe-uninitialized`, GCC 14 started failing in `CalorimeterParticleIDPostML` with a `-Wmaybe-uninitialized` diagnostic in libstdc++ (`stl_tree.h`) triggered through `std::optional<podio::LinkNavigator<...>>` in `EventLinkNavigator`.

This PR now includes a code-side workaround in `src/algorithms/interfaces/LinkTruthUtils.h`: `EventLinkNavigator` stores the navigator as `std::unique_ptr<podio::LinkNavigator<...>>` and only allocates it when input links are present. This preserves behavior while avoiding the false-positive warning path in `std::optional` destruction on GCC 14.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -98,8 +98,7 @@ jobs:
         include:
           - CXX: g++
             # TODO(#2524): remove -Wno-error=deprecated-declarations once seeding migrates to Acts Seeding2 (Acts v46 deprecations)
-            # -Wno-error=maybe-uninitialized for GCC only as it has issues with /usr/include/c++/12/bits/std_function.h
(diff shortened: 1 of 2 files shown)


wdconinc:
Also with gcc-14 there is still uninitialized behavior in std_function.h.

PR #2946, opened 2026-09-08
fix: correct ZDC photon identification units

### LINKED ISSUE #2897: ReconstructedHcalFarForwardZDCNeutrals Collection Inventing Fake Photons
When analysing DEMP (e + p -> e' + Pi + n) files from the July campaign, noticed something odd when switching from ReconstructedFarForwardZDCNeutrals (which no longer exists) to ReconstructedHcalFarForwardZDCNeutrals.

Despite only a neutron being produced in this reaction going towards the FF detectors, all events seemed to have either 0 or 2+ particles reconstructed in this collection. 

One was reconstructed with 0 mass at the origin (0,0,0) whilst the other reconstructed as a neutron at the position of the ZDC. See attached plots below.

Starting with the size of the collection, no events with 1 reconstructed particle in the ZDC seem to exist. It's either nothing or 2+.
Image

When plotting the Z reference position, would expect this to be at the ZDC (>~35000mm) - Red is for mass != 0, blue is mass == 0.
Image

When looking at the energy, the particle with mass == 0 seems to always be lower energy. Here, red is mass == 0 and blue is mass != 0 (apologies for the switcheroo on the crappy TTree draw plots).
Image

Removing the photon is relatively easy for DEMP events, but this is concerning as this particle seems to be completely fictional. No actual photon with this energy was produced or detected.

- Files from 26.07.01 campaign/container

1. Grab a DEMP file from the 26.07.01 campaign, e.g. `root://epicxrd1.sdcc.bnl.gov:1095//eic/EPIC//RECO/26.07.1/epic_craterlake/EXCLUSIVE/DEMP/DEMPgen-1.2.5/9x130/q2_3_10/pi+/DEMPgen-1.2.5_9x130_pi+_q2_3_10_ab.0001.eicrecon.edm4eic.root` or similar
2. Check the ReconstructedHcalFarForwardZDCNeutrals collection, particularly the size of this collection. 
3. Plot some basic quantities under the condition that the mass is equal or unequal to 0 to compare the difference.

Should only see a neutron (or nothing, not detecting it is fine/expected).

Get a reconstructed photon that doesn't actually exist or correspond to a real particle along with the neutron.

Why is a fake photon being added? If this is a branch for "lambda hypotheses", this should be clearer (but would also still be concerning that a photon is being created when it does not exist).

baptistefraisse:
Fixes #2897.
This PR fixes a unit inconsistency in FarForwardNeutralsReconstruction::isGamma() that caused ZDC cluster dimensions to be incorrectly scaled before comparison with the photon-identification thresholds. It also sets the reconstructed photon referencePoint to the associated cluster position instead of leaving it at the default (0, 0, 0).

### DIFF
--- a/src/algorithms/reco/FarForwardNeutralsReconstruction.cc
+++ b/src/algorithms/reco/FarForwardNeutralsReconstruction.cc
@@ -38,14 +38,13 @@ void FarForwardNeutralsReconstruction::init() {
 */
 bool FarForwardNeutralsReconstruction::isGamma(const edm4eic::Cluster& cluster) const {
 
-  double l1 = sqrt(cluster.getShapeParameters(4)) * dd4hep::mm;
-  double l2 = sqrt(cluster.getShapeParameters(5)) * dd4hep::mm;
-  double l3 = sqrt(cluster.getShapeParameters(6)) * dd4hep::mm;
+  double l1 = sqrt(cluster.getShapeParameters(4));
+  double l2 = sqrt(cluster.getShapeParameters(5));
+  double l3 = sqrt(cluster.getShapeParameters(6));
 
   // z in the local coordinates
   double z = (cluster.getPosition().z * cos(m_cfg.globalToProtonRotation) +
-              cluster.getPosition().x * sin(m_cfg.globalToProtonRotation)) *
-             dd4hep::mm;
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
Looks good to me. Units should indeed be consistently handled without additional multiplications.

PR #2953, opened 2026-09-11
BUGFIX: avoid flaky round_beam_four_momentum exceptions

veprbl:

`round_beam_four_momentum` depends on numerical tolerance window when it decides to throw. Upstream, on the side of `JEventProcessorPODIO` we can't be supplying inconsistent frames to `PODIO` - that leads to segfaults in `podio::ROOTWriter::resetBranches`.

- [x] Medium
- [ ] Low

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

  claude

### DIFF
--- a/src/algorithms/reco/HadronicFinalState.cc
+++ b/src/algorithms/reco/HadronicFinalState.cc
@@ -12,6 +12,7 @@
 #include <podio/ObjectID.h>
 #include <algorithm>
 #include <cmath>
+#include <exception>
 #include <tuple>
 
 #include "Beam.h"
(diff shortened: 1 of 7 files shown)


wdconinc (commented):
I am not sure that this is the right approach. This is not a bugfix, and it's not "flaky" rounding. Hiding all exception in debug and returning as if nothing happened is also not great. As shown in #2955, the issue this is trying to address is actually worthy of an exception: we don't have the beam momentum of the failing data set in the list of accepted beam momentum values, resulting in no physics output. That's something we would legitimately want to have an exception-induced runtime failure for.

veprbl:
Agreed.

PR #2957, opened 2026-09-12
fix: use MCBeamHadrons instead of MCBeamProtons in kinematics factories

wdconinc:
## Problem

`HadronicFinalState` and all `InclusiveKinematics*` factories took `MCBeamProtons` as their beam hadron input. For ion beams such as He3, BeAGLE emits the **struck nucleon** (a neutron) as the beam particle with `generatorStatus=4`, so `MCBeamProtons` is empty and `MCBeamNeutrons` is populated instead. This caused all kinematics algorithms to early-return with "No beam hadron found" for every event in He3 datasets.

## Fix

Switch the beam hadron input from `MCBeamProtons` to `MCBeamHadrons` in:
- `HadronicFinalState` (header + impl + factory registration in `reco.cc`)
- `InclusiveKinematicsElectron`, `DA`, `JB`, `Sigma`, `ESigma` (same three locations each)

`MCBeamHadrons` is the union of `MCBeamProtons` and `MCBeamNeutrons`, produced by the `CollectionCollector` in `beam.cc`, so it correctly covers both proton and neutron/ion beam configurations with no change to existing proton-beam event processing.

## Related

Companion to #2956 (which adds He3 per-nucleon momenta to `hadron_beam_pz_set` so `round_beam_four_momentum` does not throw for the spectator proton case).

### DIFF
--- a/src/algorithms/reco/HadronicFinalState.cc
+++ b/src/algorithms/reco/HadronicFinalState.cc
@@ -27,7 +27,7 @@ void HadronicFinalState::init() {}
 void HadronicFinalState::process(const HadronicFinalState::Input& input,
                                  const HadronicFinalState::Output& output) const {
 
-  const auto [mc_beam_electrons, mc_beam_protons, mcparts, rcparts, rcassoc] = input;
+  const auto [mc_beam_electrons, mc_beam_hadrons, mcparts, rcparts, rcassoc] = input;
   auto [hadronicfinalstate]                                                  = output;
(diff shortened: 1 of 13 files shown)


wdconinc:
I'm resolving the repeated comments about multiple beam hadrons. This is not how we model ion collisions.

veprbl (approved):

epic-capybara:
- #2960

PR #2965, opened 2026-09-15
fix(tracking): weight track truth associations per measurement

tom-bleher:
### Briefly, what does this PR introduce? Please link to any relevant presentations or discussions.

`ActsToTracks` currently increments the MC-particle association weight once for every matching RawTrackerHit -> SimTrackerHit association, even though the code describes the weight as being based on used measurements. A detector measurement split into many SimTrackerHit records can therefore outweigh several independent measurements and change a reconstructed track's truth identity.

This PR makes each fitted measurement contribute one unit of total truth weight. Within a measurement it sums positive SimTrackerHit energy deposits by MC particle and normalizes those contributions to unit sum; if no positive deposits are available it falls back to association counts. Repeated SimTrackerHit ObjectIDs within a measurement are deduplicated. The existing final normalization of track-to-particle association weights is retained, and the existing ObjectID-based MCParticle comparator keeps ordering deterministic.

Focused Catch2 tests cover the step-multiplicity regression, energy-fraction weighting within a measurement, and the count fallback.

This is complementary to #2967: #2967 fixes **which SimHits and weights belong to a digitized silicon RawHit**, while this PR fixes **how those per-measurement truth contributions are accumulated across a fitted track**. Together they prevent subthreshold deposits and Geant4 step multiplicity from biasing downstream track truth identity.

The full eic-shell build/test matrix passed on the initial implementation; the repository-generated IWYU cleanup PR #2966 has now been merged into this branch and a fresh CI run has been triggered.

### What is the urgency of this PR?
- [ ] High (please describe reason below)
- [x] Medium
- [ ] Low

### What kind of change does this PR introduce?
- [x] Bug fix (truth-association weights did not correspond to used measurements)
- [ ] New feature (issue #__)
- [ ] Optimization (issue #__)
- [ ] Updated parameters, constants (issue #__)
- [ ] Updated documentation
- [ ] other: __

### Please check if any of the following apply
- [ ] This PR requires changes to geometry (epic PR: __)
- [ ] This PR requires changes to EDM4eic (EDM PR: __)
- [ ] This PR introduces breaking changes. Please describe changes users need to make below.
- [x] This PR changes default behavior. Track-to-MC association weights are now measurement-normalized instead of SimTrackerHit-record-count normalized.
- [x] AI was used in preparing this PR. ChatGPT was used to review the association logic, implement the patch and focused tests, and prepare the PR text; the proposed behavior and code were checked against the existing `ActsToTracks` data flow and deterministic ObjectID ordering requirements.

PR #2967, opened 2026-09-15
fix(digi): make silicon truth associations match digitized charge

tom-bleher:

`SiliconTrackerDigi` applies its energy threshold while building the `RawTrackerHit` charge, but then used to create RawHit -> SimHit links by re-scanning **all** `SimTrackerHit`s with the same cell ID. A subthreshold deposit therefore contributed no digitized charge but was still presented downstream as a truth contributor to the fired cell.

This PR records the SimTrackerHits that actually pass the digitizer threshold and contribute to each RawTrackerHit, and creates links/associations only for those contributors. Link and association weights are set to each accepted SimTrackerHit's fraction of the summed accepted energy deposit in the cell, so truth weights are consistent with the charge model used to make the RawTrackerHit.

The digitization threshold behavior itself is unchanged: the threshold remains per `SimTrackerHit`, exactly as before.

Tests cover:
- a 12 keV accepted deposit plus a 9 keV subthreshold deposit in one cell: raw charge is 12 and only the accepted SimHit is associated;
- two accepted deposits of 12 and 18 keV: raw charge is 30 and association weights are 0.4/0.6.

This is complementary to #2965: this PR fixes **which SimHits and weights enter a detector measurement's truth**, while #2965 fixes **how multiple measurements are combined into a track-to-MC truth weight**. Together they make the downstream B0 truth/purity diagnostics correspond to digitized charge and measurement count rather than subthreshold deposits or Geant4 step multiplicity.

- [x] Medium
- [ ] Low

- [x] Bug fix (truth associations included deposits that did not contribute to the digitized hit)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

- [x] This PR changes default behavior. MCRecoTrackerHitAssociation/Link contents and weights now describe only SimHits that actually contributed to the RawTrackerHit charge.
- [x] AI was used in preparing this PR. ChatGPT was used to identify the inconsistency, implement the contributor bookkeeping and focused Catch2 tests, and prepare the PR text.


veprbl:
I don't understand, how do you arrive at multiple sim hits in a single cell? Shouldn't DD4hep merge those for you?

ShujieL:
In this case of SVT, the only multiple-hits scenario we noticed is two hits (primary and secondary) on the same cell with one particle going through. Because MAPS is a threshold detector and the actual charge info is not used anywhere in tracking, we left the code as-is to make the track to particle association simpler. 
It is a good idea to revisit this and make the digitization scheme more  consistent with the actual detector response. Can you remind me your use case? Is that TOF or SVT?

tom-bleher:
Thank you Shujie and Dmitry for the comments. I am looking into B0 tracking.

This PR was actually opened unintentionally by my AI coding agent, and I had noticed it only now. I did not intend to open it for the time being.

I need to develop this further and see if it leads to something substantive. I think it is better that I close this for now and reopen later if it becomes relevant.

Sorry for the trouble.

PR #2972, opened 2026-09-15
ci: make get-docs-from-open-prs into non-matrix job

veprbl:
`get-docs-from-open-prs` produces a long list of jobs, that's distracting.


wdconinc:
Better to make reusable workflow which collapses and retains parallelism.

veprbl:
> Better to make reusable workflow which collapses and retains parallelism.

Can you suggest how?

wdconinc:
> > Better to make reusable workflow which collapses and retains parallelism.
> 
> Can you suggest how?

Doesn't quite work in the "suggest on the PR changes" way, but it would involve creating a workflow_call action. See e.g. https://github.com/eic/epic/blob/main/.github/workflows/convert-to-step.yml and https://github.com/eic/epic/blob/main/.github/workflows/linux-eic-shell.yml#L203

Shows up something like this:

veprbl:
I think, this is it?

wdconinc:
Ultimate (out of scope) this is the kind of development that then also allows us to move this into a separate repo, e.g. the way we already do with list-open-prs in https://github.com/eic/actions/tree/main/.github/workflows. Single place to manage this for all... Less code duplication.

veprbl:
That is understood. We will roll this out on EICrecon and then move to eic/actions for wider adoption.

wdconinc (approved):

PR #2980, opened 2026-09-18
Fix: ZDC extend cluster time window (e.g. for background-merged events)

baptistefraisse:
### Fix: ZDC extend cluster time window (e.g. for background-merged events)

Increase the ZDC cluster time upper limit from 269 ns to 5000 ns.

The current absolute time cut rejects valid signal events produced, for example, with SignalBackgroundMerger, where events can be distributed over a longer time frame.

This allows ZDC reconstruction for merged events using longer time frames.

Closes Issue #2802.
Closes: #2892


src/detectors/ZDC/ZDC.cc
```diff
@@ -180,7 +180,7 @@ void InitPlugin(JApplication* app) {
           .MIP           = 630. * dd4hep::keV,
           .Emin_in_MIPs  = 0.5,
           .delta_in_MIPs = 0.01,
-          .tmax          = 269 * dd4hep::ns,
+          .tmax          = 5000 * dd4hep::ns,
```
veprbl: This is fragile. We either fix this or remove the broken tmax functionality altogether (will also affect insert).

baptistefraisse: Thanks. I removed `tmax` from the HEXPLIT algorithm. This updated was pushed (new commit, same branch).

